PR #10, opened 2020-02-06
Fix smartling frontend asset deployment

lehnerm:


apps/smartling/lambda/package.json
```diff
@@ -4,17 +4,18 @@
   "private": true,
   "scripts": {
     "start": "LOCAL_DEV=true ts-node src/index.ts",
-    "clean": "rimraf built",
+    "clean": "rimraf built static",
     "build-lambda": "npm run clean && tsc",
     "build-frontend": "cd ../frontend && npm run build",
+    "copy-static": "cp -R node_modules/smartling-frontend/build static",
```
davidfateh: Should we keep `v` for the logs?

```suggestion
    "copy-static": "cp -Rv node_modules/smartling-frontend/build static",
```

PR #17, opened 2020-02-07
Update extension dependencies across all apps

lehnerm:


lerna-debug.log
```diff
@@ -0,0 +1,6 @@
+0 silly argv { _: [ 'clean' ], lernaVersion: '3.20.2', '$0': 'lerna' }
```
davidfateh: I think we should gitignore this one.

PR #20, opened 2020-02-10
AI Image Tagging

kdamball:


apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
+import { CheckboxField, TextInput, Pill, Button } from '@contentful/forma-36-react-components';
+
+import { styles } from './styles';
+
+async function callAPI(url) {
+  const data = await fetch(`http://localhot:3000/tags/${url}`).then(res => res.json());
```
jelz: don't mix `async/await` with promises

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -0,0 +1,136 @@
...
+    this.setState(state => ({
+      overwrite: !state.overwrite
+    }));
+  }
+
+  addTag = async (e) => {
+    if (e.key !== "Enter") { return; }
+
+    const newTags = [e.target.value, ...this.state.tags];
+    try {
```
jelz: Why `try/catch`

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -0,0 +1,136 @@
...
+      this.setState({
+        tags: newTags
+      });
+    } catch(e) {
+    }
+  }
+
+  fetchTags =  async () => {
+    const file = await this.props.space.getAsset(this.props.entries.image.getValue().sys.id);
+    const url = file.fields.file['en-US'].url.replace(/^\/\/images.ctfassets.net\//, '');
```
jelz: You cannot assume they've got `en-US`. To get the default locale code use `locales.default` as described in https://www.contentful.com/developers/docs/extensibility/ui-extensions/sdk-reference/#locales

jelz: Instead of using `replace` I'd parse the URL w/ https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

jelz: I'd also use something like `lodash.get` to navigate these deeply nested data structures

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
jelz: In the final version we can just change it to `/tags/...` since both frontend and backend are served from the same domain

apps/ai-image-tagging/frontend/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,171 @@
...
+    const { isInstalled, allContentTypesIds, contentTypeId, contentTypeName } = this.state;
+
+    return (
+      <>
+        <div className={styles.background} />
+        <div className={styles.body}>
+          <Typography>
+            <Heading className={styles.heading}>About AI Image Tagging</Heading>
+            <Paragraph>
+              The AI Image Tagging app uses AWS Rekognition’s AI and deep learning capabilities to automatically assign tags to images. The tags are visible and accessible via the Contentful web app sidebar and tags are searchable in the search bar.
```
jelz: We shouldn't be mentioning the vendor we use (Rekognition), we're whitelabelling it with this app

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
kdamball: `async/await` pattern for doing something with errors. Not sure how to handle them yet but at least this way I dont forget

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
kdamball: On the mixing part - mostly avoiding another line of `await res.json()`. I can chnage it tho

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
jelz: yeah, but now you just swallow them so you won't notice when they happen so it's more likely to forget

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -0,0 +1,134 @@
...
+      const aiTags = await callAPI(imagePath);
+
+      // upload new tags
+      const newTags = this.state.overwrite ? aiTags : [...aiTags, ...this.state.tags];
+      await this.props.entries.imageTags.setValue(newTags)
+      this.setState({
+        tags: newTags
+      });
+    } catch(e) {
+      this.props.notifier.error(e.message)
```
jelz: We should display a crafted message rather than what is included in the `Error` instance. `Tagging image failed.` would be a good start, if you can distinguish between things that can go wrong (here I assume we may fail on HTTP in `callAPI`) then I'd differentiate messages

kdamball: Updated it

PR #21, opened 2020-02-11
Less extensions in apps

jelz:


lib/shared-sku-app/package.json
```diff
@@ -40,8 +40,7 @@
   "scripts": {
     "tsc": "tsc -p ./ --noEmit",
     "test": "contentful-extension-scripts test --env=jsdom --watch",
-    "test:coverage": "contentful-extension-scripts test --env=jsdom --coverage --passWithNoTests",
-    "test:ci": "npm run test:coverage"
```
davidfateh: 😆

apps/cloudinary/package.json
```diff
@@ -19,15 +19,8 @@
   },
   "scripts": {
     "start": "contentful-extension-scripts start --serve-only",
-    "build": "contentful-extension-scripts build",
-    "deploy": "npm run build && aws s3 sync ./build ${STATIC_APPS_BUCKET}/apps/cloudinary --acl public-read",
-    "test": "contentful-extension-scripts test --env=jsdom --watch",
-    "test:coverage": "contentful-extension-scripts test --env=jsdom --coverage --passWithNoTests",
-    "test:ci": "npm run test:coverage",
```
lehnerm: Does that also disable linting of the app?

jelz: There is no linting in the current build pipeline...

PR #22, opened 2020-02-11
Lint in CI

jelz:
There's some weird typing issue with GA, Smartling and Jira. Renamed tasks to `lint:fixme` so they don't run.


jelz:
I've let David and Taye know and he's taking a look. I'd still merge as is and fix/improve as a followup

PR #23, opened 2020-02-11
AI tagging usage

jelz:


apps/ai-image-tagging/lambda/serverless.yml
```diff
@@ -35,10 +44,30 @@ package:
...
           path: /{proxy+}
           method: ANY
+
+resources:
+  Resources:
+    UsageTable:
+      Type: AWS::DynamoDB::Table
+      Properties:
+        TableName: ${self:custom.tableName}
+        AttributeDefinitions:
```
lehnerm: For a time series, I'd suggest we have the spaceId as the primary key and the period as a number represented as a timestamp truncated to the series granularity (see for instance https://momentjs.com/docs/#/manipulating/start-of/)

apps/ai-image-tagging/lambda/usage.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const { promisify } = require('util');
+
+const AWS = require('aws-sdk');
+
+module.exports = async spaceId => {
+  const ddb = new AWS.DynamoDB();
+  const updateItem = promisify(ddb.updateItem.bind(ddb));
```
lehnerm: I'd suggest using DynamoDB DocumentClient (https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB/DocumentClient.html) as it does DynamoDB deserialization for you

The AWS SDK also supports promises with `.promise()` on the result of a request (https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/using-promises.html) so you can get rid of `promisify`

apps/ai-image-tagging/lambda/usage.js
```diff
@@ -0,0 +1,39 @@
...
+
+  const now = new Date();
+  const year = `${now.getFullYear()}`;
+  const month = `${now.getMonth()+1}`;
+  const formattedMonth = month.length === 1 ? `0${month}` : month;
+  const period = [year, formattedMonth].join('-');
```
lehnerm: The standard approach for timeseries is to express the timestamp as number of ms since epoch truncated to the granularity

apps/ai-image-tagging/lambda/serverless.yml
jelz: What the benefits would be? I think having it as it is will result in the easiest queries.

lehnerm: The timestamp representation is used for storage efficiency and access performance reasons - especially for slices of data. Also we can use standard functions and don't have to do manual date rendering which often leads to nasty bugs. Spend too much time of my life chasing them down.

apps/ai-image-tagging/lambda/usage.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const AWS = require('aws-sdk');
+
+const documentClient = new AWS.DynamoDB.DocumentClient();
+
+module.exports = async spaceId => {
+  const now = new Date();
+  const firstDay = new Date(now.getFullYear(), now.getMonth(), 1);
```
lehnerm:
```suggestion
  const firstDay = Date.UTC(now.getFullYear(), now.getMonth(), 1);
```

It pays to be really explicit with timezones. This will also directly give you msecs.

apps/ai-image-tagging/lambda/usage.js
```diff
@@ -0,0 +1,35 @@
...
+
+const AWS = require('aws-sdk');
+
+const { DateTime } = require('luxon');
+
+const documentClient = new AWS.DynamoDB.DocumentClient();
+
+module.exports = async spaceId => {
+  const now = new Date();
+  const dt = DateTime.fromObject({
```
lehnerm:
```suggestion
  const period = DateTime.utc().startOf('month').toSeconds()
```

does everything from line 10 to line 17

PR #25, opened 2020-02-13
[WIP] ecommerce consistency & fixes

giotiskl:


lib/shared-sku-app/tsconfig.json
```diff
@@ -7,7 +7,11 @@
     "rootDir": "./src",
     "experimentalDecorators": true,
     "esModuleInterop": true,
-    "lib": ["dom", "es2015"],
+    "lib": [
+      "es6",
+      "dom",
+      "es2017"
+    ],
```
davidfateh: What's with this change?

giotiskl: tests were failing because TS compiler was complaining that `.includes` isn't a method on type of `string[]`

PR #26, opened 2020-02-13
AI Image Tagging Improvements (QA feedback)

kdamball:
- BgColor on install view to match icon
- Ignore duplicate tags
- Only allow image type: png, jpg, & jpeg


apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -28,41 +28,64 @@ export class AITagView extends React.Component {
...
 
+  validateImage = async () => {
+    const imageId = get(this.props.entries.image.getValue(), 'sys.id');
+    if (!imageId) { return true; }
+
+    const file = await this.props.space.getAsset(imageId);
+    const locale = this.props.locale;
+    const fullURL = get(file, `fields.file.${locale}.url`);
+    // test if file extension is PNG/JPEG/JPG
+    const isImageValid = new RegExp(/\.png|jpeg|jpg$/, 'i').test(fullURL);
```
jelz: There is `contentType` property next to `url` which is more reliable (it's inferred from the binary rather than from the file extension).

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -118,12 +139,16 @@ export class AITagView extends React.Component {
           ))
         }
       </div>
+      {
+        this.state.unsupportedImageType && !this.state.isMissingImage &&
+        <Note noteType="warning" className={ styles.fileWarning }>Unfortunately, we can only auto-tag PNG/JPG/JPEG file types</Note>
```
jelz: I wouldn't differentiate between jpg and jpeg. I'd also allow gifs. So "we can only auto-tag PNG, GIF and JPG.

kdamball: The problem actually is that gifs can't be tagged :(

jelz: okiedokie

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -51,9 +51,9 @@ export class AITagView extends React.Component {
 
     const file = await this.props.space.getAsset(imageId);
     const locale = this.props.locale;
-    const fullURL = get(file, `fields.file.${locale}.url`);
+    const contentType = get(file, `fields.file.${locale}.contentType`);
     // test if file extension is PNG/JPEG/JPG
-    const isImageValid = new RegExp(/\.png|jpeg|jpg$/, 'i').test(fullURL);
+    const isImageValid = new RegExp(/image\/png|jpeg|jpg$/, 'i').test(contentType);
```
jelz: why RegExp constructor instead of `/.../i`?

alternatively you'd just define valid mimetypes:

`['image/png', 'image/jpg', ...].includes(contentType.trim().toLowerCase())`

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -141,7 +141,7 @@ export class AITagView extends React.Component {
       </div>
       {
         this.state.unsupportedImageType && !this.state.isMissingImage &&
-        <Note noteType="warning" className={ styles.fileWarning }>Unfortunately, we can only auto-tag PNG/JPG/JPEG file types</Note>
+        <Note noteType="warning" className={ styles.fileWarning }>Unfortunately, we can only auto-tag PNG & JPG file types</Note>
```
jelz: don't use "&" for "and"

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
kdamball: It reads better in my opinion, i.e: testing some string against a rule. `[].includes` felt a bit awkward without writing a few more lines.

PR #30, opened 2020-02-14
GA: address QA issues

taye:


apps/google-analytics/src/Timeline.tsx
```diff
@@ -93,12 +89,20 @@ export default class Timeline extends React.Component<TimelineProps, TimelineSta
     }
 
     const { dimensions, start, end, pagePath, viewId } = this.props;
+    // pathWithoutTrailingSlash
+    let filters = `ga:pagePath==${pagePath}`
+
+    // filter to path with and without trailing slash if the slug doesn't end
+    // with a trailing slash
+    if (pagePath && !pagePath.endsWith('/')) {
+      filters = `${filters},ga:pagePath==${pagePath}/`
```
davidfateh: Am I understanding correctly that the final `filters` string should contain a path with an ending slash and one without all the time?

apps/google-analytics/src/index.tsx
```diff
@@ -47,30 +53,41 @@ export class SidebarExtension extends React.Component<
...
   render() {
-    const { isAuthorized, pagePath, hasSlug, contentTypeId } = this.state;
+    const { isAuthorized, isContentTypeConfigured, pagePath, hasSlug, contentTypeName, helpText } = this.state;
     const { parameters, entry, notifier } = this.props.sdk;
     const { clientId, viewId } = parameters.installation as SavedParams;
 
+    const helpTextFrag = helpText && <>
+      <div className={styles.spaced}/>
+      <Paragraph className={styles.lightText}>
+      {helpText}. See <TextLink href={docsUrl}>this app&apos;s docs</TextLink> for help.
```
davidfateh: Let's add target blank and noopener noreferrer

apps/google-analytics/src/Timeline.tsx
lehnerm: Are the filters combined with `OR`, not with `AND`?

lehnerm: Why not just use a regex? https://developers.google.com/analytics/devguides/reporting/core/v3/reference#filters

apps/google-analytics/src/index.tsx
```diff
@@ -84,25 +101,37 @@ export class SidebarExtension extends React.Component<
...
+        {helpTextFrag}
+        </>
       );
     }
 
+    if (!isContentTypeConfigured) {
+      return <Paragraph className={styles.lightText}>The {contentTypeName}{' '}
+      content type hasn&apos;t been configured for use with this app. It must
+      have a field of type short text and must be added to the list of
+      content types in this app&apos;s configuration.</Paragraph>;
```
lehnerm: How about we add a deeplink to the app config screen?

apps/google-analytics/src/Timeline.tsx
taye: In case there's some special characters in the path that might mess up the regex

taye: If there's a trailing slash, the filter will only be for the url with a trailing slash. If there's no trailing slash, then we'll filter for both the slashed and unslashed url. I could also just make it always filter for both. Not sure why I didn't do that.

apps/google-analytics/src/index.tsx
taye: currently not possible

apps/google-analytics/src/Timeline.tsx
davidfateh: Makes sense to do both then.

PR #31, opened 2020-02-14
Build improvements

jelz:
This PR:

- [x] Cuts build 2-3 times thanks to use of concurrency=2 (matching number of vCPUs) for all tasks but bundling (experienced hanging processes), use of `npm ci`, skipping code coverage instrumentation, not reinstalling dependencies in deployment step
- [x] Separates building artifacts from testing/deploying them
- [x] Stops to default deployment stage to production (to avoid deployment to production from local environments)
- [x] Always deploys all apps to verify that the process wasn't broken in a feature branch

Note that the test deployment from all branches ends up in the same location (same bucket, same stage). It's not intended to be used for sharing because builds will eventually override each other. The sole purpose of deployment is to test if it wasn't broken.

Screen Shot 2020-02-14 at 15 01 03

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,9 @@ jobs:
       - checkout
       - run:
           name: Install dependencies
-          command: npm i
+          command: |
+            npm i
+            npm run bootstrap:ci
       - run:
           name: Lint
           command: npm run lint
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,9 @@ jobs:
           name: Lint
           command: npm run lint
       - run:
+          name: Build
+          command: npm run build
+      - run:
           name: Test
           command: npm run test
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,18 +23,21 @@ jobs:
       - run:
           name: Configure deployment
           command: |
-            static_bucket='cf-apps-static-test'
-            test "$CIRCLE_BRANCH" == 'master' && static_bucket='cf-apps-static'
-            echo "export STATIC_BUCKET=\"$static_bucket\"" >> $BASH_ENV
-            static_jira_bucket='cf-apps-jira-test'
-            test "$CIRCLE_BRANCH" == 'master' && static_jira_bucket='cf-apps-jira'
-            echo "export STATIC_JIRA_BUCKET=\"$static_jira_bucket\"" >> $BASH_ENV
+            static_s3_base="s3://cf-apps-static/apps-test-$CIRCLE_SHA1"
+            test "$CIRCLE_BRANCH" == 'master' && static_s3_base='s3://cf-apps-static/apps'
+            echo "export STATIC_S3_BASE=\"$static_s3_base\"" >> $BASH_ENV
+
+            static_jira_s3_base="s3://cf-apps-static/apps-test-$CIRCLE_SHA1/jira"
+            test "$CIRCLE_BRANCH" == 'master' && static_jira_s3_base='s3://cf-apps-jira'
+            echo "export STATIC_JIRA_S3_BASE=\"$static_jira_s3_base\"" >> $BASH_ENV
+
             stage='test'
             test "$CIRCLE_BRANCH" == 'master' && stage='prd'
             echo "export STAGE=\"$stage\"" >> $BASH_ENV
+
             source "$BASH_ENV"
-            echo "Static deployment bucket: $STATIC_BUCKET"
-            echo "Static Jira deployment bucket: $STATIC_JIRA_BUCKET"
+            echo "Static deployment S3 base: $STATIC_S3_BASE"
+            echo "Static Jira deployment S3 base: $STATIC_JIRA_S3_BASE"
             echo "API Gateway stage: $STAGE"
       - run:
           name: Install pip
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,24 +21,11 @@ jobs:
           name: Test
           command: npm run test
       - run:
-          name: Configure deployment
+          name: Exit successfully if not in master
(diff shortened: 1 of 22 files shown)


apps/smartling/lambda/package.json
```diff
@@ -4,17 +4,12 @@
...
-    "build-lambda": "npm run clean && tsc",
-    "build-frontend": "cd ../frontend && npm run build",
-    "build": "npm run clean && npm run build-lambda && npm run build-frontend",
+    "build": "rimraf built && tsc",
     "lint": "tslint --project ./tsconfig.json",
-    "lint-fix": "tslint --project ./tsconfig.json --fix",
     "pretty": "prettier --config ./.prettierrc --write '{src/*.ts,src/*.spec.ts}'",
-    "test": "LOCAL_DEV=true jest **/*.spec.ts",
-    "test:ci": "npm run build-frontend && npm run test",
-    "test-dev": "LOCAL_DEV=true jest **/*.spec.ts --watch",
```
davidfateh: `LOCAL_DEV=true` may be needed for local testing. A few things have changed here so maybe it is outdated.

jelz: We only use it here: https://github.com/contentful/apps/blob/master/apps/smartling/lambda/src/index.ts#L119-L122 to treat app file as a server.

Tests still work w/o that.

.circleci/config.yml
```diff
@@ -1,54 +1,56 @@
...
+      - run:
+          name: Build
+          command: npm run build
       - run:
           name: Test
           command: npm run test
+      - run:
+          name: Configure deployment
+          command: |
+            static_s3_base="s3://cf-apps-static/apps-test-$CIRCLE_SHA1"
```
lehnerm: Let's please not use the prd bucket for test deployments - we should get a second one for that.

.circleci/config.yml
```diff
@@ -1,54 +1,56 @@
...
-      - image: circleci/node:10-stretch
-        environment:
-          STATIC_APPS_BUCKET: s3://cf-apps-static
-    steps:
-      - checkout
+            static_jira_s3_base="s3://cf-apps-static/apps-test-$CIRCLE_SHA1/jira"
+            test "$CIRCLE_BRANCH" == 'master' && static_jira_s3_base='s3://cf-apps-jira'
+            echo "export STATIC_JIRA_S3_BASE=\"$static_jira_s3_base\"" >> $BASH_ENV
+
+            stage='test'
```
lehnerm: I disagree with the fact that we need to have lambda deployment run on every branch. Especially with this setup, in which weird  side effects will most certainly occur when deployment is run on a single stage from multiple branches.

package.json
```diff
@@ -1,14 +1,15 @@
...
   "name": "apps",
+  "private": true,
   "scripts": {
-    "lint": "lerna run lint --concurrency=1",
-    "test": "lerna run test:ci --concurrency=1",
-    "bootstrap": "lerna bootstrap --no-ci --concurrency=1",
-    "deploy": "lerna run deploy",
-    "postinstall": "npm run bootstrap"
+    "bootstrap": "lerna bootstrap --no-ci",
+    "bootstrap:ci": "lerna bootstrap --ci --concurrency=2",
```
lehnerm: I've disabled lerna concurrency due to CI flakiness for both test and bootstrap. Couldn't find much information on that - any experience with this?

.circleci/config.yml
jelz: I think benefits outweigh problems. Right now the first time we attempt to deploy from CI to Lambda is after we merge to master. This blocks further deployments until the problem is solved (likely: CI role cannot do something) or the change is reverted. By attempting to deploy in branches we mitigate this risk.

package.json
jelz: So I did some digging and:

- Using `npm ci` (so `lerna --ci`) is faster and more lightweight (according to npm: It can be significantly faster than a regular npm install by skipping certain user-oriented features) - never experienced any crashes/hanging with `--ci`
- Linting, testing and deployment can also be reliably run with concurrency of 2. 2 matches number of vCPUs we get in Circle
- The only problematic step is the build step. I don't know if it's problem with parcel (cache interference?) or it's so CPU-bound that it has hard time running in parallel but this is the only step that requires concurrency of 1 for stability

In the previous model building artifacts was part of test and deploy steps (sometimes repeated). Now we build all packages and then run tests on them and deploy in separate steps.

To sum up we build in series and everything else is done with parallelism of 2.

.circleci/config.yml
jelz: Makes sense.

.circleci/config.yml
lehnerm: As long as we just block master, I think this is fine. In case we need to ship something else fast, we can still revert the changes. If we're worried that the prd deployment might actually break something important, I'd suggest we add a dev deployment stage in the master build.

jelz: Do you mean merge to master = staging, merge to `prod` branch = deployment to production?

jelz: Other issue with the current state of affairs: frontend gets deployed, backend deployment fails. Frontend relies on new backend version. App is broken.

lehnerm: I mean that the master build would be something like
1) deploy to `dev`
2) smoke test dev (:smiling_imp:)
3) deploy to `prd`

Frontend and backend will always have to remain compatible between individual changes, since we cannot guarantee an instant client side code update due to caching anyway. We also cannot gurantee the order of deployment.

I'd also like to avoid the `prod` branch approach. I think we both agreed that it is a very problematic solution.

package.json
lehnerm: I was running into this with bootstrap: https://github.com/lerna/lerna/issues/1457

PR #33, opened 2020-02-18
AI Image tags QA Feedback

kdamball:
- Adding extra file validations (image dimensions and file size limits)
- Only show warning for content id collision on content id field and not name

### DIFF
--- a/apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
+++ b/apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
@@ -29,7 +29,7 @@ export class AITagView extends React.Component {
       overwrite: true,
       isMissingImage: !props.entries.image.getValue(),
       unsupportedImageType: false,
-      imageRequirementInvalid: false,
+      imageRequirementsNotMet: false,
       isFetchingTags: false
     }
 
@@ -47,8 +47,8 @@ export class AITagView extends React.Component {
   }
 
   validateImage = async () => {
-    const FILE_LIMIT = 5000000; // 5MB limit from AWS Rekognition
-    const FILE_DIMENSION_LIMIT = 80; // 80px limit
+    const MAX_FILE_SIZE = 5 * Math.pow(2, 20); // 5MB limit from AWS Rekognition
+    const MIN_DIMENSION_SIZE = 80; // 80px limit
     const imageId = get(this.props.entries.image.getValue(), 'sys.id');
     if (!imageId) { return ; }
 
@@ -57,14 +57,14 @@ export class AITagView extends React.Component {
     const contentType = get(file, `fields.file.${locale}.contentType`);
     const details = get(file, `fields.file.${locale}.details`);
     // test if file extension is PNG/JPEG/JPG
-    const isImageTypeValid = new RegExp(/image\/png|jpeg|jpg$/, 'i').test(contentType);
-    const isImageIncompatible = details.size > FILE_LIMIT ||
-                                details.width < FILE_DIMENSION_LIMIT ||
-                                details.height < FILE_DIMENSION_LIMIT;
+    const isImageTypeValid = new RegExp(/^image\/(png|jpe?g)$/, 'i').test(contentType);
+    const isImageIncompatible = details.size > MAX_FILE_SIZE ||
+                                details.width < MIN_DIMENSION_SIZE ||
+                                details.height < MIN_DIMENSION_SIZE;
 
     this.setState(() => ({
       unsupportedImageType: !isImageTypeValid,
-      imageRequirementInvalid: isImageIncompatible
+      imageRequirementsNotMet: isImageIncompatible
     }))
   }
 
@@ -124,7 +124,7 @@ export class AITagView extends React.Component {
   }
 
   render() {
-    let hasImageError = !this.state.isMissingImage && (this.state.unsupportedImageType || this.state.imageRequirementInvalid)
+    let hasImageError = !this.state.isMissingImage && (this.state.unsupportedImageType || this.state.imageRequirementsNotMet)
     let imageErrorMsg = this.state.unsupportedImageType ? "Unfortunately, we can only auto-tag PNG and JPG file types" : "Please make sure your image is less than 5MB and has dimensions of at least 80px for both width and height";
 
     return <div className={ styles.inputWrapper }>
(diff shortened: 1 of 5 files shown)


apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -46,17 +47,24 @@ export class AITagView extends React.Component {
   }
 
   validateImage = async () => {
+    const FILE_LIMIT = 5000000; // 5MB limit from AWS Rekognition
```
jelz: I'd call it `MAX_FILE_SIZE`. Also 5MB is `5 * Math.pow(2, 20)`

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -46,17 +47,24 @@ export class AITagView extends React.Component {
...
+    const FILE_DIMENSION_LIMIT = 80; // 80px limit
```
jelz: I'd call it `MIN_IMAGE_DIMENSION`

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -46,17 +47,24 @@ export class AITagView extends React.Component {
...
-    if (!imageId) { return true; }
+    if (!imageId) { return ; }
 
     const file = await this.props.space.getAsset(imageId);
     const locale = this.props.locale;
     const contentType = get(file, `fields.file.${locale}.contentType`);
+    const details = get(file, `fields.file.${locale}.details`);
     // test if file extension is PNG/JPEG/JPG
-    const isImageValid = new RegExp(/image\/png|jpeg|jpg$/, 'i').test(contentType);
+    const isImageTypeValid = new RegExp(/image\/png|jpeg|jpg$/, 'i').test(contentType);
```
jelz: I think this regex is missing a capture group. I'd do `/^image\/(png|jpe?g)$/i`

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -46,17 +47,24 @@ export class AITagView extends React.Component {
...
+    const isImageIncompatible = details.size > FILE_LIMIT ||
+                                details.width < FILE_DIMENSION_LIMIT ||
+                                details.height < FILE_DIMENSION_LIMIT;
 
     this.setState(() => ({
-      unsupportedImageType: !isImageValid
+      unsupportedImageType: !isImageTypeValid,
+      imageRequirementInvalid: isImageIncompatible
```
jelz: I'd call it `imageRequirementsNotMet`

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -46,17 +47,24 @@ export class AITagView extends React.Component {
   }
 
   validateImage = async () => {
+    const FILE_LIMIT = 5000000; // 5MB limit from AWS Rekognition
+    const FILE_DIMENSION_LIMIT = 80; // 80px limit
     const imageId = get(this.props.entries.image.getValue(), 'sys.id');
-    if (!imageId) { return true; }
+    if (!imageId) { return ; }
```
davidfateh: Should this function return a bool? I know `undefined` is falsey but it seems more semantic to make sure the return value is something expected.

```suggestion
    if (!imageId) { return false; }
```

In any case, do we not use prettier in this app?
```suggestion
    if (!imageId) { return; }
```

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -46,17 +47,24 @@ export class AITagView extends React.Component {
...
 
     const file = await this.props.space.getAsset(imageId);
     const locale = this.props.locale;
     const contentType = get(file, `fields.file.${locale}.contentType`);
+    const details = get(file, `fields.file.${locale}.details`);
```
davidfateh: Is `locale` guaranteed here?

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
davidfateh: Agreed, since the logic using it looks for `<`

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
kdamball: I'm not sure actually. I assumed all assets have a default locale. @jelz any gotchas I missed?

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
kdamball: It's a `void` fn. I didn't want to return a bool which is what I did before but it ended up confusing me why I was returning a value in the first place.

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
jelz: Yes it's guaranteed, we use `sdk.locales.default` which is always present and has a valid code of the default locale

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
kdamball: I wasn't sure if they are strict on binary vs decimal, hence played it safe.

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
kdamball: why do we need a capture group?

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
```diff
@@ -46,17 +47,24 @@ export class AITagView extends React.Component {
...
     // test if file extension is PNG/JPEG/JPG
-    const isImageValid = new RegExp(/image\/png|jpeg|jpg$/, 'i').test(contentType);
+    const isImageTypeValid = new RegExp(/image\/png|jpe?g$/, 'i').test(contentType);
```
jelz: Please indicate the string should start with `image` with `^image`.

Capture group is needed to limit alternative (`|`):

```
> /image\/jpg|png/.test('png')
true
> /image\/(jpg|png)/.test('png')
false
```

kdamball: Good catch

apps/ai-image-tagging/frontend/src/components/AITagView/AITagView.js
davidfateh: Gotcha, thanks for the explanation.

PR #36, opened 2020-02-18
fixing bug where saved project is not selected on app config load

davidfateh:
I also ran prettier as well so there are some unrelated style changes. I will point out the fix in the comments.

### DIFF
--- a/apps/jira/jira-app/src/components/App/Config.tsx
+++ b/apps/jira/jira-app/src/components/App/Config.tsx
@@ -163,11 +163,12 @@ export default class Config extends React.Component<Props, State> {
       const data = await JiraClient.getProjects(resource.id, this.props.token);
 
       if (data.projects.length) {
-        this.setState({
+        this.setState(prevState => ({
           // if there is only one project, automatically pick it and move on
-          checkedProject: data.projects.length === 1 ? data.projects[0].id : '',
+          checkedProject:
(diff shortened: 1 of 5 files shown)


apps/jira/jira-app/src/components/App/Config.tsx
```diff
@@ -163,11 +163,12 @@ export default class Config extends React.Component<Props, State> {
       const data = await JiraClient.getProjects(resource.id, this.props.token);
 
       if (data.projects.length) {
-        this.setState({
+        this.setState(prevState => ({
           // if there is only one project, automatically pick it and move on
-          checkedProject: data.projects.length === 1 ? data.projects[0].id : '',
+          checkedProject:
+            data.projects.length === 1 ? data.projects[0].id : prevState.checkedProject,
```
davidfateh: This is the fix. It was accidentally resetting state with this logic.

apps/jira/jira-app/src/components/Jira/Search.tsx
```diff
@@ -27,7 +27,7 @@ export default class Search extends React.Component<Props, State> {
   async componentDidUpdate(_prevProps: Props, prevState: State) {
     const { value } = this.state;
 
-    if (value.length > 4 && value !== prevState.value) {
+    if (value.length > 2 && value !== prevState.value) {
```
davidfateh: I also changed this to get the search firing faster.

PR #37, opened 2020-02-19
Jira fixes

davidfateh:
Fixes:
1. Fixing issue where selected project doesn't automatically show up on the configuration screen after installation.
2. Fixing issue where a 401 or 403 wouldn't indicate any type of error. Now a user can choose to 'sign out' and go through the auth flow again.
3. Changing the amount of characters required in the search field to start a search. Was originally 4 now it is 2. This should get the search firing a bit faster.

### DIFF
--- a/apps/jira/jira-app/src/components/Jira/Search.tsx
+++ b/apps/jira/jira-app/src/components/Jira/Search.tsx
@@ -27,7 +27,7 @@ export default class Search extends React.Component<Props, State> {
   async componentDidUpdate(_prevProps: Props, prevState: State) {
     const { value } = this.state;
 
-    if (value.length > 4 && value !== prevState.value) {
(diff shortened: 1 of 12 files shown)


apps/jira/jira-app/src/components/Jira/Search.tsx
```diff
@@ -27,7 +27,7 @@ export default class Search extends React.Component<Props, State> {
   async componentDidUpdate(_prevProps: Props, prevState: State) {
     const { value } = this.state;
 
-    if (value.length > 4 && value !== prevState.value) {
+    if (value.length > 2 && value !== prevState.value) {
```
kdamball: Any reason why start a search with fewer keystrokes? Maybe a debounce of sorts in case they're still typing?

PR #41, opened 2020-02-19
Ecommerce apps missing products bug

giotiskl:
Fixes the behavior of all ecommerce apps for missing products

### DIFF
--- a/apps/commercetools/src/AppConfig/index.spec.tsx
+++ b/apps/commercetools/src/AppConfig/index.spec.tsx
@@ -70,8 +70,8 @@ describe("AppConfig", () => {
       [/commercetools Project Key/, ""],
       [/Client ID/, ""],
       [/Client Secret/, ""],
-      [/^API Endpoint/, ""],
-      [/Auth API Endpoint/, ""],
+      [/^API URL/, ""],
+      [/Auth URL/, ""],
       [/commercetools data locale/, ""]
     ].forEach(([labelRe, expected]) => {
       const configInput = getByLabelText(labelRe) as HTMLInputElement;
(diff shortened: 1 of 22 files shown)


apps/commercetools/src/Editor/CategoryPreviews/SortableListItem.tsx
```diff
@@ -71,48 +76,47 @@ const styles = {
...
     return (
       <Card className={styles.card}>
         <>
           {isSortable && <CardDragHandle />}
           <section className={styles.description}>
             <Typography>
-              <Heading className={styles.name(category.name)}>
-                {categoryIsMissing ? category.id : category.name}
+              <Heading className={styles.heading(category)}>
+                {category.isMissing || !category.name
```
davidfateh: Optional: you use the same bool logic here as you do when you pass it into the `styles.heading` func. You could have that logic in var and reuse it.

apps/shopify/src/Pagination.js
```diff
@@ -17,7 +14,7 @@ class Pagination {
 
   variants = [];
 
-  prevSearch = "";
+  prevSearch = '';
```
davidfateh: 🤣

apps/commercetools/src/AppConfig/index.spec.tsx
```diff
@@ -70,8 +70,8 @@ describe("AppConfig", () => {
       [/commercetools Project Key/, ""],
       [/Client ID/, ""],
       [/Client Secret/, ""],
-      [/^API Endpoint/, ""],
```
jelz: you've got `singleQuote: true` in prettierrc above, why wasn't this fixed?

apps/shopify/src/productResolvers.js
```diff
@@ -30,11 +33,23 @@ export const fetchProductPreviews = async (skus, config) => {
...
+    .map(sku => {
+      try {
+        // If not valid base64 window.atob will throw
+        const unencodedId = atob(sku);
+        return { unencodedId, sku };
+      } catch (error) {
+        return null;
+      }
+    })
+    .filter(sku => sku && /^gid.*ProductVariant/.test(sku.unencodedId))
```
davidfateh: What exactly is the regex testing? It seems like this functionality takes the skus, decodes them, tests the decoded value but then uses the encoded value finally.

giotiskl: that's correct, the encoded value is what has to be passed in to the graphql query, but i need to decode them to determine which ones are validly formatted SKUs, then the rest (which are not) are put into the invalid SKUs array to be rendered as "missing products"

apps/commercetools/src/AppConfig/index.spec.tsx
giotiskl: I didn't run prettier on this file.. prettier is messed up all over the new apps repo

PR #47, opened 2020-02-25
Copy fixes for AI tagging

kdamball:


kdamball:
@davidfateh yep - edited the commit

PR #50, opened 2020-02-25
updating smartling copy

davidfateh:


apps/smartling/frontend/src/__snapshots__/index.spec.tsx.snap
```diff
@@ -398,7 +398,7 @@ Object {
                   class="Paragraph__Paragraph___2aO-9 about-p f36-margin-bottom--m"
                   data-test-id="cf-ui-paragraph"
                 >
-                  This app allows you to view the translation status of entries in the Contentful web app and easily find corresponding entries within a 
+                  The
```
kdamball: Is there a reason to have these on different lines, probably doesn't affect the rendering for sure but I'm curious.

davidfateh: I believe that's just formatting of the snapshot. I didn't generate this code.

PR #51, opened 2020-02-26
Ecommerce responsive picker

giotiskl:


lib/shared-sku-app/src/SkuPicker/styles.ts
```diff
@@ -1,5 +1,5 @@
-import { css } from 'emotion';
-import tokens from '@contentful/forma-36-tokens';
+import { css } from "emotion";
```
taye: shouldn't these be single quotes according to the prettier rule?

giotiskl: yup, run it for the file but prettier is messed up all over the apps repo when the code was migrated over unfortunately :/

PR #53, opened 2020-03-03
adding dev oauth credentials

davidfateh:


apps/jira/jira-app/src/components/Jira/IssueList.tsx
```diff
@@ -28,9 +28,7 @@ export default class IssueList extends React.Component<Props> {
     if (!issues.length) {
       return (
         <Typography>
-          <Paragraph className="paragraph-light">
-            No Jira issues are linked to this entry
-          </Paragraph>
+          <Paragraph className="paragraph-light">No Jira issues are linked to this entry</Paragraph>
```
davidfateh: Updating pretty

apps/jira/jira-app/src/constants.ts
```diff
@@ -1,13 +1,15 @@
 /* global process */
 
 let LAMBDA_URI = 'https://api.jira.ctfapps.net';
+let CLIENT_ID = 'XD9k9QU9VT4Rt26u6lbO3NM0fOqvvXan';
 
 if (process.env.nodeEnv === 'development') {
```
kdamball: On a second thought - might be better to have this as `process.env.nodeEnv !== 'development'`

davidfateh: I have a feeling changing that would require some other work. It's safer to assume it's always production for now and explicitly state when we are in development.

PR #56, opened 2020-03-05
Search for JIRA projects instead of listing all

kdamball:


kdamball:
It's definitely fixable but trying to understand exactly why it's failing.

PR #61, opened 2020-03-12
Adding Brandfolder (with final fixes)

giotiskl:


apps/brandfolder/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+  'filename',
+  'size',
+  'width',
+  'height',
+  'url',
+  'thumbnail_url',
+  'position',
+  'relationships',
+  'included',
+  'asset.id'
```
schwindtkevin: 'cdn_url' should also be allowed. Please add that, or I can create a PR for it.

davidfateh: I've added it here:
https://github.com/contentful/apps/pull/62

PR #65, opened 2020-03-18
feat: add typeform app

FLoppix:
This PR add the typeform integration:

- add typeform frontend 
- add lambda proxy backend


apps/typeform/.gitignore
```diff
@@ -0,0 +1 @@
+.envrc
```
davidfateh: How about adding `build` and `node_modules` as well to be safe?

apps/typeform/frontend/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "configure": "contentful space use && contentful space environment use",
+    "start": "contentful-extension-scripts start --serve-only",
+    "build": "contentful-extension-scripts build --no-inline",
+    "deploy": "npm run build && contentful extension update --force",
+    "login": "contentful login",
+    "logout": "contentful logout",
+    "help": "contentful-extension-scripts help",
+    "test": "contentful-extension-scripts test --env=jsdom",
+    "test:watch": "contentful-extension-scripts test --watch --env=jsdom",
+    "test:coverage": "contentful-extension-scripts test --coverage --env=jsdom"
```
davidfateh: We don't need some of these commands.
```suggestion
    "start": "contentful-extension-scripts start --serve-only",
    "build": "contentful-extension-scripts build --no-inline",
    "test": "contentful-extension-scripts test --env=jsdom",
    "test:watch": "contentful-extension-scripts test --watch --env=jsdom",
```

apps/typeform/frontend/package.json
```diff
@@ -0,0 +1,61 @@
...
+  },
+  "browserslist": [
+    "last 2 versions",
+    "IE >= 11",
+    "not IE 10"
+  ],
+  "jest": {
```
davidfateh: We can probably get rid of this.

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,231 @@
...
+    contentTypes: [],
+    compatibleFields: {},
+    workspaces: [],
+    selectedContentTypes: [],
+    selectedFields: {},
+    workspaceId: '',
+    accessToken: (window.localStorage.getItem('token') as string) || ''
+  };
+
+  async componentWillMount() {
```
davidfateh: I think we should use the did mount method.

```suggestion
  async componentDidMount() {
```

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,231 @@
...
+        selectedFields: editorInterfacesToSelectedFields(editorInterfaces, sdk.ids.app)
+      },
+      () => sdk.app.setReady()
+    );
+  }
+
+  fetchWorkspaces = async () => {
+    try {
+      const response = await fetch(
+        `${process.env.LAMBDA_ENDPOINT}/workspaces/${this.state.accessToken}`
```
davidfateh: Any reason why we don't put the access token in the authorization header instead?

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,231 @@
...
+      );
+      const result: WorkspacesResponse = await response.json();
+      this.setState({ workspaces: this.normalizeWorkspaceResponse(result) });
+    } catch (error) {
+      this.props.sdk.notifier.error(error);
```
davidfateh: We should put a custom error message here to guide the user on what went wrong. The error being sent here will likely be too technical.

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,231 @@
...
+    }
+  };
+
+  normalizeWorkspaceResponse = (response: WorkspacesResponse) => {
+    return response.workspaces.items.map(workspace => ({
```
davidfateh: Is there a guarantee that `workspaces` or `items` exists? Maybe we should add some more safety around this code.

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,231 @@
...
+      return false;
+    }
+
+    if (hasStaleWorkspaceIdSelected) {
+      this.props.sdk.notifier.error('Select a valid workspace.');
+      return false;
+    }
+
+    return {
+      parameters: { accessToken, workspaceId },
```
davidfateh: We shouldn't save the access token to installation parameters. That was the reason to pivot to using OAuth.

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,231 @@
...
+}
+
+export class AppConfig extends React.Component<Props, State> {
+  state: State = {
+    contentTypes: [],
+    compatibleFields: {},
+    workspaces: [],
+    selectedContentTypes: [],
+    selectedFields: {},
+    workspaceId: '',
```
davidfateh: Nit: Maybe we should name this `selectedWorkspaceId`?

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,231 @@
...
+                <FormLabel htmlFor="workspaceId" required>
+                  Select the Typeform workspace you want to connect
+                </FormLabel>
+                <Select
+                  id="workspaceId"
+                  name="workspaceId"
+                  onChange={(event: any) => this.setWorkSpaceId(event.currentTarget.value)}
+                  hasError={this.state.workspaces.length > 0 && !this.selectedWorkspaceIdIsValid()}
+                  value={this.state.workspaceId}
+                  data-test-id="typeform-select">
```
davidfateh: Should we add `disabled={this.state.workspaces.length === 0}`?

apps/typeform/frontend/src/Auth/TypeformOAuth.tsx
```diff
@@ -0,0 +1,50 @@
...
+  sdk: AppExtensionSDK;
+  expireSoon?: boolean;
+  setToken: (token: string) => void;
+}
+
+export function TypeformOAuth({ sdk, expireSoon, setToken, ...rest }: Props) {
+  let oauthWindow: Window | null;
+
+  useEffect(() => {
+    sdk.app?.setReady();
```
davidfateh: Are you sure we have the existential operator?

apps/typeform/frontend/src/Auth/TypeformOAuth.tsx
```diff
@@ -0,0 +1,50 @@
...
+    oauthWindow = window.open(url, 'Typeform Contentful', 'left=150,top=10,width=800,height=900');
+
+    window.addEventListener('message', handleTokenEvent);
+  };
+
+  const handleTokenEvent = ({ data }: any) => {
+    const { token, error } = data;
+
+    if (error) {
+      console.error('There was an error authenticating. Please refresh and try again.');
```
davidfateh: If there is an error, shouldn't we close the window and have that error show up somewhere in the webapp UI? I think leaving the window open won't really help much.

apps/typeform/frontend/src/Auth/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  return !expires || parseInt(expires, 10) - Date.now() <= _10Minutes;
+}
+
+function resetLocalStorage() {
+  window.localStorage.removeItem('token');
+  window.localStorage.removeItem('expireTime');
+}
+
+function geToken() {
```
davidfateh:
```suggestion
function getToken() {
```

apps/typeform/frontend/src/processTokenCallback.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (search.length) {
+    const error = searchParams.get('error');
+
+    if (error) {
+      window.opener.postMessage({ error }, '*');
+      return;
+    }
+
+    const token = searchParams.get('token') || '';
+    const expiresIn = parseInt(searchParams.get('expiresIn') || '', 10) || 3200;
```
davidfateh: Did this `3200` default come from typeform?

apps/typeform/frontend/src/Field/TypeFormField.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+const getSelectedForm = (value: string, forms: FormOption[]) => {
+  return forms.find(form => form.href === value) || initialState.selectedForm;
+};
+
+export function TypeFormField({ sdk }: Props) {
+  const { workspaceId } = sdk.parameters.installation as InstallationParameters;
+  const [state, dispatch] = useReducer(reducer, initialState);
+  const { loading, forms, value, hasStaleData, selectedForm, error, token } = state;
+  console.log(value);
```
davidfateh:
```suggestion

```

apps/typeform/frontend/src/Field/TypeFormField.tsx
```diff
@@ -0,0 +1,254 @@
...
+      default:
+        return state;
+    }
+  }
+
+  useEffect(() => {
+    const fetchForms = async () => {
+      try {
+        const response = (await (
+          await fetch(`${process.env.LAMBDA_ENDPOINT}/forms/${workspaceId}/${token}`)
```
davidfateh: Same question about why we don't have the token go into the header instead.

apps/typeform/frontend/src/Field/TypeFormField.tsx
```diff
@@ -0,0 +1,254 @@
...
+        ).json()) as TypeFormResponse;
+        const normalizedForms = normalizeFormResponse(response);
+        dispatch({
+          type: ACTION_TYPES.INIT,
+          payload: {
+            forms: normalizedForms
+          }
+        });
+      } catch (error) {
+        console.error(error);
```
davidfateh:
```suggestion

```

apps/typeform/frontend/src/Field/TypeFormField.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+  if (loading) {
+    return null;
+  }
+
+  if (error) {
+    return (
+      <Note noteType="negative">
+        We could not fetch your typeforms. Please make sure you are using a valid workspace ID and
+        access token.
```
davidfateh: I would argue that asking the user to check if they are using a valid access token is too confusing. Most users won't understand what that even means.

apps/typeform/frontend/src/Field/TypeFormField.tsx
```diff
@@ -0,0 +1,254 @@
...
+    </div>
+  );
+
+  return (
+    <React.Fragment>
+      <div className={styles.field}>
+        <img src={logo} className={styles.logo} />
+        <Select onChange={onChange} value={value} data-test-id="typeform-select">
+          <Option key="" value="">
+            {forms.length === 0 ? 'No forms available' : 'Choose typeform'}
```
davidfateh: Should this read:
```suggestion
            {forms.length === 0 ? 'No forms available' : 'Choose a typeform'}
```

apps/typeform/frontend/src/TypeFormWidget.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const el = useRef(null);
+  const { value } = sdk.parameters.invocation as Hash;
+  useEffect(() => {
+    let element = el.current;
+    typeformEmbed.makeWidget(element, value, {
+      hideFooter: true,
+      hideHeaders: true,
+      opacity: 0
+    });
+    sdk.window.updateHeight(450);
```
davidfateh: `450` should probably be a const since it reused.

apps/typeform/frontend/src/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+  </head>
```
davidfateh: Should we add a `<title>Typeform</title>`?

apps/typeform/frontend/src/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+        <TypeFormField sdk={sdk as AppExtensionSDK & FieldExtensionSDK} />,
+        document.getElementById('root')
+      );
+    }
+  });
+}
+/**
+ * By default, iframe of the extension is fully reloaded on every save of a source file.
+ * If you want to use HMR (hot module reload) instead of full reload, uncomment the following lines
+ */
```
davidfateh:
```suggestion

```

apps/typeform/frontend/src/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const isUserAuthenticated = () => {
+  return window.localStorage.getItem('token') ? true : false;
+};
+
+export function validateParameters({
+  workspaceId,
+  accessToken
+}: TypeFormParameters): string | null {
+  if (!workspaceId.length || typeof workspaceId !== 'string') {
```
davidfateh: I would switch this logic otherwise you may run into a situation where trying to access the `length` prop before checking to even see if it's a string would throw the classic "cannot read property length of undefined".
```suggestion
  if (typeof workspaceId !== 'string' || !workspaceId.length) {
```

apps/typeform/frontend/src/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+    return 'Workspace ID is invalid!';
+  }
+  if (!accessToken.length || typeof accessToken !== 'string') {
```
davidfateh: Same comment as above.

apps/typeform/frontend/src/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+    if (sdk.location.is(locations.LOCATION_APP_CONFIG)) {
+      render(<Authentication sdk={sdk as AppExtensionSDK} />, document.getElementById('root'));
+    } else if (sdk.location.is(locations.LOCATION_DIALOG)) {
+      render(
+        <TypeformPreviewWidget sdk={sdk as DialogExtensionSDK} />,
+        document.getElementById('root')
+      );
+    } else {
+      render(
+        <TypeFormField sdk={sdk as AppExtensionSDK & FieldExtensionSDK} />,
```
jelz: `AppExtensionSDK` is available only in `LOCATION_APP_CONFIG` and you handle it earlier: https://github.com/contentful/apps/pull/65/files#diff-8bd9bee2ac78785f05f6a7d94ee9f546R22-R23

apps/typeform/lambda/.gitignore
```diff
@@ -0,0 +1,6 @@
+/.serverless
+/static
+.openid
+built
+lambda/built
+lambda/.serverless
```
jelz: this looks like a mixup (`built` and `.serverless`)

apps/typeform/lambda/app.js
```diff
@@ -0,0 +1,49 @@
...
+
+const deps = {
+  fetch
+};
+
+const app = express();
+
+const FRONTEND = path.dirname(require.resolve('typeform-frontend'));
+
+app.use(cors());
```
jelz: CORS shouldn't be needed as we serve both frontend and the backend from the same domain.

apps/typeform/lambda/fetch-access-token.js
```diff
@@ -0,0 +1,26 @@
...
+
+  const response = await fetch(ENDPOINT, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded'
+    },
+    body
+  });
+
+  if (response.status !== 200) {
```
jelz: `if(!response.ok)` (will work with 201 and any other positive HTTP status)

apps/typeform/lambda/fetch-access-token.js
```diff
@@ -0,0 +1,26 @@
...
+    console.error('Typeform token exchange failed, got response:', response.status);
+    throw new Error('Typeform token exchange failed');
+  }
+
+  const result = await response.json();
```
jelz: Just return `response.json()`

apps/typeform/frontend/src/Auth/TypeformOAuth.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    return () => window.removeEventListener('message', handleTokenEvent);
+  }, []);
+
+  const executeOauth = () => {
+    const url = `https://api.typeform.com/oauth/authorize?&client_id=${
+      process.env.CLIENT_ID
+    }&redirect_uri=${encodeURIComponent(
+      process.env.OAUTH_REDIRECT_URI as string
+    )}&scope=forms:read+workspaces:read&state=${encodeURIComponent(window.location.href)}`;
```
jelz: State is not a URL, it is a random (every time) XSRF token

apps/typeform/frontend/src/Field/TypeFormField.tsx
jelz: Eliminate `process.env.LAMBDA_ENDPOINT`. It's not needed. Frontend and backend are served from the same domain, you can just use absolute URLs (`/forms/workspaceId`)

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
FLoppix: You are right! Forgot to change that

apps/typeform/lambda/app.js
FLoppix: Good catch! Was using CORS for development

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
FLoppix: Yes you can not have 0 workspaces in Typeform. You will always start with 1 workspace that you can not delete.

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
FLoppix: good catch. Was a left over from the previous implementation.

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
FLoppix: We can add that for the Typeform field but workspaces can not be empty since Typeform enforces you to have at least one workspace

apps/typeform/frontend/src/processTokenCallback.ts
FLoppix: Changed it to use the default Typeform token expiration and refactor it to a const

apps/typeform/frontend/src/utils.ts
FLoppix: removed the access token check since it is not an installation parameter anymore

apps/typeform/frontend/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "prop-types": "^15.7.2",
+    "react": "^16.9.0",
+    "react-dom": "^16.9.0",
+    "whatwg-fetch": "^3.0.0"
+  },
+  "scripts": {
+    "start": "contentful-extension-scripts start --serve-only",
+    "build": "contentful-extension-scripts build --no-inline",
+    "test": "contentful-extension-scripts test --env=jsdom",
+    "test:watch": "contentful-extension-scripts test --watch --env=jsdom"
```
davidfateh: This will help keep consistency and let ci run the test.
```suggestion
    "test:ci": "TZ=UTC contentful-extension-scripts test --env=jsdom",
    "test": "TZ=UTC contentful-extension-scripts test --watch --env=jsdom"
```

apps/typeform/frontend/package.json
davidfateh: Since we got rid of coverage, we can remove this.

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,239 @@
...
+      const response = await fetch(`/workspaces`, {
+        headers: {
+          Authorization: `Bearer ${this.state.accessToken}`
+        }
+      });
+      const result: WorkspacesResponse = await response.json();
+      this.setState({ workspaces: this.normalizeWorkspaceResponse(result) });
+    } catch (_error) {
+      this.props.sdk.notifier.error(
+        'There was a problem while fetching your Typeform workspaces. Please try again.'
```
davidfateh: Minor grammar correction:
```suggestion
        'There was a problem fetching your Typeform workspaces. Please try again.'
```

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -0,0 +1,239 @@
...
+      workspaces
+    } = this.state;
+
+    return (
+      <div>
+        <div className={styles.background('#262627')} />
+        <div className={styles.body}>
+          <div>
+            <div>
+              {}
```
davidfateh:
```suggestion

```

apps/typeform/frontend/src/AppConfig/AppConfig.tsx
davidfateh: Seems unnecessary then.

apps/typeform/frontend/src/Auth/TypeformOAuth.tsx
```diff
@@ -0,0 +1,63 @@
...
+    oauthWindow = window.open(url, 'Typeform Contentful', 'left=150,top=10,width=800,height=900');
+
+    window.addEventListener('message', handleTokenEvent);
+  };
+
+  const handleTokenEvent = ({ data }: any) => {
+    const { token, error } = data;
+
+    if (error) {
+      console.error('There was an error authenticating. Please try again.');
```
davidfateh: Should we do something more than log here?

apps/typeform/frontend/src/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  return !expires || parseInt(expires, 10) - Date.now() <= _10Minutes;
+}
+
+export function resetLocalStorage() {
+  window.localStorage.removeItem('token');
+  window.localStorage.removeItem('expireTime');
+}
+
+export function getToken() {
```
davidfateh: I would suggest ensuring that every `window.localStorage.getItem('token') || ''` line is replaced by this function. I would also do the same for expire time.

1. You handle the default case here and don't have to repeat everywhere.
2. If we ever change out the storage method, we wouldn't need to touch the rest of the code since it is abstracted away by this function.

apps/typeform/lambda/app.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
```
davidfateh: This can be removed.
```suggestion

```

apps/typeform/lambda/app.js
```diff
@@ -0,0 +1,60 @@
...
+
+const path = require('path');
+const express = require('express');
+const fetch = require('node-fetch');
+const handleForms = require('./forms-handler');
+const handleWorkspaces = require('./workspaces-handler');
+const fetchAccessToken = require('./fetch-access-token');
```
davidfateh: Nit: Why not use `import` syntax?

apps/typeform/lambda/app.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const app = express();
+
+const FRONTEND = path.dirname(require.resolve('typeform-frontend'));
+
+app.use('/forms', async (req, res) => {
+  const { authorization } = req.headers;
+  if (!authorization) {
+    res.status(401).send('Unauthorized');
```
davidfateh: Optional: Doing this will automatically do what you have above.
```suggestion
    res.sendStatus(401);
```

apps/typeform/lambda/app.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  const [, token] = authorization.split(' ');
+  const { status, body } = await handleForms(req.method, req.path, token, deps);
+  res.status(status).send(body);
+});
+
+app.use('/workspaces', async (req, res) => {
+  const { authorization } = req.headers;
+  if (!authorization) {
+    res.status(401).send('Unauthorized');
```
davidfateh: Same optional comment as above.

apps/typeform/lambda/app.js
```diff
@@ -0,0 +1,60 @@
...
+  const { status, body } = await handleWorkspaces(req.method, req.path, token, deps);
+  res.status(status).send(body);
+});
+
+app.use('/callback', async (req, res) => {
+  const { code } = req.query;
+  const { host } = req.headers;
+
+  if (!code) {
+    res.status(404).send('No Code was provided');
```
davidfateh: Status `400` would be more semantic as it indicates that the resource exists but the request was malformed.

```suggestion
    res.status(400).send('No code was provided');
```

apps/typeform/lambda/app.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  const origin = `${req.protocol}://${host}`;
+
+  const { access_token, expires_in } = await fetchAccessToken(code, origin, deps);
+
+  res.set({
+    Location: `${origin}/frontend/?token=${access_token}&expiresIn=${expires_in}`
+  });
+  res.sendStatus(302);
```
davidfateh: Optional: You can slim this down using the `redirect` method which will automatically send a 302 with a Location header.

```suggestion
  res.redirect(`${origin}/frontend/?token=${access_token}&expiresIn=${expires_in}`);
```

apps/typeform/lambda/app.js
```diff
@@ -0,0 +1,60 @@
...
+
+app.use('/callback', async (req, res) => {
+  const { code } = req.query;
+  const { host } = req.headers;
+
+  if (!code) {
+    res.status(404).send('No Code was provided');
+  }
+
+  const origin = `${req.protocol}://${host}`;
```
davidfateh: I think we should hardcode `https` to ensure security.

apps/typeform/lambda/fetch-access-token.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const { BASE_URL } = require('./constants');
+const ENDPOINT = `${BASE_URL}/oauth/token`;
+
+module.exports = async (code, origin, { fetch }) => {
+  const body =
+    `grant_type=authorization_code&code=${code}&client_id=${process.env.CLIENT_ID}&` +
+    `client_secret=${process.env.CLIENT_SECRET}&redirect_uri=${origin}/callback`;
```
davidfateh:
```suggestion
    `client_secret=${process.env.CLIENT_SECRET}&redirect_uri=${encodeURIComponent(origin)}/callback`;
```

apps/typeform/lambda/forms-handler.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  try {
+    return {
+      status: 200,
+      body: { forms: await fetchForms(method, path, token, { fetch }) }
+    };
+  } catch (err) {
+    return {
+      status: 400,
```
davidfateh: I'm wondering if we should actually pass through the status that was given to us by the fetch on line 16. Otherwise, all requests that fail will just show up as a 400 in our logs which won't be helpful for us.

apps/typeform/lambda/workspaces-handler.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  try {
+    return {
+      status: 200,
+      body: { workspaces: await fetchWorkspaces(method, path, token, { fetch }) }
+    };
+  } catch (err) {
+    return {
+      status: 400,
```
davidfateh: Same comment as above.

apps/typeform/lambda/forms-handler.js
FLoppix: Ya we should probably do that

apps/typeform/lambda/app.js
FLoppix: Okay but then it makes local development a pain when you need to change it to `http`. I mean we are serving the apps from `https` anyway on `production`

apps/typeform/lambda/app.js
FLoppix: I was following the lambda structure from `ai-image-tagging` which is just plain node and no babelconfig or any transpilation step. Also not happy with that configuration there

davidfateh: Typescript should transpile this as well

jelz: I just want to remind you that the only env we're targeting is modern node, either local or on AWS. In general you don't need to transpile to get all the new syntax/methods in node since node 8

In cases we don't use TypeScript (and IMO for projects of 200 LOC benefits of types are negligible) I'd recommend not to introduce extra build steps just to be able to `import` (and again benefits of imports in node-world are negligible, you don't need to bundle/tree shake server code).

BTW it's a little bit funny to see how people treat transpilation as a default and miss it when it's not in place ;)

apps/typeform/lambda/serverless.yml
```diff
@@ -0,0 +1,46 @@
+service:
+  name: sls-apps-typeform
+
+frameworkVersion: '>=1.0.0 <2.0.0'
+
+plugins:
+  - serverless-domain-manager
+  - serverless-express
```
jelz: I don't think it's needed. As a matter of fact you don't have the plugin installed. https://github.com/awslabs/aws-serverless-express we're using doesn't require any plugins

apps/typeform/lambda/app.js
davidfateh: That's true, I didn't realize this wasn't in typescript 😆. However, that being said I disagree that typescript is only beneficial on large projects. Types help describe many aspects of the code and can be useful.

As for import syntax, it isn't available to the node envs we use since it would require node >= 13 without any explicit flag setting like in node 12.

FLoppix:
@jelz regarding the node / transpilation discussion. My concern was more that we should we have a default way / structure of how to implement lambda functions. I agree that we should keep things simple when it is a small project and do not add any unnecessary build or transpilation steps. But seeing 3 or 4 different ways of how to implement lambda functions confuses new people on how to implement things :D Things will hopefully change when ES Modules are the default

apps/typeform/lambda/app.js
jelz: Even in node 13 ESM support is 1 - experimental hence I wouldn't use it.

When it comes to TypeScript, nothing against using it in projects of any size but for extra-small ones (especially these fitting two screens with trivial logic) in many cases I experienced that setup and maintenance (more tooling/deps to keep up-to-date) time is greater than time saved by having types. But ultimately it's your call ;)

FLoppix: For me personally it was just confusing what to use now for the lambda functions. Which structure / project setup is the default or common way on how you implement lambda functions. But my opinion is always to keep it as simple as possible

apps/typeform/lambda/app.js
FLoppix: Will keep it as it is since we are using plain JS / Node. Also the linter will cry

davidfateh:
> @jelz regarding the node / transpilation discussion. My concern was more that we should we have a default way / structure of how to implement lambda functions. I agree that we should keep things simple when it is a small project and do not add any unnecessary build or transpilation steps. But seeing 3 or 4 different ways of how to implement lambda functions confuses new people on how to implement things :D Things will hopefully change when ES Modules are the default

Yes, I agree it's confusing. We had started to talk about trying to unify our lambda projects. The first one was Jira, which as you can see should probably be slimmed down a bit to use Express like we do with the others.

PR #68, opened 2020-03-26
add start folder for media library dialog

dufkaf:
Add parameter for starting folder to show() method as per https://cloudinary.com/documentation/media_library_widget#3_set_the_configuration_options
This is global setting per space.

We have images for content in specific location in the tree so allow to start dialog in that location.
In future It would be also nice to have such setting per content type. Is there instance parameter for Contentful Apps like documented for UI extensions?


dufkaf:
made separate issue #70 for instance parameters so ignore the question here

apps/cloudinary/src/index.js
```diff
@@ -77,7 +77,7 @@ function renderDialog(sdk) {
     insertHandler: data => sdk.close(data)
   });
 
-  instance.show();
+  instance.show({folder: {path: config.startFolder}});
```
jelz, line 80: Older installations will not have `config.startFolder` and as a result we'll pass `{ folder: { path: undefined } }` to Cloudinary instance. I'm not sure how Cloudinary widget will handle it: to be on the safe side I'd check if `config.startFolder` holds a valid value and only then pass it.

apps/cloudinary/src/index.js
```diff
@@ -153,6 +153,14 @@ setup({
       "type": "Number",
       "required": false,
       "default": 10
+    },
+    {
+      "id": "startFolder",
+      "name": "Starting folder",
+      "description": "Path to starting folder for the media library dialog",
```
davidfateh, line 160: We should specify if the `path` needs to start with a `/`.

apps/cloudinary/src/index.js
davidfateh: Agreed. We should also potentially add logic to check if the path seems correct. i.e. does it start with a `/`?

davidfateh:
Hey @dufkaf thanks for contributing to the apps repo! I think this change makes sense and after review it looks like we should simply add some safety around the changed code as mentioned in the comments.

apps/cloudinary/src/index.js
dufkaf: hi, thanks for feedback, good point about missing value when upgrading app, it currently works for me as is with not setting the value in configuration at all but still check might be more prudent. As the default is set to "" in the parameter description  maybe it won't be evr set to "undefined"  but empty string instead? need to check

As for validating / - I'd leave it to user or to cloudinary. It works as it is and I don't like second guessing here what proper value should be. In fact I just checked now our installation and I have "uatmedia/-/media/Images" with no starting /  there and it goes into that folder just fine. Cloudinary docs linked above do not describe proper format. Also I can imagine empty string could go to previously used folder so I'd leave it to cloudinary to handle it in best way and avoid validation here.

BTW I'd be grateful if you'd comment the #70 too as in current state it is still not very usable for us. We'd like to have start folder and max number of linked assets set per content type (e.g. some content needs only one asset, some more).

apps/cloudinary/src/index.js
dufkaf: The [docs](https://cloudinary.com/documentation/media_library_widget#3_set_the_configuration_options) don't say and it works without.

dufkaf:
Ok, please let me check (hopefully today) if default value of "" already does not handle the undefined case but I guess I'll add the check anyway. As for path validation starting with '/ ' I'd avoid it, it works with paths without it and we are just second guessing here without really knowing. And user can fix it if what he sets there does not work for them.

apps/cloudinary/src/index.js
dufkaf: please check the fix

apps/cloudinary/src/index.js
```diff
@@ -76,8 +76,13 @@ function renderDialog(sdk) {
   const instance = cloudinary.createMediaLibrary(options, {
     insertHandler: data => sdk.close(data)
   });
+  const show_options={};
 
-  instance.show();
+  if (typeof(config.startFolder) !== 'undefined' && config.startFolder != "") {
```
davidfateh, line 81: I believe we should harden this check to ensure that a string is being passed.
```suggestion
  if (typeof config.startFolder === 'string' && config.startFolder.length) {
```

This will also rule out any null/undefined values as well.

apps/cloudinary/src/index.js
```diff
@@ -76,8 +76,13 @@ function renderDialog(sdk) {
   const instance = cloudinary.createMediaLibrary(options, {
     insertHandler: data => sdk.close(data)
   });
+  const show_options={};
```
davidfateh, line 79: Minor nit pick but let's keep style inline with the rest of the code.

```suggestion
  const showOptions = {};
```

dufkaf:
>  I have a minor suggestion about how to harden your check to ensure that only string values get added in. Other than that I think it looks good and we can move forward.

Thank you, better this way indeed.

apps/cloudinary/src/index.js
```diff
@@ -76,8 +76,13 @@ function renderDialog(sdk) {
   const instance = cloudinary.createMediaLibrary(options, {
     insertHandler: data => sdk.close(data)
   });
+  const showOptions = {};
 
-  instance.show();
+  if (typeof config.startFolder === 'string' && config.startFolder.length) {
+    show_options.folder = {path: config.startFolder};
```
davidfateh, line 82:
```suggestion
    showOptions.folder = {path: config.startFolder};
```

apps/cloudinary/src/index.js
```diff
@@ -76,8 +76,13 @@ function renderDialog(sdk) {
...
+  }
+  
+  instance.show(show_options);
```
davidfateh, line 85:
```suggestion
  instance.show(showOptions);
```

apps/cloudinary/src/index.js
dufkaf: yes noticed that too and already fixed

davidfateh:
@dufkaf Thank you again for your contribution. I've merged your feature and it will run through CI and deploy shortly. I haven't forgotten about #70; however I will have an answer later on.

dufkaf:
>  I haven't forgotten about #70; however I will have an answer later on.
@davidfateh  thank you for the review, and no problem with #70, things take time :-)

PR #71, opened 2020-04-01
fix(typeform): pass typeform client id to the frontend

FLoppix:
## Target

Since the typeform `client id` is publicly anyway in the URL it can be passed as a constant to the OAuth flow.


FLoppix:
CI is injecting only the `CLIENT_ID` and `CLIENT_SECRET` to the lambda function and not the build process of the frontend. We can change the build process to run with all the environments vars or we just pass it to the frontend.

PR #72, opened 2020-04-03
fix(gatsby): display warning when there are no content types

anho:


apps/gatsby/src/AppConfig/ContentTypesPanel.js
```diff
@@ -0,0 +1,69 @@
...
+const NoContentTypes = () => (
+  <Note noteType="warning">
+    There are no content types available in this environment. <TextLink
+    href="https://www.contentful.com/faq/basics/#how-do-i-add-a-new-content-type" target="_blank"
+    rel="noopener noreferrer">How do I add a new content type?</TextLink>
+  </Note>
+);
+
+const ContentTypesPanel = ({ contentTypes, enabledContentTypes, onContentTypeToggle }) => {
+  const ContentTypesList = () => {
```
davidfateh, line 30: Seems a bit strange that this component should live inside but the other ones don't. I get that you've done it to capture the main arguments but we could also easily make it its own standalone react component.

PR #80, opened 2020-04-20
Add Mux video uploader app 

dylanjha:
This was copied over from the [extensions repo](https://github.com/contentful/extensions/tree/master/marketplace/mux-video-uploader) and updated according to the [migration guide](https://www.contentful.com/developers/docs/extensibility/app-framework/migrating-extension-to-app/)

All the existing UI pieces from the extension stayed the same, all I added was the `/app-config` configuration page

mux-app-install_2020-04-20_16-15-41


apps/mux/src/index.tsx
```diff
@@ -0,0 +1,404 @@
...
+    return (
+      <input type="file" className="cf-file-input" onChange={this.onChange} />
+    );
+  };
+}
+
+init(sdk => {
+  const Component = sdk.location.is(locations.LOCATION_APP_CONFIG) ? Config : App;
+  render(
+    <Component sdk={sdk as FieldExtensionSDK} />,
```
davidfateh, line 396: Since you have ternary logic above, the type of the SDK won't be `FieldExtensionSDK` on the CONFIG location.

apps/mux/src/config.tsx
```diff
@@ -0,0 +1,109 @@
...
+    return {
+      // Parameters to be persisted as the app configuration.
+      parameters: this.state.parameters,
+      // Transformation of an environment performed in the
+      // installation process.
+      targetState: {
+        EditorInterface: contentTypeIds.reduce((acc, id) => {
+          // Insert the app as the first item in sidebars
+          // of all content types.
+          return { ...acc, [id]: { sidebar: { position: 0 } } }
```
davidfateh, line 102: This logic needs to change. Currently, you are adding Mux to the sidebar of every content type in the space. Mux, however, doesn't use the sidebar and will create a blank space for every single content type at the top.

Instead, you'll need to target the specific field inside of each content type where mux should show up as an `input` element. We do this in a few of our other apps.

For example, our other DAM apps share a library which helps contain this logic.

Where we set target state for DAM apps:
https://github.com/contentful/apps/blob/master/lib/shared-dam-app/src/AppConfig/AppConfig.tsx#L146

The utility functionality to help find target fields to attach to per content type:
https://github.com/contentful/apps/blob/master/lib/shared-dam-app/src/AppConfig/fields.ts

apps/mux/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "commonjs",
+    "strict": true,
+    "jsx": "react",
+    "rootDir": "./src",
+    "experimentalDecorators": true,
```
davidfateh, line 8: I didn't notice the use of this feature anywhere and I believe it can be removed.
```suggestion

```

apps/mux/src/config.tsx
```diff
@@ -0,0 +1,109 @@
...
+  async onConfigure () {
+    // Get IDs of all content types in an environment.
+    const { items: contentTypes } = await this.props.sdk.space.getContentTypes();
+    const contentTypeIds = contentTypes.map(ct => ct.sys.id)
+
+    // Return value of `onConfigure` is used to install
+    // or update the configuration.
+    return {
+      // Parameters to be persisted as the app configuration.
+      parameters: this.state.parameters,
```
davidfateh, line 95: It looks like you aren't actually checking to see if an ID or token were provided. The app cannot function without either.

I'd suggest doing some data validation here to check that the two values you are expecting actually exist before allowing the user to save this configuration.

When you `return false` in this function, we cancel the installation/save process from happening in the webapp. For example, some code might look like:
```js
if (!this.state.parameters.muxAccessTokenSecret.trim()) {
  this.props.sdk.notify.error('You must provide an access token!');
  return false;
}

File: apps/mux/src/config.tsx [hunk 5]
```diff
@@ -0,0 +1,109 @@
...
+            <Note noteType="primary" title="About the app">
+              A Contentful UI app that makes it simple to add beautiful streaming via <a href='https://mux.com'>Mux</a> to your Contentful project. Install the extension, add the component to your content model, and you're good to go! 🙌
+            </Note>
+            <TextField
+              required
+              name="mux-access-token"
+              id="mux-access-token"
+              labelText="Mux access token"
+              value={muxAccessTokenId || ''}
+              onChange={e => this.setState({ parameters: { muxAccessTokenId: e.target.value, muxAccessTokenSecret } })}
```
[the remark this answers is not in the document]
dylanjha: We would rather not - is there a way with Contentful apps to make these requests server side? Please point me to the correct documentation!

dylanjha:
@davidfateh thank you! I will get this done by end of week and also update the styles of the config screen -- is this a good example?

dropbox-example_2020-05-05_05-42-04

davidfateh:
> @davidfateh thank you! I will get this done by end of week and also update the styles of the config screen -- is this a good example?
> 
> dropbox-example_2020-05-05_05-42-04

Yep, that is perfect. All of our apps use the same kind of markup to implement the configuration screen so you should be able to copy/paste most of your way there.

apps/mux/src/config.tsx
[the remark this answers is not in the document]
dylanjha: Making this a `password` field for now. When we can have our own backend, I will revisit this.

apps/mux/src/index.tsx
dylanjha: I used `BaseExtensionSDK` as the type that gets passed to `Config`  - seemed like the most appropriate one to me, let me know if there's a better option please!

apps/mux/src/config.tsx
dylanjha: Done

mux-contentful

dylanjha:
config-screen_2020-05-06_09-50-38

Updated the styles of the config screen, I used Gatsby as an example

apps/mux/src/config.tsx
dylanjha: > You'll need to target the specific field inside of each content type where mux should show up as an input element.

This isn't exactly clear to me - would you mind helping me understand please? Currently Mux appears as a "JSON object" field

contentful-video-upload

dylanjha:
@davidfateh thanks so much for your attention here. These are the outstanding items:

- [ ] [question above](https://github.com/contentful/apps/pull/80#discussion_r420944697) I have about "targeting the specific field inside of each content type"
- [ ] linting - running `npm run lint` from the top level directory is being happy - I'll have to debug this more later
- [ ] adding an app icon - I didn't see where to do this

One last thing, not sure if anyone has ever told you this. You look JUST like Matt Damon. Crazy!

apps/mux/src/config.tsx
davidfateh: Looking at the code it appears that when you install the mux app, you'll still need the user to go into their content model and find the JSON fields where they want the mux UI to show up.  

If that is the case, we can create UI inside of the app configuration screen where the user can select these fields before installing the app. This means that a user won't have to go into the content model and manually apply mux to the JSON fields.

If you check out the Bynder app for example, we show the user a list of content types and fields within those content types where they can choose to install the Bynder app to. Our logic checks to ensure we only show JSON fields.

Screen Shot 2020-05-07 at 12 53 41

The UI above you will have to recreate yourself (you can check out the Bynder code) however, to save the user selection, we can apply this logic in the return value of your `onConfigure` handler that you pass in to the property `EditorInterface`.

```js

return {
 parameters: this.state.parameters,
 EditorInterface: {
    post: {controls: [{fieldId: 'blogImage'}], 
  }
}
```

The `EditorInterface` object's keys are the IDs of content types we want to configure for the app. In the code example above, `post` is the ID of a content type that has a JSON field with an ID of `blogImage`

Mux can use this same approach to enable users to get the mux app working for them right away after the installation without additional steps. We have a set of utility functions that we use for all our DAM apps that can help you shape these responses for the `EditorInterface` value.
 
They can be found here:

https://github.com/contentful/apps/blob/master/lib/shared-dam-app/src/AppConfig/fields.ts

I know this is a lot to cover and the docs are a bit light in this area. We are working to improve them but in the mean time please follow up with any questions and I can help you get on your way.

davidfateh:
Hey @dylanjha 

Awesome to see the new screen! The changes look good and I've provided hopefully an in-depth answer for your first question. For the others:

* For linting, I would suggest configuring it in the mux directory. Our CI will go into each app dir and run the lint there. Feel free to just copy a config from another app.
* When your app gets listed on our public marketplace, we add the information and icon then. It is totally possible for us to take the SVG you've already provided here and reuse it for our listing entry. That is something we will do on our end.

Regarding public app listings: We will have to follow up with you outside of GitHub about a few other matters such as docs, marketplace description, external links, legal links, etc. I think we already have each other's contact info so we can start that discussion soon.

Let me know if you have any questions.

apps/mux/src/config.tsx
dylanjha: @davidfateh Thanks! This all makes a lot of sense. I have this almost working - I created the "Assign to Fields" UI. I think I did this correctly - it only shows *content types* that have at least 1 JSON field, and under each content type it only shows the fields that are JSON fields.

assign-to-fields_2020-05-07_09-58-49

I'm a little stuck now using the helpers to manage the local state and construct the `EditorInterface` value.

Bynder looks like a perfect example, but checking [the code in this repo](https://github.com/contentful/apps/blob/master/apps/bynder/src/index.js) - All I see is this 138 line index file -- which looks like the code for the UI. I can't find the config code, does it live somewhere else?

dylanjha: Okay... I got closer here... but saving the `EditorInterface` value is not quite working. Here's an example where I save this

```
return {
  parameters,
  EditorInterface: {
    blogPost: {
      controls: [{ fieldId: "videoUpload" }]
    }
  }
}
```

I must be missing something really obvious here. `blogPost` is the content type ID and `videoUpload` is the field ID. This is a screenshot where I am logging the value that gets set for the `EditorInterface` key:

editorInterface_2020-05-11_15-28-01

----

I saw in another example `EditorInterface` was [nested under the key `targetState`](https://github.com/contentful/apps/blob/master/apps/gatsby/src/AppConfig/AppConfig.js#L110) so I tried that too, but it didn't work either.

-----

Two outstanding things to get this wrapped up:

* What am I doing wrong above?
* I had to put in a lot of boilerplate code converting to and from the editorInterface/contentType data structures. How can I use your helper methods? What's confusing me is:
   1. [shared-dam-app](https://github.com/contentful/apps/blob/master/apps/bynder/package.json#L15) I see this package is using it, but it [doesn't exist on NPM](https://www.npmjs.com/package/shared-dam-app). Is it private or something?
  2. I see other apps using the `setup` method from `shared-dam-app`. Should I be using that?! It looks way easier for these configuration steps...

davidfateh: Hey so I checked out your code and you are right: I missed `targetState` as the top level property on the return value. I also think the reason it wasn't working for you was due to older versions of our contentful packages.

In this PR I was able to get target state working and on refresh I can see my saved changes show up.
https://github.com/muxinc/apps/pull/1

As for your second question about the shared-dam-app packages, we use Lerna in this repo to manage local packages. So while they are not distributed via npm, Lerna will see that we have local shared packages and install them. The reason that Bynder, for example, only have UI in its config file is because of your observation about the shared libs.

I'm not sure if Mux will benefit from using it as your UX is different. The DAM apps pop a dialog modal which then iframes into their asset pickers. Do you have such an interface?

dylanjha: @davidfateh fantastic! Good find!

okay, that all makes sense.

apps/mux/extension.json
```diff
@@ -0,0 +1,24 @@
+{
```
davidfateh, line 1: This file can be completely removed. We don't use it for apps :)

apps/mux/src/config.tsx
```diff
@@ -0,0 +1,295 @@
...
+    };
+
+    // `sdk.app` exposes all app-related methods.
+    this.app = this.props.sdk.app;
+    this.space = this.props.sdk.space;
+
+    // `onConfigure` allows to configure a callback to be
+    // invoked when a user attempts to install the app or update
+    // its configuration.
+    this.app.onConfigure(() => this.onConfigure());
```
davidfateh, line 72: Could you move this to `componentDidMount`? While testing, I saw some flakiness with this call and I think it may have to do with it being here in the constructor.

apps/mux/src/config.tsx
```diff
@@ -0,0 +1,295 @@
...
+          <Typography>
+            <Heading>About Mux</Heading>
+            <Paragraph>
+              This app connects to Mux and allows you to upload videos to your
+              content in Contentful. After entering your API Credentials then
+              choose which JSON fields in your content model you would like to
+              configure for Mux Video. For those configured fields you'll get a
+              video uploader in the Contentful UI. Your videos will be
+              transcoded, stored and delivered by{' '}
+              <a href="https://mux.com">Mux</a>.
```
davidfateh, line 153: Let's use Forma-36's `<TextLink>` as well as get this to open in a new page with some security enabled.
```suggestion
              <TextLink href="https://mux.com" rel="noopener noreferrer" target="_blank">Mux</TextLink>.
```

apps/mux/src/config.tsx
```diff
@@ -0,0 +1,295 @@
...
+          </Typography>
+          <hr className="config-splitter" />
+          <Typography>
+            <Form spacing="default">
+              <Heading>API credentials</Heading>
+              <Paragraph>
+                These can be obtained by clicking 'Generate new token' in the{' '}
+                <a href="https://dashboard.mux.com/settings/access-tokens">
+                  settings on your dashboard
+                </a>
```
davidfateh, line 164:
```suggestion
                <TextLink href="https://dashboard.mux.com/settings/access-tokens" rel="noopener noreferrer" target="_blank">
                  settings on your dashboard
                </TextLink>
```

apps/mux/src/config.tsx
```diff
@@ -0,0 +1,292 @@
...
+  Typography,
+  FieldGroup,
+  TextLink,
+} from '@contentful/forma-36-react-components';
+import {
+  AppExtensionSDK,
+  AppConfigAPI,
+  SpaceAPI,
+  BaseExtensionSDK,
+} from 'contentful-ui-extensions-sdk';
```
davidfateh, line 24:
```suggestion
} from 'contentful-ui-extensions-sdk/lib/types';
```

apps/mux/src/config.tsx
```diff
@@ -0,0 +1,292 @@
...
+import {
+  editorInterfacesToSelectedFields,
+  getCompatibleFields,
+  selectedFieldsToTargetState,
+  EditorInterface,
+  ContentType,
+  CompatibleFields,
+  SelectedFields,
+} from 'shared-dam-app';
```
davidfateh, line 33:
```suggestion
} from 'shared-dam-app/src/utils';
```

apps/mux/MARKETPLACE.md
```diff
@@ -0,0 +1,39 @@
+A Contentful UI app that makes it simple to add beautiful streaming via [Mux](https://https://mux.com) to your [Contentful](https://contentful.com) project. Just install the extension, add the component to your content model, and you're good to go! 🙌🏾
```
jelz, line 1: This file should be removed, it refers to the legacy extension.

Let's collaborate on communication (docs, marketplace entry) outside of this PR.

apps/mux/.gitignore
```diff
@@ -0,0 +1 @@
+.now
```
jelz, line 1: As we use our deployment pipeline and hosting it would be nice to drop all mentions of now.

apps/mux/README.md
```diff
@@ -0,0 +1,43 @@
+![Contentful + Mux](https://banner.mux.dev/Contentful.svg)
+
+# Contentful Video Streaming Plugin
+
+A Contentful UI extension that makes it simple to add beautiful streaming via [Mux](https://https://mux.com) to your [Contentful](https://contentful.com) project. Just install the app, add the component to your content model, and you're good to go! 🙌🏾
```
jelz, line 5: Either update this file to talk about the App or remove the file

apps/mux/README_INTERNAL.md
```diff
@@ -0,0 +1,22 @@
+
+# Development
```
jelz, line 2: This could be README since instructions/marketing text will live outside of Github.

apps/mux/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "shared-dam-app": "^1.0.0"
+  },
+  "scripts": {
+    "start": "contentful-extension-scripts start --serve-only",
+    "build": "contentful-extension-scripts build",
+    "tsc": "tsc -p ./ --noEmit",
+    "test": "contentful-extension-scripts test --env=jsdom --watch",
+    "test:ci": "contentful-extension-scripts test --env=jsdom",
+    "test:dev": "TZ=UTC contentful-extension-scripts test --env=jsdom --watch",
+    "test:coverage": "contentful-extension-scripts test --env=jsdom --coverage",
```
jelz, line 54: drop, there's no coverage setup

apps/mux/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "contentful": "contentful",
```
jelz, line 55: drop, not needed

apps/mux/src/config.tsx
```diff
@@ -0,0 +1,292 @@
...
+    if (
+      !(
+        parameters.muxAccessTokenSecret &&
+        parameters.muxAccessTokenSecret.trim()
+      )
+    ) {
+      valid = false;
+    }
+
+    if (!valid) {
```
jelz, line 273: I'd simplify:

```js
const isParamValid = key => (parameters[key] || '').trim().length > 0
if (!isParamValid('muxAccessTokenId') || !isParamValid('muxAccessTokenSecret')) {
  notifier.error('boom!')
  return false;
}
```

apps/mux/src/mux-logo.svg
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jelz, line 1: You could minify it with https://jakearchibald.github.io/svgomg/

apps/mux/src/player.tsx
```diff
@@ -0,0 +1,80 @@
...
+    this.hls.on(Hls.Events.MEDIA_ATTACHED, () => this.onReady());
+  }
+
+  onReady = () => {
+    this.props.onReady();
+  };
+
+  componentDidMount() {
+    if (!this.playerRef.current) {
+      throw Error('No reference to an existing video element found.');
```
jelz, line 36: `throw new Error('...')`

apps/mux/src/config.tsx
dylanjha: I committed this change, and it looks like this is causing CI to fail

```
FAIL test/index.test.tsx
  ● Test suite failed to run

    src/config.tsx:24:8 - error TS2307: Cannot find module 'contentful-ui-extensions-sdk/lib/types'.

    24 } from 'contentful-ui-extensions-sdk/lib/types';
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dylanjha:
Thanks for the suggestions @jelz -- I made all those changes!

PR #81, opened 2020-04-21
fix(typeform): add field type hint

FLoppix:
## Purpose
Adds a field type hint to see which fields are supported for the typeform app


apps/typeform/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -213,7 +213,10 @@ export class AppConfig extends React.Component<Props, State> {
                 <Heading>Assign to content types</Heading>
                 {contentTypes.length > 0 ? (
                   <>
-                    <Paragraph>Select which content types to use with Typeform App.</Paragraph>
+                    <Paragraph>
+                      This app can only be used with <strong>Short text</strong> fields. Select
+                      which content types to use with Typeform App.
```
davidfateh, line 218:
```suggestion
                      which content types to use with the Typeform App.
```

PR #83, opened 2020-04-22
feat: add page location example

fabe:


examples/page-location/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,79 @@
...
+        .then(entries => entries.total)
+        .catch(() => 0);
+      const published = await sdk.space
+        .getPublishedEntries()
+        .then(entries => entries.total)
+        .catch(() => 0);
+      const scheduled = await sdk.space
+        .getAllScheduledActions()
+        .then(entries => entries.length)
+        .catch(() => 0);
```
davidfateh, line 43: Optional: I would use a `Promise.all` that way this makes all the requests at the same time.

examples/page-location/src/views/Page.tsx
```diff
@@ -0,0 +1,103 @@
...
+    super(props);
+
+    const invocationParams = props.sdk.parameters.invocation as InvocationParams;
+
+    this.history = createMemoryHistory({
+      initialEntries: [invocationParams.path],
+    });
+
+    this.history.listen((location) => {
+      console.log(location);
```
davidfateh, line 36: Can be removed.

PR #84, opened 2020-05-06
feat: hello world example

fabe:


examples/hello-world/README.md
```diff
@@ -0,0 +1,39 @@
...
+```bash
+npm install
+npm start
+# or
+yarn
+yarn start
+```
+
+To test, you can create an app definition in your Contentful organization settings pointing to `http://localhost:1234` and registering the `app-config` location.
+
```
FLoppix, line 19: do we have a screenshot for that ? I think that could be helpful to have the basic setup displayed for an app definition with a location. Also we should inform the users about the `insecure content` issue when we try render to the app served via `http`. Maybe some users will think that is not expected.

fabe: good point, added

PR #86, opened 2020-05-13
Cloudinary support format and quality

asisayag2:
**The purpose of this PR:**
Add support in Cloudinary's format and quality capabilities. 
Cloudinary allows global selection of the _format_/_quality_ of your media, to allow full control of your optimization level, thus improving your page load time. 

To benefit from one of Cloudinary's super-powers, one can select 'auto' for both _format_ and _quality_ and let Cloudinary deliver the best format and quality level based on the user and the media content.  

**This PR includes:**
- Addition of a _list_ parameter type with a comma-separated _value_ field - to allow selection from a predefined list of values. I added this to the `AppConfig.tsx` file using the _SelectField_ component from _forma-36_
- Added 2 parameters of type list (_format_ and _quality_)
- Initialized the Cloudinary Media Library with the default transformation (based on the user selection in the _format_ and _quality_ params
- Handled the returned value from the Media Library and kept both the original file and the derived (which includes the transformation)
- Changed the thumbnail creation to include the transformation as well 
- Changed some of the descriptions


apps/cloudinary/src/index.js
```diff
@@ -134,20 +159,20 @@ setup({
   name: 'Cloudinary',
   logo,
   description:
-    'The Cloudinary app is a widget that allows editors to select media from their Cloudinary account. Select or upload a file on Cloudinary and designate the assets that you want your entry to reference.',
+    'The Cloudinary app allows editors selecting media from their Cloudinary account. Select the asset from Cloudinary that you want your entry to reference.',
```
davidfateh, line 162:
```suggestion
    'The Cloudinary app allows editors to select media from their Cloudinary account. Select the asset from Cloudinary that you want your entry to reference.',
```

apps/cloudinary/src/index.js
```diff
@@ -134,20 +159,20 @@ setup({
...
   color: '#F4B21B',
   parameterDefinitions: [
     {
       "id": "cloudName",
       "name": "Cloud name",
-      "description": "The cloud name of the account to access.",
+      "description": "The Cloudinary cloud name that the app will connect to",
```
davidfateh, line 168:
```suggestion
      "description": "The Cloudinary cloud name that the app will connect to.",
```

apps/cloudinary/src/index.js
```diff
@@ -134,20 +159,20 @@ setup({
...
       "type": "Symbol",
       "required": true
     },
     {
       "id": "apiKey",
       "name": "API key",
-      "description": "The account API key.",
+      "description": "The Cloduinary API Key as can be found in your Cloudinary console",
```
davidfateh, line 175:
```suggestion
      "description": "The Cloduinary API Key that can be found in your Cloudinary console.",
```

apps/cloudinary/src/index.js
```diff
@@ -162,10 +187,28 @@ setup({
     {
       "id": "startFolder",
       "name": "Starting folder",
-      "description": "Path to starting folder for the media library dialog",
+      "description": "A path to a folder, which the Cloudinary Media Library will automatically browse to on load",
```
davidfateh, line 190:
```suggestion
      "description": "A path to a folder which the Cloudinary Media Library will automatically browse to on load",
```

apps/cloudinary/src/index.js
```diff
@@ -162,10 +187,28 @@ setup({
...
       "type": "Symbol",
       "required": false,
       "default": ""
+    },
+    {
+      "id": "quality",
+      "name": "Media Quality",
+      "description": "The quality level of your assets. This can be a fixed number ranging from 1-100, or you can get Cloudinary to decide the most optimized level by setting to 'auto'. More available options are auto:low/auto:eco/auto:good/auto:best. If you wish to use the original level, set to 'none'",
```
davidfateh, line 198:
```suggestion
      "description": "The quality level of your assets. This can be a fixed number ranging from 1-100, or you can get Cloudinary to decide the most optimized level by setting it to 'auto'. More options are available such as: auto:low/auto:eco/auto:good/auto:best. If you wish to use the original level, set it to 'none'.",
```

apps/cloudinary/src/index.js
```diff
@@ -162,10 +187,28 @@ setup({
...
+      "type": "List",
+      "value": "auto,none,auto:low,auto:eco,auto:good,auto:best,10,20,30,40,50,60,70,80,90,100",
+      "required": true,
+      "default": "auto"
+    },
+    {
+      "id": "format",
+      "name": "Format",
+      "description": "The format of the assets. This can be set manually to a specific format - 'jpg' as an example (all supported formats can be foudn here - https://cloudinary.com/documentation/image_transformations#supported_image_formats. By setting to 'auto' Cloudinary will decide on the most optimized format for your users. If you wish to keep the original format, set to 'none'",
```
davidfateh, line 207:
```suggestion
      "description": "The format of the assets. This can be set manually to a specific format - 'jpg' as an example (all supported formats can be found here - https://cloudinary.com/documentation/image_transformations#supported_image_formats. By setting it to 'auto', Cloudinary will decide on the most optimized format for your users. If you wish to keep the original format, set it to 'none'.", 
```

asisayag2:
Thanks @davidfateh for the review. 
I accepted your suggested grammer changes.

apps/cloudinary/src/index.js
```diff
@@ -100,12 +109,28 @@ async function openDialog(sdk, currentValue, config) {
...
+  let res = pick(asset,FIELDS_TO_PERSIST);
+  // if we have a derived images, we replace the URL with the derived URL and store the origianl URL seperatly
+  if(asset.derived){
+    res = {
+      ...res,
+      original_url: res.url,
+      original_secure_url: res.secure_url,
+      url: asset.derived[0].url,
+      secure_url: asset.derived[0].secure_url,
+      raw_transformation: asset.derived[0].raw_transformation
```
davidfateh, line 128: Hey @asisayag2 just one last question. I was looking over this code one last time to ensure that this will be backwards compatible for users that already have the app installed. What exactly is the `derived` asset doing here and do you think it will somehow affect current users?

Just for context, when we ship this, all the current users will get this change. If this somehow alters the data in their entries, it may break their front-ends.

asisayag2: Hey @davidfateh , 
When a user is selecting an image from our media library widget, you get a json back with the asset's info. 
If the asset has any transformation on it (and once we add the default f_auto/q_auto, you get this by default), there's another section in the returned json, describing the transformed image (we call it derived) 

If there is a derived section, we extract both the original image's URL and the derived URL 

As I'm still using the URL field, I don't see a reason that this will break BC

PR #87, opened 2020-05-18
feat: Backend app demo

Jwhiles:
Todo:

* Check README/setup process with someone else to make sure it makes sense.
* Add more comments to the code explaining what's going on


examples/default_values_backend_app/README.md
```diff
@@ -0,0 +1,71 @@
...
+A demo App that will prefill certain values when an entry is created.
+
+The purpose of this demo is to show how backend Apps can be setup and used.
+
+
+## Prerequisites
+This demo requires node.js, yarn, and some way of making the running server
+available to contentful. For local development it is easiest to use
+[ngrok](https://ngrok.com/)
+
```
FLoppix, line 12: Maybe we could link to directly to a download page or setup instructions for `ngrok` for Mac, Linux or Windows? I bet a lot users won't have it installed right away. And maybe add some explanations of what is `ngrok` is doing?  Adding a screenshot of what to note down when running `ngrok` would also help. If I have never used ngrok I would be a little lost

fabe: Why do we need ngrok at all here?

Jwhiles: @fabe this is for running the app locally - you need some way to make your localhost available for App events. Ngrok just seemed to be the easiest way to do this to me. If you have another suggestion I'd be happy to change it. Otherwise I'll add some more details here :)

fabe: Cool, yeah including that in the readme would be great!

examples/default_values_backend_app/README.md
```diff
@@ -0,0 +1,71 @@
...
+   from the contentful web app
+
+```shell
+export CMA_TOKEN=${a valid CMA token}
+```
+
+4. create key pair
+```shell
+# Creates a local key pair
+./tools/makekeys
```
FLoppix, line 45: Run the script in the setup step maybe

examples/default_values_backend_app/README.md
FLoppix: I think we should mention that we need to use `https` for this url

examples/default_values_backend_app/README.md
```diff
@@ -0,0 +1,71 @@
...
+not use your production environment for testing this App.
+
+3. Add a valid CMA token to the environment variables, you can obtain this token
+   from the contentful web app
+
+```shell
+export CMA_TOKEN=${a valid CMA token}
+```
+
+4. create key pair
```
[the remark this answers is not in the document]
Jwhiles: Yeah I agree. I've moved it into a general setup script, because I don't think this is the place to explain it. But let's definitely add a link to documentation when we have some.

davidfateh:
Closing for now. When we plan to merge this, we can reopen.

PR #91, opened 2020-06-02
Mux asset error handling

dylanjha:
This handles assets that are in the errored state. This state most commonly happens when someone uploads a file that is not a video file.

asset-error-handling-contentful

### DIFF
--- a/apps/mux/src/config.tsx
+++ b/apps/mux/src/config.tsx
@@ -147,7 +147,14 @@ class Config extends React.Component<ConfigProps, IState> {
               configure for Mux Video. For those configured fields you'll get a
               video uploader in the Contentful UI. Your videos will be
               transcoded, stored and delivered by{' '}
-              <TextLink href="https://mux.com" rel="noopener noreferrer" target="_blank">Mux</TextLink>.
+              <TextLink
+                href="https://mux.com"
+                rel="noopener noreferrer"
+                target="_blank"
+              >
+                Mux
+              </TextLink>
+              .
             </Paragraph>
           </Typography>
           <hr className="config-splitter" />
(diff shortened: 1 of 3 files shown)


fabe:
@dylanjha I'm not super familiar with Mux, but would it make sense to use the `accept="video/*"` attribute on the `input` element to disable uploading invalid files altogether?

dylanjha:
@davidfateh good idea. Done! --  I think these CI failures are unrelated to the code I changed:

```
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! dropbox-assets@1.0.0 build: `contentful-extension-scripts build --no-inline`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the dropbox-assets@1.0.0 build script.
```

Let me know if you need anything else from me. Thanks!

davidfateh:
> @dylanjha I'm not super familiar with Mux, but would it make sense to use the `accept="video/*"` attribute on the `input` element to disable uploading invalid files altogether?

While I think that's a good idea, I think the current approach is also a good fail-safe because someone could upload a corrupted file with the correct extension and still get problems.

dylanjha:
@fabe @davidfateh yeah, that's not a bad idea...

Mux tries to be _very_ flexible when it comes to handling videos. We will ingest any kind of video, 

Trying out `accept="video/*"` and it looks like on Safari that doesn't include mp4 files (go figure). There is [a work around](https://stackoverflow.com/questions/19107685/safari-input-type-file-accept-video-ignores-mp4-files) by doing something like this: `accept="video/mp4,video/x-m4v,video/*"`.

Upside of adding something like this is that we prevent people from selecting a non-video file and running into an error. Downside is that we accidentally prevent someone from selecting a perfectly good video file.

I would hate for the latter to happen, and even though there is a workaround for that Safari behavior it worries me a little bit that browsers treat this sort of thing differently. Let me check with some folks internally and see if we have a general recommendation for `input accept` values.

fabe:
@dylanjha Yeah, makes sense. Sounds like it would have more downsides than anything else!

dylanjha:
@fabe yeah, Safari has the same problem with `webms`. We could allow specific file extensions instead of the mime type, which would be better, but I still think it's best to keep it open and lean on Mux to error the asset for invalid video files.

Thanks!

davidfateh:
Hey @dylanjha I'm still looking into this build failure. Sorry for the delay.

PR #97, opened 2020-06-25
fix: reverting shared-sku f36 change

davidfateh:
There seems to be a parsing error in the CSS of the updated Forma-36
library which is not playing nicely with Shopify.


lib/shared-dam-app/package.json
```diff
@@ -9,7 +9,7 @@
     "@babel/plugin-transform-runtime": "7.8.3",
     "@babel/preset-env": "7.8.4",
     "@babel/preset-react": "7.8.3",
-    "@contentful/contentful-extension-scripts": "^0.19.0",
+    "@contentful/contentful-extension-scripts": "^0.15.1",
```
jelz, line 12: How about reverting only forma packages?

davidfateh: Yep could do that.

PR #124, opened 2020-08-20
Ks/bf api key config

schwindtkevin:
Add an optional config field for Brandfolder API key and then use it in the iframe if provided.


davidfateh (approved):
Thanks for the info. Looks good to me.

PR #125, opened 2020-08-24
fix ternary condition

schwindtkevin:


schwindtkevin:
@davidfateh thanks so much for reviewing my last PR and merging it in. I just tried testing live and realized I had the condition backwards.

davidfateh (approved):

davidfateh:
Good catch.

PR #126, opened 2020-08-25
Mux App: new feature: signed urls support

dylanjha:
This adds a new "advanced" feature for Mux users.

After configuring API keys, users can optionally enabled support for Mux [signed playback URLs](https://docs.mux.com/docs/security-signed-urls)

Signed playback IDs work by using a URL signing key & secret pair to generate a valid jwt. The jwt is used as a `token=` parameter on the URL when requesting a URL for images, streaming or gifs.

When signed URL support is enabled in the Mux App on the config page:

1. The app will use the Mux API to create a signing key/secret pair
1. Save the signing key/secret pair in the app SDK installation parameters (same place that API keys are saved)
1. Create assets with a `playback_policy: "signed"` (instead of what it does now: `playback_policy: "public"`)
1. The Contentful UI component will generate signatures with the saved key/secret pair in order to preview the video
1. The Mux Asset JSON object saved in Contentful will have `{ signedPlaybackId: "xxxxxx" }` instead of `{ playbackId: "xxxxxx" }`

In a practical case, a developer would enable this feature in the Contentful UI and create new entries that have Mux videos. In their application, they would use the `signedPlaybackId` that Contentful sends back and on their server generate valid json web tokens to secure playback and thumbnail generation in their applications.

Screenshots:

signed-urls_2020-08-25_12-09-37

Here is a video of the config being enabled:

https://stream.new/v/YYZzIZ3vwiZULT7F02TIeZqbu00o6SNhf006x01sEgPVcp4

This also adds a `Note` to alert the user when a `signedPlaybackId` is being used. This can be really handy if the user ends up in a situation where some assets are public and others are signed. I think this will help avoid a lot of confusion and this `Note` can also link to the Contentful Mux App Guide.

note-signed-playback_2020-08-24_22-01-50

TODO:

- [x] detailed documentation in the [Contentful Mux App Guide](https://docs.mux.com/docs/headless-cms-contentful)
- [x] thorough QA: upload a video, change the config from enabled to disabled, invalidate a signing token


apps/mux/test/index.test.tsx
```diff
@@ -50,20 +50,29 @@ test('displays an upload form before the user does anything', () => {
...
-  const wrapper = mount(<App sdk={mockedSdk as any} />);
+  const wrapper = await mount(<App sdk={mockedSdk as any} />);
+  (wrapper.instance() as App).getAsset = () => Promise.resolve();
+  // await wrapper.update()
+  console.log('debug wrapper', wrapper.state())
+  expect(wrapper.state('playbackUrl')).toEqual('https://stream.mux.com/test-playbackId123.m3u8')
+  expect(wrapper.state('posterUrl')).toEqual('https://image.mux.com/test-playbackId123/thumbnail.jpg')
   expect(wrapper.find('Player')).toHaveLength(1);
 });
+*/
```
dylanjha, line 75: TODO: fix this test

dylanjha:
Updated screenshots above, I have a section here in the detailed Mux Guide

https://docs.mux.com/docs/headless-cms-contentful#advanced-signed-urls

apps/mux/src/index.tsx
```diff
@@ -383,11 +458,27 @@ export class App extends React.Component<AppProps, AppState> {
...
+        this.state.posterUrl
+      ) {
         return (
           <div>
+            {this.isUsingSigned() && (
+              <Note>
+                Note: this mux asset is using a{' '}
+                <TextLink
+                  href="https://docs.mux.com/docs/headless-cms-contentful#advanced-signed-urls"
+                  target="_blank"
```
davidfateh, line 473: Could you please add `rel="noopener noreferrer"`?

apps/mux/src/config.tsx
```diff
@@ -255,6 +347,45 @@ class Config extends React.Component<ConfigProps, IState> {
             </Form>
           </Typography>
           <hr className="config-splitter" />
+          <Form spacing="default">
+            <Heading>Advanced: Signed URLs</Heading>
+            <Note noteType="warning" title="This is an advanced feature">
+              If you want to support signed urls you must read and understand{' '}
+              <TextLink
+                href="https://docs.mux.com/docs/headless-cms-contentful#advanced-signed-urls"
+                target="_blank"
```
davidfateh, line 356: Could you please add `rel="noopener noreferrer"`?

davidfateh (changes requested):
Looks good to me. I just noticed two external links that we should change slightly as commented below. Once done, I will approve and merge it.

apps/mux/src/config.tsx
dylanjha: done!

apps/mux/src/index.tsx
dylanjha: done!

davidfateh (approved):

PR #154, opened 2020-10-20
fix: allow TIFF file previews

VaguelySerious:
This fixes the "asset not available" error when trying to preview a TIFF asset.

Edit:
To prevent future issues I also added all other image formats supported by Cloudinary to the list.
I've tested each of them in our web app and Cloudinary is indeed capable of producing preview URLs for these formats.


Jwhiles (approved):
nice!

lehnerm (approved):

VaguelySerious:
@lehnerm Can you comment on why we need the `VALID_IMAGE_FORMATS` array at all? It seems to me assets with `resource.resource_type === 'image'` are already recognized as an image by Cloudinary and we don't have to whitelist them

PR #155, opened 2020-10-21
fix: make bootstrap work

Jwhiles:
Bootstraping the project failed for me, and it seems to be because there was an error in some of the package-lock files. I think this should fix it. But would be grateful if someone could verify the install and bootstrap process from a clean repo on their machine


andipaetzold:
@Jwhiles This does actually not work for me. I ran the following:

- `git clean -xdf`
- `npm ci`
- `npm run bootstrap`

andipaetzold (approved):
Okay it now worked after I removed and reinstalled the command line tools...

`master` errors with problems in `typeform-lambda`, so I guess that rebuild of the `package-lock.json` fixed it.

PR #158, opened 2020-10-27
chore: add .nvmrc

Jwhiles:
I noticed we didn't have one. Matched v10 as that's what we specify in the circleci config.


andipaetzold (approved):
I am actually no sure which version to pick in `.nvmrc`. I might have gone for 10.13.0 as it is the first LTS release.
Also it might make sense to run `nvm install` during CI to always use the same version in development and on CI.

shikaan (approved):
We should have a followup to make sure all the packages are latest LTS compatible and then upgrade this and CircleCI config

PR #173, opened 2021-01-06
Upgrade Node to latest LTS

kdamball:
- [x] Upgrade serverless & CCI
- [x] Optional: upgrade to latest node LTS - Can't do this now as v14 which is the latest LTS isn't supported by aws, yet
- [X] Optional: Upgrade dependencies - Will be done separately


shikaan (approved):
Looks good, as we decided to postpone upgrade of single apps

PR #180, opened 2021-01-11
Fixes after the initial publish

andipaetzold:
After initially publishing our new packages, I noticed minor issues:
* `main` and `types` is not configured correctly in `package.json`
* We include all files and not only the `lib` directory
* CI breaks on `lerna publish` because it's waiting for user interaction

### DIFF
--- a/packages/ecommerce-app-base/package.json
+++ b/packages/ecommerce-app-base/package.json
@@ -1,7 +1,8 @@
 {
   "name": "@contentful/ecommerce-app-base",
   "version": "1.0.1",
-  "main": "lib/",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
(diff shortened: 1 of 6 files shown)


packages/ecommerce-app-base/package.json
```diff
@@ -1,9 +1,13 @@
 {
   "name": "@contentful/ecommerce-app-base",
   "version": "1.0.1",
-  "main": "lib/",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
   "author": "Contentful GmbH",
   "license": "MIT",
+  "files": [
+    "lib"
```
shikaan, line 9: I think it's safer to use globs here to capture the files we want to include and reduce the package size

shikaan (approved):

PR #183, opened 2021-01-12
chore: minor package doc fixes

andipaetzold:
Fixing some example code and fix copy-paste errors in the docs

### DIFF
--- a/packages/ecommerce-app-base/docs/README.md
+++ b/packages/ecommerce-app-base/docs/README.md
@@ -50,7 +50,7 @@ ___
 
 Function that should return true when the button should be disabled.
 
-**`param`** Currently selected assets
+**`param`** Currently selected skus
 
 **`param`** App configuration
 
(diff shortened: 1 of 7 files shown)


packages/ecommerce-app-base/docs/README.md
```diff
@@ -50,7 +50,7 @@ ___
 
 Function that should return true when the button should be disabled.
 
-**`param`** Currently selected assets
+**`param`** Currently selected skus
```
FLoppix, line 53: Super nit: Maybe explain the abbreviation here? so we say stock-keeping unit (sku) or something similiar?

FLoppix (commented):
just one comment, otherwise LGTM

packages/ecommerce-app-base/docs/README.md
andipaetzold: I've added a sentence to both readmes to explain what the app actually does. Also mentioned stock keeping unit there

FLoppix (approved):

PR #187, opened 2021-01-19
chore: don't publish when changing docs or tests

andipaetzold:
We do not want to publish a new package version for every change to the docs. (e.g. https://github.com/contentful/apps/pull/186)

This can be tested locally by committing a change and running `lerna changed`.

Also, with this PR changes to tests and updating the package-lock.json (which might happen more frequently with dependabot) will be ignored.
We don't really care about bumping the app's versions as they are not published to npm.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -9,5 +9,11 @@
       "message": "chore: publish [skip ci]",
       "conventionalCommits": true
     }
-  }
+  },
+  "ignoreChanges": [
+    "**/*.md",
+    "**/*.spec.*",
+    "**/test/**",
+    "packages/*/package-lock.json"
+  ]
 }


lerna.json
```diff
@@ -9,5 +9,11 @@
       "message": "chore: publish [skip ci]",
       "conventionalCommits": true
     }
-  }
+  },
+  "ignoreChanges": [
+    "**/*.md",
+    "**/*.spec.*",
+    "**/test/**",
+    "packages/*/package-lock.json"
```
Jwhiles, line 17: What's a situation where package-lock.json changes and we don't want to publilsh a new version?

andipaetzold: The `package-lock.json` does not get published when publishing a npm package. Only the `package.json` is published.

https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json
> One key detail about package-lock.json is that it cannot be published

Jwhiles (approved):

PR #188, opened 2021-01-20
feat: add example code for tutorial

FLoppix:
Add the example code for the blog post metrics tutorial


examples/blog-post-metrics/src/components/ConfigScreen.tsx
```diff
@@ -0,0 +1,67 @@
...
+      // locations, you can just pass the currentState as is
+      targetState: currentState
+    };
+  };
+
+  render() {
+    return (
+      <Workbench className={css({ margin: '80px' })}>
+        <Form>
+          <Heading>App Config</Heading>
```
andipaetzold, line 61: That's not the headline from the tutorial

examples/blog-post-metrics/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@contentful/forma-36-react-components": "^3.75.0",
+    "@contentful/forma-36-tokens": "^0.9.4",
+    "@testing-library/jest-dom": "^5.11.9",
+    "@testing-library/react": "^11.2.3",
+    "@testing-library/user-event": "^12.6.0",
+    "@types/jest": "^26.0.20",
+    "@types/node": "^14.14.22",
+    "@types/react": "^17.0.0",
+    "@types/react-dom": "^17.0.0",
+    "contentful-ui-extensions-sdk": "^3.31.0",
```
andipaetzold, line 18: Change to `@contentful/app-sdk`

andipaetzold (approved):

examples/blog-post-metrics/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
VaguelySerious, line 1: I wasn't sure what this is, so I looked it up, but I'm still not sure if this makes any sense as an empty file?

examples/blog-post-metrics/src/components/ConfigScreen.tsx
```diff
@@ -0,0 +1,74 @@
...
+      this.props.sdk.app.setReady();
+    });
+  }
+
+  onConfigure = async () => {
+    // This method will be called when a user clicks on "Install"
+    // or "Save" in the configuration screen.
+    // for more details see https://www.contentful.com/developers/docs/extensibility/ui-extensions/sdk-reference/#register-an-app-configuration-hook
+
+    // Get current the state of EditorInterface and other entities
```
VaguelySerious, line 49:
```suggestion
    // Get the current state of EditorInterface and other entities
```

examples/blog-post-metrics/src/components/ConfigScreen.tsx
```diff
@@ -0,0 +1,74 @@
...
+    // related to this app installation
+    const currentState = await this.props.sdk.app.getCurrentState();
+
+    return {
+      // Parameters to be persisted as the app configuration.
+      parameters: this.state.parameters,
+      // In case you don't want to submit any update to app
```
VaguelySerious, line 56:
```suggestion
      // If you don't need to update the app locations,
      // you can just pass the currentState as is
```

examples/blog-post-metrics/src/components/ConfigScreen.tsx
```diff
@@ -0,0 +1,74 @@
...
+    };
+  };
+
+  render() {
+    return (
+      <Workbench className={css({ margin: "80px" })}>
+        <Form>
+          <Heading>Blog Post Metrics Config</Heading>
+          <Paragraph>
+            Welcome to your contentful app. This is your config page.
```
VaguelySerious, line 68:
```suggestion
            Welcome to your app. This is your config page.
```
This is probably a problem in `create-contentful-app` too, huh? I'll make a note to do some more renaming there.

examples/blog-post-metrics/src/components/Dialog.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+interface DialogProps {
+  sdk: DialogExtensionSDK;
+}
+
+const Dialog = (props: DialogProps) => {
+  return <Paragraph>Hello Dialog Component</Paragraph>;
+};
+
+export default Dialog;
```
VaguelySerious, line 13: Discussion: The tutorial only talks about `sidebar` and `config` locations. Should we delete unrelated components, or keep them in as a better starting point?

FLoppix: I had the same conversation with Andi :D I would prefer to leave it here. Its weird to delete files when you just created them and I think they will serve as a good starting point.

examples/blog-post-metrics/src/react-app-env.d.ts
FLoppix: There is some discussion around this file https://github.com/facebook/create-react-app/issues/6560. I would just leave it there for now

andipaetzold (approved):

VaguelySerious (approved):
Nit: your linter seems inconsistent. After the last changes we have mixed single and double quotes depending on the file.

examples/blog-post-metrics/src/components/Sidebar.spec.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import Sidebar from "./Sidebar";
+import { render } from "@testing-library/react";
+
+describe("Sidebar component", () => {
+  it("Component text exists", () => {
+    const { getByText } = render(<Sidebar />);
+
+    expect(getByText("Hello Sidebar Component")).toBeInTheDocument();
```
kdamball, line 9: How's this test passing? 🤔

andipaetzold: Because we are not running tests?

PR #189, opened 2021-01-22
Update gatsby extension base

ax-vasquez:
Update `apps/gatsby` so that:
1. It uses `gatsby-cms-extension-base` version `0.0.46`
2. Reference to `previewInstanceUrl` is renamed to `previewUrl` (as the CMS extension base expects)

### DIFF
--- a/apps/gatsby/package-lock.json
+++ b/apps/gatsby/package-lock.json
@@ -1526,26 +1526,150 @@
       }
     },
     "@emotion/core": {
-      "version": "10.0.14",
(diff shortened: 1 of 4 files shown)


andipaetzold (approved):
LGTM.
Thanks for the PR! 👍

ax-vasquez:
Awesome! Any ETA on when this change will go live? I don't mean to be impatient, I'm just trying to figure out what I should estimate :sweat_smile:

andipaetzold:
@ax-vasquez It's live now. 🚀

PR #190, opened 2021-01-26
chore: update forma

FLoppix:
Waiting for https://github.com/contentful/forma-36/issues/818


apps/gatsby/src/AppConfig/ContentTypesPanel.js
```diff
@@ -52,16 +52,18 @@ export const ContentTypesList = ({
...
-      key={sys.id}
-      labelIsLight
-      labelText={name}
-      name={name}
-      checked={enabledContentTypes.includes(sys.id)}
-      value={sys.id}
-      onChange={() => onContentTypeToggle(sys.id)}
-      id={sys.id}
-    />
+    <div>
```
andipaetzold, line 55: Why the wrapping div here? Sounds looks a hacky fix?

apps/jira/jira-app/package.json
```diff
@@ -3,27 +3,28 @@
...
-    "@babel/preset-env": "7.3.4",
-    "@babel/preset-react": "7.0.0",
+    "@babel/core": "7.8.4",
+    "@babel/plugin-proposal-class-properties": "7.8.3",
+    "@babel/plugin-transform-runtime": "7.8.3",
+    "@babel/preset-env": "^7.8.7",
+    "@babel/preset-react": "7.8.3",
     "@contentful/contentful-extension-scripts": "^0.19.1",
-    "@contentful/eslint-config-extension": "0.2.0",
+    "@contentful/eslint-config-extension": "^0.2.0",
```
andipaetzold, line 12: Why did you unpin the dependencies here? Especially dev dependencies can/should always be pinned to avoid issues.

apps/jira/jira-app/package.json
```diff
@@ -3,27 +3,28 @@
...
+    "@typescript-eslint/parser": "^4.14.2",
     "contentful-cli": "1.2.12",
     "cssnano": "4.1.10",
     "eslint": "^6.0.1",
     "fetch-mock": "^8.3.2",
     "node-fetch": "^2.6.1",
     "prettier": "1.19.1",
     "sass": "^1.23.7",
-    "typescript": "3.5.2"
+    "typescript": "^3.9.7"
```
andipaetzold, line 27: As TypeScript doesn't follow semantic versioning, we can go to the latest version straight away - unless you encountered any issues.

andipaetzold (commented):
I did not manually test all apps to see whther they work. But the changes look fine so far. Just two questions that apply for multiple apps in this repo.

apps/gatsby/src/AppConfig/ContentTypesPanel.js
FLoppix: No actually to achieve a line break. The current visuals looked not good with the new forma updates and the spacing was totally off because all the content types were on the same line.

apps/jira/jira-app/package.json
FLoppix: I was just using the latest 3.X version. But yes that is a good point :) I will update to the latest version

apps/gatsby/src/AppConfig/ContentTypesPanel.js
andipaetzold: Okay then maybe the new version changed the `display` type...

andipaetzold (approved):

PR #197, opened 2021-03-05
[NONE] Optional slug definition per content type

martin3walker:
Allows defining a slug to be appended to base path, using strings or fields, for each content type where the app is installed.


apps/gatsby/gatsby.code-workspace
```diff
@@ -0,0 +1,8 @@
+{
```
shikaan, line 1: This looks like a mistake

apps/gatsby/src/Sidebar.js
```diff
@@ -42,8 +40,72 @@ export default class Sidebar extends React.Component {
...
+    //Find the url constructor for the given contentType
+    const constructor = urlConstructors ? urlConstructors.find(
+      constructor => constructor.id === this.sdk.contentType.sys.id
+    ) : undefined;
+    // If there is no constructor set the url as the base preview
+    if (!constructor){
+      return
+    }
+
+    //QUESTION: What does the locale object in the sdk look like when there are multiple locales available. Need to account for spaces that have more than one locale, so they can have previews everywhere.
```
shikaan, line 54: https://github.com/contentful/ui-extensions-sdk/blob/861d665cb3c79d7e20261bd6b0a55cf2dd903890/lib/types.ts#L380-L394

This should be it

apps/gatsby/src/Sidebar.js
```diff
@@ -42,8 +40,72 @@ export default class Sidebar extends React.Component {
...
+
+    // Recursive helper to return slug values buried in a chain of references
+    const resolveReferenceChain = async (sdk, array, index, parentId) => {
+      console.log(sdk)
```
shikaan, line 58: Leftover

apps/gatsby/src/Sidebar.js
```diff
@@ -42,8 +40,72 @@ export default class Sidebar extends React.Component {
...
+       console.error("Gatsby Preview App: You are trying to search for a slug in a multi reference field. Only single reference fields are searchable with this app. Either change the field to a single reference, or change the field you are searching for in the slug.")
+        return ""
+      }
+
+      if (index + 2 < array.length) {
+        return resolveReferenceChain(sdk, array, (index + 1), childField.sys.id)
+      } else {
+        return childField
+      }
+    }
```
shikaan, line 73: Do we need to redefine this on every mount? 🤔 Maybe it can be just a method of the class

apps/gatsby/src/Sidebar.js
```diff
@@ -42,8 +40,72 @@ export default class Sidebar extends React.Component {
...
+          } catch {
+            console.error(`Gatsby Preview App: ${fieldArray[0]}, as defined in the slug field for this content type in the Gatsby Preview App, is not a field. Maybe you mistyped it, or maybe you meant it to be a static string in which case you need to surround it in quotes: ${`"${fieldArray[0]}"`}. The open preview button will send users to your site's base url until fixed.`)
+          }
+        })
+      )
+    )
+    //Make sure the base preview url ends with a /
+    const cleanPreviewUrl = previewUrl.charAt(previewUrl.length - 1) === "/" ? previewUrl : `${previewUrl}/`
+    
+    const fullUrl = `${cleanPreviewUrl}${slug.toString().replace(/,/i, "/")}`
```
shikaan, line 101: Shall we maybe assert the slug has the correct shape and throw if not rather than relying on the regexp-replace output here?

apps/gatsby/src/AppConfig/AppConfig.js
```diff
@@ -150,26 +157,79 @@ export class AppConfig extends React.Component {
...
-    if (enabledContentTypes.includes(ctId)) {
-      return enabledContentTypes.filter((cur) => cur !== ctId);
+  selectorTypeToggle = () => {
+    this.setState((prevState) => ({
+      selectorType: !prevState.selectorType
+    }))
+  }
+
+  disableContentType = (id) => {
+    const index = this.state.enabledContentTypes.findIndex(type => type === id)
```
shikaan, line 167: Maybe we can just `filter` and check on the output length?

shikaan (commented):
Nothing major (not sure if you want to tackle the multi-locale thing in this PR). 
Just a couple of comments to be addressed and we can ship this puppy 🐶

apps/gatsby/gatsby.code-workspace
martin3walker: Deleted 👍

apps/gatsby/src/Sidebar.js
martin3walker: Do you mean assert is the TS sense of assert? I think this is all written in JS. Also do you mean throw an error on the frontend, like a notification to the editor?

apps/gatsby/src/AppConfig/AppConfig.js
martin3walker: You mean check the output length of filtered vs. unfiltered array, and set state if the filtered array is less?

apps/gatsby/src/Sidebar.js
shikaan: I mean just checking whether the output is conform to what we expect. For example:

```
const assertSlug = (s) => {
  if (slug.length > 0) { 
    return 
  }
  
  throw new Error('Unexpected slug shape')
}
```

And then in the code

```
    const cleanPreviewUrl = previewUrl.charAt(previewUrl.length - 1) === "/" ? previewUrl : `${previewUrl}/`

    assertSlug(slug)

    const fullUrl = `${cleanPreviewUrl}${slug.join("/")}` 
```

apps/gatsby/src/AppConfig/AppConfig.js
shikaan: I have meant (variables named very poorly, just to give an idea)

```
const newEnabledContentTypes = this.state.enabledContentTypes.filter(type => type !== id)
const hasToUpdate = this.state.enabledContentTypes > newEnabledContentTypes.length

if (hasToUpdate) {
  this.setState(() => ({ enabledContentTypes: newEnabledContentTypes }))
}
```

The idea is avoid skimming through the array twice. Not a big deal, since I don't expect this to be huge

shikaan (approved):

PR #203, opened 2021-03-16
refactor: [EXT-2590] update bynder version with required changes

loweisz:
Updating bynder app according to https://developer-docs.bynder.com/ui-components migration steps. 

All fields that are not there with the new version were added through `assetFieldSelection` and then tied together with `transformAsset`, to be in the shape how it was before to ensure backwards compatibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules
 build
 dist
 .cache
+.idea
 tags
 **/.coverage/**
 lerna-debug.log
--- a/apps/bynder/src/index.html
+++ b/apps/bynder/src/index.html
@@ -24,7 +24,6 @@
     </style>
   </head>
   <body>
-    <script src="https://d8ejoa1fys2rk.cloudfront.net/5.0.5/modules/compactview/bynder-compactview-2-latest.js"></script>
     <div id="root"></div>
     <script src="./index.js"></script>
   </body>
--- a/apps/bynder/src/index.html
+++ b/apps/bynder/src/index.html
@@ -22,6 +22,7 @@
     </style>
   </head>
   <body>
+    <script src="https://d8ejoa1fys2rk.cloudfront.net/5.0.5/modules/compactview/bynder-compactview-2-latest.js"></script>
     <div id="root"></div>
     <script src="./index.js"></script>
   </body>
--- a/apps/bynder/src/index.html
+++ b/apps/bynder/src/index.html
@@ -16,7 +16,9 @@
       }
 
       .dialog-container {
-        height: 800px;
+        height: 600px;
+        display: flex;
+        justify-content: center;
         position: relative;
       }
     </style>
--- a/apps/bynder/src/index.js
+++ b/apps/bynder/src/index.js
@@ -74,6 +74,15 @@ function prepareBynderHTML({ bynderURL, assetTypes }) {
 }
 
 function transformAsset(asset) {
+  const thumbnails = {
+    "webimage": asset.files.webImage?.url,
+    "thul": asset.files.thumbnail?.url
+  }
+
+  Object.entries(asset.files)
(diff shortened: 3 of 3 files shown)


apps/bynder/src/index.html
```diff
@@ -16,12 +16,15 @@
...
-        height: 800px;
+        height: 600px;
+        display: flex;
+        justify-content: center;
         position: relative;
       }
     </style>
   </head>
   <body>
+    <script src="https://d8ejoa1fys2rk.cloudfront.net/5.0.5/modules/compactview/bynder-compactview-2-latest.js"></script>
```
realityking, line 27: Wouldn't this cause the external JS to be loaded even when the modal isn't opened? Any way to avoid that?

loweisz: Good point 👍  I changed it to be loaded dynamically now

apps/bynder/src/index.js
```diff
@@ -43,51 +66,90 @@ function makeThumbnail(resource) {
...
+    "userCreated": asset.createdBy,
+    "datePublished": asset.publishedAt,
+    "dateCreated": asset.createdAt,
+    "dateModified": asset.updatedAt,
+    "watermarked": asset.isWatermarked ? 1 : 0,
+    "limited": asset.isLimitedUse ? 1 : 0,
+    "isPublic": asset.isPublic ? 1 : 0,
+    "brandId": asset.brandId,
+    "thumbnails": {
+      "contentful": asset.files.contentful.url,
```
Jwhiles, line 98: these properties on files aren't necessarily defined, which can make this throw an error when trying to add a file. Can we use optional chaining?

Jwhiles: I'm also worried that there might be other properties that used to exist on Thumbnails that we've missed here. Were you able to find a complete list anywhere?

loweisz: not really :/ That were the only ones I found. I changed it to be generated dynamically now, however for `webImage` and `thumbnail` I still had to manually map it, as the naming has changed and to support backwards compatibility. Do you think it's save now?

.gitignore
```diff
@@ -3,6 +3,7 @@ node_modules
 build
 dist
 .cache
+.idea
```
kdamball, line 6: Where's this from? Just curious

apps/bynder/src/index.js
```diff
@@ -43,51 +66,94 @@ function makeThumbnail(resource) {
...
+    "watermarked": asset.isWatermarked ? 1 : 0,
+    "limited": asset.isLimitedUse ? 1 : 0,
+    "isPublic": asset.isPublic ? 1 : 0,
+    "brandId": asset.brandId,
+    "thumbnails": thumbnails
+  })
+}
+
+function checkMessageEvent(e) {
+  if (e.origin === 'https://app.contentful.com') {
```
kdamball, line 111: Wont this ignore other environments? i.e: we're assuming it will only work for production environments. Maybe passing in an env variable might be a better idea

loweisz: I flipped the check now, so it should work for all environments :)

.gitignore
loweisz: It's something from my IDE (Webstorm), so I just added it to not accidentally commit it

kdamball (approved):

loweisz:
**Note**:

Double checking this and added the missing field `videoPreviewURLs` as `previewUrls` as I found out it was mentioned in the migration guide:
Screenshot 2021-05-10 at 16 30 28

However there is some risk as it is not mentioned in the Assets Interface here: https://developer-docs.bynder.com/ui-components#compact-view

I've checked some examples and the fields seem to be identical. So I think we are save to update with this even though it's not mentioned in the interface.

PR #218, opened 2021-04-06
fix(Shopify): 'load more' button was wrongly displayed in some edge cases

andipaetzold:
In some rare cases the "Load more" button was displayed even though there was no new page with products/variants available. The root cause of this issue is that we get an unsorted list of products and only one of them had `hasNextPage=false`. So we now check whether any of the newly loaded products has this flag set to false and then hide the button.

### DIFF
--- a/apps/shopify/src/Pagination.js
+++ b/apps/shopify/src/Pagination.js
@@ -44,7 +44,7 @@ class Pagination {
       return {
         pagination: {
           // There is going to be a next page in the following two complimentary cases:
-          // A). The product corresponding to the last variant belongs is tagged as having a next page
+          // A). There are more products to fetch via the Shopify API
           // B). There are variants left to consume in the in-memory variants list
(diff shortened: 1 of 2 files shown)


apps/shopify/src/Pagination.js
```diff
@@ -40,13 +41,13 @@ class Pagination {
     const hasEnoughVariantsToConsume = this.variants.length >= PER_PAGE || nothingLeftToFetch;
     if (hasEnoughVariantsToConsume) {
       const variants = this.variants.splice(0, PER_PAGE);
-      const lastProduct = this.products.find(product => product.id === last(variants).productId);
+      // const lastProduct = this.products.find(product => product.id === last(variants).productId);
```
giotiskl, line 44:
```suggestion

```
let's remove the commented out line

giotiskl (approved):
LGTM with a small note

PR #219, opened 2021-04-08
feat: optionally define entry slugs

martin3walker:
feat: optionally define entry slugs


apps/gatsby/src/AppConfig/AppConfig.js
```diff
@@ -150,26 +157,80 @@ export class AppConfig extends React.Component {
...
+    const index = currentUrlConstructors.findIndex(cur => cur.id === id)
+    if (index !== -1) {
+      currentUrlConstructors[index].slug = newInput
+       constructors = currentUrlConstructors
+    } else {
+      const newConstructor = {
+        id,
+        slug: newInput
+      }
+      constructors = [...currentUrlConstructors, ...[newConstructor]]
```
shikaan, line 212: Maybe we should add some validation logic here. Tried to use the feature, but it wasn't super clear to me why things weren't working.

As a matter of fact, I have only managed to get the "string" case to work before reading the code. The rest did not and without notifying me of why (e.g. was the input in incorrect format?)

apps/gatsby/src/Sidebar.js
```diff
@@ -42,8 +41,68 @@ export default class Sidebar extends React.Component {
...
+      return ""
+    }
+    if (index + 2 < array.length) {
+       return this.resolveReferenceChain(sdk, array, (index + 1), childField.sys.id)
+    } else {
+      return childField
+    }
+  }
+
+  buildSlug = async () => {
```
shikaan, line 61: Slug should be rebuilt on value change, if the reference is a field (otherwise I see one thing in the editor, but the Gatsby button will redirect me to the old url)

apps/gatsby/src/AppConfig/AppConfig.js
martin3walker: I don't know how useful it would be. As far as I know, the only thing that I could validate is that a string is being passed. Whether that string results in the desired value is decided in the sidebar.js file. I recognize that this part of the setup is a bit confusing though. I think a good solution could be creating a minute long explainer video, that I'll link in the markup, so people have a better idea of what is going on.

apps/gatsby/src/Sidebar.js
```diff
@@ -71,15 +130,15 @@ export default class Sidebar extends React.Component {
...
   };
 
-  render = () => {
-    const { webhookUrl, previewUrl, authToken } = this.sdk.parameters.installation;
-    const contentSlug = this.sdk.entry.fields.slug;
-
+  render =  () => {
+    const { webhookUrl, authToken, previewUrl } = this.sdk.parameters.installation;
+    const { slug } = this.state
+    console.log(`Gatsby cloud preview slug: ${slug}`)
```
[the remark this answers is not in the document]
martin3walker: Removed 👍

apps/gatsby/src/AppConfig/AppConfig.js
martin3walker: Added an explainer video

apps/gatsby/src/Sidebar.js
martin3walker: I opted to rerun `buildSlug` in with `entry.onSysChanged` event listener callback. It updates too often, but it will also be far more likely to rebuild when a slug buried in a reference is changed, and it should definitely update every time a slug field directly on an entry is updated.

shikaan (approved):

PR #220, opened 2021-04-13
feat: add option to change search debounce time

Jwhiles:


andipaetzold:
Can you change the title to sth like `feat: add option to change search debounce time`.
This will be published in the changelog of `ecommerce-app-base`. People don't care about Shopify in there.

andipaetzold (approved):

PR #225, opened 2021-04-19
fix: stop simultaneous recursive chains

Jwhiles:
Sometimes the calls made in shopify can take a long time to resolve. Because of the way the pagination code is structured, this could lead to there being multiple recursive chains of calls for the function `fetchNext` happening simultaneously. Each time the function is invoked we checked whether the search term had changed, and if so, assumed that this was a new search term and that we should start fetching from scratch. If two requests became active at once, they would both repeatedly reset, as with each call it would be appear that the search term had changed.

### DIFF
--- a/apps/shopify/src/Pagination.js
+++ b/apps/shopify/src/Pagination.js
@@ -25,8 +25,13 @@ class Pagination {
     this.shopifyClient = await makeShopifyClient(this.sdk);
   }
 
-  async fetchNext(search) {
+  async fetchNext(search, recursing = false) {
     const searchHasChanged = search !== this.prevSearch;
+    const shouldStop = searchHasChanged && recursing;
+    if (shouldStop) {
+      return;
+    }
+
     if (searchHasChanged) {
       this.prevSearch = search;
       this._resetPagination();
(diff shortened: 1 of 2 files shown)


apps/shopify/src/Pagination.js
```diff
@@ -25,8 +25,13 @@ class Pagination {
     this.shopifyClient = await makeShopifyClient(this.sdk);
   }
 
-  async fetchNext(search) {
+  async fetchNext(search, recursing = false) {
```
shikaan, line 28: I am not really a fan of arbitrary flags, but that might be me. What about interrupting recursion _before_ entering the new call? So just before the `return` of this method (L63 in the new version, L58 of the old one)

shikaan (approved):
No red flags. Just one minor comment which makes - to me - the code easier to reason about

apps/shopify/src/Pagination.js
Jwhiles: Yeah, that seems like a good idea :)

Jwhiles: I'm not sure. I think that makes it more likely that we'll be returning data from a dead chain - and I think would make this change a lot more complicated.

PR #236, opened 2021-04-28
fix: only handle messages events of relevant sources [EXT-2675]

andipaetzold:
Currently, the message event handler is also triggered for messages being sent from the web app to the app iframe. This simple condition excludes those messages and ensures the handler only handles messages coming from the frontify iframe.

Question: Should we also include this condition in other apps to avoid infinite loops as they happed with frontify?

### DIFF
--- a/apps/brandfolder/src/index.js
+++ b/apps/brandfolder/src/index.js
@@ -38,23 +38,32 @@ function renderDialog(sdk) {
   const container = document.createElement('div');
   const bf_embed_url = config.bf_api_key ? BF_EMBED_URL + `&apiKey=${config.bf_api_key}&hideLogout=true` : BF_EMBED_URL;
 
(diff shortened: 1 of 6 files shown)


Jwhiles:
Looks good. I'd agree that it makes sense to do the same in the other apps.

apps/brandfolder/src/index.js
```diff
@@ -38,23 +38,32 @@ function renderDialog(sdk) {
...
-      if (event === 'selectedAttachment') {
-        sdk.close([payload]);
-      } else if (data.event === 'selectedAsset' && payload.attachments.length !== 0) {
-        const att_id = payload.attachments[0].id;
-        const attachment = payload.included.find(att => att.id === att_id);
-        if (attachment) {
-          sdk.close([attachment]);
-        }
+    if (e.source !== iframe.contentWindow) {
+      return ;
```
Jwhiles, line 56:
```suggestion
      return;
```

Jwhiles (approved):

PR #238, opened 2021-05-04
feat: [EXT-2722] use contentful hosting for image focal point app

Jwhiles:
Generally this is pretty straightforward. 

The weirdest change is probably that I've created a `test:deploy` script in each package.json. At the moment just calls the deploy script for most apps. However in the focal point app, this is now an actually different script, which will try and upload the bundle to a test org/definition. 

I've done this, because it seems tricky to manage a test deployment using our new scripts just using environment variables. It would mean making each app definition an environment variable, and then overwriting those definitions for the test script.

### DIFF
--- a/apps/image-focal-point/package.json
+++ b/apps/image-focal-point/package.json
@@ -8,6 +8,7 @@
     "@babel/plugin-transform-runtime": "7.3.4",
     "@babel/preset-env": "7.3.4",
     "@babel/preset-react": "7.0.0",
+    "@contentful/app-scripts": "^0.8.11",
     "@contentful/contentful-extension-scripts": "^0.15.1",
     "@contentful/eslint-config-extension": "0.2.0",
     "@testing-library/react": "^11.2.5",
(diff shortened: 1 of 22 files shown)


apps/image-focal-point/package.json
```diff
@@ -19,7 +20,8 @@
     "start": "contentful-extension-scripts start --serve-only",
     "build": "contentful-extension-scripts build --no-inline",
     "lint": "eslint ./ --ext .js,.jsx,.ts,.tsx",
-    "deploy": "aws s3 sync ./build ${STATIC_S3_BASE}/focal-point --acl public-read",
+    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEFINITIONS_ORG_ID} --definition-id 342Q2DqCjmsdN5BJCEPkrJ --token ${CONTENTFUL_CMA_TOKEN}",
+    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${TEST_ORG_ID} --definition-id ${TEST_APP_ID} --token ${CONTENTFUL_CMA_TOKEN}",
```
Jwhiles, line 24: my assumption here is that we'll create a new definition which we can just use to test-deploy all the apps - since it seems that the main thing we're testing here is that the upload / activation will succeed

jelz: Makes sense

jelz (approved):

andipaetzold (approved):

PR #239, opened 2021-05-06
feat: [EXT-2533] use new version of ecommerce base app

Jwhiles:


FLoppix (approved):

andipaetzold:
Why changing it to `^2.0.0` and not `2.1.0`?
If I understand lerna correctly, every change to ecommerce base package would also bump the version of the dependant apps.

PR #268, opened 2021-05-26
feat: [EXT-2533] update commercelayer and saleor to use cca

Jwhiles:
This also allows both of these apps to be updated to the latest version of ecommerce base app


shikaan (approved):
10x

PR #269, opened 2021-05-26
feat(cloudinary-asset): add transformations option

backflip:
This change adds a `defaultTransformations` setting to specify the [`default_transformations`](https://cloudinary.com/documentation/media_library_widget#3_set_the_configuration_options) option of the Cloudinary Media Library Widget. 

It takes precedence over the existing `quality` and `format` settings and allows to specify more complex transformations.

### DIFF
--- a/apps/cloudinary/src/index.js
+++ b/apps/cloudinary/src/index.js
@@ -72,17 +72,22 @@ function makeThumbnail(resource, config) {
 function renderDialog(sdk) {
   const { cloudinary } = window;
   const config = sdk.parameters.invocation;
-  let default_transformations = {};
+  let defaultTransformations = JSON.parse(config.defaultTransformations);
   
+  if (!defaultTransformations.length) {
+    const defaultTransformation = {};
 
-  // Handle format
-  if(config.format!=='none'){
-    default_transformations.fetch_format = config.format;
-  }
+    // Handle format
+    if (config.format !== 'none') {
+      defaultTransformation.fetch_format = config.format;
+    }
 
-  // Handle quality
-  if(config.quality!=='none'){
-    default_transformations.quality = config.quality;
+    // Handle quality
+    if (config.quality !== 'none') {
+      defaultTransformation.quality = config.quality;
+    }
+
+    defaultTransformations.push(defaultTransformation);
   }
 
   const options = {
@@ -92,7 +97,7 @@ function renderDialog(sdk) {
     multiple: config.maxFiles > 1,
     inline_container: '#root',
     remove_header: true,
-    default_transformations: [default_transformations]
+    default_transformations: defaultTransformations
   };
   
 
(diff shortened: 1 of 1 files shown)


apps/cloudinary/src/index.js
```diff
@@ -72,17 +72,22 @@ function makeThumbnail(resource, config) {
 function renderDialog(sdk) {
   const { cloudinary } = window;
   const config = sdk.parameters.invocation;
-  let default_transformations = {};
+  let defaultTransformations = JSON.parse(config.defaultTransformations);
```
backflip, line 75: Can we safely assume that this is not failing when we are testing it in `validateParameters` below?

apps/cloudinary/src/index.js
```diff
@@ -158,13 +163,23 @@ function validateParameters(parameters) {
...
 
+  let defaultTransformations = [];
+  try {
+    defaultTransformations = JSON.parse(parameters.defaultTransformations);
+  } catch(err) {
+    defaultTransformations = err;
+  }
+  if (!Array.isArray(defaultTransformations)) {
+    return 'Default transformations should be a valid JSON array.';
+  }
```
shikaan, line 181: I'd rather return the error string in the catch block to make this behaviour more predictable.
```suggestion
  } catch(err) {
    return 'Default transformations should be a valid JSON array.';
  }
```

apps/cloudinary/src/index.js
```diff
@@ -72,17 +72,22 @@ function makeThumbnail(resource, config) {
 function renderDialog(sdk) {
   const { cloudinary } = window;
   const config = sdk.parameters.invocation;
-  let default_transformations = {};
+  let defaultTransformations = JSON.parse(config.defaultTransformations);
   
+  if (!defaultTransformations.length) {
```
shikaan, line 77: Why overriding the other transformations when the list is empty?

apps/cloudinary/src/index.js
```diff
@@ -223,6 +238,14 @@ setup({
...
       "required": true,
       "default": "auto"
+    },
+    {
+      "id": "defaultTransformations",
+      "name": "Default Transformations",
+      "description": "Specific upload transformations. When this is specified, 'Media Quality' and 'Format' are ignored. See https://cloudinary.com/documentation/transformations_on_upload for available transformations. Needs to be a valid JSON array, e.g. '[{\"quality\": 15, \"width\": 100, \"height\": 100, \"crop\": \"crop\"}]'.",
+      "type": "Symbol",
+      "required": false,
+      "default": "[]"
```
shikaan, line 248: A free form field sounds rather risky: it will make this app hard to be backwards compatible, should Cloudinary API change, and requires a non trivial validation to be guarded against malicious injections.

What about adding the transformations you are after as the other items in this list?

apps/cloudinary/src/index.js
shikaan: I don't think so. There is no validation on the free-form symbol type

shikaan (commented):
First of all, thanks for your contribution @backflip 🙇  We profoundly appreciate you taking your time trying to add more features to our apps. 

I have left a couple of comments on the change and I would like to ask: what is it that you are after here? While we integrate with external services (as Cloudinary in this case) we tend to not allow free-form input, because handling compatibility down the line might get tricky. If you are looking for some specific transformations, my team and I will be happy to help you adding the features you need.

backflip:
Hey @shikaan, sorry about the sparse information regarding the reason for this change request! 

So the background is the following: 
- We have multiple web apps consuming media from our Cloudinary account.
- Some of them have rather specific output requirements and want to eagerly transform the corresponding media while uploading (for frontend performance reasons). Example:
  ```
  [
    {
      format: "mp4",
      video_codec: "h264",
      audio_codec: "aac",
      quality: "auto:best",
      width: 3840,
      height: 2160,
      crop: "limit",
    },
    {
      format: "mp4",
      video_codec: "h264",
      audio_codec: "aac",
      quality: "auto:best",
      width: 1920,
      height: 1080,
      crop: "limit",
    },
    ...
  ];
  ```
- Most of them have their own Contentful space, so the Contentful-Cloudinary integration seemed a good choice to specify these transitions.

backflip:
Hey @shikaan, would you consider the use case described above to be a valid one or should we rather solve this in a fork?

kdamball:
@backflip Thanks for the PR. Unfortunately due to the current cloudinary app installations, we would risk current users' workflows with this change. We suggest you go with a fork for now

PR #270, opened 2021-05-26
chore: update set-editors example

Jwhiles:
We changed the requirements for target state at some point, breaking this example
this pr get's it working again


shikaan (approved):
As far as I can tell this should have never worked 👀  `TargetState.EditorInterface` is different than the `EditorInterface` entity and previous implementation used to have `boolean`s instead of `position`

Jwhiles:
Maybe it worked at some point by evaluating to true? Regardless, it works now :)

PR #277, opened 2021-06-04
feat: [EXT-2785] show variant SKU in shopify SKU picker

Jwhiles:


apps/shopify/src/dataTransformer.js
```diff
@@ -9,12 +9,14 @@ import flatten from 'lodash/flatten';
...
   const image = get(product, ['image', 'src'], '');
   const sku = get(product, ['sku'], '');
+  const variantSku = get(product, ['variantSKU'], '');
 
   return {
     id: product.id,
     image,
     name: product.title,
-    sku
+    readableSKU: variantSku,
```
alexdonets, line 18: I wonder, do we care about consistency with capitalization when naming vars? e.g. both would be `readableSku` and `variantSku` or `readableSKU` and `variantSKU`.

alexdonets (commented):
🔥

apps/shopify/src/dataTransformer.js
Jwhiles: Yeah, good point!

alexdonets (commented):
LGTM 👍🏼 I won't be approving things for now until I have more context :)

apps/shopify/src/dataTransformer.js
```diff
@@ -9,12 +9,14 @@ import flatten from 'lodash/flatten';
...
   const image = get(product, ['image', 'src'], '');
   const sku = get(product, ['sku'], '');
+  const variantSKU = get(product, ['variantSKU'], '');
 
   return {
     id: product.id,
     image,
     name: product.title,
-    sku
+    readableSKU: variantSKU,
```
andipaetzold, line 18: By calling it `readableSKU`, we limit this to be an sku. Why not just call it something like `subtitle`? That way people can put everything they want in it e.g. color, size etc.

Jwhiles: good idea

andipaetzold (commented):
- we need to regenerate the type docs (I think, not 100% sure)
- Do we want to make this functionality available that use the whole package via setup and not only the sku picker?

packages/ecommerce-app-base/src/interfaces.ts
```diff
@@ -45,6 +45,7 @@ export interface ParameterDefinition {
 
 export interface Product {
   sku: string;
+  displaySKU?: string;
```
andipaetzold, line 48: Can this ever be `undefined`? I think it's always either `SKU: ...` or `Product ID: ...`.
If it's never `undefined`, you could also remove some  `|| ${product.sku}` from the components.

Jwhiles: I'm thinking of the other apps that use this interface at the moment, and possible external users, as it would be a breaking change to make it required - and I don't know if it is necessary.

andipaetzold (approved):

PR #286, opened 2021-06-15
feat: [] use react scripts for ai-image-tagging frontend

Jwhiles:


apps/ai-image-tagging/frontend/extension.json
```diff
@@ -1,8 +0,0 @@
-{
```
Jwhiles, line 1: I believe this file is no longer used

andipaetzold (approved):

PR #287, opened 2021-06-15
feat: [] build brandfolder with react scripts

Jwhiles:


Jwhiles:
@andipaetzold I'll react-scripts into dependencies. CRA outputs projects which include react/react-dom in their dependencies, so I think we should also keep them.

andipaetzold (approved):

PR #298, opened 2021-06-17
feat: [] Build jira app with react scripts

Jwhiles:


apps/jira/jira-app/src/components/App/Config.tsx
```diff
@@ -6,6 +6,7 @@ import InstanceStep from './Steps/InstanceStep';
 import ContentTypeStep from './Steps/ContentTypeStep';
 import JiraStep from './Steps/JiraStep';
 import { AppExtensionSDK, CollectionResponse, EditorInterface } from '@contentful/app-sdk';
+import { JiraCloudResource, CloudProject, InstallationParameters } from '../../interfaces';
```
kdamball, line 9: weird - so, they weren't used?

apps/jira/jira-app/src/components/App/Steps/InstanceStep.tsx
```diff
@@ -8,6 +8,7 @@ import {
   TextInput,
   Card
 } from '@contentful/forma-36-react-components';
+import { JiraCloudResource, CloudProject } from '../../../interfaces';
```
kdamball, line 11: And were used here but not imported?

kdamball (approved):

kdamball:
@Jwhiles left a question, weird behaviour

Jwhiles:
@kdamball previously we made all these types globally available by having them in a `typings.d.ts` file at the top level. I made the choice to move this to an interface file and to import the types explicitly when we use them - as we don't tend to do to use the previous approach, and I think it's a bit cleaner. :)

PR #302, opened 2021-06-17
feat: [] build smartling with react-scripts

Jwhiles:


apps/smartling/frontend/src/Sidebar.tsx
```diff
@@ -98,7 +98,7 @@ function sortSubs(a: Submission, b: Submission) {
     return 0;
   }
 
-  if (aStatus == 'in_progress') {
```
loweisz, line 101: Have you checked that there are no weird intended side effects expected with that check ?

Jwhiles: pretty sure this was a mistake, aStatus is `const aStatus = a.status.toLowerCase();` - so definitely a string at this point.

loweisz: makes sense 👍

loweisz (approved):

PR #403, opened 2021-08-11
chore: [] upgrade some major versions

loweisz:
upgrading some major versions in smartling and jira to solve some security issues.

None of the breaking changes seems to affect our apps or tests in there


apps/smartling/frontend/package.json
```diff
@@ -11,7 +11,7 @@
     "@types/webpack-env": "1.13.9",
     "contentful-cli": "1.2.12",
     "cross-env": "^7.0.3",
-    "cssnano": "4.1.10",
+    "cssnano": "^5.0.7",
```
Jwhiles, line 14: I would check if we even need this any more. I think react scripts may have made it unnecessary.

loweisz: Good point! Looks like it is not used anywhere. I removed it 👍

Jwhiles (approved):

PR #456, opened 2021-09-09
chore: serverless v3 deprecations

andipaetzold:
With serverless v3 there will be a few changes to the `serverless.yml`. I adjusted the code already to remove a few warnings

The explanations of each change are in the respective commit message.

I didn't do anything about the Lambda hash algorithm change warning:

```
Resolution of lambda version hashes was improved with better algorithm, which will be used in next major release.
Switch to it now by setting "provider.lambdaHashingVersion" to "20201221"
More Info: https://www.serverless.com/framework/docs/deprecations/#LAMBDA_HASHING_VERSION_V2
```

### DIFF
--- a/apps/smartling/lambda/serverless.yml
+++ b/apps/smartling/lambda/serverless.yml
@@ -1,5 +1,4 @@
-service:
-  name: sls-apps-smartling
+service: sls-apps-smartling
 
 frameworkVersion: '>=2.0.0 <3.0.0'
 
--- a/apps/smartling/lambda/serverless.yml
+++ b/apps/smartling/lambda/serverless.yml
@@ -25,12 +25,11 @@ provider:
   deploymentPrefix: sls-apps-smartling
 
 package:
-  include:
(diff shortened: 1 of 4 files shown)


apps/smartling/lambda/serverless.yml
```diff
@@ -26,12 +25,11 @@ provider:
...
 
 package:
-  include:
-    - built/**
-  exclude:
-    - .git/**
-    - "**/*.ts"
-    - node_modules/smartling-frontend/node_modules/**
+  patterns:
+    - 'built/**'
```
Jwhiles, line 29: Is it just me or is this much less readable than how it worked before

Jwhiles (approved):

apps/smartling/lambda/serverless.yml
andipaetzold: 🤷 I think it's alright. It now feels more like a glob pattern and not a custom implementation.

PR #506, opened 2021-09-29
fix: [] Fix Shopify not displaying missing SKUs for collections/products

OriginalEXE:
Hello, this PR solves three issues:

1.) For collections preview we were fetching first 20 collections meaning that for stores with more than 20 collections the preview would break. I have increased that limit to 250 (max) as Shopify unfortunately does not support fetching multiple collections by ids

2.) This PR also makes sure that missing collections/products (if they are deleted for example) still show up in the interface. This was always the case for product variants, but I missed adding it for products/collections.

3.) I have also amended the `ecommerce-app-base` to display the correct sku type in the "missing" tag.

Thanks :)

### DIFF
--- a/packages/ecommerce-app-base/src/Editor/SortableListItem.tsx
+++ b/packages/ecommerce-app-base/src/Editor/SortableListItem.tsx
@@ -121,7 +121,7 @@ const CardDragHandle = SortableHandle(() => (
 export const SortableListItem = SortableElement<Props>(
   ({ product, disabled, isSortable, onDelete, skuType }: Props) => {
     const [imageHasLoaded, setImageLoaded] = useState(false);
-    const [imageHasErrored, setImageHasErrored] = useState(!product.image);
+    const [imageHasErrored, setImageHasErrored] = useState(false);
     const productIsMissing = !product.name;
 
     return (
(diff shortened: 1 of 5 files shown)


packages/ecommerce-app-base/src/Editor/SortableListItem.tsx
```diff
@@ -118,9 +119,9 @@ const CardDragHandle = SortableHandle(() => (
 ));
 
 export const SortableListItem = SortableElement<Props>(
-  ({ product, disabled, isSortable, onDelete }: Props) => {
+  ({ product, disabled, isSortable, onDelete, skuType }: Props) => {
     const [imageHasLoaded, setImageLoaded] = useState(false);
-    const [imageHasErrored, setImageHasErrored] = useState(false);
+    const [imageHasErrored, setImageHasErrored] = useState(!product.image);
```
Jwhiles, line 124: not sure I understand the reason for this change

OriginalEXE: Hey John, without that code, when we pass a product that has no image (happens when for example the sku can no longer be found in the shop), the component unnecessarily renders and `img` with an empty sec, only to then immediately error and set this boolean to true. By immediately setting it to true, we prevent that unnecessary dance.

Just a small optimization, but not important if you think it makes the code harder to read.

Jwhiles: I might instead leave this as is, make the following change below - because it makes a bit more sense to me, and avoids `imageHasErrored` actually meaning two different things. I leave it up to you though :) 

```
          {!product.image || imageHasErrored ? (
            <div className={styles.errorImage}>
              <Icon icon={productIsMissing ? 'ErrorCircle' : 'Asset'} />
            </div>
          ) : (
            <div className={styles.imageWrapper(imageHasLoaded)}>
              <img
                style={{ display: imageHasLoaded ? 'block' : 'none' }}
                onLoad={() => setImageLoaded(true)}
                onError={() => setImageHasErrored(true)}
                src={product.image}
                alt={product.name}
                data-test-id="image"
              />
            </div>
          )}
```

Jwhiles (approved):
changes look good!

packages/ecommerce-app-base/src/Editor/SortableListItem.tsx
OriginalEXE: A great suggestion, implemented :)

PR #546, opened 2021-10-15
chore: [] Add global prettier config, run it on commit, and prettify the codebase

Jwhiles:
So we had lots of prettier config files in this code base, but no enforcement that code matched the rules, and nothing to cause prettier to be run other than the whims of whoever is working on a feature.

To give us more consistent code styles, and to boost my LOC contributed stats - I've added a global prettier config, and then run prettier on the whole codebase.

Reviewers, I suggest looking through this commit by commit - as the otherwise there is a lot of noise from reformatting. 

If we are happy with these changes - we should still hold off on merging this until the [Gatsby changes](https://github.com/contentful/apps/pull/543) are merged - as I don't want to cause the worst merge conflict of all time.


andipaetzold:
@Jwhiles Do you want to remove prettier dev dependencies from all apps/packages as well?

PR #554, opened 2021-10-19
Add duration to Mux asset json

evanfreeze:
## Description
Adds "duration" to the Mux asset JSON that comes back from Contentful for videos that are uploaded via the Mux app. This is purely additive, so should be backwards compatible.

This is helpful when you want to display a thumbnail with a duration for a Mux asset, for instance, but aren't actually playing the video (where you would get duration for free in the video player). 

I ran the app with these changes on localhost in our Contentful instance and opened a video that was uploaded with the production version of the Mux app. Here was the resulting JSON (temporarily switched the view type of this field to raw JSON to take this screenshot)...
Screen Shot 2021-10-19 at 4 43 15 PM

## List of changes
- added `duration` as a field on the `MuxContentfulObject` interface
- sets the duration for newly-uploaded videos when setting the app field initially
- when checking an asset's validity, syncs the duration to the field if it differs from what's already there

### DIFF
--- a/apps/mux/CHANGELOG.md
+++ b/apps/mux/CHANGELOG.md
@@ -16,14 +16,6 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 **Note:** Version bump only for package mux-contentful-uploader
 
 
-- Adds additional "duration" field to stored data and response from Contentful
-
-
-## [1.5.4](https://github.com/contentful/apps/compare/mux-contentful-uploader@1.5.3...mux-contentful-uploader@1.5.4) (2021-10-18)
-
-**Note:** Version bump only for package mux-contentful-uploader
-
-
 
 
 
--- a/apps/mux/CHANGELOG.md
+++ b/apps/mux/CHANGELOG.md
@@ -16,606 +16,315 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 **Note:** Version bump only for package mux-contentful-uploader
 
 
+- Adds additional "duration" field to stored data and response from Contentful
 
+## [1.5.4](https://github.com/contentful/apps/compare/mux-contentful-uploader@1.5.3...mux-contentful-uploader@1.5.4) (2021-10-18)
 
+**Note:** Version bump only for package mux-contentful-uploader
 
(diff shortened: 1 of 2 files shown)


dylanjha:
Good change! 🙌🏼  this seems like an oversight that the `duration` value wasn't populated from the beginning (especially since we have things like `"ratio"` in there.

Overall, my feedback is:

The auto-syncing is kinda nice, but also feels like an anti-pattern to be writing data simply when the field is rendered. 

I'm thinking of a situation like this:

- A developer is building an app with Contentful + Mux
- They see the `duration` field on some new videos they upload and they (understandably) assume it exists for all their historical videos
- They write code that relies on the `duration` field, it works great with all their new videos.
- But then their application is busted on some old videos that lack the `duration` field.
- If they go lookup that busted video in the Contentful dashboard then suddenly it gets the duration field and now that video is no longer broken in their application.
- This would be incredibly confusing to debug. Basically every video they view in the Contentful UI would be magically "fixed"

I don't think the above scenario is unreasonably far-fetched.

The alternative suggestion of having a "Sync" button is kind of unfortunate, but probably the better option of the two ☹️ . If the sync button only shows when the duration is lacking, maybe we put some kind of tooltip or notice that has some copy like "This record is lacking a "duration" field, click here to sync the duration".

I think the ideal scenario would be to have some kind of script the user could run that backfills duration on all their videos. We'd have to run that script server-side, though. It would be error-prone and pretty unreasonable to run on the client, especially on a large library of content. Does Contentful allow Apps to do that sort of thing? Thinking if there's a way that a user could trigger the sync as a one-time action across all their Videos. Would something like that be possible?

theelk801:
Hi, I just found out that this repo existed after contacting Mux support for a similar request. I have basically the same situation except that I'd like to get the width and height of a video rather than the duration, but it seems as though it's essentially the same process to add it so I figured I'd mention it here rather than open my own PR.

dylanjha:
>  I have basically the same situation except that I'd like to get the width and height of a video rather than the duration, but it seems as though it's essentially the same process to add it so I figured I'd mention it here rather than open my own PR.

@theelk801 curious, what is your use case for this? Any chance you're actually trying to get the aspect ratio, which is pretty common? Or are you really after the specific width and height?

theelk801:
I need the actual width and height for sizing reasons, the aspect ratio unfortunately isn't enough

andipaetzold:
I'll close this as the `duration` is being added in https://github.com/contentful/apps/pull/1249

PR #569, opened 2021-10-28
fix: [] preview button incorrectly hidden

Jwhiles:
Before the recent gatsby update, if the user didn't have a webhook URL in their installation params we would still show them the preview button. This causes issues for some users who just have a CMS preview URL.

This PR makes it so we only hide the button if there is a content Sync URL but no valid preview URL. This hopefully means that the behaviour for customers with older configs should not be impacted.

### DIFF
--- a/apps/gatsby/src/Sidebar.js
+++ b/apps/gatsby/src/Sidebar.js
@@ -261,7 +261,7 @@ export default class Sidebar extends React.Component {
 
     previewUrl = this.getPreviewUrl();
 
-    if (contentSyncUrl && !(webhookUrl || previewWebhookUrl)) {
+    if (contentSyncUrl && !previewWebhookUrl) {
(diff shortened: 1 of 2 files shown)


andipaetzold (approved):

veryspry (approved):

TylerBarnes (approved):

fedorenkodev:
Hey guys, when I have set all settings for the app(Preview Webhook, Content Sync, CMS Preview) I get the wrong error message in the sidebar - "Please add a Preview Webhook URL to your Gatsby App settings.". I resolved the issue only by clearing the Content Sync field. I believe it's wrong behavior, doesn't it?

TylerBarnes:
@fedorenkodev It is indeed, @veryspry is making a PR right now

PR #576, opened 2021-11-02
fix: bug where preview button wouldn't show up for users

veryspry:
Due to some confusing naming that is in place to ensure backwards compatibility for Gatsby users who have not yet moved to Content Sync preview, we accidentally introduced a regression through #569. This happened because we had references to a configuration value called `previewWebookUrl` which was essentially unused. The correct name for the preview webhook url in the config object is `webhookUrl` (this is the confusing backwards compatibility naming.)

This bug is causing a missing url warning to display even if the "Preview Url" field in the config is filled out since that field is mapped to the `webhookUrl` and not the `previewWebhookUrl` config value.

contentful

This PR fixes that bug and also removes references to the superfluous config item `previewWebhookUrl`. @Jwhiles sorry for any confusion here and with your recent fix! I should have caught that in your PR but had already forgotten about the historical context around `webhookUrl` (naming is important :sweat_smile:)  and I'm hoping that these changes help us to avoid a similar mistake in the future.

### DIFF
--- a/apps/gatsby/src/AppConfig/AppConfig.js
+++ b/apps/gatsby/src/AppConfig/AppConfig.js
@@ -67,6 +67,12 @@ export class AppConfig extends React.Component {
     selectorType: false,
     urlConstructors: [],
     previewUrl: '',
+    /**
+     * The Preview Webhook Url was historically named "webhookUrl" which is a *bit* confusing
+     * since you might expect it to be called "previewWebhookUrl" In order to keep backwards
+     * compatibility for users not yet using Content Sync, we must continue using the
+     * variable / keyname "webhookUrl"
+     */
     webhookUrl: '',
     contentSyncUrl: '',
     authToken: '',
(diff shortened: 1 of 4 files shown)


apps/gatsby/src/AppConfig/AppConfig.js
```diff
@@ -114,22 +119,19 @@ export class AppConfig extends React.Component {
...
       contentSyncUrl,
       webhookUrl,
-      previewWebhookUrl,
       authToken,
     } = this.state;
 
     const validPreview = !previewUrl || isValidUrl(previewUrl);
     const validContentSync = !contentSyncUrl || isValidUrl(contentSyncUrl);
     const validWebhook = !webhookUrl || isValidUrl(webhookUrl);
-    const validPreviewWebhook = !previewWebhookUrl || isValidUrl(previewWebhookUrl);
```
veryspry, line 124: As it turns out my fix in #570 was not needed and actually never added in the first place since we didn't need the variable anywhere. What we really needed was to remove `previewWebhookUrl` altogether.

KyleAMathews (approved):

zcei (approved):
Unblocking Gatsby during Berlin night hours, the change only affect their app's subfolder.

TylerBarnes:
Thank you @zcei! We appreciate you helping us out 😄

veryspry:
Thanks for merging this after hours 😀

PR #623, opened 2021-12-03
fix(jira): start auto resizer when not authenticated

andipaetzold:
When the user is not authenticated with the jira app installed, we currently leave quite some space under the button as we do not resize the iframe:

image

We only start the auto resizer once the user has authenticated. With this PR we always start the resizer when the app is rendered in the sidebar.

image


colomolo (approved):
LGTM

loweisz (approved):

PR #639, opened 2021-12-16
fix: [EXT-3369] localize name string

loweisz:
The code for displaying the name in the product list was hardcoded with `name.en`. 

I added a helper that localizes it with the selected searchLanguage and defaults to `en`


kdamball (approved):
🎉  cheers for finding the bug

PR #728, opened 2022-02-01
fix(Gatsby): Check the updated at time repeatedly to wait for a save to happen

TylerBarnes:
This PR fixes a bug where we would open a new preview tab and then refresh that tab after a few seconds once Contentful saved entry data. In some cases the refresh would happen after the user was already redirected to the frontend and it looks really buggy.

In this PR we now wait for 5 seconds, checking every 100ms to see if a save happened. As soon as a save happens we open the preview window 1 time and don't need to refresh it.
In the event that somehow Contentful's entry save took longer than 5 seconds (maybe a slow network or a busy day) the old logic is still there to wait an additional 5 seconds after opening the preview window to see if a save happened afterwards and we still need to refresh the window.

In the event that a save happened earlier than that (perhaps hours even), the user has to wait 5 seconds for the preview to open since we don't know if a save is about to happen or not. For now that's the tradeoff we have to make until there's a Contentful API where we can save entry data from our custom button. We need to wait for the latest data to be saved as we send some entry data as part of the url opened by the "Open Preview" button. If we don't wait for a save the url will have outdated data and Gatsby's preview service wont work properly.

### DIFF
--- a/apps/gatsby/src/Sidebar.js
+++ b/apps/gatsby/src/Sidebar.js
@@ -243,8 +243,8 @@ export default class Sidebar extends React.Component {
       let totalLoopsWaited = 0;
       while (waitForSave) {
         contentfulSavedRecently = this.entryWasUpdatedInLastXSeconds(5);
-        // 60 loops === 6 seconds (100 ms * 60)
-        if (totalLoopsWaited >= 60 || contentfulSavedRecently) {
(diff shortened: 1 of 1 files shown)


veryspry (approved):
My :white_check_mark: doesn't really mean anything in this repo, but this makes sense to me

Jwhiles (approved):
Looks good

PR #735, opened 2022-02-04
feat: [EXT-3435]

Jwhiles:


.github/dependabot.yml
```diff
@@ -12,3 +12,15 @@ updates:
     prefix: "fix"
     prefix-development: "chore"
     include: "scope"
+
+- package-ecosystem: npm
```
Jwhiles, line 16: I think we will have to add each new template to dependabot individually, unless we add them as packages to the top level lerna config - or add another lerna config

.circleci/config.yml
```diff
@@ -56,6 +56,9 @@ jobs:
       - run:
           name: Test
           command: npm run test
+      - run:
```
Jwhiles, line 59: As we add more templates, I think we can make a script here to loop through them and do this for each.

examples/templates/typescript-template/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
```
andipaetzold, line 2: Maybe we can use `@tsconfig/create-react-app` here, so we don't have to mess around with the config ourselves.

https://www.npmjs.com/package/@tsconfig/create-react-app

But as this is just a copy, feel free to ignore this comment

andipaetzold (commented):
I assume this is just a copy from create-contentful-app, so nothing to worry from the implementation side.

I am just not sure in which folder the templates should live. Nesting under `examples` feels a little strange. Also having an additional `[...]-template` in the folder name seems verbose.
I'd rather put it directly `/examples/typescript-template` or `/templates/typescript`. The latter would be less verbose when specifying the template in CCA.

Jwhiles:
> I assume this is just a copy from create-contentful-app, so nothing to worry from the implementation side.
> 
> I am just not sure in which folder the templates should live. Nesting under `examples` feels a little strange. Also having an additional `[...]-template` in the folder name seems verbose. I'd rather put it directly `/examples/typescript-template` or `/templates/typescript`. The latter would be less verbose when specifying the template in CCA.
> 
> wdyt?

I agree about the path, I read in the acceptance criteria that they should be in the examples folder, but let's move to just /templates

.circleci/config.yml
```diff
@@ -57,8 +57,8 @@ jobs:
           name: Test
           command: npm run test
       - run:
-          name: Build template-typescript
-          command: cd examples/templates/typescript-template && npm i && npm run build
+          name: Build TypeScript template
+          command: cd templates/typescript && npm i && npm run build
```
andipaetzold, line 61: Should we also run the tests here? If not, you can merge.

andipaetzold (approved):

PR #748, opened 2022-02-08
feat(JSTemplate): Add js template for CCA [EXT-3407]

kdamball:
Like it says in the title - support JS template


Jwhiles (changes requested):
Can we also add the new template to the dependabot config. Danke :)

Jwhiles (approved):

PR #890, opened 2022-02-24
feat: Migrate packages to f36 v4

andipaetzold:
WIP

### DIFF
--- a/packages/dam-app-base/src/AppConfig/FieldSelector.tsx
+++ b/packages/dam-app-base/src/AppConfig/FieldSelector.tsx
@@ -1,9 +1,8 @@
+import { Checkbox, Form, Subheading } from '@contentful/f36-components';
+import tokens from '@contentful/f36-tokens';
+import { css } from 'emotion';
 import * as React from 'react';
-import tokens from '@contentful/forma-36-tokens';
-import { css } from '@emotion/css';
-import { Form, Subheading, CheckboxField, Typography } from '@contentful/forma-36-react-components';
-
-import { ContentType, CompatibleFields, SelectedFields } from './fields';
+import { CompatibleFields, ContentType, SelectedFields } from './fields';
 
 interface Props {
   contentTypes: ContentType[];
@@ -33,7 +32,7 @@ export default class FieldSelector extends React.Component<Props> {
     const { compatibleFields, contentTypes, selectedFields } = this.props;
 
     return (
-      <Typography>
+      <React.Fragment>
         {contentTypes.map((ct) => {
           const fields = compatibleFields[ct.sys.id];
           return (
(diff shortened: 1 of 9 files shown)


packages/ecommerce-app-base/src/AppConfig/AppConfig.tsx
```diff
@@ -209,79 +209,78 @@ export default class AppConfig extends React.Component<Props, State> {
...
+                    : `https://app.contentful.com/spaces/${space}/environments/${environment}/content_types`
+                }
+              >
+                content model
+              </TextLink>{' '}
+              and assign it to the app from this screen.
+            </Note>
+          </>
+        )}
+        {appReady === true ? (
```
colomolo, line 273:
```suggestion
        {appReady && (
```

packages/dam-app-base/src/AppConfig/FieldSelector.tsx
```diff
@@ -33,28 +32,29 @@ export default class FieldSelector extends React.Component<Props> {
     const { compatibleFields, contentTypes, selectedFields } = this.props;
 
     return (
-      <Typography>
+      <React.Fragment>
```
colomolo, line 35: Looks like this fragment is not needed

packages/ecommerce-app-base/src/SkuPicker/ProductSelectionList/ProductSelectionListItem.tsx
```diff
@@ -97,7 +100,7 @@ export const ProductSelectionListItem = (props: Props) => {
   const productIsMissing = !product.name;
 
   return (
-    <div className={styles.productWrapper}>
+    <div className={styles.productWrapper} data-testid="ProductSelectionListItem">
```
colomolo, line 103: What confuses me is that in some places we have a `data-testid` and in other - `data-test-id`. Why is that? Maybe worth changing that?

packages/ecommerce-app-base/src/Editor/SortableListItem.tsx
```diff
@@ -125,17 +124,24 @@ export const SortableListItem = SortableElement<Props>(
     const productIsMissing = !product.name;
 
     return (
-      <Card className={styles.card}>
+      <Card
+        data-testid="SortableListItem"
```
colomolo, line 128: nit: sometimes it's pascal case and sometimes kebab. Worth unifying?

andipaetzold: I use pascal case when the property is on the top element of a component. That way the test id basically references the full component.

Gonna make it consistent anyway

packages/ecommerce-app-base/src/SkuPicker/ProductSelectionList/ProductSelectionListItem.tsx
andipaetzold: `data-testid` is the default by testing library and `data-test-id` is the forma 36 override (https://github.com/contentful/forma-36/blob/2b6d93caaf3f870919bb60745634462e3456a615/scripts/test/setupTests.js).

packages/dam-app-base/src/AppConfig/FieldSelector.tsx
andipaetzold: It is. An array is not a valid React node

colomolo: oh, you're right.

colomolo (approved):
It's hard to check all 3,5k new lines of code, but I can't see any problems with migration. Great job!

andipaetzold:
> It's hard to check all 3,5k new lines of code, but I can't see any problems with migration. Great job!

I don't expect you to review lock files 😉

fabe (approved):

PR #891, opened 2022-02-25
feat: Netlify App Event subscription

colomolo:
Fixed version of [this](https://github.com/contentful/apps/pull/865) PR.


colomolo:
> It looks like this introduces a bug in the config screen. There is an empty item at the start of the list, which no longer makes really makes sense. Hitting edit causes the app to crash.
> 
> This is on a fresh config. image
> 
> I think you probably need to look into how the original state is created [here](https://github.com/contentful/apps/blob/24463bc12a3336d2a652492c459c739bdd0ecf80/apps/netlify/frontend/src/app/index.js#L56-L58)

Fixed this here in a follow-up PR: https://github.com/contentful/apps/pull/892

shikaan (commented):
Not sure if related to this change (I would not say so), but the "Open site" link in the sidebar seems to be broken.

colomolo:
> Not sure if related to this change (I would not say so), but the "Open site" link in the sidebar seems to be broken.

Actually related. That's because of Forma v4 upgrade. Fixed it.

shikaan (commented):
Looks good to me. A couple of remarks:

* I wasn't able to test the whole flow without reading the code. We should probably make this is easier documenting (at least) what environment variables are needed to run the app

* Upon refreshing the config location, I get logged out. Is it intended? Should we maybe persist the token locally?

If we don't want to do these here, let's look at it before we merge the `netlify-events` branch

Other than that it looks good and it works 🎉

shikaan (approved):
Sorry, meant to approve

PR #927, opened 2022-03-03
chore: Link to relevant docs guide section

colomolo:


stefanjudis (approved):
That's great. is there a chance to add a screenshot of the `localhost` screen to the docs? I think it may be beneficial (only nice to have though) if folks open the URL, see exactly what they're facing on localhost to then continue.

PR #935, opened 2022-03-04
chore: [EXT-3585] Update Cloudinary logo

colomolo:


andipaetzold (approved):
You can remove the `id` and `data-name` props if you want. Other than that it's fine.

PR #982, opened 2022-03-18
refactor: [] Hooks to examples ts

loweisz:
Change the template of the typescript example to include the hooks from the `react-apps-toolkit`. 

- I also added the `App.tsx` file to be wrapped with the `SDKProvider` to not call the `init` method twice. This also allows to clean up the location mapping inside this component and make it a bit more straightforward. 
- To make use of the SDK inside the components I added the id to the text that has been rendered for all locations.
- However I don't want to make an unexpected API call, therefore I left the CMA unused and added a code comment to it.


examples/typescript/src/locations/Sidebar.tsx
```diff
@@ -1,15 +1,15 @@
...
 
-interface SidebarProps {
-  sdk: SidebarExtensionSDK;
-  cma: PlainClientAPI;
-}
+const Sidebar = () => {
+  const sdk = useSDK<SidebarExtensionSDK>()
+  /* You can remove the next line as soon as you use the CMA, or, if not needed, simply remove the hook. */
+  /*  eslint-disable @typescript-eslint/no-unused-vars */
+  const cma = useCMA();
```
Jwhiles, line 10: This pattern of having to disable some lint rule in every file feels a bit weird. Could we maybe just have the  `const cma = useCMA();` itself commented out?

andipaetzold:
Please revert `package-lock.json` to v1

Let's upgrade all lock files at once at some future point.

examples/typescript/tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx"
+    "jsx": "react"
```
andipaetzold, line 21: Any reason for this change? Shall we maybe switch to https://www.npmjs.com/package/@tsconfig/create-react-app for even less code?

examples/typescript/src/App.tsx
```diff
@@ -0,0 +1,35 @@
...
+import Page from "./locations/Page";
+import {useSDK} from "@contentful/react-apps-toolkit";
+
+const ComponentLocationSettings = {
+  [locations.LOCATION_APP_CONFIG]: <ConfigScreen />,
+  [locations.LOCATION_ENTRY_FIELD]: <Field />,
+  [locations.LOCATION_ENTRY_EDITOR]: <EntryEditor />,
+  [locations.LOCATION_DIALOG]: <Dialog />,
+  [locations.LOCATION_ENTRY_SIDEBAR]: <Sidebar />,
+  [locations.LOCATION_PAGE]: <Page />,
```
andipaetzold, line 17: Did you think about mapping to the components instead of already initializing the elements?

```typescript
const ComponentLocationSettings = {
  [locations.LOCATION_APP_CONFIG]: ConfigScreen,
  [locations.LOCATION_ENTRY_FIELD]: Field,
  [locations.LOCATION_ENTRY_EDITOR]: EntryEditor,
  [locations.LOCATION_DIALOG]: Dialog,
  [locations.LOCATION_ENTRY_SIDEBAR]: Sidebar,
  [locations.LOCATION_PAGE]: Page,
}
```

The behavior will be mostly the same, but the difference is that `React.createElement` is only called when the component is actually used. I'd find it the cleaner approach.

This needs a minor change below as well.

examples/typescript/src/App.tsx
```diff
@@ -0,0 +1,35 @@
...
+const App = () => {
+  const sdk = useSDK()
+
+  const RenderedComponent = useMemo(() => {
+    for (const [location, component] of Object.entries(ComponentLocationSettings)) {
+      if (sdk.location.is(location)) {
+        return component;
+      }
+    }
+  }, [sdk.location])
```
andipaetzold, line 29: Let's make sure we can get rid of this loop in the future - then everything is shiny :)

examples/typescript/src/locations/Sidebar.tsx
andipaetzold: +1, I also think it's not ideal to sip a `eslint-disable` with the template. But having a commented `cma` constant is also strange...

If we leave it, it should be `eslint-disable-next-line`.

andipaetzold (changes requested):

Jwhiles (approved):

examples/typescript/tsconfig.json
```diff
@@ -1,24 +1,12 @@
 {
+  "extends": "@tsconfig/create-react-app/tsconfig.json",
   "compilerOptions": {
     "target": "es5",
     "lib": [
       "dom",
       "dom.iterable",
       "esnext"
     ],
```
andipaetzold, line 9: Nit: we could also drop all that.

andipaetzold (approved):

PR #1095, opened 2022-04-12
DAM App [EXT-3561]

andipaetzold:


examples/dam-app/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+const CTA = 'Sample DAM Demo App';
+const FIELDS_TO_PERSIST = ['id', 'name', 'url'];
+
+setup({
+  cta: CTA,
+  name: 'Contentful DAM Demo App',
+  logo: 'https://images.ctfassets.net/fo9twyrwpveg/6eVeSgMr2EsEGiGc208c6M/f6d9ff47d8d26b3b238c6272a40d3a99/contentful-logo.png',
+  color: '#036FE3',
+  description:
+    'This is a sample Application to demonstrate how to make a custom DAM application on top of Contentful.',
```
loweisz, line 17: Maybe adding a "DAM (Digital Asset Management)" would not hurt here wdyt?

examples/dam-app/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+  const { result: damData } = useAsync(async () => {
+    const response = await fetch(
+      `/dam_api_response.json?api_key=${apiKey}&project_id=${projectId}`
+    );
+    return response.json();
+  }, [apiKey, projectId]);
+
+  if (!damData) {
+    return <div>Please wait</div>;
+  }
```
loweisz, line 54: Not the biggest fan of assuming a loading state when there is no data. 😬

examples/dam-app/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+function DialogLocation({ sdk }) {
+  const apiKey = sdk.parameters.installation.apiKey;
+  const projectId = sdk.parameters.installation.projectId;
+
+  const { result: damData } = useAsync(async () => {
+    const response = await fetch(
+      `/dam_api_response.json?api_key=${apiKey}&project_id=${projectId}`
+    );
+    return response.json();
+  }, [apiKey, projectId]);
```
loweisz, line 50: Do we really want to be so opinionated here and adding this whole library here? Why not just do a simple fetch in a `useEffect` ?

examples/dam-app/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+    shouldCloseOnEscapePress: true,
+    width: 400,
+    allowHeightOverflow: true,
+  });
+
+  if (!Array.isArray(result)) {
+    return [];
+  }
+
+  return result.map((asset) => pick(asset, FIELDS_TO_PERSIST));
```
loweisz, line 97: Maybe we can save another dependency here: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_pick

andipaetzold: Yes, but I am lazy and `lodash-es` gets tree shaken 😆

examples/dam-app/src/index.js
andipaetzold: Will add

examples/dam-app/src/index.js
andipaetzold: I don't use `useEffect` for async methods as it doesn't handle unmounting and then tries to update a state of an unmounted component.
But for this example, I'll drop it.

loweisz: I agree, but maybe for the sake of the keeping the example simple we can be okay with that

examples/dam-app/src/index.js
andipaetzold: Yes. Gonna remove it.

examples/dam-app/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "name": "@contentful/dam-app",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@contentful/dam-app-base": "1.3.62",
+    "@contentful/f36-components": "4.4.1",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
+    "react-scripts": "5.0.0"
+  },
```
loweisz, line 11: so_good

loweisz (approved):

PR #1096, opened 2022-04-12
feat: [] add vue as template to examples

loweisz:
This adds a vue starter template to the examples. 

It was bootstrapped with the VueCLi and uses Vue3 and the composition API + Typescript according to https://vuejs.org/guide/typescript/composition-api.html


Jwhiles (approved):

loweisz:
> This example doesn't have a localhost warning similar to what we have in the [react example](https://github.com/contentful/apps/blob/master/examples/typescript/src/components/LocalhostWarning.tsx) - might be nice to add.

Good point, I've added it 👍

PR #1097, opened 2022-04-12
E-Commerce App  [EXT-3561]

andipaetzold:


shikaan (approved):
Looks good to me. I wonder if we want to shape the `src` folder as in the cca templates to make it more approachable?

andipaetzold:
@shikaan The app bases work quite differently than regular apps as they hide quite some complexity behind the `setup` method. Yes, we can pull the Dialog location into a separate file, but I don't know whether this makes things more approachable.
I actually like that these apps are only a single file.

PR #1119, opened 2022-04-14
feat: entry list location advanced example

bgutsol:


examples/references-count/README.md
```diff
@@ -0,0 +1,17 @@
+# Entry List location example
+This example shows how to build an advanced app with an `EntryList` location. It counts a number of references for the entries.
+
+## How to use
+
+[![Install to Contentful](https://www.ctfstatic.com/button/install-small.svg)](https://app.contentful.com/deeplink?link=apps&id=50jZ722cWV7JOVj4SW10jI)
+
```
loweisz, line 7: Does that make sense in the examples folder? People will probably create their own app out of this

examples/references-count/src/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+  </head>
+  <body>
+    <script type="text/javascript" src="./index.ts"></script>
```
loweisz, line 8: Does it make sense to show a small info text here to indicate that this will never show up ?

bgutsol: I don't know, does it? 
Do you mean the comment or actual DOM node, like `<p>` with the message?

loweisz: Yes, maybe just a small text saying that this app should only be used within contentful in the entry list location. 
Just for people that are running it locally and want to visit it in the browser, then they know it is not broken.

bgutsol: Agree, done ✅

loweisz (approved):

PR #1228, opened 2022-05-04
fix: [] v4 update fixes

loweisz:
Fixing two bugs from the forma upgrade. In the ecomerce-app-base the card was added a `display: flex` to aligh the items. 
However in formav4 the styles added to the `Card` component are attached to the outer component and therefore the flex was not working anymore for the card content. : 

Here you can see the difference (the upper one is the fixed version): 

Screen Shot 2022-05-04 at 15 30 13

To fix this I added an inner element with the layout styles. 

The other thing was that in the config screen the radio buttons were not aligned correctly: 
Screen Shot 2022-05-04 at 14 16 31

### DIFF
--- a/packages/ecommerce-app-base/src/AppConfig/FieldSelector.tsx
+++ b/packages/ecommerce-app-base/src/AppConfig/FieldSelector.tsx
@@ -20,6 +20,13 @@ interface State {
   changedSkuTypes: Record<string, Record<string, boolean>>;
 }
 
+const styles = {
+  helpText: css({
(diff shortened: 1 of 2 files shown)


andipaetzold (changes requested):
This is how things looked before the v4 update:

image
image


I am very sure I adjusted it to look the same, but maybe I accidentally dropped some css... The dam-app-base package includes the adjusted styles that the field apps looks alright again.

shikaan (approved):

andipaetzold (approved):

PR #1233, opened 2022-05-06
fix: EXT-3750 introduce scope for token requests

shikaan:


kdamball (commented):
In general, will this affect only product selection or even the current data? In other words, is it breaking for everyone or just when selecting products.

shikaan:
> In general, will this affect only product selection or even the current data? In other words, is it breaking for everyone or just when selecting products.

There has been a breaking change in the API, so - while this is technically a breaking change, because it requires a parameter that did not exist before - it won't break anything which is not already broken on the server side

kdamball (approved):

PR #1234, opened 2022-05-06
fix: increase minimum forma 36 version

andipaetzold:
In earlier versions of forma 36, the auto resizer does not work because of `min-height` being part of the global styles.

This PR increases the minimum version to include the fix (https://github.com/contentful/forma-36/pull/1937)

### DIFF
--- a/packages/ecommerce-app-base/package.json
+++ b/packages/ecommerce-app-base/package.json
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "@contentful/app-sdk": "^4.0.0",
-    "@contentful/f36-components": "^4.0.0",
+    "@contentful/f36-components": "^4.0.42",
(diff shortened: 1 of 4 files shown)


packages/ecommerce-app-base/package.json
```diff
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "@contentful/app-sdk": "^4.0.0",
-    "@contentful/f36-components": "^4.0.0",
```
loweisz, line 38: Shall we not go to the latest?

andipaetzold: The goal is to maximize the version range. 4.0.42 is the first version that is working with the app-base.

loweisz (approved):

PR #1281, opened 2022-05-24
feat: migrate commercetools to app-base [EXT-3718]

andipaetzold:
Rewrite of the commercetools app using `ecommerce-app-base`


kdamball (commented):
I might have set up my acct wrong but i get this notification error whenever I load the modal first time. I think we send a search query on initial load which is always 400 in the shape of: https://api.europe-west1.gcp.commercetools.com/<acct_id>/product-projections/search?filter.query=variants.sku%3A

Screenshot 2022-06-09 at 11 01 57

martin3walker (commented):
Screen Shot 2022-06-17 at 3 42 58 PM

Every time I add a product, I'm told that its missing. Maybe this a matter of my setup being wrong?

Also is the product list supposed to provide only skus and no names?

kdamball:
Same issue reported by @martin3walker where selecting a product does not update the Entry Card. Still marked as missing even on an existing product

kdamball (commented):
Same issue reported by @martin3walker where selecting a product does not update the Entry Card. Still marked as missing even on an existing product. Categories search seem to work but no previews (not sure if this is the "breaking change")

martin3walker (approved):

kdamball (approved):

PR #1282, opened 2022-05-24
fix: E-commerce app base UI fixes

andipaetzold:
* Made "Load more" button full width
* Adjusted height of save button

## Before
SCR-20220524-pc2

SCR-20220524-pap

## After
SCR-20220524-pdk
SCR-20220524-pb4


loweisz (approved):
💯

PR #1289, opened 2022-05-30
feat: add `hideSearch` option to SkuPicker

andipaetzold:
Some ecommerce providers (e.g. commercetools) do not provide a search for all available sku types. Therefore, we need to hide the search here and only provide a list

unblocks https://github.com/contentful/apps/pull/1281


kdamball:
I think if default is to show search, then the option should be `hideSearch`.

PR #1412, opened 2022-06-29
Player Updates

blairliikala:


apps/mux/src/index.tsx
```diff
@@ -109,7 +109,7 @@ export class App extends React.Component<AppProps, AppState> {
   };
 
   async componentDidMount() {
-    this.props.sdk.window.startAutoResizer();
+    this.props.sdk.window.startAutoResizer({ absoluteElements: true });
```
andipaetzold, line 112: You will need to update the App SDK to at least v4.6.0

andipaetzold (commented):
Please take a look 👀  at the failing tests

blairliikala:
I updated the SDK to 4.8.1 and took out the `this.props.sdk.user.sys.id` since it was erroring.  The local test is passing, not quite sure what's up here?

PR #1416, opened 2022-06-30
fix: correctly reset fields, player adjustments, and iframe auto sizer changes

blairliikala:
Additionally updated Player, and enabled audio-only.


blairliikala:
Will save the startAutoResizer bit and all the other package updates for another day :)

PR #1455, opened 2022-07-08
Vite/Vitest React example

andipaetzold:
As the title says.

Same as the default typescript example just with vite & vitest instead of react-scripts and jest


loweisz (approved):
💯 💯

PR #1465, opened 2022-07-11
fix(deps): revert pubnub from 7.2.0 to 4.24.3

shikaan:
Reverts contentful/apps#1460


kdamball:
That's a massive jump in versions. What broke?

kdamball (approved):

PR #1534, opened 2022-07-21
feat: Next.js template

andipaetzold:
Adding a Next.js template


examples/nextjs/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint",
+    "test": "jest --watch",
+    "test:ci": "jest --ci"
+  },
```
loweisz, line 12: Shall we include the `"create-app-definition"` `"upload"` and `"upload-ci"` as we have in the other templates?

examples/nextjs/pages/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { useCallback, useState, useEffect } from 'react';
```
loweisz, line 1: Why are all components under `/pages/locations`? wouldn't that mean theoretically there is now a route to `/pages/locations/ConfigScreen` ?

examples/nextjs/README.md
```diff
@@ -0,0 +1,30 @@
+This project was bootstrapped with [Create Contentful App](https://github.com/contentful/create-contentful-app).
+
+## Config Location
+
+Due to a bug in Next.js the config location doesn't load in development mode. The bug has been reported in Next.js' GitHub repository: https://github.com/vercel/next.js/issues/18028.
```
loweisz, line 5:
```suggestion
Due to a bug in Next.js, the config location doesn't load in development mode. The bug has been reported in Next.js' GitHub repository: https://github.com/vercel/next.js/issues/18028.
```

examples/nextjs/package.json
andipaetzold: Won't work here. This is SSR and can't be hosted statically

examples/nextjs/pages/locations/ConfigScreen.tsx
andipaetzold: I don't understand? This is just structured the same as the React template.

examples/nextjs/package.json
loweisz: True, but the `create-app-definition` will work or not?

examples/nextjs/pages/locations/ConfigScreen.tsx
loweisz: In the other templates we only have `src/locations/<location>.tsx`, but here it is `src/pages/locations/<location>.tsx`, I just wonder why we need that `pages` folder as we don't have routing

andipaetzold: tbh, I have no experience about Next.js and just put the files into the folder that was created via create-next

loweisz: The idea is that if you put things in the `pages` folder  automatically creates the route like that.
https://nextjs.org/docs/basic-features/pages

That means we technically now have routes like `/pages/locations/ConfigScreen` available which we don't need. 
So having the same structure as the other js/ts templates should be working here as well

andipaetzold: Okay, gonna move it. Thanks

examples/nextjs/package.json
andipaetzold: yes, added that

examples/nextjs/pages/locations/ConfigScreen.tsx
andipaetzold: done

loweisz (approved):
🔥

PR #1544, opened 2022-07-25
chore: [EXT-3800] shopify API upgrade to 2022-04 version

vikaskumr:
Purpose of PR

To upgrade to new version stable version of Shopify API i.e. 2022-04.

Accepted criteria 
1). Support backward compatibility
2). Make sure future version change don't break anything
3). Test all the new features with new API version

Approach
Convert all the `gid://shopify/ProductVariant/41` format to base64 while updating the entry to database.
Convert the base64 id back to `gid://shopify/ProductVariant/41` while sending it as a parameter to Shopify API for fetching Product details , Product variant and collection.


apps/shopify/src/basePagination.js
```diff
@@ -70,7 +71,10 @@ export default class BasePagination {
    * and now want to render the next page.
    */
   async _fetchNextPage(products) {
-    return (await this.shopifyClient.fetchNextPage(products)).model;
+
+    const response = (await this.shopifyClient.fetchNextPage(products)).model;
+    const nextProducts = response.map((res) => { return { ...res, id: Buffer.from(res.id).toString('base64') } })
+    return nextProducts
```
shikaan, line 77: No idea how shopify ids are built, but you could consider converting the string to utf8, if you're trying to stay safe from a non-strictly binary strings. 

I would anyway use `btoa` here, and not rely on the polyfill for `Buffer` which seems unnecessary

apps/shopify/src/collectionPagination.js
```diff
@@ -8,12 +8,17 @@ const makePagination = async (sdk) => {
...
         first: PER_PAGE,
         sortBy: 'TITLE',
         reverse: true,
         ...(search.length && query),
       });
+
+
+      const collections = response.map((res) => { return { ...res, id: Buffer.from(res.id).toString('base64') } })
+      return collections
+
```
shikaan, line 21: Same as above

apps/shopify/src/productPagination.js
```diff
@@ -8,12 +8,16 @@ const makePagination = async (sdk) => {
...
+      const response = await this.shopifyClient.product.fetchQuery({
         first: PER_PAGE,
         sortBy: 'TITLE',
         reverse: true,
         ...(search.length && query),
       });
+      // converting the 'gid://shopify/ProductVariant/41' to base64 format as API format changed from 2022-04 version
+
+      const products = response.map((res) => { return { ...res, id: Buffer.from(res.id).toString('base64') } })
+      return products
```
shikaan, line 20: Since we seem to be doing the same over and over, should we try extract a util which encapsulates the need of base64 conversion?

apps/shopify/src/constants.js
```diff
@@ -2,6 +2,7 @@
 // returned with this unfortunate title, and there is no other way to check
 // whether the returned variant is the default one or not
 export const DEFAULT_SHOPIFY_VARIANT_TITLE = 'Default Title';
+export const DEFAULT_SHOPIFY_API_VERSION = '2022-04'
```
shikaan, line 5: I have seen this in some urls, but I don't see this passed in the Shopify client constructor. Isn't it needed there?

apps/shopify/src/dataTransformer.js
```diff
@@ -107,25 +109,24 @@ export const productsToVariantsTransformer = (products) =>
...
+        sku: id,
+        displaySKU: sku ? `SKU: ${sku}` : `Product ID: ${id}`,
+        productId: product.id,
+        name: title === DEFAULT_SHOPIFY_VARIANT_TITLE ? product.title : `${product.title} (${title})`,
+        ...(apiEndpoint &&
+          productId && {
+          externalLink: `https://${apiEndpoint}${last(apiEndpoint) === '/' ? '' : '/'
+            }admin/products/${productId}`,
         }),
+      };
```
shikaan, line 131: If possible, I would isolate commits with formatting only: it makes scanning through the changes easier

apps/shopify/src/constants.js
vikaskumr: no it is not needed in constructor as changing the version (2.15.1) in package.json will automatically pick the latest stable version

apps/shopify/src/basePagination.js
vikaskumr: Yes I have changed it and will push it in some time as it is still in progress

vikaskumr:
@shikaan this Pr https://github.com/contentful/apps/pull/1544 is still in draft :sweat_smile: and I knew all these changes.
pushing new changes in some time

apps/shopify/src/basePagination.js
```diff
@@ -70,7 +72,10 @@ export default class BasePagination {
    * and now want to render the next page.
    */
   async _fetchNextPage(products) {
-    return (await this.shopifyClient.fetchNextPage(products)).model;
+
+    const response = (await this.shopifyClient.fetchNextPage(products)).model;
+    const nextProducts = response.map((res) => { return checkAndConvertToBase64(res) })
```
colomolo, line 77: I would rewrite it like this:
```typescript
async _fetchNextPage(products) {
  const nextPage = await this.shopifyClient.fetchNextPage(products);
  return nextPage.model.map((product) => checkAndConvertToBase64(product));
}
```

apps/shopify/src/collectionPagination.js
```diff
@@ -8,12 +9,17 @@ const makePagination = async (sdk) => {
     fetchProducts: async function (search, PER_PAGE) {
       const query = { query: search };
 
-      return await this.shopifyClient.collection.fetchQuery({
+      const response = await this.shopifyClient.collection.fetchQuery({
```
colomolo, line 12: Same here. It's better to replace abstact 'response' with something more literal.

apps/shopify/src/productVariantPagination.js
```diff
@@ -89,20 +90,29 @@ class Pagination {
...
+
+    // converting the 'gid://shopify/ProductVariant/41' to base64 format as API format changed from 2022-04 version
+    const response = await this.shopifyClient.product.fetchQuery({
       first: PER_PAGE,
       sortBy: 'TITLE',
       reverse: true,
       ...(search.length && query),
     });
+
+    const products = response.map((res) => { return checkAndConvertToBase64(res) })
```
colomolo, line 102: Excessive variable and same with response.

apps/shopify/src/productVariantPagination.js
```diff
@@ -89,20 +90,29 @@ class Pagination {
...
   /**
    * This method is used when the user has already fetched a batch of products
    * and now want to render the next page.
    */
   async _fetchNextPage(products) {
-    return (await this.shopifyClient.fetchNextPage(products)).model;
+    // return (await this.shopifyClient.fetchNextPage(products)).model;
+
+    const response = (await this.shopifyClient.fetchNextPage(products)).model;
+    const nextProductVariants = response.map((res) => { return checkAndConvertToBase64(res) })
```
colomolo, line 114: Same here

apps/shopify/src/skuResolvers.js
```diff
@@ -9,6 +9,12 @@ import { productDataTransformer, collectionDataTransformer } from './dataTransfo
 
 import { validateParameters } from '.';
 import { previewsToProductVariants } from './dataTransformer';
+import { DEFAULT_SHOPIFY_API_VERSION } from './constants';
+import isBase64 from './utils/isBase64';
+import btoa from './utils/btoa'
+import atob from './utils/atob'
```
colomolo, line 15: Maybe worth putting these base64-related utils to single module? wdyt?

vikaskumr: makes sense

apps/shopify/src/skuResolvers.js
```diff
@@ -177,6 +182,26 @@ export const makeCollectionSearchResolver = async (sdk) => {
   return (search) => pagination.fetchNext(search);
 };
 
+export const filterValidIds = (skus, skuType) => {
```
andipaetzold, line 185: This filters AND converts the ids. The name of the method is not quite accurate anymore

apps/shopify/src/skuResolvers.js
```diff
@@ -177,6 +182,26 @@ export const makeCollectionSearchResolver = async (sdk) => {
...
+
+  const validIds = skus
+    .map((sku) => {
+      try {
+        // If not valid base64 window.atob will throw
+        const unencodedId = atob(sku);
```
andipaetzold, line 191: `decodedId`

apps/shopify/src/skuResolvers.js
```diff
@@ -9,6 +9,10 @@ import { productDataTransformer, collectionDataTransformer } from './dataTransfo
 
 import { validateParameters } from '.';
 import { previewsToProductVariants } from './dataTransformer';
+import { DEFAULT_SHOPIFY_API_VERSION } from './constants';
+import { isBase64, btoa, atob, checkAndConvertToBase64 } from './utils/base64'
+
+
 
 export async function makeShopifyClient(config) {
```
andipaetzold, line 17: Can we use the api version for the shopify client as well?

andipaetzold (changes requested):
Added a few comments but got a more general question:

Why do we try to detect which format the id has? It can always either be base64 or gid. From the API we always get gid and in the fields we always safe the base64

vikaskumr:
@andipaetzold Thank you for valuable comments. 
Please look it on Monday once again and we can move it to production but we can discuss the approach to release to users also as we may need to be careful still.

apps/shopify/src/skuResolvers.js
```diff
@@ -177,6 +177,24 @@ export const makeCollectionSearchResolver = async (sdk) => {
...
+  const validIds = skus
+    .map((sku) => {
+      try {
+        // If not valid base64 window.atob will throw
+        const decodedId = atob(sku);
+        return { decodedId };
+      } catch (error) {
+        return null;
+      }
+    })
```
andipaetzold, line 190: Let's use `convertBase64ToString` here to make things shorter

apps/shopify/src/skuResolvers.js
```diff
@@ -177,6 +177,24 @@ export const makeCollectionSearchResolver = async (sdk) => {
...
+    .filter((sku) => sku && new RegExp(`^gid.*${skuType}`).test(sku.decodedId))
+    /** changed to unencodedId  as API format changed and in future support for base64 will be removed so 
+        passing the gid based format to the API for fetching the productVariant for particular product */
+    .map(({ decodedId }) => decodedId);
```
andipaetzold, line 194: Why are we mapping to an object and unwrapping it here? Let's directly map to the decoded id

apps/shopify/src/productVariantPagination.js
```diff
@@ -89,20 +90,25 @@ class Pagination {
    */
   async _fetchProducts(search) {
     const query = { query: `variants:['sku:${search}'] OR title:${search}` };
-    return await this.shopifyClient.product.fetchQuery({
+
+    // converting the 'gid://shopify/ProductVariant/41' to base64 format as API format changed from 2022-04 version
```
andipaetzold, line 94: We are not doing that here

apps/shopify/src/utils/base64.js
```diff
@@ -0,0 +1,25 @@
...
+export const convertStringToBase64 = (str) => {
+  try {
+    const undecodedId = window.btoa(str);
+    return undecodedId;
+  } catch (error) {
+    return null;
+  }
+};
+
+export const convertIdToBase64 = (res) => {
```
andipaetzold, line 19: Can we improve that name. We do not really only convert the id. We convert the nested product as well

andipaetzold (changes requested):
Can we explain these comments

> converting to base64 as still storing base64 in db

and

> converting the 'gid://shopify/ProductVariant/41' to base64 format as API format changed from 2022-04 version

in the README and remove them from within the code. I think just having the current state explained once and not commenting 10 places in the code might make things more readable

apps/shopify/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+* Non-encoded object IDs in the GraphQL Storefront API instead of base 64 encoded IDs in the API response
+* (`Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0LzE=` -> `gid://shopify/Product/1`)
+
+  
+# Actions taken to make our app stable for future 
+
+* In order to make app stable and support further versions we are currently converting the non-encoded object IDs(`gid://shopify/Product/1`) to Base64 encoded IDs(`Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0LzE`) and storing it in our database.
+
+* Sending decoded version of Base64 encoded IDs(`gid://shopify/Product/1`) as arguments to fields and mutations to support coming versions as sending Base64 version will be [deprecated](https://shopify.dev/api/examples/object-ids#migrate-your-app) in coming versions.
```
andipaetzold, line 16:
```suggestion
# Persisted Object IDs

The app persists base64 encoded object IDs. These can be decoded by the users e.g. by using [`atob(...)`](https://developer.mozilla.org/en-US/docs/Web/API/atob). A decoded object ID looks the following: `gid://shopify/Product/1`.

Before Shopify API version 2022-04, the Shopify API returned and accepted encoded object IDs. To ensure backwards compatibility, the app keeps storing object IDs in base64 encoded format even though the Shopify API doesn't accept these IDs anymore (see [here](https://shopify.dev/api/release-notes/2022-04#breaking-changes)).
```

apps/shopify/src/dataTransformer.js
```diff
@@ -94,7 +95,9 @@ export const productsToVariantsTransformer = (products) =>
       const variants = product.variants.map((variant) => ({
         ...variant,
         variantSKU: variant.sku,
-        sku: variant.id,
+        id: convertStringToBase64(variant.id),
+        // converting the 'gid://shopify/ProductVariant/41' to base64 format as API format changed from 2022-04 version
```
andipaetzold, line 99: One leftover

apps/shopify/src/utils/base64.js
```diff
@@ -0,0 +1,25 @@
...
+export const convertStringToBase64 = (str) => {
+  try {
+    const undecodedId = window.btoa(str);
+    return undecodedId;
+  } catch (error) {
+    return null;
+  }
+};
+
+export const convertToBase64 = (res) => {
```
andipaetzold, line 19: Still not a good name. Without typescript it is very unclear what this does. How about `convertProductToBase64` or `convertCollectionToBase64`.

Can we maybe make this two functions if we cover multiple use cases (product vs collection vs product variant).

andipaetzold (changes requested):
I rewrote your README to make it more compact.

Just rename the util method and we should be good to go.

apps/shopify/src/utils/base64.js
vikaskumr: I was trying to make it generic as it was suggesting that we are converting the response to Base64 for whatever use cases(product vs collection vs product variant).
But yeah making separate functions will bring more clarity

apps/shopify/README.md
vikaskumr: thank you for updating as your documentation is compact 
but don't you think we should also tell how we are ensuring that our app will be stable for the coming future release  after the deprecation of Base64 IDs.

andipaetzold (approved):

PR #1586, opened 2022-08-08
feat: upgrade bynder from compactview 2 to 3

a3hill:
Updated Bynder's compact view from version 2 to 3. This update improves the integration to include [DAT](https://support.bynder.com/hc/en-us/articles/360018559260-Dynamic-Asset-Transformations-DAT-) info from Bynder. It also allows the app to be configured to use various modes the Bynder Compact View allows based on what the user needs. The new mode of `SingleSelectFile` allows a user to select a specific derivative and includes the select file info in the asset response.

This update should be backwards compatible with the previous version of the app.

Also included is a readme file for the Bynder app to include more info about how to setup and use the modes. Marketplace documentation should be updated to reflect this new mode option.

### DIFF
--- a/apps/bynder/README.md
+++ b/apps/bynder/README.md
@@ -1,0 +1,16 @@
+# Bynder App
+
+This repository contains the source code for the Bynder app that is on the [marketplace](https://www.contentful.com/marketplace/). 
+
+The app uses Bynder's [Universal Compact View](https://developer-docs.bynder.com/ui-components) to allow a user to select assets from the user's Bynder library and store the selected asset's data provided by Bynder in Contentful. 
+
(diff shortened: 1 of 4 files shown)


kdamball:
@a3hill seems like there's a mismatch between `@contentful/dam-app-base` versions in `package-lock.json` vs `package.json`. Might want to run `npm install` to update your lock file

kdamball (commented):
@a3hill seems like there's a mismatch between `@contentful/dam-app-base` versions in `package-lock.json` vs `package.json`. Might want to run `npm install` to update your lock file

a3hill:
Thanks @kdamball! Missed that when I made the fork. Ran `npm install` and committed the updated lock file.

kdamball (approved):

Issue #1592, opened 2022-08-09
[commercetools] Product preview is limited to 20 products


img-alc:
**Issue description**
Adding more than 20 products to a field leads to "Product missing" message in products preview

**Steps to reproduce**
1. Create a new entry (content type must have a field with commerce tools app installed for products widget)
2. Add more than 20 products to the field connected to commerce tools
3. Save
4. Only 20 products are properly shown, while exceeding products show the message mentioned before

**Screenshot**
![image](https://user-images.githubusercontent.com/16797534/183671953-fb27be90-c3e5-4fb4-8d62-e49199d4ed34.png)

**Expected behavior**
I expected all selected products to show correctly in the preview, the error message is misleading. The products do exist and should render (name & image).

**Root cause**
Commerce tools api response is limited to 20 entries. You have to provide the offset in order to fetch remaining results. Example of a response:
![image](https://user-images.githubusercontent.com/16797534/183673874-518d19b6-34cf-41aa-bc12-95db31c95e2f.png)

The logic to fetch the products preview is not considering the number of results, always returning at most 20 entries ([link](https://github.com/contentful/apps/blob/425a0ccace45ba724d1485e32d93de6074acc375/apps/commercetools/src/api/products.ts)).

```
export async function fetchProductPreviews(
  skus: string[],
  config: ConfigurationParameters
): Promise<Product[]> {
  if (skus.length === 0) {
    return [];
  }

  const client = createClient(config);
  const response = await client
    .productProjections()
    .search()
    .get({
      queryArgs: {
        'filter.query': [`variants.sku:${skus.map((sku) => `"${sku}"`).join(',')}`],
      },
    })
    .execute();

  if (response.statusCode === 200) {
    const products = response.body.results.map(productTransformer(config));
    const foundSKUs = products.map((product: Product) => product.sku);
    const missingProducts = skus
      .filter((sku) => !foundSKUs.includes(sku))
      .map((sku) => ({
        sku,
        image: '',
        id: '',
        name: '',
        isMissing: true,
      }));
    return [...products, ...missingProducts];
  }
  throw new Error(`Request failed with status ${response.statusCode}`);
}
```

colomolo:
@img-alc Thanks for opening the issue. We now fetching max possible amount of products. Should be enough for most cases.

PR #1604, opened 2022-08-11
feat: [] Fetch all pages of previews

colomolo:
Addressing issue: https://github.com/contentful/apps/issues/1592

### DIFF
--- a/apps/commercetools/src/api/products.ts
+++ b/apps/commercetools/src/api/products.ts
@@ -25,7 +25,7 @@ async function fetchAllProductPreviews(
   client: ReturnType<typeof createClient>,
   products: ProductProjection[] = [],
   skus: string[],
-  pagination?: { offset: number }
+  offset: number = 0
 ): Promise<ProductProjection[]> {
   if (skus.length === 0) {
     return [];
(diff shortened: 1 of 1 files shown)


apps/commercetools/src/api/products.ts
```diff
@@ -19,42 +21,71 @@ function productTransformer({ projectKey, locale }: ConfigurationParameters) {
...
-      .filter((sku) => !foundSKUs.includes(sku))
-      .map((sku) => ({
-        sku,
-        image: '',
-        id: '',
-        name: '',
-        isMissing: true,
-      }));
-    return [...products, ...missingProducts];
+    const hasNextPage = response.body.offset + response.body.count < response.body.total!;
```
kdamball, line 49: We can add `body.total` to the if-check to avoid type coercion

apps/commercetools/src/api/products.ts
```diff
@@ -19,42 +21,71 @@ function productTransformer({ projectKey, locale }: ConfigurationParameters) {
...
+    const combinedResults = [...products, ...response.body.results];
+
+    if (hasNextPage) {
+      nextPageProducts = await fetchAllProductPreviews(
```
kdamball, line 53: FWIW: this will lead to sequential requests. I would see if you can parallelise them as it would technically only take ~1s for 1000 items (assuming 500ms/req) while this sequential fetching would take ~10s

kdamball (approved):

PR #1622, opened 2022-08-16
feat: [] wistia-videos app

vikaskumr:
We need to move the Wistia app into our repo. This PR is mainly just about the deploy command being correct, and the correct files being included in the folder. If their are code quality concerns, those can be addressed in a another PR.


apps/wistia-videos/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "version": "0.1.0",
+  "homepage": "./",
+  "private": true,
+  "dependencies": {
+    "@contentful/app-scripts": "^1.1.19",
+    "@contentful/app-sdk": "4.9.0",
+    "@contentful/forma-36-react-components": "^3.100.7",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
+    "styled-components": "^5.3.3"
```
andipaetzold, line 12: Use emotion instead

apps/wistia-videos/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "wistia-videos",
+  "version": "0.1.0",
+  "homepage": "./",
+  "private": true,
+  "dependencies": {
+    "@contentful/app-scripts": "^1.1.19",
+    "@contentful/app-sdk": "4.9.0",
+    "@contentful/forma-36-react-components": "^3.100.7",
```
andipaetzold, line 9: Do not use Forma v3, upgrade and use Forma v4

apps/wistia-videos/src/components/ConfigScreen.tsx
```diff
@@ -0,0 +1,264 @@
...
+    if (response.success) {
+      Notification.success('Your connection to the Wistia Data API is working.');
+      // set the projects in the state (don't save all the projects in the config parameters to prevent
+      // the config object to become very large and reach the size limit)
+      setFetchedProjects(response.projects);
+      setShowButton(false);
+      setRequiredMessage('');
+    } else {
+      Notification.error(`Connection to Wistia Data API failed: ${response.error}`);
+      console.log(`Connection to Wistia Data API failed: ${response.error}`);
```
andipaetzold, line 126: Remove this line

andipaetzold (changes requested):
Very rough review from me, will do a more thorough one tomorrow:

Can we please split the files into `locations` and `components` as we do in our template? For consistency.

Use emotion for global styles

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -0,0 +1,177 @@
...
+      if (videosRequest.response.success === false && videosRequest.response.error) {
+        setError(videosRequest.response.error)
+        updateLoadingStatus(false)
+      }
+    })();
+      
+  }, [sdk.field, sdk.parameters.installation])
+
+  // Function to update selected video ids
+  const updateVideoIds = (id:any) => {
```
therrou: Could be possible to add to this function the parameters from the creation of the content model? Right now is possible to add unlimited videos even when we set in the content model a limit or just one.
 The same with the _required_, it is an empty array and is a required field that shouldn't be possible to publish the entry.

therrou (changes requested):

Issue #1623, opened 2022-08-16
Netlify App: Increased/uncapped the max num of configs site


FredericCatteau:
Hello,

Thank you for this app, it is very useful.
For future development, I need to configure more than 10 sites (~15 sites).
would It be possible to change (Increased or uncapped) restriction to configure more than 10 sites?

I see this restriction is defined in variable "export const MAX_CONFIGS = 10;" in apps/apps/netlify/frontend/src/constants.js.

Best regards

shikaan:
Hey @FredericCatteau, thanks for reaching out.

The reason we have the cap was to keep the case you have to fire builds for many (possibly all) the sites simple, without batching or retying in case we hit a rate limiting error. I'll file a feature request for us, but [in the meantime I'll raise the limit to unblock you](https://github.com/contentful/apps/pull/1634).

Have a great day ahead,

PR #1634, opened 2022-08-17
feat: [] increase max configs to 20

shikaan:
This is an interim solution to unblock #1623. We will followup with some retry/batching logic to prevent rate limiting issues.


kdamball (approved):

kdamball:
fixes #1623

PR #1637, opened 2022-08-19
Fix: lockfile back to version 1

martin3walker:


kdamball (approved):
Good catch 🤦🏾 I definitely missed this in the PR

PR #1640, opened 2022-08-22
fix: [EXT-3929] version ranges for node and npm

martin3walker:
Prevent people from installing deps with the wrong node or npm versions, so we avoid situations where the deployment pipeline breaks.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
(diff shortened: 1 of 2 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
kdamball, line 1: nit: newline pls

kdamball (approved):
🎉 thanks for this

PR #1656, opened 2022-08-26
feat: upgrade bynder from compactview 2 to 3 with login fixed.

a3hill:
Fixes [login issue](https://github.com/contentful/apps/pull/1641) that [original](https://github.com/contentful/apps/pull/1586) update PR created.

Not sure why, but Bynder base URL had to be `https://d8ejoa1fys2rk.cloudfront.net` instead of the URL the Bynder docs recommended to allow login.

PR #1664, opened 2022-08-29
feat: upgrade bynder from compactview 2 to 3

a3hill:
Fixes https://github.com/contentful/apps/pull/1641 that https://github.com/contentful/apps/pull/1586 update PR created. Also removes all the odd edited files that appeared [here](https://github.com/contentful/apps/pull/1656)

Not sure why, but Bynder base URL had to be https://d8ejoa1fys2rk.cloudfront.net instead of the URL the Bynder docs recommended to allow login.

### DIFF
--- a/apps/bynder/src/index.js
+++ b/apps/bynder/src/index.js
@@ -160,7 +160,7 @@ function renderDialog(sdk) {
   script.addEventListener('load', () => {
     window.BynderCompactView.open({
       language: 'en_US',
-      mode: compactViewMode ? compactViewMode : 'MultiSelect',
+      mode: compactViewMode ?? 'MultiSelect',
       assetTypes: types,
       portal: { url: bynderURL, editable: true },
       assetFieldSelection: FIELD_SELECTION,
--- a/apps/bynder/src/index.js
+++ b/apps/bynder/src/index.js
@@ -6,7 +6,7 @@ import logo from './logo.svg';
 const CTA = 'Select a file on Bynder';
 
 const BYNDER_BASE_URL = 'https://d8ejoa1fys2rk.cloudfront.net';
-const BYNDER_SDK_URL = `${BYNDER_BASE_URL}/5.0.5/modules/compactview/bynder-compactview-2-latest.js`;
+const BYNDER_SDK_URL = `${BYNDER_BASE_URL}/5.0.5/modules/compactview/bynder-compactview-3-latest.js`;
 
 const FIELDS_TO_PERSIST = [
   'archive',
@@ -80,7 +80,7 @@ function prepareBynderHTML() {
   `;
 }
 
-function transformAsset(asset) {
+function transformAsset(asset, selected) {
   const thumbnails = {
     webimage: asset.files.webImage?.url,
     thul: asset.files.thumbnail?.url,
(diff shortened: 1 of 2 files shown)


apps/bynder/src/index.js
```diff
@@ -151,15 +152,15 @@ function renderDialog(sdk) {
...
+  function onSuccess(assets, selected) {
+    sdk.close(Array.isArray(assets) ? assets.map((asset) => transformAsset(asset, selected)) : []);
     window.removeEventListener('message', checkMessageEvent);
   }
 
   script.addEventListener('load', () => {
     window.BynderCompactView.open({
       language: 'en_US',
-      mode: 'MultiSelect',
+      mode: compactViewMode ? compactViewMode : 'MultiSelect',
```
andipaetzold, line 163: nit

```suggestion
      mode: compactViewMode ?? 'MultiSelect',
```

apps/bynder/src/index.js
```diff
@@ -237,6 +238,16 @@ setup({
       default: validAssetTypes.join(','),
       required: true,
     },
+    {
+      id: 'compactViewMode',
```
andipaetzold, line 242: Do we want this setting on global level? To me it feels like users want to configure this setting differently depending on the field.

apps/bynder/src/index.js
```diff
@@ -237,6 +238,16 @@ setup({
...
+      name: 'Compact View Mode',
+      type: 'List',
+      value: 'MultiSelect,SingleSelect,SingleSelectFile',
+      default: 'MultiSelect',
+      description:
+        'Set the Compact View to allow multiple or single asset selection at a time. "SingleSelectFile" is the only mode that allows a specific derivative to be selected. See https://developer-docs.bynder.com/ui-components for more info on Modes.',
```
andipaetzold, line 248: When using SingleSelect or SingleSelectFile you can also select multiple assets by just opening the dialog again. I would've expected that view mode would replace the existing asset. Seems like a bug to me.

andipaetzold (changes requested):

apps/bynder/src/index.js
a3hill: @andipaetzold are you thinking it would be set per field? Honestly, the reason I added the choice is if teams wanted to allow an editor to select a specific DAT from Bynder. The developer in me wouldn't want an editor to choose, I would want to make that choice for them.

kdamball (approved):

apps/bynder/src/index.js
```diff
@@ -237,6 +238,16 @@ setup({
       default: validAssetTypes.join(','),
       required: true,
     },
+    {
+      id: 'compactViewMode',
+      name: 'Compact View Mode',
+      type: 'List',
+      value: 'MultiSelect,SingleSelect,SingleSelectFile',
```
andipaetzold, line 245:
```suggestion
      value: 'MultiSelect,SingleSelectFile',
```

apps/bynder/src/index.js
andipaetzold, line 248:
```suggestion
      description: "SingleSelectFile" allows you to select a specific derivative.
```

apps/bynder/src/index.js
a3hill: Updated this.

apps/bynder/src/index.js
a3hill: Updated this.

andipaetzold (approved):

PR #1702, opened 2022-09-06
feat: add color picker app [EXT-4049]

fabe:
Adds the color picker app. `package.json` probably needs some love for the deployment to work?

Preview

### Demo

[Install to Contentful](https://app.contentful.com/deeplink?link=apps&id=6PbEUk8SPxj1hFMO6ttQo)


apps/color-picker/src/index.css
```diff
@@ -0,0 +1,35 @@
...
+  box-shadow: inset 0 0 0 1px rgba(0, 0, 0, 0.1);
+}
+
+input[type='color'] {
+  -webkit-appearance: none;
+  border: none;
+  width: 32px;
+  height: 32px;
+  border-radius: 6px;
+  background: transparent;
```
loweisz, line 34: Why not use emotion for these styles as well?

apps/color-picker/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,254 @@
...
+      themes: [
+        {
+          ...parameters.themes[0],
+          colors: [
+            ...parameters.themes[0].colors,
+            { id: uuidv4(), name: '', value: '#0088cc' },
+          ],
+        },
+      ],
+    });
```
loweisz, line 50: What about extracting the "swatch part" as a new state and then only on saving add it to parameters? Then you don't need to do this spreading all the time?

apps/color-picker/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,254 @@
...
+  };
+
+  const updateSwatch = (swatch: Color) => {
+    const index = parameters.themes[0].colors.findIndex(
+      (i) => i.id === swatch.id
+    );
+    let newTheme = parameters.themes[0];
```
loweisz, line 69: nit: doesn't need to be `let` when you only change parameters

apps/color-picker/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,254 @@
...
+          colors: parameters.themes[0].colors.filter((i) => i.id !== swatch.id),
+        },
+      ],
+    });
+  };
+
+  const updateSwatch = (swatch: Color) => {
+    const index = parameters.themes[0].colors.findIndex(
+      (i) => i.id === swatch.id
+    );
```
loweisz, line 68: just for sanity, I would add a -1 check, to make it is actually found

apps/color-picker/src/locations/Field.tsx
```diff
@@ -0,0 +1,136 @@
...
+            <Flex alignItems="center" gap="spacingXs">
+              <span
+                css={css`
+                  width: 16px;
+                  height: 16px;
+                  box-shadow: inset 0 0 0 1px rgba(0, 0, 0, 0.1);
+                  background-color: ${value?.value || '#ffffff'};
+                  display: flex;
+                  border-radius: 4px;
+                `}
```
loweisz, line 63: In general, I think moving all the CSS styles into a dedicated style file or at least outside of the HTML would make it a bit easier to read 😬

apps/color-picker/src/locations/Field.tsx
```diff
@@ -0,0 +1,136 @@
...
+              />
+              <span
+                css={css`
+                  display: flex;
+                  gap: ${tokens.spacing2Xs};
+                `}
+              >
```
loweisz, line 70: You can just use the `Flex` component here: 
`<Flex gap="spacing2Xs" >`

apps/color-picker/src/locations/Field.tsx
```diff
@@ -0,0 +1,148 @@
...
+            >
+              <Flex alignItems="center" gap="spacingXs">
+                <span className={styles.colorBox(color.value)} />
+                {color.name}
+              </Flex>
+            </Menu.Item>
+          ))}
+          {allowCustomValue && (
+            <Menu.Item onClick={() => customColorPicker?.current?.click()}>
+              Other...
```
andipaetzold, line 126: Should this be "Custom" as we are selecting a custom value

apps/color-picker/src/types.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export interface Theme {
+  id: string;
+  name: string;
+  colors: Color[];
+}
+
+export interface AppInstallationParameters {
+  themes: Theme[];
```
andipaetzold, line 14: We only use one theme at the moment. Should we still store an array here? Any plans on multiple themes?

apps/color-picker/src/locations/Field.tsx
fabe: Yes, makes sense!

apps/color-picker/src/locations/Field.tsx
```diff
@@ -0,0 +1,172 @@
...
+};
+
+interface MenuButtonProps {
+  showChevron?: boolean;
+  name: string;
+  value?: FieldValue;
+  onClick?: () => void;
+}
+
+function MenuButton({ showChevron, name, value, onClick }: MenuButtonProps) {
```
loweisz, line 148: Maybe worth moving this into a new file?

loweisz (approved):
Code looks good to me

PR #1722, opened 2022-09-12
fix(docs): add double dash to treat positional arguments

AbuHafsa:
Hi, 

I ran into the issue of creating a contentful backend app through the example but the command using `npm` behaved different from other mentioned commands. Thanks to @andipaetzold, he hinted me on Slack that there was a missing double dash as npm was trying `--example` as an option rather than a positional argument.


andipaetzold (approved):
Thank you!

PR #1723, opened 2022-09-12
fix: add argument type to error on catch clause

AbuHafsa:
Hi, 

I get an error once I run `start` command. This is related to argument type being unknown on `e` as the strict property is set on `tsconfig.json`. The pull request should be fix the issue for good.
```
error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string | undefined'.

166     throw new Error(e);
```

### DIFF
--- a/examples/default-values-backend/src/index.ts
+++ b/examples/default-values-backend/src/index.ts
@@ -161,7 +161,7 @@ const addDefaultDataOnEntryCreation = async (request: Hapi.Request, h: Hapi.Resp
     } else {
       throw new Error('failed to set default values' + (await res.text()));
     }
-  } catch (e) {
+  } catch (e: any) {
     console.error(e);
     throw new Error(e);
   }


andipaetzold:
👋 

Can you check which TypeScript version is installed locally? That error should not happen with `4.3.5` which is used in the example. The error should only occur from TypeScript 4.4: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

AbuHafsa:
Hey, 

The installed version of typescript is 4.8.3. The problem in here is that you are passing `e` to `new Error()` but the constructor does not handle `unknown`. If `throw new Error(e)` is commented out, the error goes away.

andipaetzold:
Can you update your PR and upgrade the typescript version to 4.8.3? In the version specified in the package.json everything works as expected.

AbuHafsa:
You are right it does work if `4.3.5` is enforced into the package.json (I tried it right now before I push the update).

andipaetzold:
Just run `yarn` to update `yarn.lock` and we are good to go

AbuHafsa:
`app-sdk` was also automatically updated. Thank you for the review.

andipaetzold (approved):
Thank you!

PR #1738, opened 2022-09-14
feat(cloudinary): only fetch media library when required

andipaetzold:
We currently always fetch `https://media-library.cloudinary.com/global/all.js` when the app is rendered. The script is actually only required within the dialog. This PR removes the script from index.html and only loads it within the dialog.

That way we remove an unnecessary request and reduce the number of scripts that need to be executed.


colomolo (approved):
LGTM

PR #1763, opened 2022-09-21
feat: [EXT-3992] Add App actions example

colomolo:


andipaetzold (commented):
Can we use the same structure as the nextjs example? https://github.com/contentful/apps/tree/master/examples/nextjs

It looks like the example was built before it was possible to build SSR Contentful apps.

examples/app-actions/src/pages/api/translations.ts
```diff
@@ -0,0 +1,47 @@
...
+  // gives access to the CMA client and App Installation parameters
+  // through the `contentfulContext` object
+  await runMiddleware(request, response, contentfulContextMiddleware);
+
+  // Parse the body and get access to `contentfulContext` mentioned above
+  const { body: stringBody, contentfulContext } = request;
+  const body = JSON.parse(stringBody);
+
+  // Simple validation of the payload.
+  // The action we have configured for this endpoint has category `EntryList`,
```
andipaetzold, line 23:
```suggestion
  // The action we have configured for this endpoint has category `Entries`,
```

examples/app-actions/src/lib/contentful-context.ts
```diff
@@ -0,0 +1,96 @@
...
+  return async (request, response, next) => {
+    const headers = parseHeaders(request.headers);
+
+    const isVerified = verifyRequest(
+      signingSecret,
+      {
+        method: request.method as CanonicalRequest['method'],
+        path: request.url,
+        headers,
+        body: parseBody(request.body),
```
andipaetzold, line 45: Does this actually work when `body` is an object? This assumes that `JSON.parse` returns the same string which is stringified by the Contentful servers.

examples/app-actions/src/pages/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+        </Paragraph>
+        <hr className={styles.splitter} />
+
+        <Heading>Configuration</Heading>
+
+        <FormControl id="target-language">
+          <FormControl.Label>Language</FormControl.Label>
+          <Select
+            id="target-language"
+            name={'target-language'}
```
andipaetzold, line 112: nit

```suggestion
            name="target-language"
```

examples/app-actions/README.md
```diff
@@ -0,0 +1,102 @@
...
+* Extend our upcoming Workflows capabilities with apps
+* Extend our UI in a declarative way
+* Be a player in the automation story of Contentful
+
+App Actions utilize existing features of the App Framework, such as being secured through [Request Verification](https://www.contentful.com/developers/docs/extensibility/app-framework/request-verification/) and being able to access content using [App Identities](https://www.contentful.com/developers/docs/extensibility/app-framework/app-identity/).
+
+This example outlines how these pieces work together.
+
+You can find more information 
+[here](https://ctf-doc-app-branch-feat-ext-3188.netlify.app/developers/docs/references/content-management-api/#/reference/app-actions).
```
shikaan, line 17: This can be updated with a (for now) dead link to the upcoming docs

examples/app-actions/src/lib/contentful-context.ts
shikaan: I think the name `parseBody` is misleading here (just because `JSON.parse` is the opposite of what's really happening), but the verification is done comparing strings, so this should be working. 

We could however be a little safer by removing whitespaces

shikaan (approved):
Looks good to me, there are a couple of small things I would fix, but nothing major

examples/app-actions/src/lib/contentful-context.ts
colomolo: Rename it to `stringifyBody`?

andipaetzold: In my experience with request verification, you always want to create the signature based on the raw body which is not yet parsed in any way. With the `JSON.stringify` you try to recreate the raw body in case it was accidentally already parsed. This is likely to cause issues as afaik parsing and stringifying an object does not guarantee to output the same string. Also, if there is any indentation in the raw body, things will break anyway.

shikaan: Yeah, I agree: request verification heavily relies on the developer having "done their homework" and have the body in a shape that allows for verifying the signature. 

This middleware is by no means exhaustive of all the possible cases, however it runs on NextJS which does the stringification for you anyway. If we think this check makes the code more confusing, we can just get rid of it altogether

andipaetzold: I'd personally get rid of it. We build a Next.js app, so we should only do whatever is required for Next.js.
This might be different if we have an example built for request verification.

colomolo: Removed it and now passing plain body there as-is.

examples/app-actions/vercel.json
```diff
@@ -0,0 +1,8 @@
+{
+  "functions": {
+    "pages/api/translations.ts": {
```
shikaan, line 3: I am not sure this will work with the new files structure. Have you tried running it on vercel?

colomolo: Fixed. Now it's running

PR #1771, opened 2022-09-22
fix: upgrade parcel and f36 components

### LINKED ISSUE #1724: Cannot build backend app through parcel after upgrading f36 components to version 4
**Describe the bug**

I have created an app using `default-values-backend` and decided to upgrade f36 packages as it is using an older version. once I run `start` it ends unexpectedly with an error.

**To Reproduce**
- `npx create-contentful-app  --example default-values-backend`
- Configure a sample app from your account and provide public key. 
- `npm install @contentful/f36-components @contentful/browserslist-config`
- Replace imports in `public/config.js` with `@contentful/f36-components` and replace some elements like `TextField` to `TextInput`
- `npm run start`

**Expected behavior**

It builds the frontend and runs the server. 

**Screenshots**
```
> backend-app@1.0.0 buildFrontend
> parcel build public/index.html

🚨  .../backend-app/node_modules/@contentful/f36-button/dist/esm/index.js: Unexpected token: operator (?)
    at Z.get (.../backend-app/node_modules/terser/dist/bundle.min.js:1:525)
    at Object.errorToJson (.../backend-app/node_modules/@parcel/utils/src/errorUtils.js:9:20)
    at Pipeline.process (.../backend-app/node_modules/parcel-bundler/src/Pipeline.js:29:26)
    at async Object.run (.../backend-app/node_modules/parcel-bundler/src/worker.js:15:12)
    at async Bundler.loadAsset (.../backend-app/node_modules/parcel-bundler/src/Bundler.js:577:19)
    at async .../backend-app/node_modules/parcel-bundler/src/Bundler.js:610:13
    at async Promise.all (index 7)
    at async Bundler.loadAsset (.../backend-app/node_modules/parcel-bundler/src/Bundler.js:599:21)
    at async .../backend-app/node_modules/parcel-bundler/src/Bundler.js:610:13
    at async Promise.all (index 4)

Process finished with exit code 1
```

**Desktop (please complete the following information):**
 - OS: macOS @ 12.5.1
 - Browser: Chrome @  105.0.5195.102

**Additional context**

Interestingly when I have added another script that runs `parcel watch public/index.html` and ran the server separately using `ts-node` with a few changes on the code, this have rendered the configuration page without any errors.

AbuHafsa:
I upgraded `f36-components` to adapt with the current design system as the documentation has moved to the newer version and some components has been replaced e.g: `TextField`. 

That led me to upgrade `parcel` as well because I wasn't able to build anymore (It was an issue with `terser`) but it was resolved in the newer versions of parcel. 

This should resolve the issue #1724

### DIFF
--- a/examples/default-values-backend/package.json
+++ b/examples/default-values-backend/package.json
@@ -10,8 +10,8 @@
   "license": "MIT",
   "dependencies": {
     "@contentful/app-sdk": "4.9.0",
-    "@contentful/forma-36-fcss": "^0.3.3",
-    "@contentful/forma-36-react-components": "^3.93.2",
+    "@contentful/browserslist-config": "^3.0.0",
+    "@contentful/f36-components": "^4.19.1",
     "@contentful/node-apps-toolkit": "^2.0.0",
     "@hapi/hapi": "^20.1.5",
     "@hapi/inert": "^6.0.1",
--- a/examples/default-values-backend/package.json
+++ b/examples/default-values-backend/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@contentful/backend_app_demo",
   "version": "1.0.0",
-  "main": "src/index.ts",
   "scripts": {
     "start": "yarn buildFrontend && tsc && node --insecure-http-parser built/src/index.js",
     "buildFrontend": "parcel build public/index.html",
@@ -26,7 +25,7 @@
     "typescript": "^4.3.5"
   },
   "devDependencies": {
-    "parcel-bundler": "^1.12.4",
+    "parcel": "^2.7.0",
     "ts-node": "^10.2.1"
   },
   "browserslist": [
--- a/examples/default-values-backend/.gitignore
+++ b/examples/default-values-backend/.gitignore
@@ -3,3 +3,4 @@ built
 yarn-error.log
 node_modules
 .env
+.parcel-cache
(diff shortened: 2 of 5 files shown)


examples/default-values-backend/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@contentful/backend_app_demo",
   "version": "1.0.0",
-  "main": "src/index.ts",
```
AbuHafsa, line 4: I am not sure if this field is used by `lerna.json` but parcel wouldn't work with this property set. It would throw this following error. 
 ```
💡 The "main" field is meant for libraries. If you meant to output a .ts file, 
either remove the "main" field or choose a different target name.
 ```
 If we were to change it to the built file. It would throw another error related to the target file being `html` as it tries to use that property as main target.
 ```
 Target "main" declares an output file path of "built/src/index.js" which does not match 
the compiled bundle type "html".
 ```

andipaetzold: Afaik lerna doesn't use this field. Also, the `examples` folder in this repo is not managed by lerna.

So, I think it's fine to drop it: https://parceljs.org/getting-started/migration/#package.json%23main

andipaetzold (approved):
Thanks a lot for upgrading these dependencies! 👏

AbuHafsa:
Happy to help 😄

PR #1779, opened 2022-09-26
fix: rename components to use multi-words and run linter

imretoth1999:
While using the example for the Vue contentful app I noticed that by default the linter is not working. 

Multi-Word-Component-Names is set to off to not cause a linter error when having the components Field, Page, Entry, etc. 
Another option would be to rename the components to use a multi-word but that would not keep the consistency across the other template examples.

Also, the files have been linted and committed to have the example following the best standards.


colomolo:
Hi @imretoth1999 and thanks for your contribution. I think that rather than turning `vue/multi-word-component-names` off, it is better to adapt this template to follow Vue guidelines and rename components to be multi-worded. Your contribution with these changes would be more than welcome. Wdyt?

imretoth1999:
Hello @colomolo. I'm glad you think the same about the eslint rule. I thought of renaming the files but wanted to keep the consistency between all of the examples in terms of naming but I agree that following the Vue standards is better. Each component has `Location` added in the name to describe its purpose better.

colomolo (approved):
Thanks, looks good!

andipaetzold (approved):

andipaetzold:
Please adjust the title before merge as it is not accurate anymore

imretoth1999:
> Please adjust the title before merge as it is not accurate anymore

Thank you for the approve, I changed the title of my PR

PR #1850, opened 2022-10-14
fix: [] external product link built with provided url (Commercetools)

martin3walker:
The external link for the Commercetools components were being hardcoded with the address for their Western European servers. This uses the api url provided in the config page by the user to get the correct base url for the external link. 

I don't see a way to do this more cleanly with the data we have on hand, but if anyone has an idea, I'm happy to hear it :)

### DIFF
--- a/apps/commercetools/src/api/products.ts
+++ b/apps/commercetools/src/api/products.ts
@@ -5,12 +5,13 @@ import { createClient } from './client';
 
 const MAX_LIMIT = 500;
 
-function productTransformer({ projectKey, locale, apiEndpoint }: ConfigurationParameters) {
-  const baseUrl = apiEndpoint?.split('api.')[1];
+function productTransformer({ projectKey, locale, mcUrl }: ConfigurationParameters) {
   return (item: ProductProjection): Product => {
+    const merchantCenterBaseUrl =
+      mcUrl && mcUrl.length > 0 ? mcUrl : 'https://mc.europe-west1.gcp.commercetools.com';
     const id = item.id ?? '';
     const externalLink =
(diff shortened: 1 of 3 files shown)


apps/commercetools/src/api/products.ts
```diff
@@ -5,14 +5,12 @@ import { createClient } from './client';
 
 const MAX_LIMIT = 500;
 
-function productTransformer({ projectKey, locale }: ConfigurationParameters) {
+function productTransformer({ projectKey, locale, apiEndpoint }: ConfigurationParameters) {
+  const baseUrl = apiEndpoint?.split('api.')[1];
```
andipaetzold, line 9: When can `apiEndpoint` be undefined? In that case we will see an error here.

apps/commercetools/src/api/products.ts
```diff
@@ -5,14 +5,12 @@ import { createClient } from './client';
...
   return (item: ProductProjection): Product => {
     const id = item.id ?? '';
     const externalLink =
-      (projectKey &&
-        id &&
-        `https://mc.europe-west1.gcp.commercetools.com/${projectKey}/products/${id}/general`) ||
-      '';
+      (projectKey && id && `https://mc.${baseUrl}/${projectKey}/products/${id}/general`) || '';
```
andipaetzold, line 13: Are we sure that the merchant center and API URL always exactly match in that way? Is there docs on how that URL is created?

martin3walker: I don't believe so. We could implement a new field in the config for the merchant center url, and fall back to the European address.

andipaetzold (approved):

PR #1898, opened 2022-10-26
fix: add Google Analytics deprecation notice [INTEGRATE-15]

jsdalton:
Adds a deprecation notice to the top of the Google Analytics app configuration screen.

## Purpose

* Google has deprecated several libraries behind the Analytics App and/or is no longer supporting them
* We need to let customers know when they are configuring the app that it will not work for new installations.

## Approach

* Add a `Note` of type `negative` to make this clear and unambiguous
* Provide a link to relevant documentation / explanation in the help center

image

* Original ticket: https://contentful.atlassian.net/browse/INTEGRATE-8
* Help center page: https://www.contentful.com/help/deprecation-notice-google-analytics-app/
* Google notice 1: https://developers.google.com/identity/sign-in/web/sign-in
* Google notice 2: https://developers.googleblog.com/2021/08/gsi-jsweb-deprecation.html
* Google notice 3: https://support.google.com/analytics/answer/11583528
* Stackoverflow that indicates Google no longer supported (see comment): https://stackoverflow.com/questions/72961705/google-analytics-embed-api-with-new-google-login-authorization-flow-i-e-google#comment128881973_72968672

### DIFF
--- a/apps/google-analytics/src/AppConfig.tsx
+++ b/apps/google-analytics/src/AppConfig.tsx
@@ -10,6 +10,7 @@ import {
   Select,
   FormLabel,
   TextInput,
+  Note,
 } from '@contentful/forma-36-react-components';
 import styles from './styles';
 import { AppConfigParams, AppConfigState, AllContentTypes, ContentTypes } from './typings';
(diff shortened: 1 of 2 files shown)


colomolo (approved):
💎

PR #1925, opened 2022-11-02
chore: proxy requests to frontend dev server in development [INTEGRATE-41]

jsdalton:
In development mode, proxy requests to the running frontend development server rather than to the build

## Purpose

Since the frontend for this app is mounted to the backend (at `/frontend`), the front end is actually served via the build during normal operation.

This is annoying for local development, since you have to build every time before you refresh and see the result of your work.

## Approach

* Write a simple proxier that forwards requests back to the given URL and path
* Add middleware for `/frontend` (where the app is served from on the backend)
* Also add middleware for `/static` since the development server specifies at relative paths (which get served from `http://localhost:8080/static/...` in local development
* Only use the middlewares in development, otherwise fallback to the same static middleware

### DIFF
--- a/apps/smartling/lambda/src/index.ts
+++ b/apps/smartling/lambda/src/index.ts
@@ -4,6 +4,15 @@ import * as express from 'serverless-express/express';
 import * as path from 'path';
 import fetch from 'node-fetch';
 import { Issuer } from 'openid-client';
+import { URL } from 'url';
+
+function requestProxier(proxyUrl: string, proxyPathPrefix = '') {
+  return async (req: express.Request, res: express.Response) => {
+    const url = new URL(path.join(proxyPathPrefix, req.url), proxyUrl);
+    const response = await fetch(url.href);
+    response.body.pipe(res);
+  };
+}
 
 export function makeApp(fetchFn: any, issuer: any) {
   const app = express() as Express;
@@ -107,10 +116,20 @@ export function makeApp(fetchFn: any, issuer: any) {
     );
   });
 
-  app.use(
-    '/frontend',
-    express.static(path.dirname(require.resolve('@contentful/smartling-frontend')))
-  );
+  if (process.env.LOCAL_DEV === 'true' && process.env.FRONTEND_URL) {
+    // in development mode, proxy requests to the frontend development server running at FRONTEND_URL
+    app.use('/frontend', requestProxier(process.env.FRONTEND_URL));
+
+    // in development mode, static assets are expressed as relative paths, which is served
+    // under the backend and thus also needs to be proxied
+    app.use('/static', requestProxier(process.env.FRONTEND_URL, 'static'));
+  } else {
+    // in production mode, frontend requests are served from the static smartling-frontend build folder
+    app.use(
+      '/frontend',
+      express.static(path.dirname(require.resolve('@contentful/smartling-frontend')))
+    );
+  }
 
   app.use((_req, res) => res.status(404).send('Not found'));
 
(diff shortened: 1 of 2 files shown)


apps/smartling/lambda/src/index.ts
```diff
@@ -4,6 +4,15 @@ import * as express from 'serverless-express/express';
 import * as path from 'path';
 import fetch from 'node-fetch';
 import { Issuer } from 'openid-client';
+import { URL } from 'url';
+
+function requestProxier(proxyUrl: string, proxyPathPrefix = '') {
```
andipaetzold, line 9: Does that work so easy? Our frontend is using webpack which (afaik) sets up a web socket to the backend. I doubt that `fetch` could handle that kind of request.

Or is this something we kind of ignore for now as it also doesn't work with the existing setup?

jsdalton: It works just fine actually (again, for development). But let me know if you're driving at something else, since I might misunderstand your question.

apps/smartling/lambda/src/index.ts
```diff
@@ -107,10 +116,20 @@ export function makeApp(fetchFn: any, issuer: any) {
...
 
-  app.use(
-    '/frontend',
-    express.static(path.dirname(require.resolve('@contentful/smartling-frontend')))
-  );
+  if (process.env.LOCAL_DEV === 'true' && process.env.FRONTEND_URL) {
+    // in development mode, proxy requests to the frontend development server running at FRONTEND_URL
+    app.use('/frontend', requestProxier(process.env.FRONTEND_URL));
+
+    // in development mode, static assets are expressed as relative paths, which is served
```
andipaetzold, line 123: If they are relative paths, they should load from `/frontend/static`, right? Shouldn't they only load from `/static` if they are absolute?

andipaetzold (approved):

apps/smartling/lambda/src/index.ts
jsdalton: Good question. I was confused by this at first as well.

To explain by way of example:

In "production" mode the /frontend/index.html returns the following HTML document:

```html
<!doctype html><html><head><meta charset="utf-8"/><script defer="defer" src="./static/js/main.f1d912ca.js"></script><link href="./static/css/main.846e3b9c.css" rel="stylesheet"></head><body><div id="root"></div></body></html>
```

That's from the build (of smartling/frontend). The file paths are relative and thus live under the frontend path. e.g. the first JS file would be served from `/frontend/static/js/main.f1d912ca.js`.

In "development" mode the development server (which we proxy in this PR) the request would give us a different result:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
  <script defer src="/static/js/bundle.js"></script></head>
  <body>
    <div id="root"></div>
  </body>
</html>
```

Note that the HTML is totally different (not minified) and also only serves up the main static bundle file.

This file lives not at a relative path but rather an absolute path. So that's why we have to set up the secon route in development mode for `/static/*`.

andipaetzold: What confuses me is that comment

> in development mode, static assets are expressed as relative paths

Shouldn't it be

> in development mode, static assets are expressed as **absolute** paths

An absolute path is `/static`, a relative path is `./static`.

Side note: By default also in prod absolute paths would be used, but we change that with `homepage: '.'` in package.json

jsdalton: Ah good catch. The comment itself is wrong / confusing. What I meant to say was "just a path" v.s a full URL.

I.e. they are served as src="/static/somejs.js" instead of src="http://localhost:3030/static/somejs.js". Since the frontend is running on http://localhost:8080 the actual URL request is to http://localhost:8080/static/somejs.js. Hence the needs for a second proxy under the /static root path.

PR #1940, opened 2022-11-07
Add README for Smartling [#INTEGRATE-40]

jsdalton:
## Purpose

We provide little context or development support for the app integrations we manage. This adds friction for engineers who wish to contribute and also makes onboarding painful.

Having a README at the root of every app integration seems like a sensible requirement.

## Approach

* Add a basic README. Ideally we can follow a similar template for other apps. Key goal is to allow an engineer to get a Smartling development environment up and running with no outside assistance.
* Secrets and credentials are a bit of a challenge. We have an internal wiki page where these are identified and organized. IN reality it makes it hard realistically to work on this app outside Contentful.
* I have a question out to Smartling about obtaining Client Id and secret for the oauth flow. Might be something we just hand request. Either way it's worth documenting.


kdamball (commented):
Could the info here be useful? https://www.contentful.com/help/smartling-app/

jsdalton:
@kdamball: Yes! Already have it here https://github.com/contentful/apps/pull/1940/files#diff-9b936c4719fe7b437b96849b97464cce80f91d97217f603e7f6c3c5f0c43196cR53

I think it's probably a good idea to have a "Other Resources" section at hte bottom with a list of helpful links etc. (i.e. so that page is findable in two places). I will include this now

PR #1987, opened 2022-11-11
fix(ImageFocalPoint): handle multi locale []

### LINKED ISSUE #1932: "Add an image to the entry first" - On localization Content having multiple locale images
Works fine with an English locale.  When I try to set the focal point on 2nd locale, it throws an error. 
image

kdamball:
## Purpose
Image focal point does not handle multi locale. Lets fix it

## Approach
As usual, JS `this` property bit us again. `findProperLocale` would not refer to current field and not the asset's

Fixes #1932

### DIFF
--- a/apps/image-focal-point/src/index.js
+++ b/apps/image-focal-point/src/index.js
@@ -59,11 +59,11 @@ export class App extends React.Component {
   };
 
   findProperLocale() {
-    if (this.props.sdk.entry.fields[this.props.sdk.field.id].type === 'Link') {
-      return this.props.sdk.locales.default;
-    }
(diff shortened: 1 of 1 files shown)


apps/image-focal-point/src/index.js
```diff
@@ -52,18 +52,18 @@ export class App extends React.Component {
     const value = e.currentTarget.value;
     this.setState({ value });
     if (value) {
-      this.props.sdk.field.setValue(value);
+      this.props.sdk.field.setValue(value, this.findProperLocale());
```
andipaetzold, line 55: Shouldn't the app always update itself? In which scenario do we need the locale here?

andipaetzold (approved):

PR #1988, opened 2022-11-11
fix: remove duplicate white space in script

andipaetzold:
## Purpose

Removing duplicate white space characters from the `upload-ci` script in our templates.

## Approach

There should be only a single white space character between the arguments.


colomolo (approved):
👍

PR #1989, opened 2022-11-11
Update index.js

gavinmatt:
Default to single select and updated helper text

## Purpose
Fixing a customer issue when switching from multi to single select mode at install/reconfiguration.
## Approach
We dug in and found that this was confusing UI/UX for users.
None
## Deployment
None


gavinmatt:
Look mom, I'm an engineer!

andipaetzold (changes requested):
This will change the selection mode for all users who installed the Bynder app before September 1 and haven't updated the config since then. Before Sep 1 (https://github.com/contentful/apps/pull/1664), we didn't store the view mode. These installations fall back to the default. With this PR changing the default, the setting of many installations is changed which requires manual action for people to resolve. I therefore wouldn't change it.

gavinmatt:
Thank you for the context @andipaetzold! Definitely a more dangerous change than we assumed. @jsdalton FYI - we probably want to isolate just the copy updates.

PR #2021, opened 2022-11-17
chore: Upgrade apps to latest node lts

ivo-toby:
## Purpose
Update the node version of lambda packages in this mono-repo to release 16

## Dependencies  & package versions
I've updated packages to match the peerdependencies, because Lerna was not able to run successfully.
I have also updated `@contentful/app-scripts` because the <1 releases have a npm/node-version restriction and that would not run as well. 

### Note on typesafety
`smartling/lambda` is partially typed, but not entirely. Since `serverless-express` lacks ts definitions and express itself is quite old it is not really viable to introduce typings to this package as of now. Hence the `@ts-ignore` statements


apps/jira/functions/package.json
```diff
@@ -3,16 +3,6 @@
   "version": "1.4.62",
   "main": "index.js",
   "private": true,
-  "scripts": {
```
kdamball, line 6: Why is this deleted?

kdamball (commented):
Can we also update netlify lambda to v16 for node

apps/jira/functions/package.json
ivo-toby: I suspect a rebase to master, I've reinstated the script-section.

apps/graphql-playground/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.2.5",
   "private": true,
   "devDependencies": {
-    "@contentful/app-scripts": "^0.15.8",
+    "@contentful/app-scripts": "^1.2.0",
```
colomolo, line 6: Should we also pin this one, similar to other apps?

colomolo (commented):
Overall looks good to me

apps/graphql-playground/package.json
ivo-toby: Done!

colomolo (approved):
LGTM, but maybe we need another pair of 👀 to look just in case.

apps/smartling/lambda/src/index.ts
```diff
@@ -17,6 +17,7 @@ function requestProxier(proxyUrl: string, proxyPathPrefix = '') {
 export function makeApp(fetchFn: any, issuer: any) {
   const app = express() as Express;
 
+  // tslint:disable-next-line
```
kdamball, line 20: Would be great to know why these are now needed. Feel free to add a comment

ivo-toby: I fixed the issue tslint was nagging about. 
The rest of this file could need some TLC eventually and we should consider replacing TSLint with ESLint.

kdamball (approved):

PR #2079, opened 2022-11-28
fix: [] temporarily disable deploys for graphql playground

martin3walker:
## Purpose
Temporarily pinning version of graphql playground app

### DIFF
--- a/apps/graphql-playground/package.json
+++ b/apps/graphql-playground/package.json
@@ -33,7 +33,7 @@
     "analyze": "source-map-explorer 'build/static/js/*.js'",
     "start": "react-scripts start",
     "build": "GENERATE_SOURCEMAP=false react-scripts build",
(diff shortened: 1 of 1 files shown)


apps/graphql-playground/package.json
```diff
@@ -33,7 +33,7 @@
     "analyze": "source-map-explorer 'build/static/js/*.js'",
     "start": "react-scripts start",
     "build": "GENERATE_SOURCEMAP=false react-scripts build",
-    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEFINITIONS_ORG_ID} --definition-id 66frtrAqmWSowDJzQNDiD --token ${CONTENTFUL_CMA_TOKEN}",
+    "deploy": "echo: package error, not deploying",
```
kdamball, line 36: syntax error: `"deploy":"echo 'pinned bundle, skip deploy'"` or something like that

martin3walker: Thanks, fixed 👍

kdamball (approved):

PR #2136, opened 2022-12-06
chore: use vault setup [EXT-4121]

kdamball:
## Purpose
Use ctfl-vault in CI

## Approach
Finally use vault for CI secrets. This takes care of both publishing to npm & aws deploys for all apps

Read more: https://contentful.atlassian.net/browse/EXT-4121

## Deployment
We will know if it's successful once deploys go through. We should already see any warnings/errors in PR checks


vikaskumr (approved):
💎

PR #2149, opened 2022-12-06
Create initial project and app README [INTEG-66]

jsdalton:
## Purpose

Just getting the default project in the branch so that subsequent PRs are clear and easy to review.

We also want to start with a solid README that will continue to evolve as the project continues.

## Approach

* Use `create-contentful-app`. Later we will prune the unneeded files.


apps/google-analytics-4/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "google-analytics",
```
andipaetzold, line 2: Should be scoped to `@contentful`

apps/google-analytics-4/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "@tsconfig/create-react-app/tsconfig.json",
+  "compilerOptions": {
+    "target": "es5",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
```
andipaetzold, line 9: Probably we can remove those from the template... not sure why we still have them in there.

PR #2182, opened 2022-12-13
Add basic app configuration screen [INTEG-90]

jsdalton:
## Purpose

The config screen will be used to allow space admins to install the app, provide credentials and configure the app.

## Approach

* Introduce a basic configuration screen. As indicated, additional (actual) form 
* Use styles object (emotion)
* Remove boilerplate comments
* Modify boilerplate tests (mostly to extend later)


apps/google-analytics-4/src/locations/ConfigScreen.spec.tsx
```diff
@@ -9,14 +9,24 @@ jest.mock('@contentful/react-apps-toolkit', () => ({
...
+  it('can be installed', async () => {
+    const { getByText } = renderedComponent;
+
+    // install the app
     await mockSdk.app.onConfigure.mock.calls[0][0]();
 
-    expect(
-      getByText('Welcome to your contentful app. This is your config page.')
-    ).toBeInTheDocument();
+    expect(getByText('About Google Analytics for Contentful')).toBeInTheDocument();
```
andipaetzold, line 30: This test would pass if we do not trigger the install. Should there be any feedback about the installation state within the app>

apps/google-analytics-4/src/locations/ConfigScreen.tsx
```diff
@@ -1,68 +1,135 @@
-import React, { useCallback, useState, useEffect } from 'react';
+import { useCallback, useState, useEffect } from 'react';
 import { AppExtensionSDK } from '@contentful/app-sdk';
-import { Heading, Form, Paragraph, Flex } from '@contentful/f36-components';
+import { Heading, Form, Paragraph } from '@contentful/f36-components';
 import { css } from 'emotion';
 import { /* useCMA, */ useSDK } from '@contentful/react-apps-toolkit';
```
andipaetzold, line 5: Another comment-cleanup here

andipaetzold (approved):

apps/google-analytics-4/src/locations/ConfigScreen.spec.tsx
jsdalton: Yes 👍 . I will address this in a subsequent PR. For now these tests are still mainly just boilerplate.

apps/google-analytics-4/src/locations/ConfigScreen.tsx
jsdalton: Ah you mean the `/* useCMA */` bit. Good catch. Will remove on the next PR.

PR #2195, opened 2022-12-15
GA4: Add form control to allow for input of service key [INTEG-84]

jsdalton:
## Purpose

When users are installing the GA4 app, they will need to provide a set of credentials known as a "service key". This is a JSON file that is generated from the Google Developer Console. More info here: https://cloud.google.com/iam/docs/creating-managing-service-account-keys

We need to store the contents of this file as part of the app installation. We will then reuse the credentials when making calls to the Google Analytics Data API on the apps behalf.

The credentials are sensitive access data and MUST be stored as a secret. Once saved they should not be visible to the front end.

**IMPORTANT:** At this stage of development we will intentionally store these credentials insecurely as part of App Installation parameters. Prior to release or non trivial testing we will find and use a secure secret storage solution.

After we have proper secret storage, we will still need to save an identifier value in the app installation, which will serve both as a pointer to the eventual secret storage location AND as a public facing identifier to the user, so they can see what credentials are currently being used.

## Approach

* Implement a component which takes a key file string as input, parses and validates the input, then updates the "raw" secret key state as well as as the secret key id state provided by the parent. (Separating these values is intentional, as later the secret key itself will be stored outside of parameters, while secret key id will still live there)
* On app install, just show the empty form
* After install, hide the form in a collapse and show details of the installed key
* Actually checking if key is active will happen on later PR (INTEG-86)
* Actual link to help center documentation will happen later (INTEG-85)

* [x] Prevent install if no key provided (no key provided should be fine on update) (see #2205)
* [x] Add tests for the validation bit inside the form (see #2205)

### Static Screenshot

Note: This was an actual service key but it was destroyed after the screenshot and video recording 😉 

Screenshot 2022-12-15 at 09 32 36

### Animation

To get the idea of what's happening:

screen-recording

### DIFF
--- a/apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
+++ b/apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
@@ -1,0 +1,268 @@
+import React, { useCallback, useEffect, useState } from 'react';
+import {
+  Collapse,
+  Flex,
+  FormControl,
+  Note,
+  Paragraph,
+  Text,
+  Textarea,
+  TextLink,
+} from '@contentful/f36-components';
+import {
+  CheckCircleIcon,
+  ChevronDownIcon,
+  ChevronRightIcon,
+  ExternalLinkTrimmedIcon,
+} from '@contentful/f36-icons';
+import { css } from 'emotion';
+import { pick, reject } from 'lodash';
+
+import tokens from '@contentful/f36-tokens';
+import InstalledServiceAccountKey from './InstalledServiceAccountKey';
+import type { ServiceAccountKey, ServiceAccountKeyId } from '../types';
+
+interface FormControlState {
+  isInvalid: boolean;
+  errorMessage: string;
+}
+
+interface FormControlServideAccountKeyFileProps {
+  setServiceAccountKey: (serviceAccountKey: ServiceAccountKey | null) => void;
+  setServiceAccountKeyId: (serviceAccountKey: ServiceAccountKeyId | null) => void;
+  currentServiceAccountKeyId: ServiceAccountKeyId | null;
+  className?: string;
+}
+
+class AssertionError extends Error {}
+
+const styles = {
+  credentialsInput: css({
+    fontFamily: 'monospace',
+  }),
+  credentialsNote: css({
+    marginBottom: tokens.spacingM,
+  }),
+  credentialsExplanation: css({
+    marginTop: tokens.spacingL,
+  }),
+};
+
+const placeholderText = `{
+  "type": "service_account",
+  ...
+}`;
+
+const keysOfServiceAccountKey: Array<keyof ServiceAccountKey> = [
+  'type',
+  'project_id',
+  'private_key_id',
+  'private_key',
+  'client_email',
+  'client_id',
+  'auth_uri',
+  'token_uri',
+  'auth_provider_x509_cert_url',
+  'client_x509_cert_url',
+];
+
+const convertKeyFileToServiceAccountKey = (keyFile: string): ServiceAccountKey => {
+  const parsedKeyFile = JSON.parse(keyFile);
+  assertServiceAccountKey(parsedKeyFile);
+
+  // ensure key file never contains extraneous keys
+  return pick(parsedKeyFile, keysOfServiceAccountKey);
+};
+
+const convertServiceAccountKeyToServiceAccountKeyId = (
+  serviceAccountKey: ServiceAccountKey
+): ServiceAccountKeyId => ({
+  id: serviceAccountKey.private_key_id,
+  clientEmail: serviceAccountKey.client_email,
+  clientId: serviceAccountKey.client_id,
+  projectId: serviceAccountKey.project_id,
+});
+
+function assertServiceAccountKey(value: any): asserts value is ServiceAccountKey {
+  if (value?.type !== 'service_account')
+    throw new AssertionError("Key file `type` must be 'service_account'");
+
+  const missingKeys = reject(keysOfServiceAccountKey, (key) => key in value);
+  if (missingKeys.length > 0)
+    throw new AssertionError(`Key file is missing the following keys: ${missingKeys.join(', ')}`);
+
+  const notStringValues = reject(keysOfServiceAccountKey, (key) => typeof value[key] === 'string');
+  if (notStringValues.length > 0)
+    throw new AssertionError(
+      `Key file has invalid values at the following keys: ${notStringValues.join(', ')}`
+    );
+}
+
+const FormControlServiceAccountKey = ({
+  setServiceAccountKey,
+  setServiceAccountKeyId,
+  currentServiceAccountKeyId,
+  className,
+}: FormControlServideAccountKeyFileProps) => {
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const [serviceAccountKeyFile, setServiceAccountKeyFile] = useState<string>('');
+  const [formControlState, setFormControlState] = useState<FormControlState>({
+    isInvalid: false,
+    errorMessage: '',
+  });
+
+  const setValidServiceAccountKey = useCallback(
+    (newServiceAccountKey: ServiceAccountKey | null) => {
+      setFormControlState({
+        isInvalid: false,
+        errorMessage: '',
+      });
+      setServiceAccountKey(newServiceAccountKey);
+      setServiceAccountKeyId(
+        newServiceAccountKey
+          ? convertServiceAccountKeyToServiceAccountKeyId(newServiceAccountKey)
+          : null
+      );
+    },
+    [setFormControlState, setServiceAccountKey, setServiceAccountKeyId]
+  );
+
+  const setInvalidServiceAccountKey = useCallback(
+    (errorMessage: string) => {
+      setFormControlState({
+        isInvalid: true,
+        errorMessage,
+      });
+      setServiceAccountKey(null);
+      setServiceAccountKeyId(null);
+    },
+    [setFormControlState, setServiceAccountKey, setServiceAccountKeyId]
+  );
+
+  useEffect(() => {
+    if (currentServiceAccountKeyId) {
+      setServiceAccountKeyFile('');
+      setIsExpanded(false);
+    }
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -6,9 +6,10 @@
     "npm": ">= 8.0.0 < 9.0.0"
   },
   "devDependencies": {
-    "prettier": "2.8.1",
+    "@testing-library/user-event": "^14.4.3",
```
andipaetzold, line 9: I think this should rather be within the app itself. That way we could easily split up the monorepo in the future.

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
```diff
@@ -0,0 +1,275 @@
...
+  serviceAccountKey: ServiceAccountKey
+): ServiceAccountKeyId => ({
+  id: serviceAccountKey.private_key_id,
+  clientEmail: serviceAccountKey.client_email,
+  clientId: serviceAccountKey.client_id,
+  projectId: serviceAccountKey.project_id,
+});
+
+function assertServiceAccountKey(value: any): asserts value is ServiceAccountKey {
+  if (value?.type !== 'service_account')
```
andipaetzold, line 88: nit: I think we normally use parenthesis for single-line blocks in this repo. But maybe it's a mix and I just don't know.

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
```diff
@@ -0,0 +1,275 @@
...
+const convertServiceAccountKeyToServiceAccountKeyId = (
+  serviceAccountKey: ServiceAccountKey
+): ServiceAccountKeyId => ({
+  id: serviceAccountKey.private_key_id,
+  clientEmail: serviceAccountKey.client_email,
+  clientId: serviceAccountKey.client_id,
+  projectId: serviceAccountKey.project_id,
+});
+
+function assertServiceAccountKey(value: any): asserts value is ServiceAccountKey {
```
andipaetzold, line 87: Could we extract this to make it easier to test?

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
```diff
@@ -0,0 +1,275 @@
...
+        <Paragraph>
+          To use the Google Analytics app, you will need to provision a{' '}
+          <TextLink
+            icon={<ExternalLinkTrimmedIcon />}
+            alignIcon="end"
+            href="https://cloud.google.com/iam/docs/understanding-service-accounts"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            Google Cloud service account
```
andipaetzold, line 244:
```suggestion
            Google Cloud Service Account
```

apps/google-analytics-4/src/locations/ConfigScreen.tsx
```diff
@@ -64,17 +66,52 @@ const styles = {
...
+  const [parameters, setParameters] = useState<AppInstallationParameters>({
+    serviceAccountKey: null,
+    serviceAccountKeyId: null,
+  });
+
+  const [newServiceAccountKey, setNewServiceAccountKey] = useState<ServiceAccountKey | null>(null);
+  const [newServiceAccountKeyId, setNewServiceAccountKeyId] = useState<ServiceAccountKeyId | null>(
+    null
+  );
+  const [isInvalidServiceAccountKeyFile, setIsInvalidServiceAccountKeyFile] =
```
andipaetzold, line 78: Can we flip this to `isValid` to make it easier to read? Negative values need additional when reading expressions

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
jsdalton: 👍 Makes sense. I'm trying to remember if i even did this or it was Prettier 🤔? Either way I personally prefer explicit curly braces so happy to change it if prettier let's me.

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
```diff
@@ -0,0 +1,275 @@
...
+        newServiceAccountKey
+          ? convertServiceAccountKeyToServiceAccountKeyId(newServiceAccountKey)
+          : null
+      );
+      setFormControlState({
+        isInvalid: false,
+        errorMessage: '',
+      });
+    },
+    [setFormControlState, setServiceAccountKey, setServiceAccountKeyId]
```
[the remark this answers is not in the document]
jsdalton: The main reason for this is feedback from the linter which centers around enforcing these principles: https://reactjs.org/docs/hooks-rules.html

Basically this is just a convenience helper function (since a few different states are set in concert when valid vs. invalid) and since these call the set state hooks, React asks to be explicit about declaring them within useCallback and being explicit about the dependencies OR just linlining the whole thing in useEffect. 

I believe I had it inlined before which worked fine, I find it just slightly harder to reason about inlined function definitions so I preferred defining this as a callback. But I'd be happy to go with whatever seems most obvious / unsurprising in your eyes.

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
jsdalton: I kept the capitalization from Google here (lowercase "service account" throughtout) https://cloud.google.com/iam/docs/service-accounts

apps/google-analytics-4/src/locations/ConfigScreen.tsx
jsdalton: Yeah, I think it's better. I flipped it to match the input of Forma36 but i think it's much easier to reason about if we just flip it at the end when we feed it to the component.

apps/google-analytics-4/src/locations/ConfigScreen.tsx
```diff
@@ -64,17 +66,52 @@ const styles = {
...
 
-    return {
+    if (isInvalidServiceAccountKeyFile) {
+      sdk.notifier.error('Invalid service account key file. See field error for details');
+      return false;
+    }
+
+    const newServiceKeyParameters = {
+      serviceAccountKey: newServiceAccountKey,
+      serviceAccountKeyId: newServiceAccountKeyId,
```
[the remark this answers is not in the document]
jsdalton: Yes, this is exactly the reason. Before we release the `serviceAccountKey` will not be stored in params at all. When it's saved as  a secret the idea is we'll use the data in `serviceAccountKeyId` to retrieive it. I "intentionally" kept those two things separate as opposed to bundling them as one for this reason.

package.json
jsdalton: Thanks good catch. Didn't even realized I'd installed this at the top level 🤦‍♂️

apps/google-analytics-4/src/locations/ConfigScreen.tsx
jsdalton: Actually for this, if you don't mind I'll tackle the change in a subsequent PR. Since i have a few PRs stacked and this actually touches logic in a few files it's just easier to isolate the change to avoid merge conflicts. (The cleanup task is captured in INTEG-99)

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
andipaetzold: Okay, line 223 is "Service Account Key" which got me confused and write this comment.

apps/google-analytics-4/src/locations/ConfigScreen.spec.tsx
```diff
@@ -1,32 +1,73 @@
...
 
 jest.mock('@contentful/react-apps-toolkit', () => ({
   useSDK: () => mockSdk,
   useCMA: () => mockCma,
 }));
 
+// Helper to mock users clicking "save" -- return result of the callback passed to onConfigure()
+const saveAppInstallation = async () => {
+  // We manually call the LAST onConfigure() callback (this is important, as earlier calls have stale data)
+  return await mockSdk.app.onConfigure.mock.calls.slice(-1)[0][0]();
```
andipaetzold, line 15: There is now a new array method `.at(...)` to keep the code simpler: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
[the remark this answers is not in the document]
andipaetzold: All good - it's better to have more than missing one. I would've assumed that linters don't complain here anymore. 🤷

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
```diff
@@ -0,0 +1,232 @@
...
+    marginTop: tokens.spacingL,
+  }),
+};
+
+const placeholderText = `{
+  "type": "service_account",
+  ...
+}`;
+
+const FormControlServiceAccountKey = ({
```
andipaetzold, line 59: I am not really happy how state is managed in this component atm. There are four states in the parent and three states in here. Most of them are derived from `serviceAccountKeyFile`. At the moment they are kept in sync with `useEffect`s. To me, this looks very risky and easy to break if we introduce more in the future. I'd suggest pulling `serviceAccountKeyFile` in the parent component and derive all other values from that. Derived values can then use `useMemo` instead of manually updating them via `useEffect`. That should improve readability and it's easier to follow where the data comes from and what happens when you change it.

E.g. it's quite straight forward to derive the key and id from the file. Also whether it's valid or not can just be done with a `useMemo`.

andipaetzold (changes requested):

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
jsdalton: Okay, this was great advice. I moved all the state into the parent and refactored to pass event handlers down to the child. This had the nice effect of removing all the `useEffect`s and weird manipulation of state. Thanks for this helpful feedback!

andipaetzold (approved):

PR #2205, opened 2022-12-16
GA4: Require key file at install [INTEG-91]

jsdalton:
## Purpose

When installing the GA4 app, the credentials key file is required. Once the app is installed, the key file field is no longer required (since it should only be populated when replacing the key)

## Approach

* Conditionally pass an is required value down to note field is required
* Throw an error on save if the field is required but no key file is present
* Add tests for this as well as more for when saving form on already installed app

### Static Screenshot

screenshot-keyfile-require

### Screen recording

screen-recording-keyfile-require

* Builds directly off of #2195

### DIFF
--- a/apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
+++ b/apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
@@ -35,6 +35,7 @@ interface FormControlServideAccountKeyFileProps {
   setServiceAccountKey: (serviceAccountKey: ServiceAccountKey | null) => void;
   setServiceAccountKeyId: (serviceAccountKey: ServiceAccountKeyId | null) => void;
   setIsInvalid: (isInvalid: boolean) => void;
+  isRequired: boolean;
   currentServiceAccountKeyId: ServiceAccountKeyId | null;
   className?: string;
 }
@@ -60,6 +61,7 @@ const FormControlServiceAccountKey = ({
   setServiceAccountKey,
   setServiceAccountKeyId,
   setIsInvalid,
+  isRequired,
   currentServiceAccountKeyId,
   className,
 }: FormControlServideAccountKeyFileProps) => {
@@ -143,8 +145,13 @@ const FormControlServiceAccountKey = ({
   };
 
   const formControl = (
-    <FormControl marginTop="spacingL" id="accountCredentialsFile">
-      <FormControl.Label>Google Service Account Key File</FormControl.Label>
+    <FormControl
+      marginTop="spacingL"
+      id="accountCredentialsFile"
+      isInvalid={formControlState.isInvalid}
+      isRequired={isRequired}
+    >
+      <FormControl.Label isRequired={isRequired}>Google Service Account Key File</FormControl.Label>
       <Textarea
         name="accountCredentialsFile"
         placeholder={placeholderText}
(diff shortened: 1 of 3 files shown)


apps/google-analytics-4/src/locations/ConfigScreen.spec.tsx
```diff
@@ -81,7 +94,99 @@ describe('Config Screen component', () => {
       result = await saveAppInstallation();
     });
 
+    console.log(mockSdk.app.getParameters());
```
andipaetzold, line 97: Forgotten while debugging

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
```diff
@@ -186,33 +188,37 @@ const FormControlServiceAccountKey = ({
...
   const formControl = (
-    <FormControl marginTop="spacingL" id="accountCredentialsFile">
-      <FormControl.Label>Google Service Account Key File</FormControl.Label>
+    <FormControl
+      marginTop="spacingL"
+      id="accountCredentialsFile"
+      isInvalid={formControlState.isInvalid}
+      isRequired={isRequired}
+    >
+      <FormControl.Label isRequired={isRequired}>Google Service Account Key File</FormControl.Label>
```
andipaetzold, line 197: Adding `isRequired` to the parent `FormControl` is enough. It will pass the prop down to its children

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
```diff
@@ -186,33 +188,37 @@ const FormControlServiceAccountKey = ({
...
         onChange={onKeyFileTextareaChange}
-        isInvalid={formControlState.isInvalid}
       />
       {formControlState.isInvalid ? (
         <FormControl.ValidationMessage>
           Error: {formControlState.errorMessage}
         </FormControl.ValidationMessage>
+      ) : serviceAccountKeyFile ? (
+        <Flex marginTop="spacingXs" alignItems="center">
+          <Flex isInline={true} alignItems="center">
```
andipaetzold, line 212: Double flex? What are you doing here?

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
```diff
@@ -186,33 +188,37 @@ const FormControlServiceAccountKey = ({
...
+            <CheckCircleIcon variant="positive" />{' '}
+            <Text as="p" marginLeft="spacing2Xs" fontColor="gray700">
```
loweisz, line 214: If you make this a `p` why not use the `Paragraph ` element from forma directly? 🤔

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
jsdalton: Whoops, thanks for catching that 👍

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
jsdalton: ahh yes, sorry this is a bit weird. basically i ended up using the lower level HTML from the component `FormControl.ValidationMessage`. More or less recreating this component: https://github.com/contentful/forma-36/blob/main/packages/components/forms/src/ValidationMessage/ValidationMessage.tsx#L40-L45

The reason being that I wanted a component that's more or less identical to the error to show a "valid" state, with the same height and style etc., just a different icon. The built in validation message is only for errors.

It's pretty much the same nesting of Flex elements they do there. Is there a better way, do you think?

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
jsdalton: Actually this is similar to what I explained with regards to @andipaetzold 's comment.

The idea here is that we flip between the "valid" state with a green check icon and the message and the "invalid" state with error icon and similar message.

The error message we get for free with Forma36.

The "valid" message isn't provided, so I just took the lower level components from here https://github.com/contentful/forma-36/blob/main/packages/components/forms/src/ValidationMessage/ValidationMessage.tsx#L40-L45 more or less verbatim. This also helped have a smooth transition between the two states, whereas `Paragraph` introduces extra margin which would need to be fixed.

In a perfect world our component library would provide something like this perhaps. 😄

apps/google-analytics-4/src/components/FormControlServiceAccountKey.tsx
andipaetzold: Okay, it's fine then. I should probably ask the same question about double Flex in the F36 repo :D

PR #2262, opened 2023-01-03
GA4: Set up basic lambda context [INTEG-92]

jsdalton:
Installs a basic lambda context and scopes the current app to a dedicated `frontend/` directory.

## Purpose

We will need a lambda backend to proxy API request to the Google Analytics API securely. This is because the credentials are secret and are shared for a given app installation, and thus we cannot make requests directly from the frontend.

To make review of PRs easier this big setup chunk of work is done in this PR.

## Approach

* Just tried to keep this as vanilla as possible. It's just an express app with standard tooling etc.
* No serverless stuff is installed yet
* There's a health endpoint more just as a sanity check. As a concept this seems like it's not useful in a Lambda


apps/google-analytics-4/lambda/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import app from './app';
+
+const port = 3000;
+
+app.listen(port, () => {
+  return console.log(`Express server is listening at http://localhost:${port}`);
```
shikaan, line 6: nit: no need to return void here

shikaan (commented):
Looks good, you may want to add a `serverless.yaml` file (or another manifest, if you want to go in another direction) to host the lambda. Serverless in particular comes with a few utilities that help you develop locally as well, without having to tinker with the setup too much

apps/google-analytics-4/lambda/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "lambda",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
```
kdamball, line 6: Missing deploy commands for this lambda

kdamball (commented):
Some deployment commands are missing. Additionally, the frontend is missing the app def id in `upload` command. It's fine to hardcode the app def id there as it's public anyway. We do something similar in other apps

jsdalton:
THanks @kdamball and @shikaan for the reviews. I have a separate ticket for doing the lambda and deployment work you both called out. @kdamball I also created a separate task for the frontend deployment. Basically, I just hadn't gotten that far yet.

kdamball (approved):

PR #2296, opened 2023-01-05
feat: set up initial serverless configuration

jsdalton:
## Purpose

To get the backend up and running we need to set up the serverless framework. First we'll aim for a working offline, then circle back and finish the production/staging deployment.

## Approach

* Mostly copy what we're doing elsewhere, using serverless-http and serveless-offline
* Using a different offline port so the backend doesn't collide with the frontend 3000

### DIFF
--- a/apps/google-analytics-4/lambda/serverless.yml
+++ b/apps/google-analytics-4/lambda/serverless.yml
@@ -1,0 +1,22 @@
+service: google-analytics
+frameworkVersion: '3'
+
+plugins:
+  - serverless-offline
+
+custom:
+  serverless-offline:
+    httpPort: 8080
+
+provider:
+  name: aws
+  runtime: nodejs16.x
+
+functions:
+  api:
+    handler: build/index.handler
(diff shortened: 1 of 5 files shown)


apps/google-analytics-4/lambda/serverless.yml
```diff
@@ -0,0 +1,22 @@
...
+plugins:
+  - serverless-offline
+
+custom:
+  serverless-offline:
+    httpPort: 8080
+
+provider:
+  name: aws
+  runtime: nodejs16.x
```
andipaetzold, line 13: New service, let's use node 18 straight away? (if possible)

andipaetzold (approved):

apps/google-analytics-4/lambda/serverless.yml
jsdalton: Ah great, I see they just added this in November.

I will flag this for a follow up PR 👍

PR #2299, opened 2023-01-07
feat: check credentials endpoint on app configuration screen [INTEG-97]

jsdalton:
Implements the "status" check on the installed credentials components.

## Purpose

When a user has installed service account credentials in the GA4 app, they want to know whether the credentials are in fact correct and "active" -- that is whether they actually are functional against the GA APIs we will use to query analytics data.

For now, we just need the basic plumbing in place to fetch the status from the lambda backend and display it on the front end. In subsequent PRs we will make the actual calls to GA APIs etc.

## Approach

image

* Hard code a status object to be returned from the Lambda API endpoint
* Create an API client to fetch from the Lambda
* Use [zod](https://github.com/colinhacks/zod) to handle schema validation and types. Super cool! (I saw that atools is using it in a handful of their projects)
* Use [msw](https://mswjs.io/) to handle mocking of server API responses, per the react testing library [examples](https://testing-library.com/docs/react-testing-library/example-intro/). Was actually quite delightful.
* Use dotenv for the backend server.

TODO
* [ ] Add an actual frontend component test verifying the "active"

### Intentionally not included of this PR
(will be handled in subsequent PRs)

* Display the error state and message a bit nicer (though the actual error is mostly an edge case when e.g. there's a problem connecting to our API, not to be confused with any errors the API itself encounters verifying credentials
* Consider the format of the object returned from 'api/credentials' (the simple thing we have now probably doesn't make sense)
* Actually making the API call to Google APIs using the credentials, to verify that they do indeed work correctly
* Handling states other than active (e.g. the credentials invalid/expired, lack adequeate permissions, etc)
* Signing the request to the lambda

### DIFF
--- a/apps/google-analytics-4/frontend/package.json
+++ b/apps/google-analytics-4/frontend/package.json
@@ -7,15 +7,15 @@
     "@contentful/f36-components": "4.23.3",
     "@contentful/f36-tokens": "4.0.1",
     "@contentful/react-apps-toolkit": "1.2.10",
-    "@testing-library/user-event": "^14.4.3",
-    "@types/lodash": "^4.14.191",
+    "@testing-library/user-event": "14.4.3",
+    "@types/lodash": "4.14.191",
     "contentful-management": "10.21.6",
     "emotion": "10.0.27",
-    "lodash": "^4.17.21",
+    "lodash": "4.17.21",
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "react-scripts": "5.0.1",
-    "zod": "^3.20.2"
+    "zod": "3.20.2"
   },
   "scripts": {
     "start": "cross-env BROWSER=none react-scripts start",
@@ -52,7 +52,7 @@
     "@types/react": "18.0.9",
     "@types/react-dom": "18.0.3",
     "cross-env": "7.0.3",
-    "msw": "^0.49.2",
+    "msw": "0.49.2",
     "typescript": "4.9.3"
   },
   "homepage": "."
--- a/apps/google-analytics-4/frontend/package.json
+++ b/apps/google-analytics-4/frontend/package.json
@@ -14,7 +14,8 @@
     "lodash": "^4.17.21",
     "react": "17.0.2",
     "react-dom": "17.0.2",
-    "react-scripts": "5.0.1"
+    "react-scripts": "5.0.1",
+    "zod": "^3.20.2"
   },
   "scripts": {
     "start": "cross-env BROWSER=none react-scripts start",
@@ -51,7 +52,11 @@
     "@types/react": "18.0.9",
     "@types/react-dom": "18.0.3",
     "cross-env": "7.0.3",
+    "msw": "^0.49.2",
     "typescript": "4.9.3"
   },
-  "homepage": "."
+  "homepage": ".",
+  "msw": {
+    "workerDirectory": "public"
+  }
 }
--- a/apps/google-analytics-4/frontend/package.json
+++ b/apps/google-analytics-4/frontend/package.json
@@ -55,8 +55,5 @@
     "msw": "^0.49.2",
     "typescript": "4.9.3"
   },
-  "homepage": ".",
-  "msw": {
-    "workerDirectory": "public"
-  }
+  "homepage": "."
 }
--- a/apps/google-analytics-4/frontend/src/components/InstalledServiceAccountKey.tsx
+++ b/apps/google-analytics-4/frontend/src/components/InstalledServiceAccountKey.tsx
@@ -3,7 +3,7 @@ import { ExternalLinkTrimmedIcon } from '@contentful/f36-icons';
 import tokens from '@contentful/f36-tokens';
 import { css } from 'emotion';
 import { useEffect, useState } from 'react';
-import { Api, ApiError, Credentials } from '../services/api';
+import { api, ApiError, Credentials } from '../services/api';
 
 import type { ServiceAccountKeyId } from '../types';
 
@@ -42,7 +42,6 @@ const InstalledServiceAccountKey = ({ serviceAccountKeyId }: InstalledServiceAcc
 
   useEffect(() => {
     (async () => {
-      const api = new Api();
       try {
         const response = await api.getCredentials();
         setCredentialsData(response);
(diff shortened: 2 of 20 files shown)


apps/google-analytics-4/frontend/package.json
```diff
@@ -14,7 +14,8 @@
     "lodash": "^4.17.21",
     "react": "17.0.2",
     "react-dom": "17.0.2",
-    "react-scripts": "5.0.1"
+    "react-scripts": "5.0.1",
+    "zod": "^3.20.2"
```
andipaetzold, line 18: Can we keep pinning the dependencies here?

apps/google-analytics-4/frontend/src/components/InstalledServiceAccountKey.tsx
```diff
@@ -34,6 +36,31 @@ interface InstalledServiceAccountKeyProps {
 }
 
 const InstalledServiceAccountKey = ({ serviceAccountKeyId }: InstalledServiceAccountKeyProps) => {
+  const [credentialsData, setCredentialsData] = useState<Credentials | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState('');
```
andipaetzold, line 41: nit: I prefer using `undefined` or `null` as "No error" to avoid empty strings.

apps/google-analytics-4/frontend/src/services/api.ts
```diff
@@ -0,0 +1,103 @@
+import { z } from 'zod';
+import { config } from '../config';
+
+const ZCredentials = z.object({
+  status: z.string(),
+});
```
andipaetzold, line 6: Why are we validating the response in the client? We own both, the server and client, so we can rely on the shape to be fine

apps/google-analytics-4/frontend/src/services/api.ts
```diff
@@ -0,0 +1,103 @@
...
+  async getCredentials(): Promise<Credentials> {
+    return await fetchFromApi<Credentials>(this.requestUrl('/api/credentials'), ZCredentials);
+  }
+
+  private requestUrl(apiPath: string): URL {
+    const fullApiPath = this.fullApiPath(apiPath);
+    return new URL(fullApiPath, this.baseUrl);
+  }
+
+  private fullApiPath(apiPath: string): string {
```
andipaetzold, line 95: Should this be in runtime code? Feels odd to check our config/code in the users' machines. We should rather be explicit about where to put slashes, maybe we can expect a leading slash in `requestUrl` using TypeScript.

andipaetzold (commented):
We currently have quite a bit of `console.error` in the code. Should this be centralized somewhere? Is is this a follow-up where we pipe things into Sentry?

apps/google-analytics-4/frontend/src/components/InstalledServiceAccountKey.tsx
```diff
@@ -34,6 +36,31 @@ interface InstalledServiceAccountKeyProps {
...
+
+  useEffect(() => {
+    (async () => {
+      const api = new Api();
```
andipaetzold, line 45: Every time we make an api request, we have to create a new instance of `Api()`. `Api` can be initialized without any arguments. Should this be a singleton?

andipaetzold (changes requested):
.

apps/google-analytics-4/frontend/src/services/api.ts
jsdalton: I've seen many cases where even in the same team people make changes to a server while forgetting to make or deploy a change to the client, so in general it's a defensive approach. It's to avoid the `const someResult = getResult() as Result` class of runtime errors.

zod (new to me) gives both runtime schema and compile time type validation "for free." In general it would be even be more beneficial for user input of course.

My preference here would be to run with it just to learn from using it how valuable / not valuable it is. (I had in mind to refactor some handwritten type assertions I did for the credentials input.)

andipaetzold: I am totally fine with using a validation library for user input (e.g. credential validation).

I am worried about using a library to validate our server response. The front end might still work even though the validation fails. Example: we add a second property to the credentials status response. The whole application will stop working and the user cannot do anything about that.
Instead, I'd try to prevent this issue during compile time. Can we for example share TypeScript types between the lambda and the frontend? That way we can guarantee that the shape is correct.

apps/google-analytics-4/frontend/src/services/api.ts
jsdalton: Yeah I agree here. It was a last minute thing I ran into that resulted from the serverless framework adding a "stage" to the first part of the path and then realizing that the JS Url class only uses the front part of the base URL provided and strips the path. I didn't decide or determine yet whether we should just disable the stage in the URL path entirely (which would mean this code is not necessary at all, since all the base URLs in the env var would just be root) or to handle the stage in some other way. I'm pretty sure we will not include it in production.

But if I hear you we could at minimum skip the regex bit and just merge the paths by ensuring the input is correct in the config vs doing it in runtime. I will look into that for this PR at minimum.

apps/google-analytics-4/frontend/src/components/InstalledServiceAccountKey.tsx
jsdalton: Yes it's a good question / call out. In truth we can just instantiate it in the config and use it directly. I can always move it out later if there was a need to initialize it with values.

(e.g. I was thinking about if you wanted to pass an AbortController to cancel the request, it might make better sense to initialize the client with the abort controller vs. to pass it down to every Api method we implement.)

jsdalton:
@andipaetzold Just to your top level comment:

> We currently have quite a bit of console.error in the code. Should this be centralized somewhere? Is is this a follow-up where we pipe things into Sentry?

👍  I like this idea but also the first step of centralizing even if we don't get it to Sentry. I will add a follow up task to this one on my back log to replace these one off console.error calls

apps/google-analytics-4/frontend/src/services/api.ts
jsdalton: In general I think the point of validating an API response from a server is to fail loudly and explicitly vs. quietly and implicitly. Adding new attributes is not a requirement since these schemas are strict about what's included not about "extra." Removing attributes as with any migration could break things, but again I'd see this as a feature (you forgot to update your client that a given attribute was removed). It's a bit less clear the value in such a super trivial use case as this one but becomes more valuable as an API becomes more complex. The major downtime is the runtime cost of evaluating the schema on every API result, which is important to be sure.

Having said that, while I'm a strong believer in these benefits, it's not a dealbreaker requirement at this stage and I agree that we can take advantage of sharing the same codebase here to streamline this whole thing. The shared types is a nice idea I will look into now. 😄

jsdalton: Hey let me know if you'd feel comfortable punting on this issue to a follow up PR.

I created a follow up ticket (INTEG-118) to address the issue of a shared schema (we're not using any kind of schema or types at all on the backend response). Sharing types would be super valuable but is non trivial to set up from a packaging perspective. 😢  But it's important so we definitely will circle back to find a good solution here.

andipaetzold (approved):

PR #2324, opened 2023-01-17
feat(Slack): add slack app [EXT-4253]

kdamball:
## Purpose
Introducing slack app using app actions.

## Approach
Similar to other apps with a lambda behind it, we have two folders, frontend & lambda. We have a some odd structure with an underlying package.json at the root app folder but we can, in the future, move both frontend & lambda to be independent similarly to other apps. Otherwise we have to rely on `yarn workspaces` which is odd with already lerna at the top level.


ghepting:
Is this just moving https://github.com/contentful/slack-app into this repository or does this require more scrutiny?

kdamball:
Pretty much that @ghepting. Maybe running it locally and report any oddities

ghepting (approved):
LGTM – I trust testing and verification to y'all :)

loweisz (approved):

PR #2327, opened 2023-01-18
feat: Last.fm example

harshil1712:
## Purpose

Adds an example to complement the tutorial - [Add musicals to your content](https://www.contentful.com/devrel-12-days/9)

## Approach

Created an example that walks the developer through each step of creating an app. It helps the developer learn about the following location - App Configuration, Entry Field, and Sidebar.


examples/last-fm/src/locations/Field.tsx
```diff
@@ -0,0 +1,70 @@
...
+interface Album {
+  name: string;
+  image: string;
+}
+
+const Field = () => {
+  const sdk = useSDK<FieldExtensionSDK>();
+  useAutoResizer();
+  const [albumSearch, setAlbumSearch] = useState<string>('');
+  const [albumData, setAlbumData] = useFieldValue<Album | null>();
```
andipaetzold, line 22: Nit: it's probably `undefined`, not `null`.

andipaetzold (changes requested):
Can we drop the tests (they are probably failing at the moment).
The we are fine to merge

andipaetzold (approved):

PR #2335, opened 2023-01-19
Gary/adds ga4 logo

ghepting:
## Purpose
Wanted to have this source file tracked which was used as the logo in our App Config

## Approach
Track these resources in git


jsdalton (approved):
Nice. We gotta stay up with the times!

PR #2338, opened 2023-01-20
[INTENG-94] Adds signed request verification to GA4 lambda

ghepting:
## Purpose
Validate that requests from the google-analytics-4 frontend to the backend lambda are genuine

## Approach
We are using the App Framework's [Request Verification](https://www.contentful.com/developers/docs/extensibility/app-framework/request-verification/) capabilities to send signed requests to the backend where we have the same signing secret available to authenticate the origin

App Framework documentation about Request Verification: https://www.contentful.com/developers/docs/extensibility/app-framework/request-verification/ 

## Deployment
We haven't deployed out lambda to production yet, but when we prepare to, we will need to fully configure the AWS Secrets Manager and the ssm secret parameters path in the `serverless.yml` configuration file. Note: @ghepting has the signing secret in his own development vault in 1password, because he is still waiting for access to the Contentful shared 1password vaults. This will need to be moved to be shared with the rest of the team when possible.

## Screenshots
Screenshot 2023-01-20 at 3 58 04 PM


apps/google-analytics-4/.gitignore
```diff
@@ -18,6 +18,9 @@
 .env.test.local
 .env.production.local
 
+# dotenv
+/**/*/.env
```
jsdalton, line 22: I wonder if we need this? Maybe instead we should suggest instrumenting the env with `.env.local` in our README so that we don't have to exclude this file. (I've seen some projects where keeping an .env with defaults in repo is a standard, with the expectation that overrides happen wih the .local pattern, e.g. what's above in the gitignore file a few linews)

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -1,17 +1,61 @@
...
+
+      // fetch google analytics credentials for the current entry
+      fetch(req.url, req)
+        .then((res) => {
+          if (res.status === 403) throw new Error('Unauthorized'); // TODO: add error state rendering instead of throwing exception
+
+          return res.json();
+        })
+        .then((credentials) => {
+          setIsLoading(false);
```
loweisz, line 52: I guess we want to stop loading in the catch case as well, so I would put the `setIsLoading(false)` in the .finally part

apps/google-analytics-4/.gitignore
ghepting: the frontend/.env.development file is _not_ gitignored currently and is holding non-sensitive configuration data.

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -1,17 +1,61 @@
...
 const Sidebar = () => {
   const sdk = useSDK<SidebarExtensionSDK>();
-  /*
-     To use the cma, inject it as follows.
-     If it is not needed, you can remove the next line.
-  */
-  // const cma = useCMA();
+  const cma = useCMA();
+  const [isLoading, setIsLoading] = useState(true);
+  const [credentials, setCredentials] = useState({});
```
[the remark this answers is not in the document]
ghepting: removed frontend changes in this PR, this is now obsolete, sorry!

apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
```diff
@@ -0,0 +1,38 @@
+import { CreateAppSignedRequestProps } from 'contentful-management/dist/typings/entities/app-signed-request';
+import { PlainClientAPI } from 'contentful-management/dist/typings/plain/common-types';
+
+async function makeSignedRequest(
```
jsdalton, line 4: Small naming thing but when I look at this function name and what the function does it does not make sense. I expect this to "make" (i.e. return) a signed Request. Instead We're actually _doing_ the signed request.

I'd suggest something like `fetchWithSignedRequest` or something that makes it clear that this fires off a (fetch) request.

apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
```diff
@@ -0,0 +1,38 @@
...
+      method: req.method,
+      headers: req.headers,
+      path: new URL(req.url).pathname,
+    }
+  );
+  Object.assign(req.headers, additionalHeaders);
+
+  // fetch google analytics credentials for the current entry
+  return fetch(req.url, req).then((res) => {
+    if (res.status === 403) throw new Error('Unauthorized');
```
jsdalton, line 32: This might be good for testing but I recommend we either:

* Just return the request without any handling. The 403 can get handled by the user of this function.
* Raise a more specific error that allows us to intentionally handle failed signature matches. Right now we have more than one behavior in our backend that oculd return a 403, e.g. you're not authorized to hit a Google API. This logic would swallow those errors. To avoid I think we should look at the message body (for an errorType indicating failed signature) and then raise a custom error, which can get handled.

apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
```diff
@@ -0,0 +1,38 @@
+import { CreateAppSignedRequestProps } from 'contentful-management/dist/typings/entities/app-signed-request';
+import { PlainClientAPI } from 'contentful-management/dist/typings/plain/common-types';
+
+async function makeSignedRequest(
+  url: URL,
+  appDefinitionId: string,
+  cma: PlainClientAPI,
+  method: CreateAppSignedRequestProps['method'] = 'GET',
+  headers: {} = {} // TOOD: improve HTTP request headers typing
```
jsdalton, line 9: If we haven't already let's capture this TODO in a subtask (because I think we should fix this and don't want to file it as "someday")

apps/google-analytics-4/lambda/config/serverless-env.dev.yml
```diff
@@ -0,0 +1 @@
+SIGNING_SECRET: IFnWcnKn-oOjWcERvh-GtcU62TotiPDvH59bD7ru2Jwb8BWn4X3iNXsRaWVxI9K0
```
jsdalton, line 1: Hmmm, this should be excluded via our .gitignore? 🤔

apps/google-analytics-4/lambda/src/app.spec.ts
```diff
@@ -8,12 +8,20 @@ import {
...
 import { GoogleApi } from './services/google-api';
+import * as NodeAppsToolkit from '@contentful/node-apps-toolkit';
 
 chai.use(chaiHttp);
 
 describe('app', () => {
   let mockClient: SinonStubbedInstance<AnalyticsAdminServiceClient>;
 
+  beforeEach(() => {
+    // TODO: set headers and fully test signature verification later
```
jsdalton, line 19: Still valid TODO? Let's also make sure to pull into subtask and not make it a someday.

apps/google-analytics-4/lambda/src/app.ts
```diff
@@ -8,21 +9,46 @@ import {
...
+  res.header('Access-Control-Allow-Origin', '*');
+  res.header('Access-Control-Allow-Methods', 'GET,PUT,POST,DELETE,OPTIONS');
+  res.header(
+    'Access-Control-Allow-Headers',
+    'Content-Type, Authorization, Content-Length, X-Requested-With'
+  );
+  res.send(200);
+});
+// validate signed requests
+app.use(['/api/credentials', '/api/account_summaries'], verifySignedRequestMiddleware);
```
jsdalton, line 31: Is wildcard syntax allowed here? e.g. `/api/*`?

apps/google-analytics-4/lambda/src/app.ts
```diff
@@ -8,21 +9,46 @@ import {
...
+  InvalidSignature: (e: InvalidSignature) => new ApiError(e.message, e.name, 403),
+  UnableToVerifyRequest: (e: UnableToVerifyRequest) => new ApiError(e.message, e.name, 422),
 };
 
 app.get('/health', function (_req, res) {
   res.status(204).send();
 });
 
 app.get('/api/credentials', (_req, res) => {
+  // throw new AuthorizationError(
```
jsdalton, line 48: We can safely remove the comments now

apps/google-analytics-4/lambda/src/app.ts
```diff
@@ -8,21 +9,46 @@ import {
...
+app.use(['/api/credentials', '/api/account_summaries'], verifySignedRequestMiddleware);
+
 // Maps an error class name -> a handler function that takes the error of that type as input and returns
 // a correctly configured API error as output.
 const errorClassToApiErrorMap: ApiErrorMap = {
   GoogleApiError: (e: GoogleApiError) => new ApiError(e.details, e.name, e.httpStatus),
   GoogleApiClientError: (e: GoogleApiClientError) => new ApiError(e.details, e.name, e.httpStatus),
   GoogleApiServerError: (e: GoogleApiServerError) => new ApiError(e.details, e.name, e.httpStatus),
+  InvalidSignature: (e: InvalidSignature) => new ApiError(e.message, e.name, 403),
+  UnableToVerifyRequest: (e: UnableToVerifyRequest) => new ApiError(e.message, e.name, 422),
```
jsdalton, line 40: Per my comment on the frontendcode, both of these errorTypes should be handled explicitly inside our `makeSignedRequest` function (I believe)

apps/google-analytics-4/lambda/src/lib/errors/middlewares.ts
```diff
@@ -11,7 +11,7 @@ export type ApiErrorMap = Record<string, (e: any) => ApiError<Record<string, unk
 
 export const apiErrorHandler: ErrorRequestHandler = (error, _request, response, next) => {
   if (error) {
-    console.log(JSON.stringify(error));
+    console.error(JSON.stringify(error));
```
jsdalton, line 14: We need a proper logger! 🙈  (but not this very second)

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    sinon.stub(NodeAppsToolkit, 'verifyRequest').get(() => verifyRequestStub);
+  });
+
+  describe('given an unsigned request', () => {
+    before(() => {
+      verifyRequestStub.returns(false);
+    });
+
+    it('returns 403 Unauthorized', async () => {
+      const response = await chai.request(app).get('/api/credentials');
```
jsdalton, line 23: We will probably want to refactor this test or figure out another way to integration test it. E.g. here I'd like to remove the /api/credentials endpoint, but the presence of this line will block that.

(I ran into the same testing question when I considred writing integration tests for our mapper)

jsdalton: We could do what I did in `apiErrorMapper` and test the middleware directly btw.

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -0,0 +1,87 @@
...
+import { InvalidSignature } from '../lib/errors/invalid-signature';
+import { UnableToVerifyRequest } from '../lib/errors/unable-to-verify-request';
+import { IncomingHttpHeaders } from 'http';
+
+export const verifySignedRequestMiddleware = (
+  req: express.Request,
+  _res: express.Response,
+  next: express.NextFunction
+) => {
+  const signingSecret = (process.env.SIGNING_SECRET || '').trim();
```
jsdalton, line 17: I think we should (in a follow up PR, not now) formalize our env var usage via a config instead of hard coding here. This is what I did for the frontend https://github.com/contentful/apps/blob/apps/ga4/apps/google-analytics-4/frontend/src/config.ts

Specifically we should fail at start if the env var is missing rather than attempting to fail gracefully as we're doing here. (This approach will just nicely give customers signing errors instead of telling us developers that we messed up our deploy 😉 )

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  next();
+};
+
+const makeCanonicalReq = (req: express.Request) => {
+  const requiredHeaders = requestSigningHeaders(req.headers);
+
+  return <CanonicalRequest>{
+    method: req.method,
+    path: `/dev${req.originalUrl}`, // TODO: wtf express? how do i get the real request path?!
```
jsdalton, line 45: This actually needs to be fixed before we merge I think.

We could make the stage an env var? Not sure. Pretty annoying how this all works!

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -0,0 +1,87 @@
...
+  'x-contentful-signed-headers': string;
+  'x-contentful-signature': string;
+  'x-contentful-user-id': string;
+  'x-contentful-space-id': string;
+  'x-contentful-environment-id': string;
+  'x-contentful-app-id': string;
+  [key: string]: string;
+}
+
+// TODO: can we refactor to sth like this?
```
jsdalton, line 66: I'd ditto capturing the TODO as a task or maybe leaving it out otherwise. (No strong feelings either way)

jsdalton (changes requested):
Nice! Some small clean up stuff I think we should tackle. Other refactor kinds of suggestions I'd be more than happy to push to a follow up PR so we can optimize toward merging this.

apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
ghepting: done

apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
ghepting: refactored to not parse the response or throw error.

apps/google-analytics-4/lambda/config/serverless-env.dev.yml
ghepting: ah, damnit! regenerated a fresh one that isn't compromised.

apps/google-analytics-4/lambda/src/lib/errors/middlewares.ts
ghepting: Yeah, what are other folks using? I'd love to see us adopt Datadog for *everything* (logging, metrics, tracing, RUM, etc)

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
ghepting: Agreed!

ghepting: Created a ticket: https://contentful.atlassian.net/browse/INTEG-128

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
ghepting: i know! yeah i'll take another pass at this now

apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
ghepting: Okay I think this is better:

image

image

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
ghepting: oh i left this as a question i wanted to ask @pkeavenycontentful for input/pairing on! I couldn't get this working for the life of me 😅

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.spec.ts
ghepting: Yeah i had it like that at first, doing a mock app instance, but i wasn't a fan of that because we have 3 middlewares all working together on the full response expectation for this integration test. Seems valuable to just use a real route and change the path if that endpoint goes away.

ghepting: discussed offline and will keep like this for now, but consider adding more isolated unit tests for compartmentalization later on

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
ghepting: (removed for now and will pair with Patrick to see what he thinks!)

apps/google-analytics-4/lambda/src/app.ts
```diff
@@ -8,21 +9,46 @@ import {
...
+
 const app = express();
 
+// allow all OPTIONS requests
+app.options('/*', function (req, res, _next) {
+  res.header('Access-Control-Allow-Origin', '*');
+  res.header('Access-Control-Allow-Methods', 'GET,PUT,POST,DELETE,OPTIONS');
+  res.header(
+    'Access-Control-Allow-Headers',
+    'Content-Type, Authorization, Content-Length, X-Requested-With'
```
jsdalton, line 26: Just noticing this on my second pass here but I'm not sure the value we're passing here is correct for us. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers

Specifically:

* I believe we'd need to whitelist explicitly the headers we do want to accept (e.g. X-Environment-Id etc.)
* Conent-Type is not needed here since it's auto safelisted https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header

jsdalton (approved):
Just one small question about our OPTION response but I think we can deal with it in a follow up. Looking good otherwise! 😎

apps/google-analytics-4/lambda/src/app.ts
ghepting: ah, good catch! this was definitely not complete and I totally overlooked this, too

ghepting: - [] fixme

apps/google-analytics-4/lambda/src/app.spec.ts
ghepting: 👍

apps/google-analytics-4/lambda/src/app.ts
ghepting: it's documented so, but I believe we couldn't get it to work and punted on it. i can test once more tomorrow before i merge if you'd like

apps/google-analytics-4/lambda/src/app.ts
ghepting: fixed

apps/google-analytics-4/lambda/src/app.ts
ghepting: i believe this is obsolete now – discussed on slack w/u :)

apps/google-analytics-4/lambda/src/app.ts
```diff
@@ -8,14 +9,44 @@ import {
...
+      'X-Contentful-Timestamp',
+      'X-Contentful-Signed-Headers',
+      'X-Contentful-Signature',
+      'X-Contentful-User-ID',
+      'X-Contentful-Space-ID',
+      'X-Contentful-Environment-ID',
+      'X-Contentful-App-ID',
+    ].join(', ')
+  );
+  res.send(200);
```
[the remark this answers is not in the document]
ghepting: 👌 will update, thx :)

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+  'x-contentful-signed-headers': string;
+  'x-contentful-signature': string;
+  'x-contentful-user-id': string;
+  'x-contentful-space-id': string;
+  'x-contentful-environment-id': string;
+  'x-contentful-app-id': string;
+  [key: string]: string;
+}
+
+function requestSigningHeaders(headers: IncomingHttpHeaders): Partial<ContentfulSignedHeaders> {
```
andipaetzold, line 66: Isn't this already part of node-apps-toolkit?

https://github.com/contentful/node-apps-toolkit/blob/master/src/requests/utils.ts#L54

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+  req: express.Request,
+  _res: express.Response,
+  next: express.NextFunction
+) => {
+  const signingSecret = (process.env.SIGNING_SECRET || '').trim();
+  const canonicalReq = makeCanonicalReq(req);
+  let isValidReq = false;
+
+  try {
+    isValidReq = NodeAppsToolkit.verifyRequest(signingSecret, canonicalReq, 60); // 60 second TTL
```
andipaetzold, line 22: nit: extract 60 as a constant

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.error(e);
+    throw new UnableToVerifyRequest('Unable to verify request', {
+      cause: e,
+    });
+  }
+
+  if (!isValidReq) {
+    throw new InvalidSignature(
+      'Request does not have a valid request signature. ' +
+        'See: https://www.contentful.com/developers/docs/extensibility/app-framework/request-verification/'
```
andipaetzold, line 33: Why are we logging this? This is not a library where we'd need to inform people what to do.

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+    );
+  }
+
+  next();
+};
+
+const makeCanonicalReq = (req: express.Request) => {
+  const requiredHeaders = requestSigningHeaders(req.headers);
+
+  return <CanonicalRequest>{
```
andipaetzold, line 43: It's not recommended to cast types in TypeScript. I'd set the return value type in line 40. You probably don't get type warnings here.

apps/google-analytics-4/lambda/src/app.ts
```diff
@@ -8,14 +9,44 @@ import {
   GoogleApiServerError,
 } from './services/google-api';
 import { ServiceAccountKeyFile } from './types';
+import { verifySignedRequestMiddleware } from './lib/verify-signed-request-middleware';
+import { InvalidSignature } from './lib/errors/invalid-signature';
+import { UnableToVerifyRequest } from './lib/errors/unable-to-verify-request';
```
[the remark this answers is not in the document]
ghepting: yeah not a bad idea, this is messy!

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+) => {
+  const signingSecret = (process.env.SIGNING_SECRET || '').trim();
+  const canonicalReq = makeCanonicalReq(req);
+  let isValidReq = false;
+
+  try {
+    isValidReq = NodeAppsToolkit.verifyRequest(signingSecret, canonicalReq, 60); // 60 second TTL
+  } catch (e) {
+    console.error(e);
+    throw new UnableToVerifyRequest('Unable to verify request', {
```
andipaetzold, line 25: Don't you call `next(error)`? That will then be caught in the error handler. But I am not an express expert.

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
ghepting: Ah, yeah just for us if we are doing/forming anything wrong :)

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
ghepting: oh yes, let me try that. i'm very new to express and I was just tying this into the custom error handling middleware added by @jsdalton in a likely very ignorant way :) – thanks for the note!

andipaetzold: I think that should still be possible with `next(new UnableToVerifyRequest(...))`. But maybe both works and both is valid.

ghepting: oh! nifty! I'll give that a try shortly :) 

Thanks @andipaetzold! I always like to learn the right way, when I'm doing something stupid/implementing an anti-pattern like this. Thanks for the time & energy you're spending to level me up 😄

PR #2353, opened 2023-01-24
refactor: [] page location example

loweisz:
## Purpose
Update the page location example. Refactored the components using the vite template. 
Changed the class components to functional components, replaced the space API with cma adapter and used the newest version of react router v6.


colomolo:
Why does it have two package.json and package-lock.json files?

loweisz:
> Why does it have two package.json and package-lock.json files? Same with tsconfig, .gitignore and other files.

Good call, was a copy past error

andipaetzold (commented):
We don't have any tests, should we still add the test setup?

andipaetzold:
> We don't have any tests, should we still add the test setup?

We can remove two deps: vitest, happy-dom. Shorten the config and remove the test script

loweisz:
> > We don't have any tests, should we still add the test setup?
> 
> We can remove two deps: vitest, happy-dom. Shorten the config and remove the test script

Makes sense 👍

PR #2355, opened 2023-01-25
Feature/google api error states

jsdalton:
## Purpose

* Our calls to Google API can sometimes fail (e.g. bad credentials, account not authorized, etc.). If that happens we need to reply back to the client with a sensible, well structured error message describing the problem.
* In general any error that might occur needs to fail sensible from an API client perspective

## Approach

* Via testing determine some known errors from GoogleAPI
* Generally try to sort "client errors" and "server errors" into two buckets to help reason about responses
* Add a middleware to throw errors. NOTE: This is WIP, committing part 1 now to unblock the team in our other work

TODO

* [ ] Actually map the errors raised by the Google API client to the new ApiError class with correct status etc.


ghepting (approved):
LGTM!

PR #2398, opened 2023-01-30
adds eslint-plugin-react-hooks

ghepting:
## Purpose
get better linting to catch useEffect / other react hook dependencies, etc.

## Approach
This is the Way™!

https://www.npmjs.com/package/eslint-plugin-react-hooks

### DIFF
--- a/apps/google-analytics-4/lambda/package.json
+++ b/apps/google-analytics-4/lambda/package.json
@@ -29,6 +29,7 @@
     "chai": "^4.3.7",
     "chai-http": "^4.3.0",
     "eslint": "^8.30.0",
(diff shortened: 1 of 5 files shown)


apps/google-analytics-4/lambda/package.json
```diff
@@ -29,6 +29,7 @@
     "chai": "^4.3.7",
     "chai-http": "^4.3.0",
     "eslint": "^8.30.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
```
andipaetzold, line 32: Why add a react lint plugin to the lambda? we don't have react here.

PR #2401, opened 2023-01-31
address nits from INTENG-94 PR

ghepting:
## Purpose
Addressing post hoc PR feedback

### DIFF
--- a/apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
+++ b/apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
@@ -1,4 +1,4 @@
-import * as NodeAppsToolkit from '@contentful/node-apps-toolkit';
+import { verifyRequest } from '@contentful/node-apps-toolkit';
 import {
   CanonicalRequest,
   ContentfulContextHeader,
@@ -17,9 +17,10 @@ export const verifySignedRequestMiddleware = (
   const signingSecret = (process.env.SIGNING_SECRET || '').trim();
   const canonicalReq = makeCanonicalReq(req);
   let isValidReq = false;
+  const TTL = 60; // TTL for signed requests (in seconds)
 
   try {
-    isValidReq = NodeAppsToolkit.verifyRequest(signingSecret, canonicalReq, 60); // 60 second TTL
+    isValidReq = verifyRequest(signingSecret, canonicalReq, TTL);
   } catch (e) {
     console.error(e);
     throw new UnableToVerifyRequest('Unable to verify request', {
--- a/apps/google-analytics-4/lambda/src/app.ts
+++ b/apps/google-analytics-4/lambda/src/app.ts
@@ -34,7 +34,7 @@ app.options('/*', function (req, res, _next) {
       'X-Contentful-App-ID',
     ].join(', ')
   );
-  res.send(200);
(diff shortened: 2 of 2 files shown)


apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -8,6 +8,7 @@ import express from 'express';
 import { InvalidSignature } from '../lib/errors/invalid-signature';
 import { UnableToVerifyRequest } from '../lib/errors/unable-to-verify-request';
 import { IncomingHttpHeaders } from 'http';
+import { HttpMethod } from 'googleapis-common';
```
jsdalton, line 11: This might have been unintentional auto importing. I don't think we want to get the HttpMethod from the google apis lib though, it creates an unnecessary dependency between our  middleware and the fact that we happen to use google api libs elsewhere.

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
```diff
@@ -37,13 +39,13 @@ export const verifySignedRequestMiddleware = (
...
 };
 
-const makeCanonicalReq = (req: express.Request) => {
+const makeCanonicalReq = (req: express.Request): CanonicalRequest => {
   const requiredHeaders = requestSigningHeaders(req.headers);
 
-  return <CanonicalRequest>{
-    method: req.method,
+  return {
+    method: req.method as HttpMethod,
```
jsdalton, line 46: Would it just except `string` instead of relying on the google api lib? (per my comment above)

jsdalton (approved):
A few small comments from me, but nothing major. 👍

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.ts
ghepting: ah gross, you're totally right :(

ghepting: proposal: since this app is _for google analytics_ is it reasonable to punt on this nit? i _absolutely_ get your point, but feel it's low value to redeclare this type interface _just in the name of "purity"_

apps/google-analytics-4/lambda/src/lib/verify-signed-request-middleware.spec.ts
```diff
@@ -1,26 +1,59 @@
...
+  const timestamp = '1675203823305';
+  const spaceId = '30x8uoqewkkz';
+  const environmentId = 'master';
+  const userId = '6iUaAfJg2ZoKe4oEYF7kxe';
 
   beforeEach(() => {
-    // TODO: set headers and fully test signature verification later
-    sinon.stub(NodeAppsToolkit, 'verifyRequest').get(() => verifyRequestStub);
+    process.env.STAGE = 'dev'; // don't change this without recomputing the signed request signature
+    process.env.SIGNING_SECRET = 'mFjgMwlcE5iHKl855rGk9adWTP9ZagSamQctAElpptMtpGQmNIuGoRWOZZ_lhU_4';
```
andipaetzold, line 18: Wouldn't the clean solution be to pass in the env variables via a function parameter. Why not do that instead?

PR #2430, opened 2023-02-03
Feature/service account key api passthrough [INTEG-130]

jsdalton:
## Purpose

One of the final pieces of "plumbing" needed to make correctly authenticated requests against Google APIs using the credential ID the app stores in config is to actually pass those values through from the front end to the backend. Then the backend needs to take those values and use them to make the API calls.

## Approach

* On the frontend, pull out the API instantiation into a hook to make it easier to use with correctly parameterized key from installation
* A lot of work on the frontend was needed to pass through the service account key, even though later we will remove it (we will never have access to this value once it's in secret storage)
* Encode and pass values down as base64 encoded headers
* *IMPORTANT*: For development we are passing both the service account key id AND the service account key file. This is totally incorrect / insecure handling of the key file and is only being done for development. Once we introduce secret storage only the key id will be passed down and the other code will be removed (and replaced with a fetch on the backend to the secret storage using the id found in the service account key id)
* needed to rework our signature verification middleware to accommodate headers that don't need to be signed.
* 

In addition to the code being removed there is some other refactor work needed here:
* the shared types and mocks are getting pretty glaring and should really be moved into a shared package!
* the fetchAPI should likewise be pulled out separate from the API. There's already a ticket for this: INTEG-137


apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
```diff
@@ -6,16 +6,16 @@ async function fetchWithSignedRequest(
   appDefinitionId: string,
   cma: PlainClientAPI,
   method: CreateAppSignedRequestProps['method'] = 'GET',
-  headers: Record<string, string> = {}
+  unsignedHeaders: Record<string, string> = {}
```
ghepting, line 9: this is still a smell to me. i'm going to work next week on making all of this not an issue on the backend. the signing validation logic should be the only thing responsible for "knowing" what headers matter and don't. Thanks for getting it working with your new additions temporarily either way though 🙇

apps/google-analytics-4/lambda/src/middlewares/service-account-key-provider.ts
```diff
@@ -0,0 +1,84 @@
...
+    );
+  }
+  const serviceAccountKeyId = decodeServiceAccountHeader(serviceAccountKeyIdHeaderValue);
+  assertServiceAccountKeyId(serviceAccountKeyId);
+  req.serviceAccountKeyId = serviceAccountKeyId;
+
+  // Pulling the account key from the header will be removed later and replaced with fetching
+  // the code directly from secret storage!
+  const serviceAccountKeyHeaderValue = req.headers['x-contentful-serviceaccountkey'];
+  if (!serviceAccountKeyHeaderValue || typeof serviceAccountKeyHeaderValue !== 'string') {
```
ghepting, line 21: I think this typeof check is superfluous unless you were dealing with express framework typecasting header values?

apps/google-analytics-4/lambda/src/middlewares/service-account-key-provider.ts
```diff
@@ -0,0 +1,84 @@
...
+    throw new InvalidServiceAccountKey('Improperly formatted service account key', { cause: e });
+  }
+};
+
+function assertServiceAccountKeyId(value: unknown): asserts value is ServiceAccountKeyId {
+  const requiredFields = ['clientEmail', 'projectId', 'id'];
+  for (const requiredField of requiredFields) {
+    if (!value || typeof value !== 'object' || !(requiredField in value)) {
+      throw new InvalidServiceAccountKey(
+        `'${requiredField}' is missing from Service Account Key header`
```
ghepting, line 47: nit: remove ` header` here so this still makes sense once this is being extracted from secrets manager:

```
`'${requiredField}' is missing in Service Account Key data`
```

apps/google-analytics-4/lambda/src/middlewares/service-account-key-provider.ts
```diff
@@ -0,0 +1,84 @@
...
+      );
+    }
+
+    // @ts-expect-error we just confirmed above that requiredField is in value
+    if (typeof value[requiredField] !== 'string') {
+      throw new InvalidServiceAccountKey(
+        `'${requiredField}' must be a string in Service Account Key header`
```
ghepting, line 54: (same here: reword to exclude ` header`)

apps/google-analytics-4/lambda/src/middlewares/service-account-key-provider.ts
ghepting: Also, this same feedback in the other instances of this messaging below, too

apps/google-analytics-4/lambda/src/middlewares/service-account-key-provider.ts
jsdalton: yeah, that's the issue. express says this could be a string _or_ an array of strings. e.g if your request is -h "foo=bar" -h "foo=baz" then your header `request.headers['foo']` is going to be `['bar', 'baz']`.

apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
jsdalton: Yeah that's fair. I think we can/shoiuld fold it into a quick (but significant) refactor of this fetching bits. This is what INTEG-137 is about for me. We can strategize about specifics when one of us picks it up.

apps/google-analytics-4/lambda/src/middlewares/service-account-key-provider.ts
jsdalton: Sure, can do. (FWIW I believe the secret storage bit will be entirely new / non backward compatible error messages, which i consider fine for this development stage 😉 )

apps/google-analytics-4/lambda/src/middlewares/service-account-key-provider.ts
ghepting: yep, k

apps/google-analytics-4/frontend/src/helpers/signed-requests.ts
ghepting: k, sg

ghepting (approved):

PR #2462, opened 2023-02-06
chore: add semver []

kdamball:
## Purpose
Due to failures in releases in master - we should use the correct vault orb as other public repos

## Approach
semantic-release-ecosystem is the preferred way to go forward.

### DIFF
--- a/.contentful/vault-secrets.yaml
+++ b/.contentful/vault-secrets.yaml
@@ -5,6 +5,6 @@ services:
       - dependabot
   circleci:
     policies:
-      - semantic-release
+      - semantic-release-ecosystem
       - aws-push-artifacts:
           account-id: "017078452822"


colomolo (approved):
Thanks!

andipaetzold:
You need to update config.yml as well, right?

https://github.com/contentful/apps/blob/master/.circleci/config.yml#L11

kdamball:
@andipaetzold you can only update that once `vault.yaml` changes are in master

kdamball:
And `package.json`

omasopust:
@kdamball We are seeing permission errors now in Vault for this project, is the usage change on the way?

kdamball:
@omasopust follow up here: https://github.com/contentful/apps/pull/2467

PR #2486, opened 2023-02-08
chore(lerna): avoid pushing git tags []

kdamball:
## Purpose
It's becoming impractical to catch up to all git tag conflicts and in turn we can't deploy any changes. We only need publishing for app packages which do not upgrade as often. The conflicts are usually in apps directory which doesn't need any tagging.

## Approach
Remove lerna git tagging and instead use default lerna publishing. 

- [x] Qn: will this still publish without tags?

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "test": "lerna run test:ci --concurrency=1 --stream",
     "deploy": "lerna run deploy --concurrency=3",
     "deploy:test": "lerna run deploy:test --concurrency=3",
-    "publish-packages": "lerna version --conventional-commits --create-release github --yes && lerna publish from-git --yes",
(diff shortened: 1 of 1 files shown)


andipaetzold (commented):
* Why did you remove the `--yes` and `--conventional-commits`?
* When running `lerna publish`, `lerna version` is run in the background (https://github.com/lerna/lerna/tree/main/libs/commands/publish#readme). Don't we want to run `lerna publish from-packages` here?

kdamball:
@andipaetzold removed `--conventional-commits`. I assumed `github` was tied to the `--yes` command. Will readd those two. Will also swap to `from-packages`. Didn't know of that option

andipaetzold (approved):
I think this works. But I am not 100% sure. Let's 🤞.

PR #2575, opened 2023-02-22
- [INTEG-163] adds sentry to the frontend of GA4 app

ghepting:
## Purpose
Adds observability of unhandled exceptions to the frontend of the GA4 app via @sentry/react SDK

## Approach
adds and configures the @sentry/react SDK

- https://github.com/getsentry/sentry-javascript/tree/master/packages/react 
- semi-related pitch from Ecosystem Extensibility team: https://contentful.atlassian.net/wiki/spaces/ECO/pages/4141350915/Pitch+-+App+Monitoring 

## Deployment
- [x] Verify both environment (node env) and release (git commit hash) are set correctly in CI
- [x] enable localhost filter in Sentry project settings

### DIFF
--- a/apps/google-analytics-4/frontend/src/config.ts
+++ b/apps/google-analytics-4/frontend/src/config.ts
@@ -8,4 +8,7 @@ function getEnvironmentVariable(environmentVariableName: string) {
 
 export const config = {
   backendApiUrl: getEnvironmentVariable('REACT_APP_BACKEND_API_URL'),
+  version: getEnvironmentVariable('REACT_APP_VERSION'),
+  release: getEnvironmentVariable('REACT_APP_RELEASE'),
+  sentryDSN: getEnvironmentVariable('REACT_APP_SENTRY_DSN'),
 };
--- a/apps/google-analytics-4/frontend/src/index.tsx
+++ b/apps/google-analytics-4/frontend/src/index.tsx
@@ -1,3 +1,5 @@
+import * as Sentry from '@sentry/react';
+import { BrowserTracing } from '@sentry/tracing';
 import React from 'react';
 import { render } from 'react-dom';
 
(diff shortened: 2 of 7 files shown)


apps/google-analytics-4/frontend/src/config.ts
```diff
@@ -8,4 +8,7 @@ function getEnvironmentVariable(environmentVariableName: string) {
 
 export const config = {
   backendApiUrl: getEnvironmentVariable('REACT_APP_BACKEND_API_URL'),
+  version: getEnvironmentVariable('REACT_APP_VERSION'),
+  release: getEnvironmentVariable('REACT_APP_RELEASE'),
```
ryunsong-contentful, line 12: I only comment here to pick your brain and learn. So I see this env variable is set by the scripts in package.json. What are the trade-offs of putting this env variable in the .env file and doing something like this `https://stackoverflow.com/questions/34650527/how-to-use-environment-variables-in-package-json` versus what you have now?

ryunsong-contentful (approved):

apps/google-analytics-4/frontend/src/config.ts
ghepting: i put everything except the git commit hash in the dotenv files – unless I'm missing some knowledge, i don't think you can interpolate arbitrary bash commands within a dotenv file? I would be thrilled to learn I'm wrong and be taught a way as I sunk 20 minutes into try to do exactly that! 😅

apps/google-analytics-4/frontend/src/index.tsx
```diff
@@ -6,6 +8,19 @@ import { SDKProvider } from '@contentful/react-apps-toolkit';
...
+
+Sentry.init({
+  dsn: config.sentryDSN,
+  integrations: [new BrowserTracing()],
+  // Set tracesSampleRate to 1.0 to capture 100%
+  // of transactions for performance monitoring.
+  // We recommend adjusting this value in production
+  tracesSampleRate: 1.0,
+  environment: process.env.NODE_ENV,
+  // TODO: setup Sentry as part of release pipeline (see: https://docs.sentry.io/platforms/javascript/sourcemaps/?_ga=2.56533545.342806665.1676988870-873194326.1675171780#uploading-source-maps-to-sentry)
```
ghepting, line 21: ticket filed for this. will remove this TODO when it's been completed

ghepting: @mgoudy91 might be a good one for you to pick up for us while you're in the build stuff https://contentful.atlassian.net/browse/INTEG-171

lilbitner (approved):

PR #2530, opened 2023-02-15
INTEG-133/ ga-4 domain management and API gateway

mgoudy91:
## Purpose
[JIRA ticket](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?modal=detail&selectedIssue=INTEG-133)
Sets up serverless configuration so that we can deploy GA-4 to production, securely.

## Approach
Mostly constructed off existing slack/netlify serverless.yml files. In the future, we may implement dynamodb to host some of the secrets we'll use here.

## Dependencies and/or References
[AWS notes from discussion with Extensibility](https://docs.google.com/document/d/1PQhyT2gnQ5Q-BO19x2xfZlE282KESRiFvjzhT0RsdUE/edit?usp=sharing)

## Deployment
Deployment of deployment configuration files is always a bit more complex, no known concerns at this time though.

PR #2577, opened 2023-02-22
Revert "INTEG-133/ ga-4 domain management and API gateway (#2530)"

mgoudy91:
This reverts commit 5aacd356665935f5ebc2f30aad978f693d78564f.

## Purpose
We broke our dev process in addition to this not working for our deploy needs. We'll use this code later but need to un-stuck the team now.

## Approach
Doing a single revert (the squashed merge commit) for now. We'll revert-revert this later once we're in a better place.


jsdalton (approved):
👍

pkeavenycontentful (approved):

PR #2585, opened 2023-02-23
Expose Wistia hashed_id for searching instead of only title

caomicc:
## Purpose
My team needs to be able to search wistia videos by `hashed_id` instead of only by `name` for migration to our new site as our current site only references hashes

## Approach
This was a quick add - just an `or` that says serve the videos matching `name` OR `hashed_id`

N/A

## Deployment
No changes


ghepting:
Hey @caomicc – sounds like a useful and pretty straight forward tweak, but I wonder what kind of confusion this would cause for the average user getting unexpected results typing in a few random characters. Have you considered just forking and installing your own instance of a custom app definition as a temporary solution vs trying to update the upstream app logic?

PR #2592, opened 2023-02-24
[INTEG-164]: Content Editors can view additional analytics data in their sidebar

whitelisab:
## Purpose
Content editors need to be able to open a Contentful entry where the GA4 app is installed, and see a sidebar app. This sidebar app has a chart of view data alongside other data above and below the chart, including metric showing page views, a dropdown to select a date range, and a link to view the data in GA4. This PR adds the components displaying the information above and below the chart. The sample data used to render these components was also updated to utilize the date dimension to organize views by date.

## Approach
I created a SidebarAnalytics component that is rendered by the Sidebar location. This component renders Header and Footer components, and eventually will render the Chart component once that is added. The Sidebar Analytics component uses the updated JSON sample data.

[Google Analytics Data API (GA4) Dimensions](https://developers.google.com/analytics/devguides/reporting/data/v1/api-schema#dimensions)


apps/google-analytics-4/frontend/src/components/sidebar/SidebarAnalytics.tsx
```diff
@@ -0,0 +1,46 @@
...
+      setRunReportResponse(await response.json());
+    }
+
+    fetchData();
+  }, []);
+
+  const pageViews =
+    runReportResponse &&
+    runReportResponse.rows.reduce((acc, val) => {
+      const metric = Number(val.metricValues[0].value);
```
pkeavenycontentful, line 28: You can use the unary Typescript operator instead of the `Number` class to convert the string value to a number more performantly: 
```
const metric = +val.metricValues[0].value
```

apps/google-analytics-4/frontend/src/components/sidebar/SidebarAnalytics.tsx
```diff
@@ -0,0 +1,46 @@
...
+      return acc + metric;
+    }, 0);
+
+  const metricName = runReportResponse && runReportResponse.metricHeaders[0].name;
+
+  return (
+    <>
+      <ChartHeader
+        metricName={metricName !== undefined ? metricName : ''}
```
pkeavenycontentful, line 37: TypeScript truthy checks should include `null, undefined, 0, false, and empty string`, is there a reason for the extra check for `!== undefined`?

apps/google-analytics-4/frontend/src/types.ts
```diff
@@ -22,3 +22,30 @@ export interface ServiceAccountKeyId {
...
   clientId: string;
 }
+
+export interface Row {
+  dimensionValues: { value: string; oneValue: string }[];
+  metricValues: { value: string; oneValue: string }[];
+}
+
+export interface RunReportResponse {
+  dimensionHeaders: { name: string }[];
```
pkeavenycontentful, line 32: 💯 on this interface

apps/google-analytics-4/frontend/src/types.ts
```diff
@@ -22,3 +22,30 @@ export interface ServiceAccountKeyId {
   projectId: string;
   clientId: string;
 }
+
+export interface Row {
```
ryunsong-contentful, line 26: nit/sure: Does this need to be exported or will it only ever be used by RunReportResponse and RunReportResponse is what will be used by other components?

apps/google-analytics-4/frontend/src/components/sidebar/ChartHeader.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface Props {
+  metricName: string;
+  metricValue: string;
+}
+
+export default function ChartHeader(props: Props) {
+  const { metricName, metricValue } = props;
+
+  const [dateSelection, setDateSelection] = useState('lastWeek');
```
ryunsong-contentful, line 12: nit: missing typing on useState ie - `useState<string>('lastWeek');`

ryunsong-contentful: Or better yet - is there a Interface/Type for this? You definitely know this better than I do but are the date range options limited to like 
```
'last-week'
'current-week'
'one-month'
```
If so then we can just create a type/interface of DateRangeType so that you can only change to those values

apps/google-analytics-4/frontend/src/components/sidebar/SidebarAnalytics.tsx
```diff
@@ -0,0 +1,46 @@
+import { useAutoResizer, useFieldValue } from '@contentful/react-apps-toolkit';
+import ChartFooter from 'components/sidebar/ChartFooter';
+import ChartHeader from 'components/sidebar/ChartHeader';
+import { useEffect, useState } from 'react';
+import { RunReportResponse } from 'types';
+import { config } from '../../config';
+
+export default function SidebarAnalytics() {
+  const [runReportResponse, setRunReportResponse] = useState<RunReportResponse | undefined>();
+  const [slugValue] = useFieldValue<string | undefined>('slug');
```
ryunsong-contentful, line 10: I'm pretty sure you can get rid of the `undefined` typing because your slugValue will never be undefined since you set the value here and never update it anywhere else

apps/google-analytics-4/frontend/src/components/sidebar/SidebarAnalytics.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const metricName = runReportResponse && runReportResponse.metricHeaders[0].name;
+
+  return (
+    <>
+      <ChartHeader
+        metricName={metricName !== undefined ? metricName : ''}
+        metricValue={pageViews !== undefined ? pageViews.toString() : ''}
+      />
+      <ChartFooter
+        slugName={slugValue !== undefined ? slugValue : ''}
```
ryunsong-contentful, line 41: No need to check for undefined since it'll never be undefined (same idea on line 10)

apps/google-analytics-4/frontend/src/components/sidebar/SidebarAnalytics.tsx
```diff
@@ -0,0 +1,46 @@
+import { useAutoResizer, useFieldValue } from '@contentful/react-apps-toolkit';
+import ChartFooter from 'components/sidebar/ChartFooter';
+import ChartHeader from 'components/sidebar/ChartHeader';
+import { useEffect, useState } from 'react';
+import { RunReportResponse } from 'types';
+import { config } from '../../config';
+
+export default function SidebarAnalytics() {
+  const [runReportResponse, setRunReportResponse] = useState<RunReportResponse | undefined>();
```
ryunsong-contentful, line 9: I have an idea here: so if you set this value like
`useState<RunReportResponse>({} as RunReportResponse);`
then you can do the following on lines 37 and 38.

apps/google-analytics-4/frontend/src/components/sidebar/SidebarAnalytics.tsx
ryunsong-contentful, line 37: Based on the comment on line 9. If you make that change you can get rid of line 32 where you set metric name and you can just do

```
metricName={runReportResponse.metricName ?? ''}
metricValue={runReportResponse.pageViews ? pageViews.toString() : ''}
```

ryunsong-contentful:
Overall looks good, just mostly small optimization comment related things. There's only one real blocker I saw which was this here:

If you run `npm run test` there's an unresolved error - there's actually three total but two of them come from the api.ts file which have nothing to do with you or this PR so this comment is only referencing the one error related to this PR.

``` 
FAIL  src/locations/Sidebar.spec.tsx
  ● Sidebar component › Component text exists

    TypeError: (0 , _reactAppsToolkit.useFieldValue) is not a function or its return value is not iterable

       8 | export default function SidebarAnalytics() {
       9 |   const [runReportResponse, setRunReportResponse] = useState<RunReportResponse | undefined>();
    > 10 |   const [slugValue] = useFieldValue<string | undefined>('slug');
         |                                    ^
      11 |
      12 |   useAutoResizer();
```

ryunsong-contentful:
One more thing: Is it possible to get a screenshot? You should definitely show off what you build :)

pkeavenycontentful (approved):
Couple changes, otherwise looks good.

apps/google-analytics-4/frontend/src/components/sidebar/SidebarAnalytics.tsx
ryunsong-contentful, line 41: You can make this slugName={slugValue ?? ''} if you make the change above in line 9

apps/google-analytics-4/frontend/src/types.ts
whitelisab: It is currently only used by `RunReportResponse`, so I will remove the export here.

apps/google-analytics-4/frontend/src/components/sidebar/ChartFooter.tsx
```diff
@@ -0,0 +1,37 @@
...
+  viewUrl: string;
+}
+
+const styles = {
+  slug: css({
+    color: tokens.gray600,
+    fontSize: tokens.fontSizeS,
+    marginBottom: tokens.spacingM,
+  }),
+};
```
ghepting, line 17: my custom style radar is going off – can you include screenshots of this with/without? it's possible there is a f36 way of achieving the desired result without custom one-off like this?

whitelisab: Yes, here are screenshots with the styling and without, as well as a screenshot of the legacy GA app so you can see that I was trying to mimic that styling on the slug name. I can achieve the same look by removing the custom styling and using: `<Text fontColor="gray600" fontSize="fontSizeS" marginBottom="spacingM">{slugName}</Text>`. So I can update it to that if you thing I should try to keep the same look as the legacy app. Let me know what you think!

Screenshot 2023-02-24 at 10 59 12 AM
Screenshot 2023-02-24 at 10 59 39 AM
Screenshot 2023-02-24 at 11 59 52 AM

apps/google-analytics-4/frontend/src/components/sidebar/ChartFooter.spec.tsx
```diff
@@ -0,0 +1,16 @@
+import ChartFooter from './ChartFooter';
+import { render, screen } from '@testing-library/react';
+
+describe('Chart Footer for the sidebar app', () => {
+  it('can render the slug name', () => {
+    render(<ChartFooter slugName="my-page" viewUrl="https://contentful.com" />);
+
+    expect(screen.getByText('my-page')).toBeInTheDocument();
```
lilbitner, line 8: Love all of the thorough testing 💯 curious about your choice of `toBeInTheDocument` vs `toBeVisible` ? Just in terms of testing what the user exactly sees, I believe but could be incorrect, that the former asserts it exists but not that it is visible to the user?

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -1,17 +1,13 @@
...
 const Sidebar = () => {
-  const sdk = useSDK<SidebarExtensionSDK>();
   /*
      To use the cma, inject it as follows.
      If it is not needed, you can remove the next line.
   */
   // const cma = useCMA();
 
-  return <Paragraph>Hello Sidebar Component (AppId: {sdk.ids.app})</Paragraph>;
+  return <SidebarAnalytics />;
```
lilbitner, line 10: Love that you made this into its own component! Great separation of concerns here. I am wondering if its possible this main analytic app content, like the dropdown, chart etc, will ever exist somewhere else besides the `sidebar`? I have absolutely no idea around the context of if this app could exist in other locations, but I get nervous about pigeon holing the file structure, and placement of components in directories based on their location in the app i.e sidebar directory. What are your thoughts?

lilbitner (commented):
Just a few comments I am curious about! Looking good!

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
whitelisab: I was following the pattern of the `ConfigScreen` components being in a directory called `config-screen`, so since these are in the `Sidebar` I called the directory and component `sidebar`. I am happy to rename the `sidebar` directory to `main-app` and rename this component from `SidebarAnalytics` to `AnalyticsApp`, does that sound good?

apps/google-analytics-4/frontend/src/components/sidebar/ChartFooter.spec.tsx
whitelisab: My understanding is that `toBeInTheDocument` tests whether an element is present in the document regardless of visibility, but it is a good suggestion to update these to `toBeVisible` to that it is asserting whether the element is actually visible to the user. I can make this update

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
lilbitner: Yeah the logic totally makes sense. I think for now perhaps `sidebar` suffices as the directory name since that is the immediate location, and we can adjust if needed!

apps/google-analytics-4/frontend/src/components/sidebar/SidebarAnalytics.tsx
whitelisab: For this one, no, so I have updated this to remove the extra check for `undefined`

whitelisab:
> 

This was a good catch @ryunsong-contentful ! I forgotten that there was the default test example in the `Sidebar.spec.tsx` file. For the time being, I actually removed that file, following the pattern of what we have for `ConfigScreen.tsx` where we are treating these location components as an App.tsx for each location iframe we are building out.

apps/google-analytics-4/frontend/src/components/main-app/ChartHeader.spec.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { render, screen } from '@testing-library/react';
+
+const mockMetricName = 'pageviews';
+const mockMetricValue = '150';
+
+describe('Chart Header for the analytics app', () => {
+  it('can render the metric value', () => {
+    render(<ChartHeader metricName={mockMetricName} metricValue={mockMetricValue} />);
+
+    expect(screen.getByText('150')).toBeVisible();
```
ryunsong-contentful, line 11: nit: can use the variable `mockMetricValue` here

ryunsong-contentful (approved):

PR #2593, opened 2023-02-24
INTEG-149 customize and style Chart.js to match Contentful styles

lilbitner:
## Purpose
In order to provide the same data visualization that the soon-to-be deprecated Google Analytics application provided, it is necessary to create a native Line Chart component to our new GA4 application. 

## Approach
This component utilizes the `Chart.js` library as well as the React-specific library, `react-chartjs-2`, both of which were decided on by the team after discussion. As you might notice, the component itself is not tied to the exact JSON data that is provided by GA4. I felt it was important that this component does not rely on very specific data structures to assure the `LineChart` component is only aware of the exact data it needs to render the chart effectively - both for reusability and to maintain decoupled components as we build our component library. 

I imagine there will be a service of some sort of bridge the gap between component and JSON, as part of a separate subtask currently in the works.

Example of what it looks like locally within the Sidebar thus far (I am sure dimensions might need to be adjusted when combining the other metric components). 
Screenshot 2023-02-24 at 12 38 13 PM

Would appreciate any and all suggestions around code or design! **_Meeting with Hande next week to confirm the look and feel of the chart itself._** 

Follow-up tasks: 
1. Opening this PR now for momentum and parallel work reasons, but would like to work on the fit and feel of it within the web app - specifically around responsiveness. I am seeing some edge case non-ideal behavior of the chart when adjusting screen size. 

[Chart.js](https://www.chartjs.org/docs/latest/)
[react-chartjs-2](https://react-chartjs-2.js.org/)


apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+        backgroundColor: tokens.gray900,
+        bodyColor: tokens.colorWhite,
+        padding: 8,
+        titleMarginBottom: 4,
+        titleFont: {
+          size: 9,
+        },
+        bodyFont: {
+          size: 8,
+          weight: '700',
```
lilbitner, line 59: many of these typography specific choices are subjective for what looks good within the application (imo). Will most likely tweak these slightly as design is verified and the component is placed in the app fully.

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Line } from 'react-chartjs-2';
+import tokens from '@contentful/f36-tokens';
+
+const ACCESSIBILITY_LABEL = 'Analytics line chart';
+
+ChartJS.register(CategoryScale, LinearScale, PointElement, LineElement, Title, Tooltip);
+
+ChartJS.defaults.font.size = 12;
+ChartJS.defaults.font.family = tokens.fontStackPrimary;
+ChartJS.defaults.font.weight = '500';
```
lilbitner, line 20: avoided using a token here because of type conversion. Happy to convert a token of type **number** to a **string**,  but felt it was perhaps unnecessary for the benefit of using a token.

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+        },
+      },
+    },
+  };
+
+  const props = {
+    'aria-label': accessibilityLabel,
+  };
+
+  return <Line data={data} options={options} fallbackContent={accessibilityLabel} {...props} />;
```
lilbitner, line 73: not a big fan of spreading props, but believe this is the only way to spread html props down to the `canvas` element, to support accessibility.

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+  LinearScale,
+  PointElement,
+  LineElement,
+  Title,
+  Tooltip,
+} from 'chart.js';
+import { Line } from 'react-chartjs-2';
+import tokens from '@contentful/f36-tokens';
+
+const ACCESSIBILITY_LABEL = 'Analytics line chart';
```
pkeavenycontentful, line 14: Is there a reason to make this into a var instead of an inline string value in the template below?

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+ChartJS.register(CategoryScale, LinearScale, PointElement, LineElement, Title, Tooltip);
+
+ChartJS.defaults.font.size = 12;
+ChartJS.defaults.font.family = tokens.fontStackPrimary;
+ChartJS.defaults.font.weight = '500';
+ChartJS.defaults.borderColor = tokens.gray200;
+ChartJS.defaults.datasets.line.borderColor = tokens.colorPrimary;
+
+interface LineChartProps {
+  dataValues: Array<string | number>;
```
pkeavenycontentful, line 25: Nit: TS best practices usually have array values as `string[]` instead of the `Array<string>` class.

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const data = {
+    labels: xAxisLabels,
+    datasets: [
+      {
+        data: dataValues,
+      },
+    ],
+  };
+
+  const options = {
```
pkeavenycontentful, line 46: Nit: Looks like chartJS provides a `CoreChartData` type that we might be able to use here: https://github.com/chartjs/Chart.js/blob/cfefb3d/src/types/index.d.ts#L1461 Might help for type safety so we aren't passing invalid values to any of these properties.

pkeavenycontentful (approved):
Verified changes locally. Looks good, couple small comments. Great job!

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
lilbitner: GREAT catch! Thank you! Will absolutely add this 💯

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
lilbitner: Oh yes good point, will adjust!

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
lilbitner: I have previously followed practices that generally suggest all static constants should be defined above the component, to separate them from more interactive code. Personally, I find it helps with readability within JSX, but am certainly open to including them within the component itself! Do you tend to have a preference or practice with that?

pkeavenycontentful: Not particularly. My preference for variables in the template is more for repeatability. I.E. if a template string is used more than once, it deserves a variable. Your approach is also good, I wouldn't say either one is right or wrong, only personal preference. 

No need to change anything here 👍

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
ghepting: i think we should consistently use the f36 tokens wherever possible. 12px font size should be `tokens.fontSizeS` and weight 500 should be `tokens.fontWeightMedium` – you can use base 16px to convert rems to px for the font size and typecast 500.toString() as you mentioned for the weight.

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
ghepting: likely some legibility concerns around 8px font size here, but fine for a first pass. if we can without it looking ridiculous i'm a fan of trying to not go below the `fontSizeS` (0.75rem / 12px) whenever possible simply for accessibility

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
ghepting: seems reasonable to me!

ghepting (approved):

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,82 @@
...
+    ],
+  };
+
+  const options: ChartOptions<'line'> = {
+    responsive: true,
+    plugins: {
+      tooltip: {
+        backgroundColor: tokens.gray900,
+        bodyColor: tokens.colorWhite,
+        padding: 8,
```
ghepting, line 54: use `tokens.spacingXs`. it may be useful to create a function that parses the `px` off for unit-less integer formatting

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,82 @@
...
+        titleMarginBottom: 4,
```
ghepting, line 55: use `tokens.spacing2Xs`

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,82 @@
...
+        titleFont: {
+          size: 9,
+        },
+        bodyFont: {
+          size: 8,
+          weight: '700',
```
ghepting, line 61: I was surprised to find that there wasn't a `fontWeightBold` documented at https://f36.contentful.com/tokens/typography#font-weights so I posted to in the forma36 channel in slack to ask if there is a reason that token is missing? https://contentful.slack.com/archives/C961BF4RX/p1677508234933369

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,82 @@
...
+ChartJS.defaults.datasets.line.borderColor = tokens.colorPrimary;
+
+interface LineChartProps {
+  dataValues: number[];
+  xAxisLabels: string[];
+  tooltipMetricLabel: string;
+  accessibilityLabel: string;
+}
+
+const LineChart = ({
```
ryunsong-contentful, line 33: nit: Just a note here for convention consistency.
Change `LineChartProps to Props` and then destructure your props below the function definition.

```
const LineChart = (props: Props) => {
  const { dataValues, xAxisLabels, tooltipMetricLabel, accessibilityLabel } = props
  // rest of the code 
}
```

ryunsong-contentful:
Just want to backup Gary's comments of using the spacing values from Forma36. 
You probably already have a plan for this but just make sure the commits are squashed before merging!

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
lilbitner: great catch here, have added a ToDo item for when they do update the token library to include this font weight!

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
lilbitner: Thats a fair point! I have adjusted this to better utilize the tokens 👍

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
lilbitner: Thanks for pointing this out! Have adjusted!

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.tsx
lilbitner: Great idea! Have created a helper function to parse px and rem off the token values.

ryunsong-contentful (approved):
Approved - FYI you will need to rebase, you shouldn't have any merge conflicts though

PR #2620, opened 2023-02-28
[INTEG-181] Add lambda proxy endpoint for run-report through data API

ryunsong-contentful:
## Purpose
[JIRA Ticket](https://contentful.atlassian.net/browse/INTEG-181)
The Run Report function in GA4 is what gets the GA4 data for a given web page. This endpoint is needed so that we get that data.

## Approach
Just following the existing pattern of fetching account summaries.

Adds Data GA4 dependency (changes shown in package.json file)

### DIFF
--- a/apps/google-analytics-4/lambda/package-lock.json
+++ b/apps/google-analytics-4/lambda/package-lock.json
@@ -10,6 +10,7 @@
       "license": "ISC",
       "dependencies": {
         "@google-analytics/admin": "^4.3.1",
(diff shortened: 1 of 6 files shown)


ghepting (approved):

jsdalton (approved):
I think it's looking great. 💯 

I'm 90% sure you'll have to make one change  (see my comment) but everything else is 👌 . Approving now to speed things along.

PR #2625, opened 2023-03-01
refactors app.ts, adds controllers and middleware for better organization

ghepting:
## Purpose
improves readability of app.ts

## Approach
adds controllers and middleware for better organization, moves large amount of dependencies out of app.ts and into more focused modules


apps/google-analytics-4/lambda/src/controllers/concerns/error-class-to-api-error-map.ts
```diff
@@ -0,0 +1,27 @@
+// Maps an error class name -> a handler function that takes the error of that type as input and returns
```
jsdalton, line 1: Just the minorest of nitpick comments:

Not sure about introducing a "Rails" concept name here 😅 . I feel like it ends up feeling awkward or weird to folks who aren't familiar with that ecosystem.

I wish I had a better idea for the name / location of this bit of logic. It's definitely application specific -- it's our own mapping that's aware of multiple aspects of the code base (the error middleware, specific error types our controllers emit, how we want the server to respond).

For now i guess I don't have a better idea where this could go!

jsdalton (commented):
I think this is progress. I like how it cleans up the app.ts file and makes thing more orderly.

My main concern is I'm not sure how "standard" this style of decomposing the app file is -- specifically around separating routes from controllers.

What I've seen elsewhere (here's the best link I could find in 10 seconds of googling: https://www.geeksforgeeks.org/how-to-separate-routers-and-controllers-in-node-js/) is more of a bundling of routes and controllers. So routes and controllers are coupled together and then "mounted" to the main app via `app.use()`.

The advantage of this approach is you'd naturally end up writing tests for a bundle of controllers and routes at that level, as opposed to keeping the tests at the top level (app.spec.ts) but still having a lot that tests controller specific stuff. In other words, with this PR we've left this "monolith" of controller tests at the top but decoupled it all at the code level.

apps/google-analytics-4/lambda/src/controllers/api-controller.ts
```diff
@@ -0,0 +1,57 @@
+import { NextFunction, Request, Response } from 'express';
+import { RunReportParamsType } from '../types';
+import { GoogleApi } from '../services/google-api';
+
+const ApiController = {
+  credentials: (_req: Request, res: Response) => {
+    console.log('headers', _req.headers);
```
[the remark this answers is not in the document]
ryunsong-contentful: Can't remember if these console.logs will go away if you rebase

apps/google-analytics-4/lambda/src/services/google-api.ts
```diff
@@ -153,39 +157,44 @@ export class GoogleApi {
...
+    const DEFAULT_DIMENSIONS = ['date'];
+    const DEFAULT_METRICS = ['screenPageViews', 'totalUsers', 'screenPageViewsPerUser'];
 
     try {
       const [response] = await this.betaAnalyticsDataClient.runReport({
         property: property,
         dateRanges: [
           {
-            startDate: startDate ?? (new Date(Date.now() - ONE_WEEK)).toISOString().split('T')[0],
+            startDate: startDate ?? new Date(Date.now() - ONE_WEEK).toISOString().split('T')[0],
```
ryunsong-contentful, line 177: Maybe put a comment here for other people and our future-selves to say that this formats to `yyyy-mm-dd`

apps/google-analytics-4/lambda/src/controllers/concerns/error-class-to-api-error-map.ts
ghepting: moving to middlewares 👍

ghepting: actually i just deleted this file and moved it as a const into the apiErrorHandler middleware ✅

apps/google-analytics-4/lambda/src/middlewares/apiErrorHandler.ts
```diff
@@ -10,21 +10,21 @@ import { isApiError, ApiError } from './api-error';
 export type ApiErrorMap = Record<string, (e: any) => ApiError<Record<string, unknown>>>;
 
 export const apiErrorHandler: ErrorRequestHandler = (error, _request, response, next) => {
+  console.log('apiErrorHandler', error);
```
ryunsong-contentful, line 13: delete?

apps/google-analytics-4/lambda/src/controllers/apiController.ts
```diff
@@ -0,0 +1,57 @@
+import { NextFunction, Request, Response } from 'express';
+import { RunReportParamsType } from '../types';
+import { GoogleApi } from '../services/google-api';
+
+const ApiController = {
+  credentials: (_req: Request, res: Response) => {
+    console.log('headers', _req.headers);
```
ryunsong-contentful, line 7: delete or are you intentionally keeping this?

apps/google-analytics-4/lambda/src/middlewares/allowCorsOptionsRequests.ts
```diff
@@ -0,0 +1,22 @@
+import { RequestHandler } from 'express';
```
ryunsong-contentful, line 1: Add TODO to use express' cors library or is that something we won't do for this app?

ryunsong-contentful (commented):
Left nits - I personally think they're useful but let me know if you think otherwise

apps/google-analytics-4/lambda/src/middlewares/allow-cors-options-requests.ts
```diff
@@ -0,0 +1,22 @@
+import { RequestHandler } from 'express';
```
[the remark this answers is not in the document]
ghepting: done

apps/google-analytics-4/lambda/src/lib/errors/middlewares.ts
```diff
@@ -10,21 +10,21 @@ import { isApiError, ApiError } from './api-error';
 export type ApiErrorMap = Record<string, (e: any) => ApiError<Record<string, unknown>>>;
 
 export const apiErrorHandler: ErrorRequestHandler = (error, _request, response, next) => {
+  console.log('apiErrorHandler', error);
```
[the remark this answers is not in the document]
ghepting: removed

apps/google-analytics-4/lambda/src/controllers/api-controller.ts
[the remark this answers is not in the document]
ghepting: removed

ryunsong-contentful (approved):

PR #2627, opened 2023-03-01
INTEG-154 add frontend error handling for GA4 app

lilbitner:
## Purpose
In the case that there is an API error thrown when fetching the necessary chart data for the GA4 app, we want to make it clear to the user what went wrong. 

## Approach
Following what I derived as Contentful convention (could be completely wrong about this!), an error note will appear in place of the chart if an **error** is thrown. Additionally, if there is no error but also no data to render, a **warning** note will instead appear. Could see an argument for not having a warning banner for that second case, but just a simple text component. Examples show below: 

Screenshot 2023-03-01 at 1 00 04 PM

Screenshot 2023-03-02 at 12 09 53 PM

I still render the other analytic components (dropdown for example), in the case that changing the date actually throws the error. Ideally then the user will have a better idea of what happened/how to resolve. 

I added a test file for the `AnalyticsApps` file as it was becoming a bit too complex to lack testing for now, imo. I think we will need to iron out some convention around mocking etc. 

Per usual, always open to any and all thoughts here around design and code!


apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
```diff
@@ -0,0 +1,51 @@
+import AnalyticsApp from './AnalyticsApp';
+import { render, screen } from '@testing-library/react';
+import runReportResponseHasViews from '../../../../../lambda/public/sampleData/runReportResponseHasViews.json';
+import runReportResponseNoView from '../../../../../lambda/public/sampleData/runReportResponseNoViews.json';
```
lilbitner, line 4: was not sure how best to do this. I could move this mock to the shared file that Patrick created so we do not have crazy relative imports?

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      animation: `${load} 1s linear infinite`,
+    },
+  }),
+  wrapper: css({
+    height: '100px',
+    width: '100px',
+    margin: 'auto',
+  }),
+};
```
lilbitner, line 42: just moved this file under its own directory

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+      );
+
+    render(<AnalyticsApp />);
+
+    const dropdown = await findByTestId(SELECT_TEST_ID);
+    const errorNote = getByTestId(NOTE_TEST_ID);
+
+    const noteText = await findByText('mock Api error');
+
+    expect(dropdown).toBeVisible();
```
pkeavenycontentful, line 69: Nice test case here 👍

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
```diff
@@ -0,0 +1,73 @@
+import AnalyticsApp from './AnalyticsApp';
+import { render, screen } from '@testing-library/react';
+import runReportResponseHasViews from '../../../../../lambda/public/sampleData/runReportResponseHasViews.json';
```
pkeavenycontentful, line 3: The thought behind the `/shared` folder was to work around imports like this, so by putting them in that folder you can access them here with `@shared/runReportResponseHasViews.json`, and the lambda could do the same. But right now these files are in a statically served `public` directory from express, so not sure the best way to handle this. My vote would be to also statically serve the `/shared` folder but that might break some node/express best practice. 

@ghepting @jsdalton thoughts?

pkeavenycontentful (approved):
Some comments, but looks good overall 👍

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -3,30 +3,41 @@ import ChartFooter from 'components/main-app/ChartFooter';
...
-import { styles } from './AnalyticsApp.styles'
-import ChartContent from './ChartContent';
-import { Flex } from '@contentful/f36-components';
+import DateRange from 'helpers/dateRange.enum';
+import { config } from 'config';
+import { styles } from './AnalyticsApp.styles';
+import ChartContent from 'components/main-app/ChartContent';
+import { Flex, Note } from '@contentful/f36-components';
 
+const DEFAULT_ERR_MSG = 'Oops! Cannot display the analytic data at this time.';
```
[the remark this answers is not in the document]
lilbitner: Thanks for finding this!

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
ghepting: it's fine to import these in specs like this IMO. once the frontend isn't using them then we can just move them into a lambda/test/fixtures directory and this will be cleaner.

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
ghepting: i commented on the above – i think this is only temporary so we shouldn't worry about it for now, but yeah this is gross if it were to stay like this  :)

ghepting (approved):

PR #2628, opened 2023-03-01
adds sentry/node to ga4 lambda

ghepting:
## Purpose
Adds Sentry for observability on exceptions to GA4 backend lambda

## Approach
Adds sentry/node SDK and configures frontend to pass x-contentful-* headers to obtain the same level of 

sentry/node

## Deployment
- [x] configure git commit hash as release in `Sentry.init`
- [x] resolve cognitive dissonance with serverless `stage` (dev/prd/test) vs NODE_ENV (development/production/test)

Contentful Context, Git commit hash, and node env:
Screenshot 2023-03-04 at 10 56 32 AM


apps/google-analytics-4/frontend/src/App.tsx
```diff
@@ -23,6 +23,33 @@ const ComponentLocationSettings = {
 const App = () => {
   const sdk = useSDK();
 
+  const location = useMemo(() => {
+    for (const [location] of Object.entries(ComponentLocationSettings)) {
+      if (sdk.location.is(location)) {
+        return location;
+      }
+    }
+  }, [sdk]);
```
ghepting, line 32: - [ ] bug: this is returning `undefined` – resolve before merging

image

ghepting: resolved: 
image

apps/google-analytics-4/frontend/src/services/api.ts
```diff
@@ -133,9 +139,12 @@ export class Api {
     return await fetchFromApi<Credentials>(
       this.requestUrl('api/credentials'),
       ZCredentials,
-      this.appDefinitionId,
+      this.contentfulContext.app!,
```
ghepting, line 142: @jsdalton / @ryunsong-contentful maybe worth pairing on this `!` typescript piece to see if one of you two have a suggestion for me 😄

apps/google-analytics-4/lambda/serverless.yml
```diff
@@ -18,7 +18,10 @@ functions:
   api:
     environment:
       SIGNING_SECRET: ${self:custom.environment.SIGNING_SECRET}
+      SENTRY_DSN: ${self:custom.environment.SENTRY_DSN}
       STAGE: ${opt:stage, 'dev'}
+      NODE_ENV: ${env:NODE_ENV}
+      RELEASE_VERSION: ${env:CIRCLE_SHA1}
```
ghepting, line 24: TODO:

- [ ] configure `CIRCLE_SHA1` environment variable in development using dotenv (`.env` file)

apps/google-analytics-4/frontend/src/services/api.ts
jsdalton: I guess I wonder why `app` is an optional field in https://github.com/contentful/apps/blob/master/apps/google-analytics-4/frontend/src/types.ts#L8 ? That's where the smell is coming from in my opinion. I _think_ all of those should be "required".

apps/google-analytics-4/lambda/serverless.yml
jsdalton: Let's discuss that TODO. @mgoudy91 and I ended up removing `dotenv` because we ended up with two many systems that were managing per environment data. It probably makes sense to move these to the serverless environment configs.

But CIRCLE_SHA1 does present some difficulties with that idea. 🤔

apps/google-analytics-4/frontend/src/services/api.ts
ghepting: yeah so I will revisit this but I was basing this on:

```
export interface ContentfulContext extends IdsAPI {
  app?: string;
  contentType?: string;
  entry?: string;
  environment: string;
  environmentAlias?: string;
  field?: string;
  location?: string;
  organization: string;
  space: string;
  user: string;
}
```

and had actually attempted to make this `extend` IdsAPI originally.

ghepting: The problem is that when you attempt to use the `sdk.ids` and have this misaligned – sdk.ids is a type `Omit<IdsAPI, EntryScopedIds>` – happy to pair w/u a bit on this detail as I was going in circles w/this – there's probably a "safe" coercion we can apply at the right spot here

ghepting: Also:

> I guess I wonder why app is an optional field in https://github.com/contentful/apps/blob/master/apps/google-analytics-4/frontend/src/types.ts#L8 ? That's where the smell is coming from in my opinion. I think all of those should be "required".

it's optional because it's either `app` (current) or `extension` (legacy)

jsdalton: Gotcha. I think we should attempt to clean this all up a bit, but I suggest doing it as a follow up refactor, so as to unblock this PR.

apps/google-analytics-4/lambda/serverless.yml
jsdalton: Should we set a default fallback value here? Maybe something explicit like 'undefined' or 'development' -- or even fall back to NODE_ENV? That way if we see it in the logs we know where it came from as opposed to something being busted with our whole configuration

apps/google-analytics-4/lambda/src/middlewares/setSentryContext.ts
```diff
@@ -0,0 +1,33 @@
...
+  'X-Contentful-Organization',
+  'X-Contentful-Space',
+  'X-Contentful-User',
+];
+
+export const setSentryContext = (req: Request, _res: Response, next: NextFunction) => {
+  Sentry.setUser({ id: req.header('X-Contentful-User') as Sentry.User['id'] });
+
+  const contentfulContextHeaders = Object.keys(req.headers).filter((key) => {
+    const downcasedHeaders = ContentfulContextHeaders.map((h) => h.toLowerCase());
```
jsdalton, line 21: Small thing but we're rebuilding this list of downcasedHeaders on every iteration of the loop. Instead let's build it once outside the loop.

jsdalton (approved):
A few minor changes and suggestions!

apps/google-analytics-4/frontend/src/services/api.ts
ghepting: Yeah, sgtm

apps/google-analytics-4/lambda/serverless.yml
ghepting: updated to:

```
      NODE_ENV: ${env:NODE_ENV, 'development'}
      RELEASE_VERSION: ${env:CIRCLE_SHA1, 'undefined'}
```

ghepting: Re: removing dotenv – I'm fine with that, I can just add the CIRCLE_SHA1 to the npm scripts commands

ghepting:
```
"scripts": {
    "debug": "NODE_ENV=development CIRCLE_SHA1=$(git rev-parse --short HEAD) SLS_DEBUG=* node inspect ./node_modules/serverless/bin/serverless.js offline",
    "build": "NODE_ENV=production CIRCLE_SHA1=$(git rev-parse --short HEAD) rimraf ./build && tsc",
    "lint": "eslint .",
    "start:dev": "NODE_ENV=development CIRCLE_SHA1=$(git rev-parse --short HEAD) nodemon",
    "test": "NODE_ENV=test CIRCLE_SHA1=$(git rev-parse --short HEAD) TS_NODE_TRANSPILE_ONLY=1 mocha",
    "test:debug": "NODE_ENV=test CIRCLE_SHA1=$(git rev-parse --short HEAD) TS_NODE_TRANSPILE_ONLY=1 mocha -- --inspect --inspect-brk",
    "test:watch": "NODE_ENV=test CIRCLE_SHA1=$(git rev-parse --short HEAD) TS_NODE_TRANSPILE_ONLY=1 mocha --watch --watch-files src --watch-files test",
    "deploy:test": "NODE_ENV=test CIRCLE_SHA1=$(git rev-parse --short HEAD) sls deploy --stage test",
    "deploy": "CIRCLE_SHA1=$(git rev-parse --short HEAD) sls deploy --stage $STAGE"
  },
```

PR #2644, opened 2023-03-02
INTEG-133 Google Analytics 4 serverless infra structure

mgoudy91:


apps/google-analytics-4/lambda/package.json
```diff
@@ -11,13 +11,14 @@
     "start:dev": "nodemon",
     "test": "TS_NODE_TRANSPILE_ONLY=1 mocha",
     "test:debug": "TS_NODE_TRANSPILE_ONLY=1 mocha -- --inspect --inspect-brk",
-    "test:watch": "TS_NODE_TRANSPILE_ONLY=1 mocha --watch --watch-files src --watch-files test"
+    "test:watch": "TS_NODE_TRANSPILE_ONLY=1 mocha --watch --watch-files src --watch-files test",
+    "deploy:test": "sls deploy --stage test",
+    "deploy": "sls deploy --stage $STAGE"
```
mgoudy91, line 16: The most important line to review. Please double check this makes sense to you too!

apps/google-analytics-4/lambda/config/serverless-env.test.yml
```diff
@@ -1 +1,2 @@
-SIGNING_SECRET: ${env:SIGNING_SECRET}
+SIGNING_SECRET: ${ssm:/aws/reference/secretsmanager/ci/apps/google-analytics-4-action/test/signing-secret}
+domainName: google-analytics-4-test.ctfapps.net
```
ghepting, line 2: it's pretty common convention for env vars is to be uppercased. is there a reason we're doing these in camelCase?

apps/google-analytics-4/lambda/package.json
```diff
@@ -34,14 +34,15 @@
     "mocha": "^10.2.0",
     "nodemon": "^2.0.20",
     "rimraf": "^3.0.2",
-    "serverless": "^3.26.0",
-    "serverless-domain-manager": "^6.4.3",
-    "serverless-offline": "^12.0.4",
+    "serverless": "2.72.3",
```
ghepting, line 37: 😭

ghepting (approved):

apps/google-analytics-4/lambda/package.json
ghepting: would love to see some notes/comments left as to why we needed to revert to an older version. this concerns me from a tech debt/vulnerability standpoint and I want to make sure it's clear to future-us why we did this.

apps/google-analytics-4/lambda/config/serverless-env.test.yml
```diff
@@ -1 +1,2 @@
-SIGNING_SECRET: ${env:SIGNING_SECRET}
+SIGNING_SECRET: ${ssm:/aws/reference/secretsmanager/ci/apps/google-analytics-4-action/test/signing-secret}
```
jsdalton, line 1: See my other comment, for conistency let's do `signedSecret` here and in the other templates.

jsdalton (approved):
🚀 🚀 🚀

apps/google-analytics-4/lambda/package.json
ghepting: my main concern about this by the way is dependabot issues this has a cascading affect on. there's several `critical` dep VULNs that cannot be resolved due to pinned serverless in other apps lamba's already so this will surely have the same problem very soon if not immediately :/

PR #2653, opened 2023-03-03
fixes prettier violations in preparation for more aggressive assertion on code quality/formatting consistency

ghepting:
## Purpose
preparation for more aggressive assertion on code quality/formatting consistency

## Approach
ran `prettier --write` in the apps/google-analytics-4 directory to auto correct all existing violations

PR #2655, opened 2023-03-03
normalizes casing of headers (upperfirst+camelcasing)

ghepting:
## Purpose
be consistent with our contentful header casing

## Approach
adds some useful abstraction and types for contentful context (`sdk.ids` & `sdk.location`) and contentful context headers (`X-Contentful-<Whatever>`)

### DIFF
--- a/apps/google-analytics-4/frontend/src/helpers/contentfulContext.ts
+++ b/apps/google-analytics-4/frontend/src/helpers/contentfulContext.ts
@@ -1,0 +1,19 @@
+import { ContentfulContext } from '../types';
+import { KnownSDK, locations } from '@contentful/app-sdk';
+
+const Locations = [
+  locations.LOCATION_APP_CONFIG,
+  locations.LOCATION_ENTRY_FIELD,
+  locations.LOCATION_ENTRY_EDITOR,
+  locations.LOCATION_DIALOG,
+  locations.LOCATION_ENTRY_SIDEBAR,
+  locations.LOCATION_PAGE,
+  locations.LOCATION_HOME,
+];
+
+export const contentfulContext = (sdk: KnownSDK): ContentfulContext => {
+  return {
+    ...sdk.ids,
+    location: Locations.find((location) => sdk.location.is(location)),
+  };
+};
--- a/apps/google-analytics-4/frontend/src/services/api.ts
+++ b/apps/google-analytics-4/frontend/src/services/api.ts
@@ -31,9 +31,9 @@ type Headers = Record<string, string>;
 
 export type Credentials = z.infer<typeof ZCredentials>;
 
-export class ApiError extends Error { }
-export class ApiServerError extends ApiError { }
-export class ApiClientError extends ApiError { }
+export class ApiError extends Error {}
+export class ApiServerError extends ApiError {}
+export class ApiClientError extends ApiError {}
 
 export async function fetchFromApi<T>(
   apiUrl: URL,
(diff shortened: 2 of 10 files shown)


apps/google-analytics-4/frontend/src/services/api.ts
```diff
@@ -31,9 +31,9 @@ type Headers = Record<string, string>;
 
 export type Credentials = z.infer<typeof ZCredentials>;
 
-export class ApiError extends Error { }
-export class ApiServerError extends ApiError { }
-export class ApiClientError extends ApiError { }
+export class ApiError extends Error {}
+export class ApiServerError extends ApiError {}
+export class ApiClientError extends ApiError {}
```
ghepting, line 36: more prettier formatting noise – will be addressed holistically for ga4 app in: https://github.com/contentful/apps/pull/2653

apps/google-analytics-4/frontend/src/helpers/contentfulContext.ts
```diff
@@ -0,0 +1,31 @@
...
+  locations.LOCATION_PAGE,
+  locations.LOCATION_HOME,
+];
+
+export const location = (sdk: KnownSDK) => {
+  let currentLocation = undefined;
+
+  Locations.forEach((location) => {
+    if (sdk.location.is(location)) {
+      currentLocation = location;
```
jsdalton, line 19: Maybe just `return location` to terminate early on a match?

apps/google-analytics-4/frontend/src/helpers/contentfulContext.ts
```diff
@@ -0,0 +1,31 @@
...
+  locations.LOCATION_DIALOG,
+  locations.LOCATION_ENTRY_SIDEBAR,
+  locations.LOCATION_PAGE,
+  locations.LOCATION_HOME,
+];
+
+export const location = (sdk: KnownSDK) => {
+  let currentLocation = undefined;
+
+  Locations.forEach((location) => {
```
jsdalton, line 17: i'm not super opinionated here but if we're doing a procedural loop in wonder if `for...in` is a better choice?

jsdalton (approved):
One or two small suggestions, but otherwise good to go 💯

apps/google-analytics-4/frontend/src/helpers/contentfulContext.ts
ghepting: ah good catch, updating now

apps/google-analytics-4/frontend/src/helpers/contentfulContext.ts
ryunsong-contentful: This is going to cause a bug Im pretty sure. In a `forEach`, when you return, you dont return out of the entire function (similar how a .map return will return out of the current context but not the context of the entire function.

You can use the `.find()` higher order array function.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Aka. 

```
return Locations.find((location) => sdk.location.is(location))
```

This will either return undefined (when it can't find it) or the matching object that fulfills the condition as true

ryunsong-contentful: Another benefit is that it turns all of this logic into one line :)

ryunsong-contentful (changes requested):
Requesting changes as there is a bug here (see comment above)

apps/google-analytics-4/frontend/src/helpers/contentfulContext.ts
ghepting: refactored to just be: ```Locations.find((location) => sdk.location.is(location))``` – i was being an idiot :D

ryunsong-contentful (approved):

PR #2657, opened 2023-03-03
Fix sentry org naming

mgoudy91:
## Purpose
We misspelled conten`t`ful in the org name setting for sentry, unsurprisingly this causes the job to fail.

## Approach
I have added a missing 't' to Contentful.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ commands:
   publish:
     steps:
       - vault/get-secrets:
-          template-preset: "semantic-release-ecosystem"
+          template-preset: 'semantic-release-ecosystem'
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -157,12 +157,12 @@ jobs:
...
           name: Test
           command: cd examples/javascript && npm run test
-  
+
   notify-sentry-deploy:
     docker:
       - image: cimg/base:stable
     environment:
-      SENTRY_ORG: contenful
+      SENTRY_ORG: contentful
```
mgoudy91, line 165: here

ghepting (approved):

PR #2692, opened 2023-03-06
SLS 3 

mgoudy91:
## Purpose
In order to not restrict our dependabot and general module upgrades, we need to upgrade serverless to v3. This is known to have some file pathing issues in repos like ours with multiple package.json files, but is worth testing out anyway.

## Approach
Upgrades serverless, serverless-domain-manager, and serverless-offline to their most recent versions. Also removes some compatibility flags for preparing sls 2 to be upgraded.

[JIRA](https://contentful.atlassian.net/browse/INTEG-199)


jsdalton (approved):
👍

ryunsong-contentful (approved):

PR #2701, opened 2023-03-06
INTEG-205 Fix GA4 FE url

mgoudy91:
## Purpose
The GA4 frontend needs the correct production url for app configuration to work. Now that we know what that is we can set it.

## Approach
N/A

[jira](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?modal=detail&selectedIssue=INTEG-205)


jsdalton (approved):
:shipit:

ghepting (approved):

PR #2710, opened 2023-03-07
add env support for sentry in higher envs

mgoudy91:
## Purpose
need to check in sentry env values for test and prd


jsdalton:
I _think_ this can be closed now since we've moved the variable to the main serverless.yml file now.

PR #2716, opened 2023-03-07
Add some Chart + Error handling styling fixes

lilbitner:
## Purpose
This PR adds some styling fixes (as per some suggestions from Hande) to the `LineChart` component as well as the `AnalyticsApp` component in the case of errors or warnings that might arise, as I noticed some unhandled text overflow. 

I am a bit blocked on the more pressing ticket right now hence why this small styling fix PR fix.

#### `LineChart` component: 

- Made the font sizes of the tooltip larger (at least 12px so they are readable). Also removed unnecessary noise within the Tooltip to save space. 
- Hande emphasized that this chart component does not have to be 'perfect' because ideally the user would click the link to view it more extensively in the Google Analytics app. I think this is an interesting point to keep in mind. 

### **Before**: 
Screenshot 2023-03-07 at 3 53 12 PM

### **After**: 
Screenshot 2023-03-07 at 3 06 45 PM

### `AnalyticsApp` component: 

- Handled overflow of error or warning messages by hiding the overflow and adding ellipsis of the text content. 
- This is a good example of a component we an make ourselves to add to the Storybook component library as I feel the F36 Note component doesn't add a lot of ability to customize styling. For now this styling is just in the `AnalyticsApp` component, but would like to migrate it eventually to its own file. 

### **Before**: 
Screenshot 2023-03-06 at 1 57 07 PM

### **After**: 
Screenshot 2023-03-07 at 2 08 27 PM

Would appreciate any and all thoughts as usual!! 💡 🚀

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
@@ -26,8 +26,8 @@ const ChartContent = (props: Props) => {
       <LineChart
         dataValues={parseRowViews()}
         xAxisLabels={parseRowDates()}
-        tooltipMetricLabel='Page views'
-        accessibilityLabel='Analytics line chart'
+        tooltipMetricLabel="Page views"
+        accessibilityLabel="Analytics line chart"
       />
     </>
   );
--- a/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
@@ -26,8 +26,8 @@ const ChartContent = (props: Props) => {
       <LineChart
         dataValues={parseRowViews()}
         xAxisLabels={parseRowDates()}
-        tooltipMetricLabel="Page views"
-        accessibilityLabel="Analytics line chart"
+        tooltipMetricLabel='Page views'
+        accessibilityLabel='Analytics line chart'
       />
     </>
   );
--- a/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
@@ -26,8 +26,8 @@ const ChartContent = (props: Props) => {
       <LineChart
         dataValues={parseRowViews()}
         xAxisLabels={parseRowDates()}
-        tooltipMetricLabel={''}
-        accessibilityLabel={'Analytics line chart'}
+        tooltipMetricLabel="Page views"
+        accessibilityLabel="Analytics line chart"
       />
     </>
   );
--- a/apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.styles.ts
+++ b/apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.styles.ts
@@ -39,4 +39,15 @@ export const styles = {
     width: '100px',
     margin: 'auto',
   }),
+  note: css({
+    overflow: 'hidden',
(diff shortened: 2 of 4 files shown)


apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.styles.ts
```diff
@@ -39,4 +39,15 @@ export const styles = {
...
     margin: 'auto',
   }),
+  note: css({
+    overflow: 'hidden',
+    marginBottom: tokens.spacingM,
+  }),
+  noteContent: css({
+    textOverflow: 'ellipsis',
+    overflow: 'hidden',
+    display: '-webkit-box',
```
mgoudy91, line 49: I should know this but I'll just be honest and say I don't: do we need to provide fallbacks for non webkit browsers?
Like
```
display: -webkit-box;
display: -ms-flexbox;
display: flex;
```

mgoudy91 (approved):
🥇 lgtm

One question that is unlikely an issue

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.styles.ts
lilbitner: that is actually a great question! My previous understanding is that if the browser is not a webkit browser, it will default to `inherit` the display property instead. But I could be incorrect here! tricky to test as it seems like most iOS browsers use webkit - unless you know one one? will keep looking into it !

apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
```diff
@@ -26,8 +26,8 @@ const ChartContent = (props: Props) => {
       <LineChart
         dataValues={parseRowViews()}
         xAxisLabels={parseRowDates()}
-        tooltipMetricLabel={''}
-        accessibilityLabel={'Analytics line chart'}
+        tooltipMetricLabel="Page views"
```
ryunsong-contentful, line 29: nit: when you run `prettier --write .` is this what happens, I thought we we're single quoting

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.styles.ts
```diff
@@ -39,4 +39,15 @@ export const styles = {
     width: '100px',
     margin: 'auto',
   }),
+  note: css({
```
ryunsong-contentful, line 42: nice -  good fix

ryunsong-contentful (approved):

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.styles.ts
mgoudy91: Sounds good, I don't think this is really holding anything up then! Appreciate the thoughts

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.styles.ts
lilbitner: great point! a good reason to have this as a component of its own so we only have to change in one spot moving forward 👍

apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
lilbitner: good catch here - I am not sure how it got thru, will reconfirm my prettier/eslint configuration. Thanks for noticing!

PR #2718, opened 2023-03-07
Enable GA4 tests in CI

jsdalton:
## Purpose

* Tests haven't been running in production all along!
* We have failing tests as a result
* A few things were in need of minor cleanup

## Approach

* Add a lambda `test:ci` and fix the misnamed frontend "test-ci"
* .gitignore directories were not playing nicely with tooling, which expects relative path
* The number of env vars was getting to be a headache. So use dotenv _just for testing_
* For now just use the same strategy as the frontend config, i.e. fail loudly when variables are not set
* Use a sandbox to in app.spec.ts to prevent stubs from spilling across tests
* Introduce a mock request library since it's super hard to create realistic express request objects!
* Refactor service account provider tests to use the mock request, since these were failing
* Refactor verify signed requests middleware to test directly, i.e. remove the depedency on our specific test
* Fix broken GoogleApiService tests. These probably could use some more love!
* Add a hard --exit at the end of our tests. Our test runner hangs on app.spec.ts! TODO: debug this

* https://sinonjs.org/releases/latest/sandbox/
* https://github.com/chaijs/chai-http/issues/178#issuecomment-336602457


apps/google-analytics-4/lambda/package.json
```diff
@@ -10,6 +10,7 @@
     "lint": "eslint .",
     "start:dev": "NODE_ENV=development CIRCLE_SHA1=$(git rev-parse --short HEAD) nodemon",
     "test": "NODE_ENV=test CIRCLE_SHA1=$(git rev-parse --short HEAD) TS_NODE_TRANSPILE_ONLY=1 mocha",
+    "test:ci": "NODE_ENV=test CIRCLE_SHA1=$(git rev-parse --short HEAD) TS_NODE_TRANSPILE_ONLY=1 CI=true mocha",
```
ghepting, line 13: 🤦‍♂️

ryunsong-contentful (approved):

apps/google-analytics-4/lambda/package.json
```diff
@@ -5,13 +5,14 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "debug": "NODE_ENV=development CIRCLE_SHA1=$(git rev-parse --short HEAD) SLS_DEBUG=* node inspect ./node_modules/serverless/bin/serverless.js offline",
-    "build": "NODE_ENV=production CIRCLE_SHA1=$(git rev-parse --short HEAD) rimraf ./build && tsc",
+    "debug": "NODE_ENV=development SLS_DEBUG=* node inspect ./node_modules/serverless/bin/serverless.js offline",
+    "build": "NODE_ENV=production rimraf ./build && tsc",
```
ghepting, line 9: why was the `CIRCLE_SHA1` removed here?

jsdalton: Good question. i _believe_ we don't need this env var during the debug or build modes, but only during the actual deploy step. We fall back to "undefined" in our template serverless.yml. So removing it to get rid of cruft in our package.json file.

Do we need it or is it desirable? We can find a way to include it if so.

ghepting: I _believe_ that it needs to be included in the build otherwise the compiled [frontend] files will have an `undefined` sentry release configured

apps/google-analytics-4/lambda/src/config.ts
```diff
@@ -0,0 +1,12 @@
...
+  if (environmentVariableValue === undefined) {
+    throw new Error(`Missing environment variable: '${environmentVariableName}'`);
+  }
+  return environmentVariableValue;
+}
+
+export const config = {
+  signingSecret: getEnvironmentVariable('SIGNING_SECRET'),
+  stage: getEnvironmentVariable('STAGE'),
+};
```
ghepting, line 12: 👍

ghepting (approved):

mgoudy91 (approved):

PR #2725, opened 2023-03-08
[Build-Fix] Fixes error mapper duplication bug

ryunsong-contentful:
## Purpose
Fixes a duplication bug in the app.ts and errorHandler in the middleware


jsdalton (approved):

jsdalton (approved):
But we don't need the `import { apiErrorMap } from '../apiErrorMap';`

PR #2726, opened 2023-03-08
fix request signing verification

ghepting:
## Purpose
fix request signing verification

## Approach
coerce all headers back to strings to undo what serverless does with headers like `Content-Length`, etc and let the node-apps-toolkit do the header filtering as usual

### DIFF
--- a/apps/google-analytics-4/lambda/src/middlewares/verifySignedRequests.ts
+++ b/apps/google-analytics-4/lambda/src/middlewares/verifySignedRequests.ts
@@ -1,13 +1,8 @@
 import * as NodeAppsToolkit from '@contentful/node-apps-toolkit';
-import {
-  CanonicalRequest,
-  ContentfulContextHeader,
-  ContentfulHeader,
-} from '@contentful/node-apps-toolkit/lib/requests/typings';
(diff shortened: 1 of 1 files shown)


apps/google-analytics-4/lambda/src/middlewares/verifySignedRequests.ts
```diff
@@ -37,47 +32,17 @@ export const verifySignedRequestMiddleware = (req: Request, _res: Response, next
...
 const makeCanonicalReq = (req: Request) => {
-  const requiredHeaders = requestSigningHeaders(req.headers);
+  // coerce all headers to strings
+  Object.keys(req.headers).forEach((key) => {
+    req.headers[key] = req.headers[key]?.toString();
+  });
 
   // TODO: make this stage prefixing logic better? (yuck)
-  const pathPrefix = process.env.STAGE !== 'prd' ? `/${process.env.stage}` : '';
+  const pathPrefix = process.env.STAGE !== 'prd' ? `/${process.env.STAGE}` : '';
```
ghepting, line 41: @jsdalton 😅

jsdalton (approved):

PR #2727, opened 2023-03-08
Spike/slack app suggested changes

pkeavenycontentful:
## Purpose
The Slack App was recently moved into this repository from an external repository, as a result, the setup steps need some adjusting. This PR's purpose is to suggest some ways to tweak the directory and the README so it's less of a pain in the future.


apps/slack/.nvmrc
```diff
@@ -0,0 +1 @@
+v16.19.0
```
jsdalton, line 1: I suggest we create two of these, one for the frontend and one for the backend.

The backend serverless.yml says we're running there on node14, so I'd suggest on the lambda folder we do a v14 to align our dev environment with the actual serverless environment

jsdalton: Btw I see this conflicts with what they say in the README 🤔 .

But here's the link: https://github.com/contentful/apps/blob/master/apps/slack/lambda/serverless.yml#L14

So the danger here is you could introduce a feature only available in node 16 that works locally but fails in production in the node 14 environment.

apps/slack/Dockerfile
```diff
@@ -4,12 +4,15 @@ RUN apk add --update --no-cache openssh build-base python3 bash git && rm -rf /v
 
 WORKDIR /app
 
-COPY package.json yarn.lock frontend/package.json lambda/package.json /app/
+#package.json was not brought over from the other repo, in fact it's part of a yarn workspaces set up that we might want to move away from
+COPY package.json package-lock.json frontend/package.json lambda/package.json /app/
```
jsdalton, line 8: Ah right I see. So package.json doesn't exist. Did you mean to introduce `package-lock.json` here as well? (which also doesn't exist)

apps/slack/README.md
```diff
@@ -101,10 +104,14 @@ This sections explains how to run the Slack app locally.
...
   - Create a key pair. Save the private key at `lamba/private-key.pem`
-- Configure Slack
+- Configure Slack (https://api.slack.com/apps)
   - Store Client ID and Client Secret (Basic Information -> App Credentials) in `lambda/config/serverless.dev.yml` (`oauthCredentials` -> `client_id` / `client_secret`)
   - Store Signing Secret (Basic Information -> App Credentials) in `lambda/config/serverless.dev.yml` (`slackSigningSecret` -> `signing_secret`)
 - Add the ngrok URL to in the `lambda/config/serverless.dev.yml` (`customDomain` -> `domainName`). Remove the `https://` from the URL. Add a trailing `/dev` to the URL (e.g.: `domainName: be25-95-91-246-99.ngrok.io/dev`)
+- `cd` into lambda
+- run `npm install && npm run start` in /lambda
+
+// Ignore below
```
jsdalton, line 114: Yeah the docker file is wrong as are these instructions.

FYI if you intended this as an invisible comment you do you need to do `[//]`?

jsdalton (commented):
I had a few minor suggestions here. I didn't even realize how outdated some of this stuff is.

I think it's helpful to move ahead with your PR even if you don't take all my suggestions.

PR #2729, opened 2023-03-09
[INTEG-169] Render view list for content type, slug, and url path

whitelisab:
## Purpose
Space Admins need a way to select which content types will show the GA4 app in the sidebar. They should also be able to specify a slug field and a URL prefix.

## Approach
The `AssignContentTypePage` component contains the logic for accessing all content types and saving selected content types as installation parameters. The `AssignContentTypeCard` component displays the input fields for each content type (Content type, Slug field, and URL Prefix) and allows users to remove and add content types. Any changes to content types are saved when the user clicks "Install" or "Save" for the app.

Empty state - no content types selected
Screenshot 2023-03-08 at 4 30 13 PM

Content types selected
Screenshot 2023-03-08 at 4 29 56 PM

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
@@ -1,0 +1,138 @@
+import {
+  Box,
+  Button,
+  Card,
+  FormControl,
+  Select,
+  Stack,
+  TextInput,
+  TextLink,
+} from '@contentful/f36-components';
+import { css } from 'emotion';
+import { AllContentTypes, ContentTypeEntries } from 'types';
+
+interface Props {
+  allContentTypes: AllContentTypes;
+  contentTypeEntries: ContentTypeEntries;
+  onContentTypeChange: (prevKey: string, newKey: string) => void;
+  onContentTypeFieldChange: (key: string, field: string, value: string) => void;
+  onAddContentType: () => void;
+  onRemoveContentType: (key: string) => void;
+}
+
+const styles = {
+  contentTypeItem: css({
+    flexBasis: '100%',
+  }),
+  hidden: css({
+    visibility: 'hidden',
+  }),
+};
+
+const AssignContentTypeCard = (props: Props) => {
+  const {
+    allContentTypes,
+    contentTypeEntries,
+    onContentTypeChange,
+    onContentTypeFieldChange,
+    onAddContentType,
+    onRemoveContentType,
+  } = props;
+
+  return (
+    <Card>
+      {Object.keys(contentTypeEntries).length ? (
+        <Stack marginBottom="none" spacing="spacingXs">
+          <Box className={styles.contentTypeItem}>
+            <FormControl marginBottom="none">
+              <FormControl.Label>Content type</FormControl.Label>
+            </FormControl>
+          </Box>
+          <Box className={styles.contentTypeItem}>
+            <FormControl marginBottom="none">
+              <FormControl.Label>Slug field</FormControl.Label>
+            </FormControl>
+          </Box>
+          <Box className={styles.contentTypeItem}>
+            <FormControl marginBottom="none">
+              <FormControl.Label>URL prefix</FormControl.Label>
+            </FormControl>
+          </Box>
+          <Box>
+            <TextLink className={styles.hidden}>Remove</TextLink>
+          </Box>
+        </Stack>
+      ) : null}
+      {Object.entries(contentTypeEntries).map(([key, { slugField, urlPrefix }], index) => {
+        return (
+          <Stack spacing="spacingXs" paddingBottom="spacingS" key={key}>
+            <Box className={styles.contentTypeItem}>
+              <Select
+                id={`contentType-${index}`}
+                name={`contentType-${index}`}
+                onChange={(event: React.ChangeEvent<HTMLSelectElement>) =>
+                  onContentTypeChange(key, event.target.value)
+                }
+                value={key}>
+                <Select.Option value="" isDisabled>
+                  Select content type
+                </Select.Option>
+                {Object.entries(allContentTypes)
+                  .filter(([type]) => type === key || !contentTypeEntries[type])
+                  .map(([type, { name: typeName }]) => {
+                    return (
+                      <Select.Option value={type} key={`type-${type}`}>
+                        {typeName}
+                      </Select.Option>
+                    );
+                  })}
+              </Select>
+            </Box>
+            <Box className={styles.contentTypeItem}>
+              <Select
+                id={`slugField-${index}`}
+                name={`slugField-${index}`}
+                isDisabled={!key}
+                isInvalid={key ? !slugField : false}
+                onChange={(event: React.ChangeEvent<HTMLSelectElement>) =>
+                  onContentTypeFieldChange(key, 'slugField', event.target.value)
+                }
+                value={slugField}>
+                <Select.Option value="" isDisabled>
+                  Select slug field
+                </Select.Option>
+                {key &&
+                  allContentTypes[key]?.fields?.map((field) => (
+                    <Select.Option key={`${key}.${field.id}`} value={field.id}>
+                      {field.name}
+                    </Select.Option>
+                  ))}
+              </Select>
+            </Box>
+            <Box className={styles.contentTypeItem}>
+              <TextInput
+                id={`urlPrefix-${index}`}
+                name={`urlPrefix-${index}`}
+                isDisabled={!key}
+                onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+                  onContentTypeFieldChange(key, 'urlPrefix', event.target.value)
+                }
+                value={urlPrefix}
+              />
+            </Box>
+            <Box>
+              <TextLink onClick={() => onRemoveContentType(key)}>Remove</TextLink>
+            </Box>
+          </Stack>
+        );
+      })}
+      <Button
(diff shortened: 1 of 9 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+interface Props {
+  allContentTypes: AllContentTypes;
+  contentTypeEntries: ContentTypeEntries;
+  onContentTypeChange: (prevKey: string, newKey: string) => void;
+  onContentTypeFieldChange: (key: string, field: string, value: string) => void;
+  onAddContentType: () => void;
+  onRemoveContentType: (key: string) => void;
+}
+
+const styles = {
```
mgoudy91, line 23: Thinking about how this could be done without having a hidden `remove` header cell and came up with this:
```
const styles = {
  contentTypeItem: css({
    flex: 4,
  }),
  removeItem: css({
    flex: 1,
  }),
  hidden: css({
    visibility: 'hidden',
  }),
};
```

Where the input field cells and header cells are `contentTypeItem`, and the last (now empty)`Box` in the header and all the remove Box cells are `removeItem`.


...I'm not convinced my approach is any better, I'll let you decide what you like more. 😅

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypePage.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const sdk = useSDK<AppExtensionSDK>();
+
+  const onConfigure = useCallback(async () => {
+    const currentState = await sdk.app.getCurrentState();
+
+    const newParameters = Object.assign(
+      {},
+      parameters,
+      omitBy(
```
mgoudy91, line 33: I know this is far from the only place we use lodash, so I don't mean to force this discussion here, 

BUT

work like this could be handled by some of the [ES6 arrow functions ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) on the keys of the object.

Probably just interested in checking if people are interested in making the switch, doesn't need to happen in this PR

mgoudy91 (approved):
A few thoughts on possible changes, mostly subjective.

As written, I still think this is good to deploy 🏅

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
whitelisab: Thank you for this! I also didn't love having the hidden header cell, so I will update to your approach.

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypePage.tsx
whitelisab: It's a good question -- this will be refactored in some of the pending work for this config page and we will revisit then!

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypePage.tsx
```diff
@@ -0,0 +1,80 @@
...
+        onContentTypeChange={handleContentTypeChange}
+        onContentTypeFieldChange={handleContentTypeFieldChange}
+        onAddContentType={handleAddContentType}
+        onRemoveContentType={handleRemoveContentType}
+      />
+    </Stack>
+  );
+};
+
+export default AssignContentTypePage;
```
jsdalton, line 80: Silly naming nitpick -- maybe we're just following a convention being used elsewhere in our app (?) but this component isn't really a "Page" so this name doesn't make so much sense to me. Is it a `AssignContentTypeSection` maybe?

jsdalton (approved):
Nice, I think this looks awesome! 💯 

I had a small naming question that can safely be ignored or deferred to another PR. 

Everything else looks great. 100% agree that things will be better when we find a new home for the logic aside from `useKeyService`, so we can tackle that shortly.

Excellent contribution! ⭐

apps/google-analytics-4/frontend/src/hooks/useKeyService.tsx
```diff
@@ -1,21 +1,31 @@
 import { useCallback, useState, useEffect } from 'react';
 import { AppExtensionSDK } from '@contentful/app-sdk';
 import { useSDK } from '@contentful/react-apps-toolkit';
-import { AppInstallationParameters, ServiceAccountKey, ServiceAccountKeyId } from 'types';
+import {
```
ryunsong-contentful, line 4: Just beware that Lillian's changes also have a change to this hook. I'm pretty sure you guys won't conflict just take a look at her pr/discuss with her in person because whoever's pr gets in first will create a merge conflict here.

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypePage.tsx
ryunsong-contentful: This is my fault. I named all these sections pages, renaming these things to sections makes more sense now.

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+    contentTypeEntries,
+    onContentTypeChange,
+    onContentTypeFieldChange,
+    onAddContentType,
+    onRemoveContentType,
+  } = props;
+
+  return (
+    <Card>
+      {Object.keys(contentTypeEntries).length ? (
```
ryunsong-contentful, line 44: The only thing I would consider here is two things (made in two separate comments). 
You don't have to go crazy on refactoring the repeatability here but I think there's some low hanging fruit like
```
          <Box className={styles.contentTypeItem}>
            <FormControl marginBottom="none">
              <FormControl.Label>URL prefix</FormControl.Label>
            </FormControl>
          </Box>
```

You can either make this a util component or just have it in this file that takes in some label params and whatever else it needs. 

I have a comment below that shows what I think isn't low hanging fruit and probably isn't worth refactoring although you're totally free to if you think it's useful

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  .map(([type, { name: typeName }]) => {
+                    return (
+                      <Select.Option value={type} key={`type-${type}`}>
+                        {typeName}
+                      </Select.Option>
+                    );
+                  })}
+              </Select>
+            </Box>
+            <Box className={styles.contentTypeItem}>
```
ryunsong-contentful, line 89: These dropdowns is the example where I think refactoring here isn't exactly low hanging fruit but could be worth it if you find something elegant.

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+          </Box>
+          <Box className={styles.contentTypeItem}>
+            <FormControl marginBottom="none">
+              <FormControl.Label>URL prefix</FormControl.Label>
+            </FormControl>
+          </Box>
+          <Box className={styles.removeItem}></Box>
+        </Stack>
+      ) : null}
+      {Object.entries(contentTypeEntries).map(([key, { slugField, urlPrefix }], index) => {
```
ryunsong-contentful, line 64: Second comment is all these `Object.entries/Object.keys`. 

I think the JSX would be much easier if we had some state, and then have a useEffect that set these states with the types of arrays you want.

That way the JSX can read something more like 
```
{some useful variable name}.map(some blah blah JSX)
```
and then you know by looking at the JSX immediately what you are looping over.

ghepting (approved):

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypePage.tsx
whitelisab: Makes sense, I will update the names of these components from pages to sections

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
whitelisab: yup, this is a good callout, thank you! Updating this now

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
whitelisab: This is a great suggestion; I have made this update

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeCard.tsx
whitelisab: Agreed that this will take bit more effort, so I am going to leave it as is for now. I have refactored the header labels to reduce repeatability.

ryunsong-contentful (approved):

PR #2730, opened 2023-03-09
INTEG-191 Connect sidebar chart with real GA4 data

lilbitner:
## Purpose
The purpose of this PR is to connect the frontend implementation of the Analytics app (currently rendered in the Side bar), with the endpoint communicating with the Google API. 

This is fully functioning with a hard coded `propertyId` and `slug`. 

### **Things to note:**
I wanted to open this PR ASAP to get feedback, questions, comments, but am still working thru the following quirks: 
1. For some reason the dates returned from the endpoint are not sent chronologically. First thought is this might be resolved by passing an appropriate dimension to the `runReports` Google endpoint. if not, I will resort to sorting within our API. Still working through this, but otherwise the dates are being handled and represented appropriately (I believe).
2. Any changes in the `useKeyService` file are likely going to be adjusted with Lisa's work as well as a ticket already laid out to refactor. The changes here are to quickly support this functionality. 

## Approach

The API parameters are passed as query params. I went back and forth a lot with this and do feel that `propertyId` and `slug` should be path params instead, but this is awkward given the `properties/{number} `format of the `propertyId`. Any ideas to resolve this are welcome!

## How to test

Pull down this code and hardcode the `reportSlug` and `propertyId` props within the `Sidebar` component. As per the demo, render the app within the side bar of a content entry, and this should work appropriately but let me know if it does not!

I am still experiencing issues with a `verifiySignedRequests` file so this might inhibit communication (?) (as a warning - I had to comment out some of the code locally).


apps/google-analytics-4/frontend/src/apis/api.ts
```diff
@@ -82,9 +92,9 @@ export class Api {
 
   // TODO: When we actually hook this up to the sidebar chart, we will need to update this type and schema (similar to the listAccountSummaries pattern)
   // It's currently typed like this to provide maximum flexibility until we actually integrate with the chart
-  async getRunReportData(): Promise<RunReportData> {
+  async runReports(params?: RunReportParamsType): Promise<RunReportData> {
```
lilbitner, line 95: made this name change as it seemed to be a bit more succinct and follows the naming of the google endpoint itself but am for sure open to reverting the change

apps/google-analytics-4/frontend/src/apis/api.ts
```diff
@@ -65,9 +66,18 @@ export class Api {
     );
   }
 
-  private requestUrl(apiPath: string): URL {
+  encodeParam = (param: any) => encodeURIComponent(param);
+
+  appendQueryParams = (url: URL, queryParams: any) => {
+    Object.keys(queryParams).forEach((key) => url.searchParams.append(key, queryParams[key]));
```
lilbitner, line 72: would appreciate any opinions on encoding these query strings (using line 69).

apps/google-analytics-4/frontend/src/hooks/useKeyService.tsx
```diff
@@ -37,60 +39,71 @@ export default function useKeyService(props: Props): KeyServiceInfoType {
...
+  const sdk = useSDK<any>();
 
-    if (!serviceAccountKeyFileIsValid) {
-      sdk.notifier.error('Invalid service account key file. See field error for details');
-      return false;
-    }
+  init((sdkLocation) => {
+    if (sdkLocation.location.is(locations.LOCATION_APP_CONFIG)) setIsConfigureScreen(true);
+    else setIsConfigureScreen(false);
+  });
```
lilbitner, line 47: All of these changes in this file will likely be adjusted with a refactor of this service. Just needed to add some conditional statements to assure this works in the config screen AND the side bar.

apps/google-analytics-4/lambda/src/controllers/apiController.ts
```diff
@@ -31,7 +31,10 @@ const ApiController = {
     try {
       const serviceAccountKeyFile = req.serviceAccountKey;
       const { propertyId, slug, startDate, endDate, dimensions, metrics } =
-        req.params as unknown as RunReportParamsType;
+        req.query as unknown as RunReportParamsType;
```
lilbitner, line 34: query vs param as mentioned in PR description. For now they are all queries, open to thoughts and suggestions!

apps/google-analytics-4/lambda/src/controllers/apiController.ts
```diff
@@ -31,7 +31,10 @@ const ApiController = {
...
+
+      const formatArrays = (param: string | string[]) =>
+        Array.isArray(param) ? param : param.split(',');
```
lilbitner, line 37: also wondering if there is an easier way to format arrays sent as query params. I dug thru the internet but am sure there is something a bit more succinct than this.

apps/google-analytics-4/frontend/src/apis/api.ts
ryunsong-contentful: I'll chime my opinion here, I think for a proxy architecture - we should just copy the endpoints name (like you did here with Google since the proxy/lambda is just literally forwarding the request for us). 

In a standard REST frontend/backend structure, we should use standard CRUD names like `create`, `get`, `update`, `delete` in our controller naming (which is not the case here).

apps/google-analytics-4/frontend/src/apis/apiTypes.ts
```diff
@@ -23,8 +23,45 @@ const ZAccountSummary = z.object({
 export const ZAccountSummaries = z.array(ZAccountSummary);
 export type AccountSummaries = z.infer<typeof ZAccountSummaries>;
 
-// TODO: Update once we know exactly the shape of the run report GA data
-export const ZRunReportData = z.object({});
+const ZRow = z.object({
```
ryunsong-contentful, line 26: nice, thank you for filling this out!

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
```diff
@@ -1,29 +1,44 @@
 import AnalyticsApp from './AnalyticsApp';
 import { render, screen } from '@testing-library/react';
+import { mockSdk, mockCma } from '../../../../test/mocks';
 import runReportResponseHasViews from '../../../../../lambda/public/sampleData/runReportResponseHasViews.json';
 import runReportResponseNoView from '../../../../../lambda/public/sampleData/runReportResponseNoViews.json';
+import { ServiceAccountKey, ServiceAccountKeyId } from '@/types';
```
ryunsong-contentful, line 6: nit: `/types`

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -1,38 +1,54 @@
...
-import { config } from 'config';
+import { useEffect, useState, useMemo } from 'react';
+import { useApi } from 'hooks/useApi';
+import getRangeDates from 'helpers/handle-date-range/handle-date-range';
+import ChartContent from '../ChartContent';
+import { DateRangeType, ServiceAccountKeyId, ServiceAccountKey } from 'types';
 import { styles } from './AnalyticsApp.styles';
-import ChartContent from 'components/main-app/ChartContent';
 import { Flex, Note } from '@contentful/f36-components';
+import { RunReportData } from '@/apis/apiTypes';
```
ryunsong-contentful, line 11: same import question as above

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -1,38 +1,54 @@
...
+  serviceAccountKeyId: ServiceAccountKeyId;
+  serviceAccountKey: ServiceAccountKey;
+  propertyId: string;
+  reportSlug: string;
+}
+const AnalyticsApp = (props: Props) => {
+  const { serviceAccountKeyId, serviceAccountKey, propertyId, reportSlug } = props;
+  const [runReportResponse, setRunReportResponse] = useState<RunReportData>({} as RunReportData);
   const [dateRange, setDateRange] = useState<DateRangeType>('lastWeek');
+  const [startEndDates, setStartEndDates] = useState<any>(getRangeDates('lastWeek')); // TYPE
```
ryunsong-contentful, line 26: Can we create an Object that maps the string like 
```
const DATE_RANGES = {
  lastWeek: 'lastWeek',
  lastMonth: 'lastMonth',
}
```
Then we can just call those values instead of writing those strings in every time

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -1,38 +1,54 @@
 import { useAutoResizer, useFieldValue } from '@contentful/react-apps-toolkit';
 import ChartFooter from 'components/main-app/ChartFooter';
 import ChartHeader from 'components/main-app/ChartHeader';
-import { useEffect, useState } from 'react';
-import { DateRangeType, Row, RunReportResponse } from 'types';
-import DateRange from 'helpers/dateRange.enum';
-import { config } from 'config';
+import { useEffect, useState, useMemo } from 'react';
```
ryunsong-contentful, line 4: I find this name to be confusing because of the term `App`. Why is the entire directory structure called app? 
Doesn't this file and the directory more closely resemble the rendering of the sidebar? 
What about the name `/sidebar/graph/SidebarAnalyticGraph.tsx` or something like that - I'm sure you can come up with something better than I.

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -1,38 +1,54 @@
...
 
+  useEffect(() => {
     async function fetchData() {
       try {
-        const response = await fetch(`${baseUrl}/sampleData/runReportResponseHasViews.json`);
-        if (response.ok) {
-          setRunReportResponse(await response.json());
-        }
+        const reportData = await api.runReports(reportRequestParams);
+        setRunReportResponse(reportData);
```
ryunsong-contentful, line 51: I don't think you have to solve this problem here, but I think we should add a TODO to handle the error similar to the `DisplayServiceAccountPage.tsx`

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -1,38 +1,54 @@
...
+      propertyId,
+      dimensions: ['date'],
+      metrics: ['screenPageViews'],
+      slug: reportSlug,
+    }),
+    [startEndDates.start, startEndDates.end, reportSlug, propertyId]
+  );
 
+  useEffect(() => {
     async function fetchData() {
```
ryunsong-contentful, line 48: nit: fetchRunReportData? Or something more descriptive

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -41,30 +57,20 @@ const AnalyticsApp = () => {
...
-          break;
-      }
-      return { ...runReportResponse, rows: [...newRows] };
-    };
     if (runReportResponse.rowCount) {
-      const newData = sliceByDateRange(dateRange);
-      setPageViewData(newData);
+      setStartEndDates(getRangeDates(dateRange));
+      setRunReportResponse(runReportResponse);
+      setLoading(false);
```
ryunsong-contentful, line 66: I don't think there will be a bug here per-se atm but you have two different useEffects setting the loading state to false. So if one useEffect completes, then the component will render below. 
1. Either you need two different loading states or marriage the useEffects together to settle the loading state

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -41,30 +57,20 @@ const AnalyticsApp = () => {
...
     }
   }, [dateRange, runReportResponse]);
 
+  useEffect(() => {
```
ryunsong-contentful, line 70: I think we can get rid of this useEffect and just do the logic in the `handleDateRangeChange` function because the only time the dateRange changes is when the user does an action right?

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -41,30 +57,20 @@ const AnalyticsApp = () => {
...
-        case DateRange.LastMonth:
-          newRows = runReportResponse.rows.slice(0, 28);
-          break;
-      }
-      return { ...runReportResponse, rows: [...newRows] };
-    };
     if (runReportResponse.rowCount) {
-      const newData = sliceByDateRange(dateRange);
-      setPageViewData(newData);
+      setStartEndDates(getRangeDates(dateRange));
```
ryunsong-contentful, line 64: If the runReportResponse changes, why do we need to reset the dateRange again? I'm probably missing something.

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -85,15 +91,15 @@ const AnalyticsApp = () => {
           <p className={styles.noteContent}>{error?.message || DEFAULT_ERR_MSG}</p>
         </Note>
       );
-    } else if (!pageViewData.rowCount) {
+    } else if (!runReportResponse.rowCount) {
```
ryunsong-contentful, line 94: I know you didn't do this but in line 91, you don't need the `?` because you're already checking if `error` is undefined/null in line 88

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
ryunsong-contentful: Found it, it's in `handle-date-range.ts`

apps/google-analytics-4/frontend/src/helpers/handle-date-range/handle-date-range.ts
```diff
@@ -0,0 +1,31 @@
+import { DateRangeType } from 'types';
```
ryunsong-contentful, line 1: another nitty file naming complaint - i feel like the term `handle` should be reserved for function naming. This feels more like a DateRangeUtils or DateRangeHelpers type of file. Also, I think Gary is pushing for camel casing file names and not using dashes.

apps/google-analytics-4/frontend/src/hooks/useKeyService.tsx
```diff
@@ -25,6 +25,8 @@ interface Props {
 export default function useKeyService(props: Props): KeyServiceInfoType {
   const { onSaveGoogleAccountDetails } = props;
 
+  const [isConfigureScreen, setIsConfigureScreen] = useState<boolean>(false); //TYPE
```
ryunsong-contentful, line 28: Extraneous comment?

apps/google-analytics-4/frontend/src/hooks/useKeyService.tsx
ryunsong-contentful: Yes, this will definitely need to be refactored but looks good for now!

ryunsong-contentful:
Added minor comment changes but overall, this is great!

apps/google-analytics-4/frontend/src/hooks/useKeyService.tsx
lilbitner: good catch! 🦅

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
lilbitner: good point, will make a ticket!

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
lilbitner: will rename that file to be more generalized ^^^ will make it easier to find!

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
lilbitner: I think the intention here is that this component can potentially exist not just in the side bar location - I think as we build out more apps we should perhaps build them agnostic to location until it is necessary to define a location for them. Since it is the entire "app" itself that the user sees, I believe that is why this file was named `AnalyticsApp`. What are your thoughts on that?

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
lilbitner: I think you are right some of this useEffect logic can absolutely be consolidated. Will adjust 💯

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
lilbitner: ah yes I believe this is remnant of the previous logic we had slicing the dates manually instead of letting the API handle it. Thanks for pointing this out, I am going to remove one of these loadings states !

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
lilbitner: absolutely, love the idea of consolidating this 💯

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
ryunsong-contentful: Ahhh I see, I think not calling it sidebar makes sense. I still think the word app is overloaded but it's no big deal. I can commit to this np

ryunsong-contentful (approved):

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -1,7 +1,21 @@
 import AnalyticsApp from 'components/main-app/analytics-app/AnalyticsApp';
+import useKeyService from 'hooks/useKeyService';
+
+// TO DO: Add propertyId and reportSlug logic
 
 const Sidebar = () => {
-  return <AnalyticsApp />;
+  const { parameters } = useKeyService({});
```
jsdalton, line 7: Hm, I think we should _not_ get the parameters from the `useKeyService` and instead just fetch them directly from the app sdk. Then you can avoid all those refactorings you're needing to do in that hook.

apps/google-analytics-4/frontend/src/hooks/useKeyService.tsx
```diff
@@ -37,60 +39,71 @@ export default function useKeyService(props: Props): KeyServiceInfoType {
...
     newServiceAccountKey,
     parameters,
     onSaveGoogleAccountDetails,
+    isConfigureScreen,
   ]);
 
   useEffect(() => {
-    sdk.app.onConfigure(() => onConfigure());
-  }, [sdk, onConfigure]);
+    if (isConfigureScreen) sdk.app.onConfigure(() => onConfigure());
```
jsdalton, line 97: This is a bit of a smell and more evidence that the hook is the wrong place to organize all this code. cc @ryunsong-contentful 

I have a suggestion down below that we avoid this hook altogether in your PR, since we have targeted it to be refactored away.

ryunsong-contentful: Sounds good - need two points of clarification @jsdalton 
1. When you say suggestion below is that in some PR somewhere else? I am unable to locate the suggestion.
2. By my PR do you mean the PR for mapping the account to properties pr?

lilbitner: Great idea. I think this is a good call to just pull the parameters off the sdk and avoid using `theKeyService` given the future refactor. Will adjust that right now.

apps/google-analytics-4/lambda/src/types.ts
```diff
@@ -25,4 +25,16 @@ export interface RunReportParamsType {
   endDate: string;
   dimensions: string | string[];
   metrics: string | string[];
+  sortResponse: boolean;
```
ryunsong-contentful, line 28: this should be deleted right?

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -1,7 +1,21 @@
 import AnalyticsApp from 'components/main-app/analytics-app/AnalyticsApp';
+import { useSDK } from '@contentful/react-apps-toolkit';
+
+// TO DO: Add propertyId and reportSlug logic
 
 const Sidebar = () => {
-  return <AnalyticsApp />;
+  const sdk = useSDK<any>();
+
+  const { serviceAccountKey, serviceAccountKeyId } = sdk.parameters;
```
jsdalton, line 9: Now that we've cleaned this up it leads me to think we _could_ -- instead of passing these low level keys which are only used to set up the API -- just initialize the api here (via useApi) and pass it down directly.

But I'd save for a later refactor if you agree 😄

jsdalton (approved):
Looking great, thanks for those small changes! 💯

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
lilbitner: oh I do like that idea very much, a good separation of concerns. 👍

PR #2737, opened 2023-03-10
Feature/integ 152 Add Wistia Pagination Support

pkeavenycontentful:
## Purpose
The current Wistia integration does not support pagination past 100 videos, this PR adds pagination for up to 2 pages or 200 videos. Pagination for n pages will come in a follow-up. 

## Approach
The problem initially was that the client would only search the first 100 videos of the data set (this being the default max for each request: https://wistia.com/support/developers/data-api#paging-and-sorting-responses 

While backend consideration was considered, in order to have all of the videos available in the index for searching, I decided on frontend pagination so that there isn't additional data needed from the Wistia API for searching. 

https://wistia.com/support/developers/data-api#paging-and-sorting-responses


apps/wistia-videos/src/components/Field.tsx
```diff
@@ -31,37 +33,63 @@ const StyledImageContainer = styled.div`
...
   useEffect(() => {
+    const fetchInitialVideos = async () => {
+      try {
+        const parameters: any = sdk.parameters.installation;
+        const projectIds = await getProjectIds(
+          parameters.excludedProjects,
+          parameters.apiBearerToken
+        );
+        let videos = await fetchPagedVideos(projectIds, parameters);
+        if (videos.length === 100) {
```
ghepting, line 62: does their API not return a `total` or something – this seems very brute forced :(

mgoudy91: It does not! Patrick showed me yesterday your only options are a starting value, and a page count, which can be 100 at max. You also don't get any metadata around the total number of records in your request

pkeavenycontentful: I took another look at this, and I think I found a workaround we can use that's less brute forced 👍

apps/wistia-videos/src/components/ConfigScreen.tsx
```diff
@@ -112,24 +102,26 @@ const Config = (props: ConfigProps) => {
...
-      Notification.error(`Connection to Wistia Data API failed: ${response.error}`);
-      console.log(`Connection to Wistia Data API failed: ${response.error}`);
-      setFetchedProjects([]);
-      if (response?.code && response?.error) {
-        setRequiredMessage(response?.error);
+      setLoadingStatus(false);
+    } catch (error) {
+      if (error instanceof WistiaError) {
+        Notification.error(`Connection to Wistia Data API failed: ${error.message}`);
+        console.log(`Connection to Wistia Data API failed: ${error.message}`);
```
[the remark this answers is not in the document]
pkeavenycontentful: This was present in the code before so wasn't sure whether to remove it. Since we have the `Notification` above, I don't really think we need it so will remove.

apps/wistia-videos/src/functions/getVideos.ts
```diff
@@ -0,0 +1,71 @@
...
+        await fetch(`https://api.wistia.com/v1/projects/${id}.json`, {
+          headers: {
+            Authorization: `Bearer ${bearerToken}`,
+          },
+        })
+      ).json();
+      let videos = project.medias;
+      if (project.mediaCount > 100) {
+        // start at 2 since the first page is already gotten
+        for (let i = 2; i <= getTotalPages(project.mediaCount); i++) {
```
ryunsong-contentful, line 48: nit: I feel it's more natural to do 0 based index logic here
```
        for (let i = 1; i < getTotalPages(project.mediaCount); i++)
```

apps/wistia-videos/src/functions/getVideos.ts
```diff
@@ -0,0 +1,71 @@
...
+          const additionalProject = await (
+            await fetch(`https://api.wistia.com/v1/projects/${id}.json?page=${i}`, {
```
ryunsong-contentful, line 50: I think your `promise.All` call is the right move here with regards to your `projectIds` array. But for these loops when there's > 100 media count, is there no way to just get all of them in one api call and then you can chop them up yourself?

Does calling `https://api.wistia.com/v1/projects/${id}.json` only give you the first page?
If so, why not follow the same pattern and use `Promise.All` here as well?

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -31,37 +33,47 @@ const StyledImageContainer = styled.div`
 
 const Field = (props: FieldProps) => {
   const { sdk } = props;
-  const [data, updateData] = useState<WistiaItem[] | []>([]);
+  const [data, updateData] = useState<WistiaVideo[] | []>([]);
```
ryunsong-contentful, line 36: nit: You can get rid of the basic `[]` typing and just make it 
```
  const [data, updateData] = useState<WistiaVideo[]>([]);
```
and do the following below

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -31,37 +33,47 @@ const StyledImageContainer = styled.div`
...
   const [selectedIds, setIds] = useState<number[]>([]);
-  const [dropdownData, filterDropdownData] = useState<WistiaItem[] | []>([]);
+  const [dropdownData, filterDropdownData] = useState<WistiaVideo[] | []>([]);
```
ryunsong-contentful, line 38: same comment as above

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -31,37 +33,47 @@ const StyledImageContainer = styled.div`
...
   const [loading, updateLoadingStatus] = useState(true);
+  const [end, setEnd] = useState(100);
```
ryunsong-contentful, line 40: maybe make this a constant? Since i see this 100 used in other places.

Maye somethign like `VIDEOS_PER_PAGE`?

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -115,30 +143,43 @@ const Field = (props: FieldProps) => {
...
-                      >
-                        <StyledImageContainer>
-                          <img src={item.thumbnail.url} alt={item.name} />
-                        </StyledImageContainer>
-                      </Card>
-                      <Paragraph>{item.name}</Paragraph>
-                    </GridItem>
-                  ))}
-                </Grid>
+              <Flex
```
ryunsong-contentful, line 146: nit: If you made those state changes in the comments above , on line 138 you can now just make this 
```
data.length
```

ryunsong-contentful (approved):
Overall looks good, I think the biggest thing is the looping of fetch calls. But I'm approving now because I'm going to assume you'll address the comments by making changes or responding via comments and more importantly to get this moving along.

apps/wistia-videos/src/functions/getVideos.ts
```diff
@@ -35,26 +37,33 @@ export const fetchVideos = async (
...
-        }
+      let videos: WistiaVideo[] = project.medias;
+
+      if (project.mediaCount > VIDEOS_PER_PAGE) {
+        // create additional array so we can use promise.all and pattern match with above
+        const additionalProjectMedias: WistiaVideo[] = await Promise.all(
+          Object.keys(new Array(getTotalPages(project.mediaCount)).fill(0)).map(
+            async (page: string) => {
+              const response: Response =
+                await // start at 2nd page since 1st page is already gotten before this
```
ryunsong-contentful: Could be the formatting that's confusing me but I think this `await` is not needed

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -107,38 +137,51 @@ const Field = (props: FieldProps) => {
         </Paragraph>
       ) : (
         <>
-          {data?.length > 0 && !error ? (
+          {data?.length && !error ? (
```
ryunsong-contentful, line 140: You dont need the `?` now. You can just make it `data.length`

apps/wistia-videos/src/components/Field.tsx
ryunsong-contentful: Resolving this beacause it's the same comment on line 140

apps/wistia-videos/src/functions/getVideos.ts
```diff
@@ -0,0 +1,82 @@
...
+      ).json();
+      let videos: WistiaVideo[] = project.medias;
+
+      if (project.mediaCount > VIDEOS_PER_PAGE) {
+        // create additional array so we can use promise.all and pattern match with above
+        const additionalProjectMedias: WistiaVideo[] = await Promise.all(
+          Object.keys(new Array(getTotalPages(project.mediaCount)).fill(0)).map(
+            async (page: string) => {
+              // start at 2nd page since 1st page is already gotten before this
+              const response: Response = await await fetch(
```
ryunsong-contentful, line 55: You only need one `await` here right?

apps/wistia-videos/src/functions/getVideos.ts
```diff
@@ -0,0 +1,82 @@
...
+  const mappedProjects = await Promise.all(
+    projectIds.map(async (id: string) => {
+      const project: WistiaProject = await (
+        await fetch(`https://api.wistia.com/v1/projects/${id}.json`, {
+          headers: {
+            Authorization: `Bearer ${bearerToken}`,
+          },
+        })
+      ).json();
+      let videos: WistiaVideo[] = project.medias;
```
ryunsong-contentful, line 47: move this variable down to line 68 and make this a const

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -75,6 +89,22 @@ const Field = (props: FieldProps) => {
...
+    excludedProjects: Project[],
+    bearerToken: string
+  ): Promise<string[]> => {
+    const projects = await fetchProjects(bearerToken);
+    console.info('Succesfully fetched the Wistia projects.');
+    // Map through projects and return ids to retrieve all the videos from each project. Filter out the projects selected to be excluded
+    return projects
+      .map((item: WistiaProject) => item.hashedId)
+      .filter((id: string) => {
+        const include =
```
ryunsong-contentful, line 102: Maybe I'm thinking about this wrong, but why can't you just call `excludedProjects.includes(id)` which will return  a boolean if the id is in that array or not

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -75,6 +89,22 @@ const Field = (props: FieldProps) => {
     setNewValues(updatedIds);
   };
 
+  const getProjectIds = async (
+    excludedProjects: Project[],
+    bearerToken: string
+  ): Promise<string[]> => {
+    const projects = await fetchProjects(bearerToken);
+    console.info('Succesfully fetched the Wistia projects.');
```
ryunsong-contentful, line 97: nit: delete?

apps/wistia-videos/src/components/Field.tsx
```diff
@@ -29,39 +31,51 @@ const StyledImageContainer = styled.div`
...
+      } finally {
+        updateLoadingStatus(false);
+      }
+    };
+
     const fieldValues = sdk.field.getValue();
     setIds(
       fieldValues !== undefined && fieldValues.items.length > 0
-        ? fieldValues.items.map((item: WistiaItem) => item.id)
+        ? fieldValues.items.map((item: WistiaProject) => item.id)
```
ryunsong-contentful, line 74: I know you didn't write this but in line 73, if `fieldValues` is null, it'll pass this check. We should just do 
`fieldValues && fieldValues.items ? etc...`

jsdalton (commented):
As I mentioned in our side conversation, I think we need to solve for `fetchProjects()` here in addition to what you've already solved for with fetchVideos. Otherwise if the customer has > 100 projects we will not correctly fetch their videos (or allow them to exclude unneeded projects).

apps/wistia-videos/src/components/Field.tsx
pkeavenycontentful: Fixed 👍

apps/wistia-videos/src/functions/getVideos.ts
pkeavenycontentful: Yeah, not sure how that other one got in there, it's removed now.

apps/wistia-videos/src/functions/getVideos.ts
pkeavenycontentful: I actually just deleted this var and concatted the new array to the existing `project.medias`

apps/wistia-videos/src/components/Field.tsx
pkeavenycontentful: Good question, I didn't write this code initially (pasted it from what was there elsewhere in the file). Do you want me to change to use `includes`? I know it's not ideal where it is now, but it is working.

apps/wistia-videos/src/components/Field.tsx
pkeavenycontentful: Fixed 👍

apps/wistia-videos/src/functions/getVideos.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      if (project.mediaCount > VIDEOS_PER_PAGE) {
+        // create additional array so we can use promise.all and pattern match with above
+        const additionalProjectMedias: WistiaVideo[] = await Promise.all(
+          Object.keys(new Array(getTotalPages(project.mediaCount)).fill(0)).map(
```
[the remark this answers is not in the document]
pkeavenycontentful: Done 👍

apps/wistia-videos/src/components/Field.tsx
ryunsong-contentful: I think it makes it simpler but if you are worried about regression, I think it is fine to leave as is!

ryunsong-contentful (approved):

apps/wistia-videos/src/functions/getVideos.ts
```diff
@@ -0,0 +1,97 @@
...
+  accumulated: WistiaProject[]
+): Promise<WistiaProject[]> => {
+  let projects = await (
+    await fetch(`https://api.wistia.com/v1/projects.json${page ? '?page=' + page : ''}`, {
+      headers: {
+        Authorization: `Bearer ${bearerToken}`,
+      },
+    })
+  ).json();
+  accumulated = [...accumulated, ...projects];
```
jsdalton, line 24: Small thing that might not matter, but I'd consider appending new results as opposed to prepending them as you do here, in case there is ordering from the API that matters (e.g. some specified sort order).

apps/wistia-videos/src/functions/getVideos.ts
```diff
@@ -0,0 +1,97 @@
...
+  if (projects.error) {
```
jsdalton, line 25: This clause probably belongs just after the fetch call, since we'd probably want to throw right away after fetching as opposed to adding the accumulated resultss?

jsdalton (approved):
Cool, I think this will work. 👍 Mostly small suggestions from me for you to consider.

apps/wistia-videos/src/functions/getVideos.ts
pkeavenycontentful: Not sure I follow. `accumulated` are the results from previous calls, while `projects` are the results from a newer call based on pagination. Wouldn't this pattern follow what you're describing?

apps/wistia-videos/src/functions/getVideos.ts
pkeavenycontentful: Good call, this is moved up 👍

apps/wistia-videos/src/functions/getVideos.ts
jsdalton: Yes, you are right. That's just me being dumb! 🤦

PR #2740, opened 2023-03-10
adds post-deploy step

ghepting:
## Purpose
adds sentry CLI for uploading source maps and creating the release to GA4 app

## Approach
adds post-deploy script hook


.circleci/config.yml
```diff
@@ -74,12 +74,19 @@ jobs:
             sudo apt-get update
             sudo apt-get install python3-pip python3-dev
             sudo pip3 install awscli
+      - run:
+          name: Install Sentry CLI
+          command: |
+            npm install @sentry/cli
```
jsdalton, line 80: Since sentry is now specific to our app i think we should _not_ install it globally. Instead can we include it in GA4 and/or do `npx sentry-cli ...`?

.circleci/config.yml
```diff
@@ -74,12 +74,19 @@ jobs:
...
       - run:
           name: Build
           command: |
             REACT_APP_BACKEND_BASE_URL=$BACKEND_BASE_URL_TEST \
             REACT_APP_SLACK_CLIENT_ID=$SLACK_CLIENT_ID_TEST \
             npm run build
+      - run:
+          name: Post Deploy
+          command: npm run post-deploy
```
jsdalton, line 89: I notice this is happening _before_ the actual deploy, don't we want it after?

jsdalton: Also in keeping with `npm run deploy:test` let's do `npm run post-deploy:test` for this action specifically, and rename in our project and the main apps/package.json accordingly

jsdalton (changes requested):

.circleci/config.yml
ghepting: but the release is not – it's specific to this whole repository – and the sentry project is as well.

.circleci/config.yml
ghepting: fixed the order, thanks for catching that 😅

.circleci/config.yml
ghepting: I'm picturing that we do this globally as soon as we want to add the next app and then each app is just responsible for their own source and source map files part. open to thoughts about this

jsdalton: my opinion here is that we should keep the installation of the dependency coupled to the place where it's used. so if we have a script in GA4 then it _should_ be declared in GA4 and not implicitly depend on an outside context. inverse would be true too, why install a global dependency when it's only used in one place?

ghepting: yeah fair enough. we can rearrange it later on then. i'll adjust – thanks for the input

jsdalton (approved):
Looks good, thanks! Consider making it a dev dependency.

ghepting:
> Looks good, thanks! Consider making it a dev dependency.

👍 done

mgoudy91 (approved):
Yep I love this as an establishing pattern around our deployments: provide lots of hooks at the highest level for build configuration and tooling, and leave it up to each app to decide to implement

PR #2743, opened 2023-03-10
[INTEG-168] Render GA4 accounts and properties selection in the Configuration Page

ryunsong-contentful:
## Purpose
[JIRA Ticket](https://contentful.atlassian.net/browse/INTEG-168)
Adding in the Account to Properties mapping.

## Approach
Went with a simple dropdown approach. This is to accomplish the main goal of making sure everything mechanically works before improving the UX/UI

This should be all tested with Lillians and Lisa's PRs to make sure everything is working e2e

Screenshot 2023-03-10 at 1 58 14 PM
Screenshot 2023-03-10 at 1 58 18 PM


apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
```diff
@@ -74,7 +75,10 @@ const DisplayServiceAccountCard = (props: Props) => {
       setAdminApiError(undefined);
     } catch (e: any) {
       if (isApiErrorType(e)) handleApiError(e);
-      else throw new Error('Unhandled error exception');
+      else {
+        setUnknownError(e);
+        throw new Error('Unhandled error exception');
```
mgoudy91, line 80: I love having more error handling, but did you ever hit these in development?

If so we should probably try to know what happened ha. Otherwise, this works

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  return (
+    <Stack spacing="spacingL" marginBottom="none" flexDirection="column" alignItems="flex-start">
+      <Box>
+        <Subheading>Configuration</Subheading>
+        <Paragraph>Choose your Account and associated Property</Paragraph>
+        {!loadingProperties &&
+        !loadingParameters &&
+        accountsSummaries.length &&
+        flattenedProperties ? (
```
mgoudy91, line 62: I think this is a little bit of a grey area, but I personally think when render logic has this many moving pieces it's worth pulling out into its own method for readability. It would also make it easier to progressively display loading state depending on the timing of these requisite objects.

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
ryunsong-contentful: So this is some tribal knowledge I'll need to share with the team. So the way this works is that you'll notice in line 77. We run a check in the `isApiErrorType` function. What this does is basically tell us if this error is something we know about (ie Admin API failure, Data API Failure, bad credentials with the service keys etc...) So anything that we know about in the lambda, we will map it also in the client side here.

However, what about errors that we never accounted for, or things we never considered. They all go into this bucket, to which then we'll need to debug and add to our list of known errors described above. 

So the whole point of this is to capture the literally unknown errors we never accounted for. I guess I could print e.message in this case but hopefully you understand the greater architecture context.

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+      },
+    ],
+    name: 'accountSummaries/258082603',
+    account: 'accounts/258082603',
+    displayName: 'contentful-fake-test-2',
+  },
+];
+
+describe('Empty Account Properties mapping dropdown', () => {
+  it('does not renders the dropdown', async () => {
```
whitelisab, line 51: nit: renders should be render

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
ryunsong-contentful: you mean something like this?
```
  const shouldRenderDropdown = () => {
    return (
      !loadingProperties && !loadingParameters && accountsSummaries.length && flattenedProperties
    );
  };

  return (
    <Stack spacing="spacingL" marginBottom="none" flexDirection="column" alignItems="flex-start">
      <Box>
        <Subheading>Configuration</Subheading>
        <Paragraph>Choose your Account and associated Property</Paragraph>
        {shouldRenderDropdown() ? (
          <Select
```

whitelisab (commented):
Looks good to me! I missed this on my PR initially as well, but [this part](https://github.com/contentful/apps/blob/master/apps/google-analytics-4/frontend/src/components/config-screen/api-access/ApiAccessSection.spec.tsx#L93-L97) of `ApiAccessSection.spec.tsx` needs to be updated with the new `savedPropertyId` property in order for that test to pass.

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.spec.tsx
ryunsong-contentful: english is hard

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
mgoudy91: Yeah that's one option that I think could work fine, another would be something like a `renderDropdown` that we would call in the general render, that could handle everything from rendering the dropdown on the happy path, or things like spinners / error messages for other states. 

Right tool for the job right? what you proposed looks great and I think makes the presentational aspect of this way better 👍

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
mgoudy91: Yeah definitely, I figured this was to handle errors we haven't seen, I just saw this block twice and figured I'd ask

mgoudy91 (approved):
SEND IT

PR #2750, opened 2023-03-13
[INTEG-242] Fix incorrect parsing of server errors

ryunsong-contentful:
## Purpose
[JIRA TICKET](https://contentful.atlassian.net/browse/INTEG-242)
Fixes bug with run_report throwing parsing errors and updates UI to handle unexpected errors

## Approach
Issue was with a change that was made with the run_report controller without testing the config page. This fix makes it so that the parsing only occurs when it is called from the sidebar, or anytime the params are passed. 

The config page passes no params because it only checks to see if the run_report endpoint throws some API related error.


lilbitner (approved):
LGTM, thanks for fixing 👍

ghepting (approved):

PR #2762, opened 2023-03-13
[INTEG-209 & INTEG-249]  Error types are misnamed (they all say error) and throw generic error e for Sentry

ryunsong-contentful:
## Purpose
[JIRA Link](https://contentful.atlassian.net/browse/INTEG-209) - Fix error mapper
[JIRA Link](https://contentful.atlassian.net/browse/INTEG-249) - Throw errors instead of modifying the message so that sentry prints out what the actual error was.

- Also update ERROR TYPE MAP to use PascalCasing
- Updates zod error typing to only check for strings in fetchApi validate response instead of specific error types

## Approach
The two tickets were small enough related tasks to throw into one PR.
Using `constructor.name` did not work for the error map. It still printed out Error, so string literals were used instead

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
@@ -47,9 +47,6 @@ const DisplayServiceAccountCard = (props: Props) => {
 
   const handleApiError = (error: ApiErrorType) => {
     switch (error.errorType) {
-      case ERROR_TYPE_MAP.unknown:
-        setUnknownError(error);
-        throw new Error(error.message);
       case ERROR_TYPE_MAP.invalidServiceAccount:
         setInvalidServiceAccountError(error);
         break;
@@ -65,7 +62,7 @@ const DisplayServiceAccountCard = (props: Props) => {
         break;
       default:
         setUnknownError(error);
-        throw new Error('Unhandled error exception');
+        throw error;
     }
   };
 
@@ -79,7 +76,7 @@ const DisplayServiceAccountCard = (props: Props) => {
       if (isApiErrorType(e)) handleApiError(e);
       else {
         setUnknownError(e);
-        throw new Error('Unhandled error exception');
+        throw e;
       }
     } finally {
       setIsLoading(false);
@@ -104,7 +101,7 @@ const DisplayServiceAccountCard = (props: Props) => {
       if (isApiErrorType(e)) handleApiError(e);
       else {
         setUnknownError(e);
-        throw new Error('Unhandled error exception');
+        throw e;
       }
     } finally {
       setIsLoading(false);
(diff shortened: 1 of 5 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
```diff
@@ -65,7 +62,7 @@ const DisplayServiceAccountCard = (props: Props) => {
         break;
       default:
         setUnknownError(error);
-        throw new Error('Unhandled error exception');
+        throw new Error(JSON.stringify(error));
```
jsdalton, line 65: Hm, I have a slightly different idea about what we could/should do here. (But it would be interesting to see how Sentry handles both cases.)

I believe we should just do:

```js
throw error
```

Since all we're doing here is re-throwing the error we caught above, since we weren't able to handle it.

An alternative would be

```js
throw new Error('Something something something', { cause: err })
```

`cause` is a newish language feature (but well supported: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) which preserves the original error but allows you to throw something new if you feel like it.

Honestly we have nothing to add in this case, so I strongly recommend going with `throw error`. 😄

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
```diff
@@ -79,7 +76,7 @@ const DisplayServiceAccountCard = (props: Props) => {
       if (isApiErrorType(e)) handleApiError(e);
       else {
         setUnknownError(e);
-        throw new Error('Unhandled error exception');
+        throw new Error(e);
```
jsdalton, line 79: Per above let's just `throw e`

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
```diff
@@ -104,7 +101,7 @@ const DisplayServiceAccountCard = (props: Props) => {
...
```
jsdalton, line 104: Per above let's just `throw e`

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
```diff
@@ -181,7 +178,7 @@ const DisplayServiceAccountCard = (props: Props) => {
       return (
         <RenderSimpleBadgeNote
           badgeLabel="Unknown Error"
-          noteMessage={'An unexpected error has occurred.'}
+          noteMessage={'Unknown error occurred. Please contact support for help.'}
```
jsdalton, line 181: Cool, I think we can (should?) evolve this but it's actually a just fine start for now. Just a small copy suggestion:

> An unknown error occurred. You can try the action again in a few minutes, or contact support if the error persists.

jsdalton (changes requested):

jsdalton (approved):
Nice 💯

PR #2768, opened 2023-03-14
Integ 221 sidebar querying should make use of content types and property saved in app installation parameters

lilbitner:
## Purpose
This PR simply ties the logic of `propertyId` and `contentType` selection to the fetching of analytics chart data within the `Sidebar` itself. 

## Approach
Pleas refer to the loom video below for an overview of thoughts here. 

https://www.loom.com/share/aec61cf50f204226a53cfff94304ffeb 

You will notice the data is grabbed from the `sdk` parameters directly within the `Sidebar` component. The `Api` instance is also passed down to the `AnalyticsApp` component as per a suggestion on a previous PR by Jim. 

Additionally I re-added a test that was being flaky on Friday with some adjustments.

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
@@ -62,19 +62,6 @@ describe('AnalyticsApp', () => {
     expect(noteText).toBeVisible();
   });
 
-  it('mounts with error message when error thrown', async () => {
-    mockApi.mockRejectedValue(() => new Error('api error'));
-    renderAnalyticsApp();
-
-    const dropdown = await findByTestId(SELECT_TEST_ID);
-    const warningNote = getByTestId(NOTE_TEST_ID);
-    const noteText = getByText('api error');
-
-    expect(dropdown).toBeVisible();
-    expect(warningNote).toBeVisible();
-    expect(noteText).toBeVisible();
-  });
-
   it('renders nothing when it has no response', async () => {
     renderAnalyticsApp();
 
--- a/apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
@@ -1,5 +1,6 @@
 import AnalyticsApp from './AnalyticsApp';
 import { act, render, screen } from '@testing-library/react';
+import { Api } from 'apis/api';
 import { mockSdk, mockCma, validServiceKeyFile, validServiceKeyId } from '../../../../test/mocks';
 import runReportResponseHasViews from '../../../../../lambda/public/sampleData/runReportResponseHasViews.json';
 import runReportResponseNoView from '../../../../../lambda/public/sampleData/runReportResponseNoViews.json';
@@ -12,11 +13,11 @@ jest.mock('@contentful/react-apps-toolkit', () => ({
 }));
 
 const mockApi = jest.fn();
-jest.mock('hooks/useApi', () => ({
-  useApi: () => ({
-    runReports: mockApi,
-  }),
-}));
+// jest.mock('hooks/useApi', () => ({
+//   useApi: () => ({
+//     runReports: mockApi,
+//   }),
+// }));
 
 const { findByTestId, getByTestId, getByText, findByText } = screen;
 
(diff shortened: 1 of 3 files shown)


apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
```diff
@@ -14,24 +14,24 @@ const DEFAULT_ERR_MSG = 'Oops! Cannot display the analytics data at this time.';
...
-  const [slugValue] = useFieldValue<string>('slug');
+  const [slugValue] = useFieldValue<string>(slugField);
   const [loading, setLoading] = useState<boolean>(true);
   const [error, setError] = useState<Error>();
 
-  const api = useApi(serviceAccountKeyId, serviceAccountKey);
-
   useAutoResizer();
 
+  const reportSlug = `${urlPrefix || ''}${slugValue || ''}`;
```
lilbitner, line 33: The legacy code sends an array of these slugs in the form of 'filters', to account for a trailing '/' and a non-trailing '/'. You can see the code [here](https://github.com/contentful/apps/blob/85a95f9c800caedcbc20c7a99ce115b9c742fd48/apps/google-analytics/src/Timeline.tsx#L102). I am not sure why this is the case but feel they must have added it for a reason ?. Right now the accepted `slug` type is only a string at the root of the `googleApiService`. Since it is a larger change I haven't implemented it until validating that the methodology behind these slugs and page paths is correct, and that this formatting is warranted. Keeping an eye on it 👍

lilbitner:
https://www.loom.com/share/aec61cf50f204226a53cfff94304ffeb

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -1,20 +1,22 @@
 import AnalyticsApp from 'components/main-app/analytics-app/AnalyticsApp';
 import { useSDK } from '@contentful/react-apps-toolkit';
-
-// TO DO: Add propertyId and reportSlug logic
+import { useApi } from 'hooks/useApi';
 
 const Sidebar = () => {
   const sdk = useSDK<any>();
```
whitelisab, line 6: You can use the `SidebarExtentionSDK` type here instead of any (imported from `import { SidebarExtensionSDK } from '@contentful/app-sdk';`)

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
```diff
@@ -12,11 +13,6 @@ jest.mock('@contentful/react-apps-toolkit', () => ({
 }));
 
 const mockApi = jest.fn();
-jest.mock('hooks/useApi', () => ({
-  useApi: () => ({
-    runReports: mockApi,
-  }),
-}));
 
 const { findByTestId, getByTestId, getByText, findByText } = screen;
```
whitelisab, line 17: `findByText` is not being used in this file, it can be removed.

whitelisab (approved):
This looks great to me! I made some small comments for you to review

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
lilbitner: oh thank you for this suggestion 👍 have adjusted!

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.spec.tsx
lilbitner: nice catch, have removed! 🦅

ryunsong-contentful (approved):

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
jsdalton: Yeah! so I our URL is:

prefix + slug

and both are user input. (one in the app config screen and the other on the entry itself).

for prefix especially people might put:

* `about/blog`
* `about/blog/`
* `/about/blog/`
* `/about/blog`

we either need to validate those at entry or normalize them in your request.

and i'm not sure how google treats this either 🤔 

if you want to punt on this to a follow up PR, maybe drop a comment here and add a subtask?

jsdalton (approved):
💯  Let's go with this and tackle the few comments in follow up!

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -1,20 +1,23 @@
...
-
-// TO DO: Add propertyId and reportSlug logic
+import { SidebarExtensionSDK } from '@contentful/app-sdk';
+import { useApi } from 'hooks/useApi';
 
 const Sidebar = () => {
-  const sdk = useSDK<any>();
+  const sdk = useSDK<SidebarExtensionSDK>();
+
+  const { serviceAccountKey, serviceAccountKeyId, savedPropertyId, contentTypes } =
```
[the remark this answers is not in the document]
lilbitner: Good thought, will also follow-up on this 👍

apps/google-analytics-4/frontend/src/components/main-app/analytics-app/AnalyticsApp.tsx
lilbitner: https://contentful.atlassian.net/browse/INTEG-256

PR #2795, opened 2023-03-15
Integ 230 optimize date mapping for run report data

lilbitner:
## Purpose
This PR fixes three primary issues: 
1. Fixes an off-by-one issue with the list of dates appearing on the x-axis of the chart.
2. Orders the dates in ascending order with a simply `orderBys` parameter straight to the google `runReports` api. 
3. Hydrates the missing data gaps returned from the `runReports` proxy. The GA4 endpoint is not including dates for which the metric data is zero (this case most often occurs on weekends). This caused strange gaps in time with how our dates were laid out in the chart. Therefore, I have implemented a way to supplement this data so that our axes are consistently spread.

## Approach
1. The dates that are provided from the GA4 app were being translated to local strings, sometimes causing the dates to be shifted one day earlier or later. There is no need to translate these dates to local time as they are direct date references to data, therefore I just formatted them to UTC time, as you can see in the `ChartContent` component. 

2. Replaced our custom sorting logic with the Google logic to order the dates appropriately using the `orderBys` parameter. 

3. Spent a while trying to figure out the best solution for the missing dates provided. The legacy embedded app and endpoint provided empty data, so I dug through lots of documentation to see if there were parameters we could provide to assure all data was provided as the default with the new GA4 endpoint, but to no avail. I felt the most viable solution was to fill the 'date' gaps with the missing dates and their corresponding '0' value. I confirmed with many examples that this supplemental data still properly reflects the data in google analytics, but will continue to validate this logic. 

### Before: 

Screenshot 2023-03-16 at 8 25 05 AM

You will notice both the strange gaps in data (Feb 18th and 19th missing) and the off-by-one with the end-date. 

### After: 
Screenshot 2023-03-16 at 8 30 04 AM

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
@@ -1,5 +1,6 @@
 import { Row, RunReportResponse } from '@/types';
 import LineChart from './line-chart/LineChart';
+import { parseDayAndMonth } from 'helpers/DateHelpers/DateHelpers';
 
 interface Props {
   pageViewData: RunReportResponse;
@@ -14,11 +15,8 @@ const ChartContent = (props: Props) => {
 
   const parseRowDates = (): string[] => {
     return pageViewData.rows.map((r: Row) => {
-      const d = r.dimensionValues[0].value;
-      const utcDate = new Date(
-        `${d.substring(0, 4)}-${d.substring(4, 6)}-${d.substring(6, 8)}`
-      ).toUTCString();
-      const [day, month] = utcDate.substring(5, 11).split(' ');
+      const date = r.dimensionValues[0].value;
+      const { month, day } = parseDayAndMonth(date);
       return `${month} ${day}`;
     });
   };
--- a/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
@@ -28,7 +28,7 @@ const ChartContent = (props: Props) => {
       <LineChart
         dataValues={parseRowViews()}
         xAxisLabels={parseRowDates()}
-        tooltipMetricLabel="Page views"
+        tooltipMetricLabel="Page views:"
         accessibilityLabel="Analytics line chart"
       />
     </>
--- a/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
@@ -15,9 +15,11 @@ const ChartContent = (props: Props) => {
   const parseRowDates = (): string[] => {
     return pageViewData.rows.map((r: Row) => {
       const d = r.dimensionValues[0].value;
-      return new Date(
+      const utcDate = new Date(
         `${d.substring(0, 4)}-${d.substring(4, 6)}-${d.substring(6, 8)}`
-      ).toLocaleDateString('en-us', { month: 'short', day: 'numeric' });
+      ).toUTCString();
+      const [day, month] = utcDate.substring(5, 11).split(' ');
+      return `${month} ${day}`;
     });
   };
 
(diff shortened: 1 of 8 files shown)


apps/google-analytics-4/frontend/src/components/main-app/ChartContent.tsx
```diff
@@ -15,9 +15,11 @@ const ChartContent = (props: Props) => {
   const parseRowDates = (): string[] => {
     return pageViewData.rows.map((r: Row) => {
       const d = r.dimensionValues[0].value;
-      return new Date(
+      const utcDate = new Date(
         `${d.substring(0, 4)}-${d.substring(4, 6)}-${d.substring(6, 8)}`
-      ).toLocaleDateString('en-us', { month: 'short', day: 'numeric' });
+      ).toUTCString();
```
ryunsong-contentful, line 20: I know you have more context, but I'd seriously consider here and use toISOString() and parse that to get the UTC month and day you need.
https://stackoverflow.com/questions/58847869/utc-vs-iso-format-for-time

ryunsong-contentful: Additionally if you do something like this 
```
date.toISOString().split("T")[0] // outputs something like '2023-03-16'
```

You get this and maybe you can write a function that returns a tuple of day and month based on this parsing of the ISO string since this substring logic is done twice

apps/google-analytics-4/lambda/src/services/googleApiService.ts
```diff
@@ -44,14 +45,51 @@ export class GoogleApiService {
...
   }
 
-  sortReportResponse = (rows: ReportRowType[]) => {
-    rows.filter(
-      (row: ReportRowType) => row.dimensionValues?.length && row.dimensionValues[0].value
+  formatDate = (dateValue: Date) => dateValue.toISOString().slice(0, 10).split('-').join('');
+
+  deriveDate = (dateValue: string) =>
+    new Date(
+      `${dateValue.substring(0, 4)}-${dateValue.substring(4, 6)}-${dateValue.substring(6, 8)}`
```
ryunsong-contentful, line 52: see above comment about ISOString

apps/google-analytics-4/lambda/src/services/googleApiService.ts
```diff
@@ -44,14 +45,51 @@ export class GoogleApiService {
...
     );
 
-    rows.sort((v1: ReportRowType, v2: ReportRowType) => {
-      return Number(v1.dimensionValues[0].value) - Number(v2.dimensionValues[0].value);
+  mapDates = (startDate: string, endDate: string) => {
```
ryunsong-contentful, line 55: nit: I don't think this name reflects what the logic is doing here. This is more building a prefix of dates keys. I think something like `prebuildDateKeyArray` would be more appropriate

apps/google-analytics-4/lambda/src/services/googleApiService.ts
```diff
@@ -89,10 +127,22 @@ export class GoogleApiService {
             },
           },
         },
+        keepEmptyRows: true,
+        orderBys: [
```
ryunsong-contentful, line 131: nice!

apps/google-analytics-4/lambda/src/services/googleApiService.ts
lilbitner: really like this idea, will do!

apps/google-analytics-4/lambda/src/services/googleApiService.ts
```diff
@@ -44,14 +45,51 @@ export class GoogleApiService {
...
+
+    while (date <= end) {
+      dates.push(new Date(date));
+      date.setUTCDate(date.getUTCDate() + 1);
+    }
+
+    return dates;
+  };
+
+  supplementDates = (rows: ReportRowType[], _startDate?: string, _endDate?: string) => {
```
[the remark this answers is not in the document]
lilbitner: yeah good point about naming. And the startDate and endDate are just passed in here to then pass into the date `prebuildDateKeyArray` function. But perhaps I can call the function at a different spot to avoid the repetitiveness.

ryunsong-contentful (approved):

PR #2807, opened 2023-03-16
[INTEG-216] Refactor useKeyService into more usable components

ryunsong-contentful:
## Purpose
[JIRA Link](https://contentful.atlassian.net/browse/INTEG-216)
The current implementation of the useKeyService hook has several issues
1. Bloated and ugly to understand
2. Misnamed and has more responsibilities than its name implies
3. Is the incorrect architecture especially in cases where there could be more components that need to use this hook

Fixing these issues will at worst ease and at best unblock future work when it comes to handling the save mechanisms regarding the Configuration page.

Important Side Note: Intentionally adds a regression where in the UI, the content type assignment will not persist. @whitelisab and I talked about this and she will take care of this in her PR to set all the error handling since she has the best context on this.

## Approach
Parent controls the state of the parameters and save onConfigure execution. The children can only do the following
1. use the `mergeSdkParameters` function to update the state
2. update its respective `isValid` state so that onConfigure will fail/return false
3. have read access to the parameters

This architecture makes things more flexible and correctly separates the responsibilities of the parent and children.

Note: Jim noted that there is a need for a callback function (Gary's idea) implementation in the future when it comes to storing/updating secrets.


apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
```diff
@@ -10,17 +10,82 @@ import { AccountSummariesType } from 'types';
...
-  const sdk = useSDK<AppExtensionSDK>();
+  useEffect(() => {
+    const fetchParametersFromSdk = async () => {
+      const _parameters = await sdk.app.getParameters();
+      if (_parameters) {
+        setParameters(_parameters);
+
+        if (_parameters.serviceAccountKey && _parameters.serviceAccountKeyId) {
+          setIsValidServiceAccount(true);
+        }
```
ghepting, line 39: why wouldn't the child components be responsible for this logic?

jsdalton (approved):
Looking good to me! 😎

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
```diff
@@ -1,22 +1,27 @@
...
-  } = useKeyService({});
+import { useSDK } from '@contentful/react-apps-toolkit';
+interface Props {
+  mergeSdkParameters: Function;
+  onIsValidContentTypeAssignment: Function;
+  parameters: KeyValueMap;
+}
+
+const AssignContentTypeSection = (props: Props) => {
+  const { mergeSdkParameters, onIsValidContentTypeAssignment } = props;
```
whitelisab, line 22: We should also destructure the `parameters` prop and this component should have a `useEffect` where the `parameters.contentTypes` is setting the `contentTypes` state. Right now, the saved `contentTypes` in the installation parameters are not being rendered.

whitelisab (commented):
This looks great! I just have one comment on the AssignContentTypeSection component

PR #2812, opened 2023-03-16
INTEG-224: Add DynamoDB config

gordo-contetnful:
## Purpose

This adds the initial conifguration for DynamoDB integration for storing shared credentials related to https://contentful.atlassian.net/browse/INTEG-83

## Approach

- Utilizes `docker-compose.yml` to spin up all resources (services) containerized to use DynamoDB, DynamoDB admin client, DB migrator to add a table in the dev environment 

- Adds serverless config for allocating resources in AWS (DynamoDB, IAM role)

## Dependencies and/or References

closes https://contentful.atlassian.net/browse/INTEG-224

## Deployment

This is being implemented for incremental adoption across a few PRs

PR #2813, opened 2023-03-17
INTEG-225: Add DynamoDB Client

ghost:
## Purpose

Follow up PR to #2812

## Approach

-  Adds DynamoDB client using the `@aws-sdk/client-dynamodb` and `@aws-sdk/lib-dynamodb` 
-  Adds encrypt/decrypt methods for serializing and storing service keys

closes: https://contentful.atlassian.net/browse/INTEG-225, https://contentful.atlassian.net/browse/INTEG-226

## Deployment

Part of an incremental adoption plan over successive PRS


apps/google-analytics-4/lambda/src/services/awsApiService.ts
```diff
@@ -0,0 +1,59 @@
...
+    try {
+      const command = new PutCommand({
+        TableName: input.tableName,
+        Item: {
+          sharedCredentialsId: input.sharedCredentialsId,
+          value: hash, // Add encrypt/decrypt methods
+        },
+      });
+
+      const data = await this.dynamoDbClient.send(command);
```
jsdalton, line 52: I don't really see a huge problem with the way you've done this, but any reason not to use the `put()` method per the documentation as opposed to the low level `send()`? https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB/DocumentClient.html#put-property

apps/google-analytics-4/lambda/src/services/dynamoDbService.ts
```diff
@@ -0,0 +1,63 @@
...
+    client = new DynamoDBClient({
+      region: config.awsRegion,
+      endpoint: config.dynamoDbEndpoint,
+    })
+  ) {
+    this.dynamoDBDocumentClient = DynamoDBDocumentClient.from(client);
+    this.tableName = config.dynamoDbTableName;
+  }
+
+  async getSharedCredentials(input: SharedCredentialsInput) {
```
jsdalton, line 25: Take this as a follow up PR but I'm not seeing the value of sharing `input` between these two methods.

This is a pretty common pattern for getters and setters where you have:

get(key)
set(key, value)

And i think ideally what you're doing is much the same:

getSharedCredentials(sharedCredentialsId)
saveSharedCredentials(sharedCredentialsId, sharedCredentials)

jsdalton (approved):
Looking good, it's progress. I called out a small thing for a subsequent PR but nothing blocking.

apps/google-analytics-4/lambda/src/services/dynamoDbService.ts
gordo-contetnful: You are spot on, I will add this cleanup in the client side PR coming tomorrow 👍

apps/google-analytics-4/lambda/src/controllers/apiController.ts
```diff
@@ -12,6 +13,25 @@ const ApiController = {
...
+
+    // if (!sharedCredentials) {
+    //   const data = await dynamoDB.saveSharedCredentials({
+    //     sharedCredentialsId,
+    //     serviceKey: req.serviceAccountKey!,
+    //   });
+    //   console.log({ data });
+    // } else {
+    //   console.log({ sharedCredentials });
+    // }
```
ghepting, line 33: remove all of this unused psuedo code. we have the final product in another branch

PR #2821, opened 2023-03-17
remove query string from request signing to work around AWS gateway/serverless differences vs development

ghepting:
## Purpose
remove query string from request signing verification to ensure deterministic frontend and backend comparison in production serverless environment contexts

## Approach
- omit query strings from request verification signing
- adds logging so we can revisit once we can verify what format we're receiving in production after passing through aws gateway -> events -> lambda -> node (express)


jsdalton (approved):
🚀

PR #2823, opened 2023-03-17
[INTEG-259] No more flashing statuses on Service Account Card

ryunsong-contentful:
## Purpose
[JIRA Ticket](https://contentful.atlassian.net/browse/INTEG-259)
Fixes bug where statuses flash before loading the Configuration Page finishes

https://user-images.githubusercontent.com/124832189/225998498-22fb4bda-9ee9-421d-b5cd-bf1d81725d6a.mov

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
@@ -34,7 +34,9 @@ const DisplayServiceAccountCard = (props: Props) => {
     isAppInstalled,
   } = props;
 
-  const [isLoading, setIsLoading] = useState(true);
+  const [isLoadingAdminApi, setIsLoadingAdminApi] = useState(true);
+  const [isLoadingDataApi, setIsLoadingDataApi] = useState(true);
+
   const [adminApiError, setAdminApiError] = useState<ApiErrorType>();
   const [dataApiError, setDataApiError] = useState<ApiErrorType>();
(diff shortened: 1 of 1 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
```diff
@@ -240,7 +242,13 @@ const DisplayServiceAccountCard = (props: Props) => {
       </FormControl>
       <FormControl marginBottom="none">
         <FormControl.Label marginBottom="none">Status</FormControl.Label>
-        <Paragraph>{isLoading ? <Spinner variant="primary" /> : <RenderStatusInfo />}</Paragraph>
+        <Paragraph>
+          {isLoadingAdminApi && isLoadingDataApi ? (
+            <Spinner variant="primary" />
```
ghepting, line 247: i'd recommend using a [Skeleton](https://f36.contentful.com/components/skeleton) component vs a Spinner here, but otherwise LGTM

ghepting (approved):

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
ghepting: actually i guess this only makes sense when there is going to be an error. in the case where it's only `active` perhaps this doesn't make sense 🤔 ignore this for now.

pkeavenycontentful (approved):

PR #2824, opened 2023-03-17
[INTEG-267] Handle incomplete content types in sidebar

whitelisab:
## Purpose
This PR adds additional handling to the sidebar component to deal with situations where there are no content types saved or the saved content types are incomplete (no slug field is saved). In these situations we were seeing errors: "Cannot destructure property 'slugField' of 'slugFieldInfo' as it is undefined." and "Missing `fieldId`. The `fieldId` can only be omitted when your app is renderd in field location."

## Approach
If `contentTypes` in installation parameters returns an empty object, then pass an empty string for the slug field and the URL prefix. 

The `useFieldValue` hook throws an error if a `fieldId` is not passed when used in the Sidebar location [see here](https://github.com/contentful/create-contentful-app/blob/master/packages/contentful--react-apps-toolkit/src/useFieldValue.tsx#L54). If no slug field is selected, we want to use an empty string for the value and catch the thrown error. This PR addresses this by calling `useFieldValue` in a custom hook.

Additional error handling and displaying messages the user will be addressed by @lilbitner in INTEG-219. This PR just allows the Sidebar app to be in a working state if saved content types are incomplete.


pkeavenycontentful (approved):

apps/google-analytics-4/frontend/src/hooks/useGetFieldValue.tsx
```diff
@@ -0,0 +1,16 @@
+import { useFieldValue } from '@contentful/react-apps-toolkit';
+
+const useGetFieldValue = (fieldId: string) => {
```
ryunsong-contentful, line 3: I don't think this needs to be a hook actually. Can't you just make this a regular `.ts` file since there's no XML comopnents here. This means you can just make this a normal function that's exported.

One key pattern of a hook is that it almost always has an useEffect inside the hook itself. If it doesn't have an useEffect, then it either has some other hook like memo, callback, or reducer etc...

ryunsong-contentful:
Nice work to fix those sentry errors!

apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -11,7 +11,7 @@ const Sidebar = () => {
     .installation as AppInstallationParameters;
 
   const currentContentType = sdk.contentType.sys.id;
-  const slugFieldInfo = contentTypes[currentContentType];
+  const slugFieldInfo = contentTypes[currentContentType] ?? { slugField: '', urlPrefix: '' };
```
lilbitner, line 14: 💯

lilbitner (approved):

apps/google-analytics-4/frontend/src/hooks/useGetFieldValue.tsx
```diff
@@ -0,0 +1,16 @@
...
+  let fieldValue = '';
+
+  try {
+    const [value] = useFieldValue<string>(fieldId);
+    fieldValue = value ?? '';
+  } catch (e) {
+    // TODO: address handling this error in INTEG-219
+    console.log(e);
```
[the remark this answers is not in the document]
whitelisab: I can change this to `console.error` for now and then the next task will address the rest!

ryunsong-contentful (approved):

apps/google-analytics-4/frontend/src/hooks/useGetFieldValue.tsx
whitelisab: @ryunsong-contentful and I synced about this -- since the `useGetFieldValue` hook is calling `useFieldValue`, this needs to be in a custom hook or else we see the error `React Hooks must be called in a React function component or a custom React Hook function`.

PR #2826, opened 2023-03-17
[Cleanup] Removes Console Logs

ryunsong-contentful:
## Purpose
Just cleaning up console logs in our app

## Approach
Asked the man @ghepting himself


ghepting:
Actually I think we want to leave these for a while! cc @jamessam

ryunsong-contentful:
Alright closing this pr then!

PR #2827, opened 2023-03-17
Zend 3321 Slack config fix

mgoudy91:
## Purpose
The slack app is experiencing an issue where it fails to find Contentful-hosted configuration information, because it uses a resolved alias environment name instead of the canonical environment name when writing configuration details to a dynamoDB instance. This information is transmitted used the default CM API headers, which do not not yet expose the canonical environment name. 

## Approach
For now, we need to append the canonical environment name and then receive and store that name when doing first-time configuration of the slack app. Eventually this information may (and should) be exposed by the SDK, which would make the front end component of this work no longer needed.

[support link](https://contentful.atlassian.net/browse/ZEND-3321)

### DIFF
--- a/apps/slack/Dockerfile
+++ b/apps/slack/Dockerfile
@@ -2,16 +2,14 @@ FROM node:14-alpine AS base
 
 RUN apk add --update --no-cache openssh build-base python3 bash git && rm -rf /var/cache/apk/*
 
-WORKDIR /app
+WORKDIR /app/lambda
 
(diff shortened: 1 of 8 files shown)


mgoudy91:
@ghepting There's still something going on here, notifications aren't triggering from the aliased space. It's probably a similar issue, I'll look into it monday

ghepting:
> @ghepting There's still something going on here, notifications aren't triggering from the aliased space. It's probably a similar issue, I'll look into it monday

Ack. Let's pair some more Monday morning

PR #2839, opened 2023-03-20
INTEG-219 The sidebar app should only be visible for configured content types that have a slug field selected

lilbitner:
## Purpose
This PR adds handling for cases when content types may be incompatible with rendering the GA4 app itself. 

## Approach
As per the legacy code and ticket, there are three cases handled here: 
1. The content type has not been configured in the app config, but the app has been added to the content type manually
2. The content type does not have a valid slug field configured
3. The entry hasn’t been published

The main logic to determine when these incompatibilities exist is located in the new `useSidebarSlug` hook. I felt it was necessary to clean-up some of the ever growing files and separating things into hooks or smaller components such as `AnalyticsMetricDisplays` and the `Note` component (was going to wait until we had Storybook for this, but felt custom styling this atom/molecules and having it live in its own file was warranted now).

I decided to render the warning note in place of all metric displays in these cases, as it seems unnecessary to display the header and footer when no date is associated. In the cases of an API error or empty data, the note still appears in place of the chart only as before. Ex: 
Screenshot 2023-03-21 at 9 29 57 AM

I also decided to not call the runReports endpoint if we do not have the required information and if there is a content related incompatibility. You will notice this change in the `AnalyticsApp` `fetchRunReportDate` method. 

Additionally, I added a lot of testing to the new components and to some components that may not have had tests. I also changed some sub directory names under `main-app` to follow the camel case structure that we have throughout that directory itself - I feel some agreed upon consistency is necessary, so fine either or but figured I would change this for now to see how we like it. Happy to revert any name changes.

Additionally part 2, this PR resolves the issue Gavin posted yesterday around the warning note inappropriately rendering within the Sidebar, despite there existing data. Was a simple adjustment of the conditional that determines its render. 

## Questions/notes

One thing I noticed is that if the content has not yet been published, and the user publishes the content, the change is not broadcasted given the current code structure, and I am looking into if there is a native hook we can use to monitor these changes - **working on implementing use of the `onSysChange` callback as part of SDK - might be in follow-up PR.**


apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
```diff
@@ -0,0 +1,10 @@
+export const DEFAULT_ERR_MSG = 'Oops! Cannot display the analytics data at this time.';
+export const EMPTY_DATA_MSG = 'There are no page views to show for this range.';
+export const DEFAULT_CONTENT_MSG = 'Oops! Something went wrong with the slug field configuration.';
+
+export const getContentTypeSpecificMsg = (contentTypeName: string) => ({
+  noSlugConfigMsg: `The ${contentTypeName} content type has not been configured for use with this app. It must have a field of type short text and must be added to the list of content types in
+    this app's configuration.`,
+  noSlugContentMsg: `This ${contentTypeName} entry does not have a valid slug field.`,
+  notPublishedMsg: `This ${contentTypeName} entry has not yet been published.`,
+});
```
lilbitner, line 10: I feel like this file will be a good home for very specific error messages and any other specific copy that we display. Happy to rename or relocate.

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsMetricDisplays/AnalyticsMetricDisplay.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const AnalyticsMetricDisplay = (props: Props) => {
+  const { handleDateRangeChange, runReportResponse, reportSlug, error, metricName, pageViews } =
+    props;
+  return (
+    <>
+      <ChartHeader
+        metricName={metricName ? metricName : ''}
+        metricValue={pageViews || pageViews === 0 ? pageViews.toString() : ''}
```
rusticpenguin, line 23: would `pageViews !== undefined` work here?

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsMetricDisplays/AnalyticsMetricDisplay.tsx
```diff
@@ -0,0 +1,37 @@
...
+        handleChange={handleDateRangeChange}
+      />
+
+      <ChartContent error={error} pageViewData={runReportResponse} />
+
+      <ChartFooter
+        slugName={`Page path: ${reportSlug}` || ''}
```
rusticpenguin, line 30: If I'm understanding the sentiment of this correctly, did you mean to do: ```reportSlug ? `Page path: ${reportSlug}` || ''```?

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
```diff
@@ -63,8 +52,9 @@ const AnalyticsApp = (props: Props) => {
       setLoading(false);
     }
 
-    fetchRunReportData();
-  }, [api, reportRequestParams]);
+    if (reportSlug && propertyId && !isContentTypeWarning) fetchRunReportData();
```
lilbitner, line 55: not calling this endpoint if we do not have required slug or propertyId params, and if there is a content related warning.

apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.tsx
```diff
@@ -21,16 +24,24 @@ const ChartContent = (props: Props) => {
...
   };
 
-  return (
-    <>
+  const renderChartContent = () => {
+    if (error) {
+      return <Note body={error?.message || DEFAULT_ERR_MSG} variant="negative" />;
+    } else if (pageViewData.rows && !pageViewData.rows.length) {
+      return <Note body={EMPTY_DATA_MSG} variant="warning" />;
+    }
```
rusticpenguin, line 32: Since these two if statements aren't related to each other, I feel that separating them would help make this easier to skim:
```suggestion
    if (error) {
      return <Note body={error?.message || DEFAULT_ERR_MSG} variant="negative" />;
    }
    
    if (pageViewData.rows && !pageViewData.rows.length) {
      return <Note body={EMPTY_DATA_MSG} variant="warning" />;
    }
```

lilbitner: Great suggestion, will adjust 👍 !

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsMetricDisplays/AnalyticsMetricDisplay.tsx
lilbitner: ah yes this is actually leftover code from pervious PR's that I just moved to this file, so not a direct change but I will adjust to enforce the conditional logic. Great catch!

apps/google-analytics-4/frontend/src/components/main-app/SlugWarningDisplay/SlugWarningDisplay.tsx
```diff
@@ -0,0 +1,40 @@
...
+    switch (true) {
+      case !slugFieldIsConfigured:
+        return noSlugConfigMsg;
+      case !contentTypeHasSlugField:
+        return noSlugContentMsg;
+      case !isPublished:
+        return notPublishedMsg;
+      default:
+        return DEFAULT_CONTENT_MSG;
+    }
```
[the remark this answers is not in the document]
lilbitner: ah yeah I always go back and for with switch statements and a string of 'if' statements. I am certainly open to changing it here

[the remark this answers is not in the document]
rusticpenguin: I feel that, it's really up to the scenario. I personally like using switches in places where you can match on something like an enum

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsMetricDisplays/AnalyticsMetricDisplay.tsx
lilbitner: yeah this is a good callout, though I think the way this is written (previously written code I moved into this file) is super explicit about what `pageViews` (a number that can be 0), whereas `pageViews !== undefined` is not as clear to me as to why we are explicitly checking for undefined. Just a personal preference though!

apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.tsx
```diff
@@ -21,16 +24,26 @@ const ChartContent = (props: Props) => {
...
   };
 
-  return (
-    <>
+  const renderChartContent = () => {
+    if (error) {
+      return <Note body={error?.message || DEFAULT_ERR_MSG} variant="negative" />;
+    }
+
+    if (pageViewData.rows && !pageViewData.rows.length) {
```
whitelisab, line 32: I think that this needs to be based on `rowCount` instead of `rows`, perhaps? The old `renderChartContent` function in the `AnalyticsApp` used `rowCount` to determine whether to render the chart content. What I am seeing locally is that if run_report returns no data (rowCount is 0, but there are still some rows, all of which have 0 as the value), then I am still seeing the sidebar chart. I shouldn't be seeing the sidebar chart if there is no data, right?

apps/google-analytics-4/frontend/src/components/main-app/SlugWarningDisplay/SlugWarningDisplay.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+      slugFieldIsConfigured: false,
+      contentTypeHasSlugField: false,
+      isPublished: true,
+      ...useSidebarSlugMock,
+    }));
+
+    render(<SlugWarningDisplay slugFieldInfo={{ slugField: '', urlPrefix: '' }} />);
+
+    const warningMsg = await findByText(
+      "The Category content type has not been configured for use with this app. It must have a field of type short text and must be added to the list of content types in this app's configuration."
```
whitelisab, line 33: Maybe import the warning message from constants as you've done with the other tests?

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const useSidebarSlug = (slugFieldInfo: ContentTypeValue) => {
+  const sdk = useSDK<SidebarExtensionSDK>();
+
+  const { slugField, urlPrefix } = slugFieldInfo;
+  const slugFieldValue = useGetFieldValue(slugField);
+
+  const reportSlug = `${urlPrefix || ''}${slugFieldValue || ''}`;
+  const slugFieldIsConfigured = Boolean(slugField);
+  const contentTypeHasSlugField = slugField in sdk.entry.fields;
+  const isPublished = Boolean(sdk.entry.getSys().publishedAt);
```
lilbitner, line 15: currently working on implementing a way to listen for changes in the 'published' status of the entry. Might add to a follow-up PR if it is a significant change.

apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
whitelisab: I like it; I utilized the same pattern for my content type warnings 😄

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsMetricDisplays/AnalyticsMetricDisplay.tsx
rusticpenguin: I like the way you put that, I completely agree!

apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
```diff
@@ -0,0 +1,10 @@
+export const DEFAULT_ERR_MSG = 'Oops! Cannot display the analytics data at this time.';
+export const EMPTY_DATA_MSG = 'There are no page views to show for this range.';
+export const DEFAULT_CONTENT_MSG = 'Oops! Something went wrong with the slug field configuration.';
+
+export const getContentTypeSpecificMsg = (contentTypeName: string) => ({
+  noSlugConfigMsg: `The ${contentTypeName} content type has not been configured for use with this app. It must have a field of type short text and must be added to the list of content types in
```
jsdalton, line 6: I'd like to suggest a slight rewording here. Since we don't pass through specific details of _how_ the content type is misconfigured, I think we can be a bit more clear about heading to the app config screen to figure out what's wrong and how to fix it.

How about:

> The Google Analytics 4 sidebar item cannot be displayed on this entry because the '${contentTypeName}' content type has not been correctly configured. Head over to to the <link>app configuration screen</link> to fix the issue.

And then we can use `sdk.navigator.openAppConfig()` to on the link to send the user over there.

wdyt?

apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
```diff
@@ -0,0 +1,10 @@
...
+    this app's configuration.`,
+  noSlugContentMsg: `This ${contentTypeName} entry does not have a valid slug field.`,
```
whitelisab, line 8: Should this situation have some actionable next steps in the warning message? Like saying that a new slug field must be selected in the app's configuration?

jsdalton (commented):
Looking good! Just one small comment on the messaging of the error.

whitelisab (commented):
Looking great! Made some comments for you to review.

apps/google-analytics-4/frontend/src/components/main-app/SlugWarningDisplay/SlugWarningDisplay.spec.tsx
lilbitner: ah yes when I had originally tried this, the tests failed only because of the line break from formatting. But now that I have reworded the message I am able to import 👍 thanks for pointing out!

apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.tsx
lilbitner: great catch! This was a tricky one as my solution here was to originally solve a race condition that was occurring of rendering, but I believe the refactor of components ultimately solved this, so I am able to revert back to `rowCount`, which as point out more accurately reflects the data!

apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
lilbitner: Great point - I have adjusted the copy as per your suggestion and to better nudge the user towards action as well. To clarify your last point, are you envisioning a link somewhere below the message navigating to the app config page? I think something along those lines would be very useful for the user

lilbitner: Screenshot 2023-03-22 at 11 09 15 AM

Have update this here. Please let me know what you think @jsdalton !

jsdalton (approved):
Looks great!

PR #2858, opened 2023-03-21
adds optgroups for GA4 account + property Select options

ghepting:
## Purpose
image

## Approach
use `<optgroup>`s to organize GA4 properties by account

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
@@ -18,7 +18,7 @@ export default function MapAccountPropertySection(props: Props) {
   const { accountsSummaries, parameters, mergeSdkParameters, onIsValidAccountProperty } = props;
 
   const [selectedPropertyId, setSelectedPropertyId] = useState<string>('');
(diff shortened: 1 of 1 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
```diff
@@ -85,15 +79,19 @@ export default function MapAccountPropertySection(props: Props) {
...
-                  key={flattenedProperty.propertyId}
-                  value={flattenedProperty.propertyId}>
-                  {`${flattenedProperty.accountName} (${flattenedProperty.accountId}) > ${flattenedProperty.propertyName} (${flattenedProperty.propertyId})`}
-                </Select.Option>
-              );
-            })}
+            {sortedAccountSummaries.map((accountSummary: AccountSummariesType) => (
+              <optgroup
+                label={`${accountSummary.displayName} (${getIdOnly(accountSummary.account)})`}
+                key={getIdOnly(accountSummary.account)}>
```
ghepting, line 85: note the `<optgroup>` addition here is the biggest change to clean this UI up for our users☝️

jsdalton (commented):
I like it a lot, it's a big improvement for me. 💯 

I suggest for the dropdown default text we do:

> Select an Analytics Property option...

PR #2859, opened 2023-03-21
[INTEG-292] fix categories not respecting MC url

mgoudy91:
## Purpose
[Zendesk ticket for barkbox](https://contentful.atlassian.net/jira/software/c/projects/ZEND/boards/254?modal=detail&selectedIssue=ZEND-3359&quickFilter=1321)

CommerceTools uses "Merchant Center URL"s as region-specific storefronts. While we persist this information for an app installer, we were only using it when creating links for products, and not categories.

## Approach
I used the same approach products use and adapted it to work with categories as well.

### DIFF
--- a/apps/commercetools/src/api/categories.ts
+++ b/apps/commercetools/src/api/categories.ts
@@ -3,16 +3,20 @@ import { Pagination, Product } from '@contentful/ecommerce-app-base';
 import { ConfigurationParameters } from '../types';
 import { createClient } from './client';
 
-function categoryTransformer({ projectKey, locale }: ConfigurationParameters) {
+function categoryTransformer({ projectKey, locale, mcUrl }: ConfigurationParameters) {
   return (item: Category): Product => {
+    const merchantCenterBaseUrl =
+      mcUrl && mcUrl.length > 0 ? mcUrl : 'https://mc.europe-west1.gcp.commercetools.com';
     const id = item.id ?? '';
+    const externalLink =
+      (projectKey && id && `${merchantCenterBaseUrl}/${projectKey}/categories/${id}/general`) || '';
     return {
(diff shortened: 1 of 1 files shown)


apps/commercetools/src/api/categories.ts
```diff
@@ -3,16 +3,20 @@ import { Pagination, Product } from '@contentful/ecommerce-app-base';
 import { ConfigurationParameters } from '../types';
 import { createClient } from './client';
 
-function categoryTransformer({ projectKey, locale }: ConfigurationParameters) {
+function categoryTransformer({ projectKey, locale, mcUrl }: ConfigurationParameters) {
   return (item: Category): Product => {
+    const merchantCenterBaseUrl =
+      mcUrl && mcUrl.length > 0 ? mcUrl : 'https://mc.europe-west1.gcp.commercetools.com';
```
gordo-contetnful, line 9: is there a chance for `mcUrl` to come through undefined? might be a little more readable with:

```typescript
const merchantCenterBaseUrl =
      mcUrl ?? 'https://mc.europe-west1.gcp.commercetools.com'; 
```

gordo-contetnful (approved):
Just a small Q but nothing blocking, feel free to disregard LGTM

apps/commercetools/src/api/categories.ts
mgoudy91: yeah, sure. I just took exactly what was there in `product.ts` since I still don't know this app well. I think it can actually be both undefined and an empty string actually.

mgoudy91: Yeah just confirmed that, so we'll need to keep it like this. Good thinking though

PR #2860, opened 2023-03-21
feat: handle prefixes and slug joining gracefully [INTEG-256]

jsdalton:
## Purpose

We had a follow up task from a previous PR to make sure our joining of prefixes to URL slugs was robust (and not just naively joining).

## Approach

* Join paths irrespective of trailing or forward `/`
* Use the same output for query as for display (assumption: this works?). display matches the output of how GA4 displays page paths.

image

### DIFF
--- a/apps/google-analytics-4/frontend/src/utils/pathJoin.ts
+++ b/apps/google-analytics-4/frontend/src/utils/pathJoin.ts
@@ -1,0 +1,6 @@
+export const pathJoin = (...parts: (string | undefined)[]): string => {
+  return parts
+    .map((part) => (part || '').trim().replace(/(^[/]*|[/]*$)/g, ''))
+    .filter((part) => part.length)
+    .join('/');
+};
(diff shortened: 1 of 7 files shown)


ryunsong-contentful (approved):

rusticpenguin (approved):
LGTM!

apps/google-analytics-4/frontend/src/utils/pathJoin.ts
```diff
@@ -0,0 +1,6 @@
+export const pathJoin = (...parts: (string | undefined)[]): string => {
+  return parts
+    .map((part) => (part || '').trim().replace(/(^[/]*|[/]*$)/g, ''))
+    .filter((part) => part.length)
+    .join('/');
+};
```
lilbitner, line 6: nice! clean and simple 💯

lilbitner (approved):

PR #2862, opened 2023-03-21
[INTEG-218] Sync saved content types to sidebar location and handle errors

whitelisab:
## Purpose
This PR addresses two items for the content type configuration:
1.  Automatically adds/removes the app from content type sidebar locations based on the content types saved in the app configuration. 
2. Adds warning and error messages for situations where the content types saved in installation parameters get out of sync with Contentful data.

## Approach
On configure, the Editor Interface for the app state is updated so that the app is added/removed to the sidebar location for each of the configured content types. It is automatically added to the second position in the sidebar. If a user moved the app to another position, it will retain that new location.

Error states: error icon is displayed, content type selection is invalid, a tooltip provides a useful message, and an error notification appears on save and prevents the configuration from being saved.
- The user is attempting to save an empty content row
- A previously configured content type has been deleted

Example error state
Screenshot 2023-03-21 at 1 26 41 PM

Warning states: warning icon is displayed and a tooltip provides a useful message.
- User has selected a content type without selecting a slug field
- The app has been manually removed from the sidebar for a configured content type
- A previously configured slug field for a content type has been deleted

Example warning state
Screenshot 2023-03-21 at 1 25 49 PM


apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeRow.tsx
```diff
@@ -24,44 +29,98 @@ const AssignContentTypeRow = (props: Props) => {
...
+      setContentTypeOptions(contentTypeOptions);
+
+      if (isSaved) {
+        setIsContentTypeInOptions(
+          contentTypeOptions.find((option) => option[0] === contentTypeId) ? true : false
+        );
+        if (slugField !== undefined) {
+          setIsSlugFieldInOptions(
+            allContentTypes[contentTypeId]?.fields.find((field) => field.id === slugField)
+              ? true
```
jsdalton, line 76: This and the following line aren't really needed. 😉

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeRow.tsx
```diff
@@ -24,44 +29,98 @@ const AssignContentTypeRow = (props: Props) => {
...
+              : false
+          );
+        }
+      }
+    };
+
+    getContentTypeOptions();
```
jsdalton, line 83: This isn't an async function so we don't need to do wrap the logic in a separate function and call it ... we can just stick it at the main useEffect scope.

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/ContentTypeWarning.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const [warningType, setWarningType] = useState<ContentWarningTypes>(WarningTypes.Empty);
+  const [tooltipContent, setTooltipContent] = useState<string>('');
+
+  useEffect(() => {
+    const getTooltipContent = () => {
+      let content = '';
+
+      // Error states
+      if (!contentTypeId) {
+        setWarningType(WarningTypes.Error);
```
jsdalton, line 43: We might consider moving this and the one below to the bottom, so that error states take precedence on the warning type. (Imagine you have an error and a warning, i think we want the error icon to show.)

jsdalton (commented):
I just had a few small questions and comments, otherwise looking good ✨

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeRow.tsx
```diff
@@ -24,44 +29,98 @@ const AssignContentTypeRow = (props: Props) => {
...
-        </Box>
-      );
+  const [isSaved, setIsSaved] = useState<boolean>(false);
+  const [contentTypeOptions, setContentTypeOptions] = useState<AllContentTypeEntries>([]);
+  const [isContentTypeInOptions, setIsContentTypeInOptions] = useState<boolean>(true);
+  const [isSlugFieldInOptions, setIsSlugFieldInOptions] = useState<boolean>(true);
+  const [isInSidebar, setIsInSidebar] = useState<boolean>(false);
+
+  useEffect(() => {
+    const originalContentTypeIds = Object.keys(originalParameters.contentTypes);
```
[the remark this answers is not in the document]
whitelisab: Yes, this is a good callout! I am making some adjustments to ensure that everything still works when you don't have any content types saved in parameters.

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/ContentTypeWarning.tsx
whitelisab: Agreed, that makes sense!

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeRow.tsx
whitelisab: Changed `.find()` to `.some()` to remove the ternary 🙂

jsdalton (approved):
Looking awesome! 💯

PR #2888, opened 2023-03-22
Integ 269 graph jitter bug and error flashing before load

lilbitner:
## Purpose
This PR fixes two issues: 
1. When the analytics chart is mounting, there is a weird stuttering that occurs, that is very uncomfortable to look at. 😅 
2. An incorrect date was being provided at the end of the runReports data. (Feb 21st instead of March 21st for example). 

## Approach
1. To solve the first issue, despite diving head first into some animation adjusting, this bug actually just ended up being an issue of the chart not properly having a height and width defined, so on resize it would flicker until it properly inherited parent height and width. Therefore I added a height and width definition to its parent. This also stabilized the container itself, so rendering the warning note instead of the chart does not cause a lot of shifting around. 
2. The incorrect date was being added because the date mapping occurring on the API side was incorrectly using 'day' as a way to compare if the data already exists. So Feb 21st would be a match to March 21st. I changed this conditional to use `valueOf()` to appropriately determine the equality of dates. 

## How to test
Pull down this code and render the GA4 app in the sidebar of an entry. Ideally, notice on render and when changing the date range, there is no flickering of the chart! Additionally, the dates provided should also be an accurate representation of the date range selected. 

Video showing no jitter bug: https://www.loom.com/share/679f315fd9604988a03b165b57007e7c

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.styles.ts
+++ b/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.styles.ts
@@ -33,7 +33,7 @@ export const styles = {
     },
   }),
   wrapper: css({
-    height: '150px',
+    height: '160px',
     width: '100px',
     margin: 'auto',
   }),
--- a/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.styles.ts
+++ b/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.styles.ts
@@ -33,7 +33,7 @@ export const styles = {
     },
   }),
   wrapper: css({
-    height: '100px',
+    height: '150px',
     width: '100px',
     margin: 'auto',
   }),
--- a/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.styles.ts
+++ b/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.styles.ts
@@ -33,6 +33,7 @@ export const styles = {
     },
   }),
   wrapper: css({
+    // height of this loading wrapper matches height of ChartContent
     height: '160px',
     width: '100px',
     margin: 'auto',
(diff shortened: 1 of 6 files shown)


ryunsong-contentful (approved):

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.styles.ts
```diff
@@ -33,7 +33,7 @@ export const styles = {
     },
   }),
   wrapper: css({
-    height: '100px',
+    height: '160px',
```
ghepting, line 36: where is this magic number coming from?

apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.styles.ts
```diff
@@ -0,0 +1,11 @@
+import { css } from 'emotion';
+
+export const styles = {
+  root: css({
+    width: '100%',
+    height: '10rem',
```
ghepting, line 6: where is this magic number coming from?

lilbitner: sure, this number is derived from the height of the current chart component in the app given the space. Adding this constraint here also made the container height consistent so when the warning message appears instead of the chart, it is centered and does not shift around the contents. Everything maintains the 10rem. Happy to add a note within the code here?

apps/google-analytics-4/frontend/src/components/main-app/LineChart/LineChart.styles.ts
```diff
@@ -0,0 +1,7 @@
+import { css } from 'emotion';
+
+export const styles = {
+  root: css({
+    width: '100%',
```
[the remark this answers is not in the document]
lilbitner: yeah good point, this was just an effort on my end to adjust the alignment of the graph to be exactly left aligned on render: 

origin: 
Screenshot 2023-03-23 at 8 06 58 AM

vs with the container width of 100%: 
Screenshot 2023-03-23 at 8 07 19 AM

but perhaps there is a better way to do that?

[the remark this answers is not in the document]
lilbitner: Good point, the main reason I added this was to make the chart itself left aligned, but perhaps it looks better not left aligned now that I am staring at it again....

original without this container: 
Screenshot 2023-03-23 at 8 06 58 AM

with the container: 
Screenshot 2023-03-23 at 8 07 19 AM

PR #2889, opened 2023-03-22
[INTEG-261] Service Account Checklist

ryunsong-contentful:
## Purpose
[JIRA Ticket](https://contentful.atlassian.net/browse/INTEG-261)

Adds a UI checklist to the service account with appropriate error state handling and messaging

Screenshot 2023-03-22 at 1 27 52 PM
Screenshot 2023-03-22 at 1 28 15 PM
Screenshot 2023-03-22 at 1 30 18 PM

## Approach
Iterative discussions about the UI/UX

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
@@ -25,8 +25,13 @@ export default function GoogleAnalyticsConfigPage() {
     {} as Partial<EditorInterface>
   );
   const [originalParameters, setOriginalParameters] = useState<KeyValueMap>({});
+  const [hasServiceCheckErrors, setHasServiceCheckErrors] = useState<boolean>(true);
 
   const sdk = useSDK<AppExtensionSDK>();
+
+  const handleHasServiceCheckErrorsChange = (hasErrors: boolean) => {
+    setHasServiceCheckErrors(hasErrors);
+  };
 
   /** SDK Parameters handlers/effects **/
   const mergeSdkParameters = (_parameters: KeyValueMap) => {
@@ -169,8 +174,9 @@ export default function GoogleAnalyticsConfigPage() {
           isInEditMode={isInEditMode}
           onInEditModeChange={handleInEditModeChange}
           onIsValidServiceAccount={handleIsValidServiceAccount}
+          onHasServiceCheckErrorsChange={handleHasServiceCheckErrorsChange}
         />
-        {isAppInstalled && !isInEditMode && (
+        {isAppInstalled && !isInEditMode && !hasServiceCheckErrors && (
           <>
             <Splitter />
             <MapAccountPropertySection
--- a/apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
@@ -176,7 +176,7 @@ export default function GoogleAnalyticsConfigPage() {
           onIsValidServiceAccount={handleIsValidServiceAccount}
           onHasServiceCheckErrorsChange={handleHasServiceCheckErrorsChange}
         />
-        {isAppInstalled && !isInEditMode && !hasServiceCheckErrors && (
+        {isAppInstalled && (!hasServiceCheckErrors || parameters.propertyId) && (
           <>
             <Splitter />
             <MapAccountPropertySection
--- a/apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ChecklistSplitter.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ChecklistSplitter.tsx
@@ -1,5 +1,0 @@
-import tokens from '@contentful/f36-tokens';
-
-export default function ChecklistSplitter() {
-  return <div style={{ borderBottom: `1px solid ${tokens.gray300}` }} />;
-}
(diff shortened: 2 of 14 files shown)


whitelisab (commented):
Looks awesome! 🚀 Just added a few small comments

jsdalton (changes requested):
For now just some high level UX comments. Will do a proper review later:

* We definitely need the call to action "button" or "link" in this PR (since without this it's a step backwards from what we have today, which at least provides details about the error)
* I'd probably have some concrete suggestions around the messaging wording but can save those for later.
* Ditto for design. Probably just needs a little polish. Overall the "temperature" feels right.
* I think we'll likely "need" an info button for details about the error. We can put a somewhat generic "Success!" / "The Admin API is not enabled" / "There was an error checking the Admin API status" but think it'll be too much to provide details here. Nevertheless some details will be important.
* For me there's a bug on installation. When I do a fresh install with one API disabled it shows up as "Active". But when I press "test" it (correctly) reveals the fail.

Mostly just notes to discuss directly later.

apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
```diff
@@ -169,8 +174,9 @@ export default function GoogleAnalyticsConfigPage() {
           isInEditMode={isInEditMode}
           onInEditModeChange={handleInEditModeChange}
           onIsValidServiceAccount={handleIsValidServiceAccount}
+          onHasServiceCheckErrorsChange={handleHasServiceCheckErrorsChange}
         />
-        {isAppInstalled && !isInEditMode && (
+        {isAppInstalled && !isInEditMode && !hasServiceCheckErrors && (
```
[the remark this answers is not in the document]
ryunsong-contentful: From what I remember from our conversations the decision was to not let the user even proceed past the service account key step until everything has passed.

However, there is an edge case we kicked down the road for when everything was installed correctly with the properties and content types being setup. But then later the user updates the service key incorrectly for example. What should we do then since we'll have existing properties and content types? Right now it won't show them, but there's a good argument to do show them as well along witht eh errors.

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ServiceAccountChecklist.tsx
```diff
@@ -0,0 +1,227 @@
...
+        return '';
+    }
+  };
+
+  const renderServiceAccountCheck = () => {
+    return invalidServiceAccountError ? (
+      <GenericCheckRow
+        icon={<ErrorCircleIcon marginLeft="spacingXs" marginRight="spacingXs" variant="negative" />}
+        title={'Service Account'}
+        description={'Invalid service account and service account key'}
```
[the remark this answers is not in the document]
ryunsong-contentful: I like it

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ChecklistSplitter.tsx
```diff
@@ -0,0 +1,5 @@
+import tokens from '@contentful/f36-tokens';
+
+export default function ChecklistSplitter() {
+  return <div style={{ borderBottom: `1px solid ${tokens.gray300}` }} />;
```
lilbitner, line 4: nit: seems like we are following a pattern of having a separate styles file instead of inline styling. But not imperative to adjust.

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ServiceAccountChecklist.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+          parameters={{ propertyId: '123' }}
+        />
+      );
+    });
+
+    await screen.findByText('Service Account');
+    await screen.findByText('Admin api');
+    await screen.findByText('Data api');
+    await screen.findByText('GA4 Account Properties');
+    await screen.findAllByText(/Success!/);
```
lilbitner, line 46: Do you need all of the `awaits` here, or just one? Additionally I am noticing in this file there are not assertions being made like `expect().toBe()`, is there a reason for that?

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/DisplayServiceAccountCard.tsx
```diff
@@ -192,64 +208,94 @@ const DisplayServiceAccountCard = (props: Props) => {
...
+              <Box paddingRight="spacingXs" paddingTop="spacingXs">
+                <TextLink
+                  testId="editServiceAccountButton"
+                  as="button"
+                  variant="primary"
+                  onClick={() => onInEditModeChange(true)}>
+                  Edit
+                </TextLink>
+              </Box>
+              <Box style={{ minWidth: '60px', minHeight: '30px' }}>
```
lilbitner, line 231: nit: more inline styling here, just want to make sure we are following a pattern one way or another.

lilbitner (commented):
just left a few non-blocking comments :) Great work, looks really awesome 💯 !

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ChecklistSplitter.tsx
ryunsong-contentful: done

apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ServiceAccountChecklist.spec.tsx
ryunsong-contentful: Removed the awaits. I didn't do any `expect` calls because I don't know what else to expect other than these checklist statuses being rendered which is already checked by calling `findBy`

lilbitner: ah okay I am following your logic, do the tests ever fail when you do not have assertions? or does it just timeout? I am just curious now 😄

ryunsong-contentful: Nope, everything seems to be in a happy path

PR #2896, opened 2023-03-23
docs: Update README.md

b-wbvreeuwijk:
The Bynder entry was pointing to the Brandfolder directory.

## Purpose
Fix in the documentation

## Approach
Just replace the link

## Dependencies and/or References
None

PR #2897, opened 2023-03-23
fix: remove unneeded secrets in app-test build step

jsdalton:
## Purpose

We're unnecessarily loading vault secrets during apps-testing, which are not needed there. As a result we can't take outside PRs to this repository (like this one #2896).

## Approach

* Remove it


andipaetzold (approved):
Please merge using `chore:` and not `fix:` to prevent releases (I am scared of lerna)

PR #2919, opened 2023-03-24
Integ 231 improve error handling within analytics app component

lilbitner:
## Purpose
This PR accomplishes the following: 

1. Adds more robust **error handling** for the common errors when rendering the GA4 app in the Sidebar. ⛔ 
2. Fixes some small **UI issues**. ✨ 

## Error handling:
I added a component here called the `ErrorDisplay` component that is rendered in place of the Chart itself when an error is thrown. This component handles the anticipated two most common errors, providing an invalid argument somehow which is unlikely (invalid `propertyId`.), or a disabled Data api error. If the error does not belong to one of these error types, then the provided error message is displayed. I added a comment on this code to perhaps adjust the logic to display an actionable default error message if the error is not properly identified by the frontend, instead of the "mystery" error message. 

### Example error message for invalid arguments:
Screenshot 2023-03-24 at 4 35 48 PM

## Small UI fixes:

Note content not wrapping appropriately: 

### Before: 
Screenshot 2023-03-24 at 9 03 30 AM

### After: 
Screenshot 2023-03-24 at 9 03 15 AM

Outline of Date range Select menu cut-off on right side:

### Before: 
Screenshot 2023-03-27 at 8 27 54 AM

### After: 
Screenshot 2023-03-27 at 8 27 35 AM

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/common/HyperLink/HyperLink.tsx
+++ b/apps/google-analytics-4/frontend/src/components/common/HyperLink/HyperLink.tsx
@@ -1,0 +1,32 @@
+import React, { MouseEventHandler } from 'react';
+import { TextLink } from '@contentful/f36-components';
+
+interface Props {
+  body: string;
+  substring: string;
+  onClick?: MouseEventHandler<HTMLAnchorElement>;
+  hyperLinkHref?: string;
+}
+
+const HyperLink = (props: Props) => {
+  const { body, substring, onClick = () => {}, hyperLinkHref } = props;
+  const link = (
+    <TextLink onClick={onClick} href={hyperLinkHref} target="_blank" rel="noopener noreferer">
+      {substring}
+    </TextLink>
+  );
+
+  const formatLink = () => {
+    const bodyWithTextLink = body.split(substring).reduce((prev: any, current, i) => {
+      if (!i) {
+        return [current];
+      }
+      return prev.concat(link, current);
+    }, []);
+    return bodyWithTextLink;
+  };
+
+  return formatLink();
+};
+
+export default HyperLink;
--- a/apps/google-analytics-4/frontend/src/components/common/HyperLink/HyperLink.tsx
+++ b/apps/google-analytics-4/frontend/src/components/common/HyperLink/HyperLink.tsx
@@ -23,7 +23,7 @@ const HyperLink = (props: Props) => {
       }
       return prev.concat(link, current);
     }, []);
-    return bodyWithTextLink;
+    return bodyWithTextLink as JSX.Element;
   };
 
   return formatLink();
--- a/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
@@ -45,6 +45,7 @@ const AnalyticsApp = (props: Props) => {
       try {
         const reportData = await api.runReports(reportRequestParams);
         setRunReportResponse(reportData);
+        setError(undefined);
       } catch (e) {
         setError(e as Error);
       }
--- a/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
@@ -46,6 +46,7 @@ const AnalyticsApp = (props: Props) => {
         const reportData = await api.runReports(reportRequestParams);
         setRunReportResponse(reportData);
       } catch (e) {
+        console.error(e);
         setError(e as Error);
       }
 
--- a/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
@@ -46,7 +46,6 @@ const AnalyticsApp = (props: Props) => {
         const reportData = await api.runReports(reportRequestParams);
         setRunReportResponse(reportData);
       } catch (e) {
-        console.error(e);
         setError(e as Error);
       }
 
--- a/apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.tsx
@@ -2,7 +2,8 @@ import { Row, RunReportResponse } from 'types';
 import Note from 'components/common/Note/Note';
 import LineChart from 'components/main-app/LineChart/LineChart';
 import { parseDayAndMonth } from 'helpers/DateHelpers/DateHelpers';
-import { DEFAULT_ERR_MSG, EMPTY_DATA_MSG } from '../constants/noteMessages';
+import ErrorDisplay from 'components/main-app/ErrorDisplay/ErrorDisplay';
+import { EMPTY_DATA_MSG } from '../constants/noteMessages';
 import { styles } from './ChartContent.styles';
 
 interface Props {
@@ -27,7 +28,7 @@ const ChartContent = (props: Props) => {
 
   const renderChartContent = () => {
     if (error) {
-      return <Note body={error?.message || DEFAULT_ERR_MSG} variant="negative" />;
+      return <ErrorDisplay error={error} />;
     }
 
     if (!pageViewData.rowCount) {
--- a/apps/google-analytics-4/frontend/src/components/main-app/ErrorDisplay/ErrorDisplay.tsx
+++ b/apps/google-analytics-4/frontend/src/components/main-app/ErrorDisplay/ErrorDisplay.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import { useSDK } from '@contentful/react-apps-toolkit';
 import { SidebarExtensionSDK } from '@contentful/app-sdk';
 import Note from 'components/common/Note/Note';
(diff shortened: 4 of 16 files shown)


apps/google-analytics-4/frontend/src/components/main-app/ErrorDisplay/ErrorDisplay.tsx
```diff
@@ -0,0 +1,52 @@
...
+          break;
+        default:
+          setErrorDisplay(e.message);
+      }
+    };
+
+    if (isApiErrorType(error)) handleApiError(error);
+    else {
+      setErrorDisplay(error.message || DEFAULT_ERR_MSG);
+    }
```
lilbitner, line 46: If error is not of `ApiErrorType` then display the provided error message or a default error message.

apps/google-analytics-4/frontend/src/components/common/HyperLink/HyperLink.tsx
```diff
@@ -0,0 +1,32 @@
...
+      if (!i) {
+        return [current];
+      }
+      return prev.concat(link, current);
+    }, []);
+    return bodyWithTextLink as JSX.Element;
+  };
+
+  return formatLink();
+};
```
lilbitner, line 30: Created this component, which can hopefully be used elsewhere, to easily "embed" a hyperlink (`TextLink` component) into a string dynamically. Ex: 
Screenshot 2023-03-24 at 4 35 48 PM

apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
```diff
@@ -1,13 +1,13 @@
-export const DEFAULT_ERR_MSG = 'Oops! Cannot display the analytics data at this time.';
-export const EMPTY_DATA_MSG = 'There are no page views to show for this range.';
-export const DEFAULT_CONTENT_MSG = 'Oops! Something went wrong with the slug field configuration.';
+export const DEFAULT_ERR_MSG = 'An unknown error has occurred. Please try again or contact support';
```
lilbitner, line 1: is there an optimal link to provide the user tp contact support? Thoughts on adding a hyperlink to https://www.contentful.com/support/?utm_source=webapp&utm_medium=help-menu&utm_campaign=in-app-help ?

apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.tsx
```diff
@@ -27,7 +28,7 @@ const ChartContent = (props: Props) => {
 
   const renderChartContent = () => {
     if (error) {
-      return <Note body={error?.message || DEFAULT_ERR_MSG} variant="negative" />;
+      return <ErrorDisplay error={error} />;
```
whitelisab, line 31: I don't have a strong opinion on this, but wanted to ask the question of whether it makes sense to have the chart header and footer still showing when there are errors, or if we want to handle it more like the content type warnings , where errors are displayed at the `AnalyticsApp` level rather than the `ChartContent` level?

whitelisab (commented):
Looks great to me! Just added a couple small comments/questions!

apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.tsx
lilbitner: Yes great question! I have been going back and forth with this, originally just rendered the error message in place of all analytic components like you said, but then it made me a bit nervous to not show the data range select menu in the case that an error is thrown on selection of one date range and not the others - felt in this case the components would shift around a lot of the error message replaced everything, and the user would lose some context that they could piece together. But I am fine with either way definitely 👍

apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
```diff
@@ -1,13 +1,13 @@
...
+export const EMPTY_DATA_MSG =
```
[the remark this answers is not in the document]
lilbitner: yes great catch 🦅 !

apps/google-analytics-4/frontend/src/components/common/HyperLink/HyperLink.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { MouseEventHandler } from 'react';
+import { TextLink } from '@contentful/f36-components';
+
+interface Props {
+  body: string;
+  substring: string;
+  onClick?: MouseEventHandler<HTMLAnchorElement>;
+  hyperLinkHref?: string;
```
ryunsong-contentful, line 8: nit: What is the benefit of making the link optional?

apps/google-analytics-4/frontend/src/components/main-app/SlugWarningDisplay/SlugWarningDisplay.tsx
```diff
@@ -18,48 +17,40 @@ const SlugWarningDisplay = (props: Props) => {
...
-    const content = { bodyMsg: DEFAULT_CONTENT_MSG, hyperLink: <></> };
+    const content: { bodyMsg: string | JSX.Element } = { bodyMsg: supportHyperLink };
     if (!slugFieldIsConfigured) {
-      content.bodyMsg = noSlugConfigMsg;
-      content.hyperLink = linkToOpenConfigPage;
+      content.bodyMsg = appConfigPageHyperLink(noSlugConfigMsg);
     } else if (!contentTypeHasSlugField) {
-      content.bodyMsg = noSlugContentMsg;
-      content.hyperLink = linkToOpenConfigPage;
+      content.bodyMsg = appConfigPageHyperLink(noSlugContentMsg);
```
ryunsong-contentful, line 43: nit: I think you can combine these two if checks since they do the same thing like
```
   if (!slugFieldIsConfigured || !contentTypeHasSlugField) {
      content.bodyMsg = appConfigPageHyperLink(noSlugConfigMsg);
 }
```

apps/google-analytics-4/frontend/src/components/main-app/ErrorDisplay/ErrorDisplay.tsx
```diff
@@ -0,0 +1,67 @@
...
+} from '../constants/noteMessages';
+import HyperLink from 'components/common/HyperLink/HyperLink';
+
+interface Props {
+  error: Error;
+}
+
+const ErrorDisplay = (props: Props) => {
+  const { error } = props;
+  const [errorDisplay, setErrorDisplay] = useState<string | JSX.Element>('');
```
ryunsong-contentful, line 19: I think storing components in a state to be an anti-pattern and potentially dangerous down the road. Right now it will work but I think this is an important React principle we should follow. Generally I think this errorDisplay (potential rename involved here as well) should be some defined object via `type/interface` and then you render your JSX logic down below in your return based on that errorDisplay object and it's data fields.

apps/google-analytics-4/frontend/src/components/main-app/ErrorDisplay/ErrorDisplay.tsx
```diff
@@ -0,0 +1,52 @@
...
+            />
+          );
+          break;
+        case ERROR_TYPE_MAP.disabledDataApi:
+          setErrorDisplay(PERMISSION_DENIED_MSG);
+          break;
+        default:
+          setErrorDisplay(e.message);
+      }
+    };
```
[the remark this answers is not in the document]
ryunsong-contentful: Yup left a comment in the review down below since this was "outdated" by the time I got to it

apps/google-analytics-4/frontend/src/components/main-app/ErrorDisplay/ErrorDisplay.tsx
ryunsong-contentful: Screenshot 2023-03-27 at 4 22 12 PM

ryunsong-contentful (changes requested):
Everything looked good as far as the stuff you mentioned UI/Error Handling and overall great work on much needed fixes!

Two things however
1. Storing the React component in state.
2. I found two unhandled error cases (which produces the screenshots below)
a. One where you kill the lambda (ie no server/server error/500s(
b. Invalid service key 

Screenshot 2023-03-27 at 4 25 50 PM
Screenshot 2023-03-27 at 4 26 53 PM


Now this PR only says to improve the error handling. So if this is just a foundation PR for handling Errors, I'm cool with you creating a ticket and handling those fixes separately as your next task (I'm also fine with you fixing them in this PR).
The second point will be a relatively annoying refactor for you but I think it'll prevent a lot of headaches down the road and is worth fixing in this PR.

apps/google-analytics-4/frontend/src/components/main-app/ChartContent/ChartContent.tsx
whitelisab: That makes sense to me to keep it as you have it in this PR, so that users have the additional context in the header and footer 👍

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
```diff
@@ -46,6 +46,7 @@ const AnalyticsApp = (props: Props) => {
         const reportData = await api.runReports(reportRequestParams);
         setRunReportResponse(reportData);
       } catch (e) {
+        console.error(e);
```
ryunsong-contentful, line 49: nit: Double checking, was this intended?

apps/google-analytics-4/frontend/src/components/common/HyperLink/HyperLink.tsx
lilbitner: my intention here is that sometimes we will want a component where an href link is provided and sometimes we have an onClick event to pass. For example, the hyperlink currently used to open the app config page passes an onClick JS function to open the config page. In other cases, a link such as Google Analytics will be opened. Thoughts?

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
lilbitner: good eye, was using to test. Will remove :)

ryunsong-contentful (approved):

PR #2944, opened 2023-03-27
Fix/integ 192

rusticpenguin:
## Purpose
We have hit Shopify's 250 limit for some customers using the shopify app.

## Approach
For where we can paginate using the API, we are using the shopify-buy pagination queries. When we don't have access to pagination, we loop through through the requested ID's and batch request 250 at a time.

This PR also makes the App's config insensitive to if the user has a url with an `https:` protocol or a trailing `/`
INTEG-192

### DIFF
--- a/apps/shopify/src/constants.js
+++ b/apps/shopify/src/constants.js
@@ -2,7 +2,7 @@
 // returned with this unfortunate title, and there is no other way to check
 // whether the returned variant is the default one or not
 export const DEFAULT_SHOPIFY_VARIANT_TITLE = 'Default Title';
-export const SHOPIFY_API_VERSION = '2022-04';
+export const SHOPIFY_API_VERSION = '2023-01';
 
 export const SKU_TYPES = [
   {
--- a/apps/shopify/src/constants.js
+++ b/apps/shopify/src/constants.js
@@ -3,6 +3,8 @@
 // whether the returned variant is the default one or not
 export const DEFAULT_SHOPIFY_VARIANT_TITLE = 'Default Title';
 export const SHOPIFY_API_VERSION = '2023-01';
+export const SHOPIFY_ENTITY_LIMIT = 250;
+export const SHOPIFY_ENTITY_LIMIT_AS_INDEX = SHOPIFY_ENTITY_LIMIT - 1;
 
 export const SKU_TYPES = [
   {
--- a/apps/shopify/src/constants.js
+++ b/apps/shopify/src/constants.js
@@ -4,7 +4,6 @@
 export const DEFAULT_SHOPIFY_VARIANT_TITLE = 'Default Title';
 export const SHOPIFY_API_VERSION = '2023-01';
 export const SHOPIFY_ENTITY_LIMIT = 250;
-export const SHOPIFY_ENTITY_LIMIT_AS_INDEX = SHOPIFY_ENTITY_LIMIT - 1;
 
 export const SKU_TYPES = [
   {
--- a/apps/shopify/src/dataTransformer.js
+++ b/apps/shopify/src/dataTransformer.js
@@ -1,8 +1,13 @@
 import get from 'lodash/get';
-import last from 'lodash/last';
 import flatten from 'lodash/flatten';
 import { DEFAULT_SHOPIFY_VARIANT_TITLE } from './constants';
 import { convertBase64ToString, convertStringToBase64 } from './utils/base64';
+
+/**
+ * Removes the protocol and trailing slash from a URL
+ * This is a QOL for users who copy-paste the URL from the browser
+ */
+const removeHttpsAndTrailingSlash = (url) => url.replace(/^https?:\/\//, '').replace(/\/$/, '');
 
 /**
  * Transforms the API response of Shopify collections into
@@ -21,9 +26,9 @@ export const collectionDataTransformer = (collection, apiEndpoint) => {
         collectionIdDecoded && collectionIdDecoded.slice(collectionIdDecoded.lastIndexOf('/') + 1);
 
       if (apiEndpoint && collectionId) {
-        externalLink = `https://${apiEndpoint}${
-          last(apiEndpoint) === '/' ? '' : '/'
-        }admin/collections/${collectionId}`;
+        externalLink = `https://${removeHttpsAndTrailingSlash(
+          apiEndpoint
+        )}/admin/collections/${collectionId}`;
       }
     } catch {}
   }
@@ -54,9 +59,9 @@ export const productDataTransformer = (product, apiEndpoint) => {
         productIdDecoded && productIdDecoded.slice(productIdDecoded.lastIndexOf('/') + 1);
 
       if (apiEndpoint && productId) {
-        externalLink = `https://${apiEndpoint}${
-          last(apiEndpoint) === '/' ? '' : '/'
-        }admin/products/${productId}`;
+        externalLink = `https://${removeHttpsAndTrailingSlash(
+          apiEndpoint
+        )}/admin/products/${productId}`;
       }
     } catch {}
   }
@@ -125,9 +130,9 @@ export const previewsToProductVariants =
       name: title === DEFAULT_SHOPIFY_VARIANT_TITLE ? product.title : `${product.title} (${title})`,
       ...(apiEndpoint &&
         productId && {
-          externalLink: `https://${apiEndpoint}${
-            last(apiEndpoint) === '/' ? '' : '/'
-          }admin/products/${productId}`,
+          externalLink: `https://${removeHttpsAndTrailingSlash(
+            apiEndpoint
+          )}/admin/products/${productId}`,
         }),
     };
   };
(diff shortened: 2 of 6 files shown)


apps/shopify/src/dataTransformer.js
```diff
@@ -1,9 +1,29 @@
...
+ * Decodes the ID of a Shopify resource
+ * The ID is encoded in base64 and the actual ID is the last part of the string
+ * e.g. gid://shopify/Product/1234567890 -> 1234567890
+ */
+export const decodeId = (sku) => {
+  try {
+    const decodedId = convertBase64ToString(sku);
+    return decodedId && decodeId.slice(decodedId.lastIndexOf('/') + 1);
+  } catch {
+    return null;
```
pkeavenycontentful, line 16: Try/catch is a good practice, but just returning the `null` in the catch is a bit of a code smell to me. Is `convertBase64ToString()` a BE call that could fail? If not, why not return line 14 without the try/catch?

apps/shopify/src/dataTransformer.js
```diff
@@ -1,9 +1,29 @@
...
+  }
+};
+
+/**
+ * Removes the protocol and trailing slash from a URL
+ * This is a QOL for users who copy-paste the URL from the browser
+ */
+export const removeHttpsAndTrailingSlash = (url) =>
+  url.replace(/^https?:\/\/(www\.)?/, '').replace(/\/$/, '');
```
pkeavenycontentful, line 25: Nice 👍

apps/shopify/src/skuResolvers.js
```diff
@@ -26,37 +29,41 @@ export async function makeShopifyClient(config) {
...
   }
 
   const validIds = filterAndDecodeValidIds(skus, 'Collection');
-
   const shopifyClient = await makeShopifyClient(config);
 
-  const response = await shopifyClient.collection.fetchAll(250);
+  const response = await shopifyClient.collection.fetchAll(SHOPIFY_ENTITY_LIMIT);
+  if (response.length > 0) {
+    while (response[response.length - 1].hasNextPage) {
```
pkeavenycontentful, line 51: Interesting. Is `.hasNextPage` a field that Shopify gives us? Pretty cool if so.

apps/shopify/src/skuResolvers.js
```diff
@@ -26,37 +29,41 @@ export async function makeShopifyClient(config) {
...
+      const nextPage = await shopifyClient.collection.fetchAll(
+        response[response.length - 1].nextPage
```
pkeavenycontentful, line 53: Love this while loop 💯

apps/shopify/src/skuResolvers.js
```diff
@@ -26,37 +29,41 @@ export async function makeShopifyClient(config) {
...
+
   const collections = response.map((res) => convertCollectionToBase64(res));
 
   return validIds.map((validId) => {
     const collection = collections.find(
       (collection) => collection.id === convertStringToBase64(validId)
     );
     return collection
-      ? collectionDataTransformer(collection, config.apiEndpoint)
+      ? collectionDataTransformer(collection, removeHttpsAndTrailingSlash(config.apiEndpoint))
```
pkeavenycontentful: Really love the scout rule here 👍

apps/shopify/src/constants.js
```diff
@@ -2,7 +2,8 @@
 // returned with this unfortunate title, and there is no other way to check
 // whether the returned variant is the default one or not
 export const DEFAULT_SHOPIFY_VARIANT_TITLE = 'Default Title';
-export const SHOPIFY_API_VERSION = '2022-04';
+export const SHOPIFY_API_VERSION = '2023-01';
```
mgoudy91, line 5: This doesn't appear to be a huge jump, but any known breaking changes with this version bump?

apps/shopify/src/skuResolvers.js
```diff
@@ -69,25 +76,29 @@ export const fetchCollectionPreviews = async (skus, config) => {
...
 export const fetchProductPreviews = async (skus, config) => {
   if (!skus.length) {
     return [];
   }
-
   const validIds = filterAndDecodeValidIds(skus, 'Product');
   const shopifyClient = await makeShopifyClient(config);
-  const response = await shopifyClient.product.fetchMultiple(validIds);
+
+  const requests = [];
```
pkeavenycontentful, line 87: I believe this should be a `let` since you're mutating the array below.

apps/shopify/src/skuResolvers.js
```diff
@@ -69,25 +76,29 @@ export const fetchCollectionPreviews = async (skus, config) => {
...
+  for (let i = 0; i < validIds.length; i += SHOPIFY_ENTITY_LIMIT) {
+    requests.push(
+      shopifyClient.product.fetchMultiple(validIds.slice(i, i + (SHOPIFY_ENTITY_LIMIT - 1)))
+    );
+  }
+  const response = (await Promise.all(requests)).reduce((acc, curr) => [...acc, ...curr], []);
```
pkeavenycontentful: If you change `response` to a `let` above, then you can reassign without having to do `const` again.

apps/shopify/src/skuResolvers.js
```diff
@@ -143,11 +148,31 @@ export const fetchProductVariantPreviews = async (skus, config) => {
...
+
+  const validIds = filterAndDecodeValidIds(skus, 'ProductVariant');
+
+  const requests = [];
+  for (let i = 0; i < validIds.length; i += SHOPIFY_ENTITY_LIMIT) {
+    response.push(
+      _fetchProductVariantPreviews(validIds.slice(i, i + (SHOPIFY_ENTITY_LIMIT - 1)), config)
+    );
+  }
+  const response = (await Promise.all(requests)).reduce(
```
pkeavenycontentful: Where is `requests` being assigned? I see it's initialized above with an empty array, so wouldn't this be equivalent to `Promise.all([])`?

pkeavenycontentful (commented):
A few comments, some nits, some might need to be solved before merging.

apps/shopify/src/skuResolvers.js
rusticpenguin: Yup! It's a field attached to the the response nodes! It's sick

apps/shopify/src/dataTransformer.js
rusticpenguin: This was to keep sentiment with the old code which used a try/catch for this. `convertBase64ToString()` uses btoa which _could_ throw an error given it's handed a non-latin symbol.

I'm cool with just removing the try/catch, Shopify probably will only send us latin characters

mgoudy91: ➕ 

I might rather have specific guardrails here around an empty `sku`, or a poorly-formatted one (not ever sure if we know what that would be)

apps/shopify/src/skuResolvers.js
rusticpenguin: `const` blocks the reassignment of a variable, mutating an array or object is looked at as fine in JS

eg.
```
const numbersArr = [1]
numbersArr.push(2) 
// Success

const failedNumbersArr = [1]
failedNumbersArr = [...failedNumbersArr, 2]
// Uncaught TypeError: Assignment to constant variable
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

mgoudy91: I think this is actually okay `technically` since the array identifier isn't changing, just the contained values which can still mutate.

I had never run into this before, but it looks like this isn't totally a settled issue, and there are opinions on both sides.

apps/shopify/src/skuResolvers.js
rusticpenguin: I prefer to write code with solely consts, I find it's easier for developers to read code that's declarative vs imperative. Leads to less cognitive load when trying to keep track of what's happening.

I'm down to discuss this though!

apps/shopify/src/skuResolvers.js
rusticpenguin: Looks like you're on an older commit here. I accidentally pushed up that when refactoring to use Promise.all.  This is where the requests are being updated in the current code:

https://github.com/contentful/apps/pull/2944/files#diff-4a726df8342beeae0cec7ae3ccab3fed5adb3c4abeb148635af505d319365bd3R166

mgoudy91 (approved):
Looks like there's some active discussion around some parts of the code still that should probably be resolved before this goes out, but as written I think this looks good enough to ship.

Nice work! 🎉

apps/shopify/src/skuResolvers.js
pkeavenycontentful: Hmmmm, I take your point, as long as we aren't re-declaring `const` values then we're good.

pkeavenycontentful (approved):
Pulled down locally, verified backwards compatibility ✅ . Difficult to verify the specific use case of more than 250 collections, but as long as the previous functionality is retained, we can merge/deploy this and verify in prod.

apps/shopify/src/dataTransformer.js
rusticpenguin: Tested it locally with an empty sku/malformed, and looks like it handles it gracefully and the results still load successfully

PR #2948, opened 2023-03-27
fixes infinite loop for collections

rusticpenguin:
## Purpose
The "hasNextPage" is permanently set to true, causing us to infinitely call shopify for collections

## Approach
We are checking if the current page has 250 results, if it does we can safely assume there will be a next page


pkeavenycontentful (approved):

apps/shopify/src/skuResolvers.js
```diff
@@ -48,11 +48,14 @@ export const fetchCollectionPreviews = async (skus, config) => {
...
   if (response.length > 0) {
-    while (response[response.length - 1].hasNextPage) {
+    let hasNextPage = response.length === SHOPIFY_ENTITY_LIMIT;
+    while (hasNextPage) {
       const nextPage = await shopifyClient.collection.fetchAll(
         response[response.length - 1].nextPage
       );
       response.push(...nextPage);
+
+      hasNextPage = nextPage.length === SHOPIFY_ENTITY_LIMIT;
```
ryunsong-contentful, line 58: Won't this leave a bug where lets say you have 300 elements. 250 get covered and pushed to the response array. But you're going to miss the other 50 right?

apps/shopify/src/skuResolvers.js
```diff
@@ -48,11 +48,14 @@ export const fetchCollectionPreviews = async (skus, config) => {
 
   const response = await shopifyClient.collection.fetchAll(SHOPIFY_ENTITY_LIMIT);
   if (response.length > 0) {
-    while (response[response.length - 1].hasNextPage) {
+    let hasNextPage = response.length === SHOPIFY_ENTITY_LIMIT;
+    while (hasNextPage && response[response.length - 1].hasNextPage) {
```
ryunsong-contentful, line 52: If `response` is an array (which I think should be renamed to something more descriptive), why don't we just `.map` this and return the array of promises. And then resolve those promises. Didn't you do this promise/resolution pattern somewhere as well in this file?

ryunsong-contentful: Like you did here https://github.com/contentful/apps/blob/master/apps/shopify/src/skuResolvers.js#L89-L94

apps/shopify/src/basePagination.js
```diff
@@ -34,7 +34,9 @@ export default class BasePagination {
       pagination: {
         hasNextPage: this.hasNextProductPage,
       },
-      products: products.map(this.dataTransformer),
+      products: products.map((product) => {
+        return this.dataTransformer(product);
+      }),
```
rusticpenguin, line 39: this was originally causing the dataTransformer's `apiEndpoint` to be populated with the index of the current product

apps/shopify/src/skuResolvers.js
rusticpenguin: This was changed significantly in the current commit. Mind taking a look?

rusticpenguin:
The force push was from reverting the revert on main and rebasing off that to fix git history

apps/shopify/src/utils/base64.js
```diff
@@ -19,8 +19,8 @@ export const convertStringToBase64 = (str) => {
 export const convertProductToBase64 = (res) => {
   return {
     ...res,
-    id: convertStringToBase64(res.id),
-    product: res.product && { ...res.product, id: convertStringToBase64(res.product.id) },
+    id: convertStringToBase64(res?.id),
+    product: res?.product && { ...res?.product, id: convertStringToBase64(res?.product.id) },
```
rusticpenguin, line 23: This adds tolerance to if the product/collection/variant were to be deleted when selected, it'll show up as product missing:
image

mgoudy91 (approved):
Appreciate the comments and screenshots on the PR, answered a few questions I think I would have had 👍 

Anyway, I don't have any actual comments on the code itself. Nice work

pkeavenycontentful:
LGTM. Fixes an edge case and seems to solve the infinite loop case. Nice work 👍 

Let's be mindful of the deploy and be sure to test it out in prod since the live data seems to vary from what we intend. LMK when this is merged and I can monitor.

PR #2950, opened 2023-03-27
Integ 316 copy updates

gavinmatt:
## Purpose
Resolve ambiguous references to Google Analytics and remove an & in favor of "and"

## Approach
Targeted copy replacement following QA

N/A

## Deployment
N/A

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/api-access/ApiAccessSection.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/api-access/ApiAccessSection.tsx
@@ -33,7 +33,7 @@ const ApiAccessSection = (props: Props) => {
       <div>
         <Subheading marginBottom="spacingXs">API access</Subheading>
         <Paragraph marginBottom="none">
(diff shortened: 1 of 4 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/api-access/ApiAccessSection.tsx
gavinmatt, line 1: I know that this one is pedantic but feels right to update. Technically the Data API is only available for GA4. Still, might as well eliminate a place for mistakes.

matthew-gordon (approved):
LGTM 👍

jsdalton (approved):

apps/google-analytics-4/frontend/src/components/config-screen/api-access/ApiAccessSection.tsx
jsdalton: no i think it's totally legit 👍

ryunsong-contentful (approved):

PR #2951, opened 2023-03-27
Add date filter to GA4 external link

whitelisab:
## Purpose
When a user is viewing GA4 data in the sidebar and clicks on the 'Open in Google Analytics' link, the report they are viewing in GA4 should have the same date filter as the sidebar. This PR adds date filter parameters to the external URL that matches the dates provided to the` run_report` endpoint.

## Approach
There are some default report date filters (e.g. Today, Last 7 Days, etc.). Instead of using these, I decided to create a custom date filter that exactly matches the dates provided to the `run_report` endpoint so that the dates in the sidebar and in the linked report will be the same.

Example sidebar showing March 20-27 and corresponding GA4 report (you can see the custom date filter in the top right)

Screenshot 2023-03-27 at 4 39 00 PM
Screenshot 2023-03-27 at 4 41 09 PM

### DIFF
--- a/apps/google-analytics-4/frontend/src/helpers/externalUrlHelpers/externalUrlHelpers.ts
+++ b/apps/google-analytics-4/frontend/src/helpers/externalUrlHelpers/externalUrlHelpers.ts
@@ -1,46 +1,59 @@
 import { StartEndDates } from 'types';
 
-export const getExternalUrl = (
-  propertyId: string,
-  pagePath: string,
-  startEndDates: StartEndDates
-) => {
-  // Base Google Analytics URL
-  const BASE_URL = 'https://analytics.google.com/analytics/web/#';
+export const BASE_URL = 'https://analytics.google.com/analytics/web/#';
+const PAGE_PATH_VIEW = 'unifiedPagePathScreen';
 
-  if (!propertyId || !pagePath || !startEndDates.start || !startEndDates.end) {
+export const getExternalUrl = (
+  propertyId?: string,
+  reportFilterOptions?: {
+    pagePath?: string;
+    startEndDates?: StartEndDates;
+  }
+) => {
+  if (!propertyId) {
     return BASE_URL;
   }
 
   const propertyIdSegment = `/p${propertyId}/reports/explorer?params=`;
 
-  const explorerCardParams = 'unifiedPagePathScreen';
   const explorerCardSegment = encodeURIComponent(
-    `_u..nav=maui&_r.explorerCard..seldim=["${explorerCardParams}"]&`
+    `_u..nav=maui&_r.explorerCard..seldim=["${PAGE_PATH_VIEW}"]`
   );
 
-  const dataFiltersParams = {
-    type: 1,
-    fieldName: 'unifiedPagePathScreen',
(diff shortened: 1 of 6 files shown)


apps/google-analytics-4/frontend/src/helpers/externalUrlHelpers/externalUrlHelpers.ts
```diff
@@ -43,3 +57,13 @@ export const encodeDataFiltersParameters = (str: string) => {
 
   return encodedStr;
 };
+
+const formatDateParams = (dateString: string) => {
+  // Dates need to be formatted from YYYY-MM-DD to YYYYMMDD and month must be zero-padded
```
whitelisab, line 62: I tested this with single digit months and dates (e.g. 2023-1-3) and learned that the month needs to be zero-padded (e.g. 2023-01-3). For the date, it's ok that it's a single digit and it still works.

apps/google-analytics-4/frontend/src/helpers/externalUrlHelpers/externalUrlHelpers.ts
```diff
@@ -1,36 +1,50 @@
...
+  propertyId: string,
+  pagePath: string,
+  startEndDates: StartEndDates
+) => {
   // Base Google Analytics URL
   const BASE_URL = 'https://analytics.google.com/analytics/web/#';
 
-  if (propertyId && pagePath) {
-    const propertyIdSegment = `/p${propertyId}/reports/explorer?params=`;
+  if (!propertyId || !pagePath || !startEndDates.start || !startEndDates.end) {
```
mgoudy91, line 11: I don't know the ga4 url structure super well, but is there a world where someone would pass the properyId and pagePath, but not dates, and reasonably expect a url constructed with those two parameters?

ryunsong-contentful: You're also missing a check for `!startEndDates`
If `startEndDates` is null/undefined, your code will just break during runtime when it tries to call `!startEndDates.start`

ryunsong-contentful (approved):

apps/google-analytics-4/frontend/src/helpers/externalUrlHelpers/externalUrlHelpers.ts
whitelisab: Thank you both! I made some updates to these helpers to better handle these situations

apps/google-analytics-4/frontend/src/helpers/externalUrlHelpers/externalUrlHelpers.ts
```diff
@@ -1,36 +1,50 @@
-export const getExternalUrl = (propertyId: string, pagePath: string) => {
+import { StartEndDates } from 'types';
+
+export const getExternalUrl = (
+  propertyId: string,
+  pagePath: string,
+  startEndDates: StartEndDates
```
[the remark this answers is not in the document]
whitelisab: 👍 I made them optional

PR #2952, opened 2023-03-27
Revert "Fix/integ 192"

pkeavenycontentful:
Reverts contentful/apps#2944

Ran into an issue on prod, need to verify the fix locally and rollback until then.


rusticpenguin (approved):
uh oh D: thank you for the eyes on this!

PR #2990, opened 2023-03-29
SAP build scripts

mgoudy91:
## Purpose
Adds SAP package json hooks for our lerna build and deploy steps:

- `test:ci` and `build` jobs
- Snapshot and typescript fixes
- Add missing dep


apps/sap-commerce-cloud/src/components/Dialog/ProductList.tsx
```diff
@@ -20,7 +20,7 @@ export class ProductList extends React.Component<Props> {
 
   render() {
     const isFieldTypeSymbol =
-      get(this.props.sdk.parameters.invocation, 'fieldType', '') === 'Symbol';
+      (get(this.props.sdk.parameters.invocation, 'fieldType', '') as string) === 'Symbol';
```
pkeavenycontentful, line 23: Interesting use of `as string` and `as string[]` above. Did this code throw an error without the explicit casting to string?

pkeavenycontentful (approved):
One question, otherwise looks good to me.

apps/sap-commerce-cloud/src/components/Dialog/ProductList.tsx
mgoudy91: yeah it had some bizarre type error in the build and a quick google provided this an answer. I think it looks kinda dumb but it's also fairly inert and seems to make typescript happy 🤷

PR #2992, opened 2023-03-29
Feature/integ 305 Add SAP Analytics Code to Request Headers

pkeavenycontentful:
## Purpose
SAP Requires the SAP Analytics code sent as a request header for official partners. This PR adds the request header for all calls to SAP. 

## Approach
Trying out a new approach to secrets storage for this PR. We discovered that AWS Function URLs allow for a single-use lambda endpoint without setting up an entire lambda for secret storage. (Shout out to @mgoudy91 for finding this 👍 )

https://docs.aws.amazon.com/lambda/latest/dg/lambda-urls.html

For this PR, it queries the AWS Function URL set up in our AWS space, which queries our Secret Storage for the SAP Application Interface key. It then uses this key in Request calls. 

An argument could be made for using the AWS Secrets Manager client SDK if the above approach doesn't working for whatever reason. 

Instructions for adding the Application Interface Key: https://partneredge.sap.com/en/library/assets/partnership/dgl/76/37/SAP807637.html (login required, credentials are in our password vault)

Also, SAP rejects calls from localhost when a request header is present. Currently we've validated the key works in Postman/RestNinja. 

## Deployment
Env vars added for determining the dev environment. 

## Follow-up / out of scope tasks

- API Gateway on the Function URL to make it look nicer
- Moving Function URL to the correct region in AWS 
(both requires permissions I don't have at the moment)


apps/sap-commerce-cloud/src/api/fetchBaseSites.ts
```diff
@@ -1,12 +1,27 @@
...
+import { config } from '../config';
 
-export async function fetchBaseSites(config: any): Promise<string[]> {
-  const response = await fetch(config.installation.apiEndpoint + '/occ/v2/basesites');
-  if (response.ok) {
-    const responseJson = await response.json();
-    const baseSites = responseJson['baseSites'].map(baseSiteTransformer());
-    return baseSites;
+export async function fetchBaseSites(
+  parameters: any,
```
matthew-gordon, line 5: are we concerned at all with implicit/explicit `any` types at all? Wondering if we should define these explicitly? Just for my own curiosity nothing blocking by any means 🙏

matthew-gordon (approved):

apps/sap-commerce-cloud/src/api/fetchBaseSites.ts
pkeavenycontentful: Personally I'm not a fan of `any`, and would rather we have explicit types/interfaces defined. A lot of `any`'s in this app are leftovers from the external team that originally built this, so seems like more of a shortcut to less TS errors. I can create a type for these parameters while I'm in that code (better to do it now than later imo). 

Good callout 👍

pkeavenycontentful: @matthew-gordon updated.

PR #2998, opened 2023-03-29
[INTEG-255] Handle error cases for property selection

whitelisab:
## Purpose
Space admins configuring the GA4 app need useful feedback when there are error cases with the property selection. 

## Approach
There are several cases where no property ids are returned from the account_summaries endpoint (either due to an error fetching or if the service account doesn't have access to any properties). If there are no properties, we want to show a helpful message instead of any empty dropdown. Below are two example states where there are no properties to display.

Screenshot 2023-03-29 at 2 33 29 PM

Screenshot 2023-03-29 at 2 34 43 PM

There is also a case where a space admin may have previously selected a property id, but that id is no longer available. In this case, we want to let the space admin know that there is an error and how to fix it. Below are screenshots for this situation.

Screenshot 2023-03-29 at 2 39 00 PM

Screenshot 2023-03-29 at 2 39 07 PM

I also updated the MapAccountPropertySection component so that it won't render until the API access section is done loading. So when a user clicks the test button, then the property id dropdown will also load until the test is complete. This fixes the problematic state where we had seen indefinite loading on the property id dropdown.

Finally, I created a shared Warning Display component that is used by both the property id section and the content types section.


apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertyDropdown.tsx
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (!content) {
+      setWarningType('');
+    }
+
+    setTooltipContent(content);
+  }, [selectedPropertyId, originalPropertyId, isPropertyIdInOptions]);
+
+  const validateSelection = () => {
```
jsdalton, line 51: Hm I was a tiny bit confused by this at first but I _think_ it's because this doesn't need to be a function. Shouldn't it just be a variable?

```js
const selectionValue = (selectedPropertyId === originalPropertyId && !isPropertyIdInOptions) ? '' : selectedPropertyid
```

and then update below in the component prop

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
```diff
@@ -1,45 +1,61 @@
...
+
+      setIsPropertyIdInOptions(isInOptions);
+    }
+  }, [sortedAccountSummaries, originalPropertyId, parameters.propertyId]);
 
   useEffect(() => {
     if (parameters.propertyId) {
       setSelectedPropertyId(parameters.propertyId);
-      onIsValidAccountProperty(true);
+      // Prevent saving if selected property id is not in list of available properties
```
jsdalton, line 48: Hm, do we want to do this?

I'm thinking that we might prefer to do what we do for content types -- that is leave them in an invalid state (withv visible error state) but allow you to save regardless. Keep in mind this would block you from saving other important changes you make to content types. It also might take some time for someone to get access, so they could be blocked for a while.

To be a bit more clear we could allow them to keep their existing value as is (even though it's not in the list) and save would go through.

WDYT?

apps/google-analytics-4/frontend/src/components/config-screen/WarningDisplay/constants/warningMessages.ts
```diff
@@ -13,6 +14,13 @@ export const getSlugFieldDeletedMsg = (contentTypeId: string, slugField: string)
   Please select a new slug field in order for the app to render correctly in the sidebar. `;
 };
 
+// Property selection warning and error messages
+export const NO_PROPERTIES = `There are no properties to display.`;
```
jsdalton, line 18: Copy suggestions:

> The service account this app is linked to has not been given access to any analytics properties.

or something similar? this suggestion tries to describe the problem and suggest how they could solve it.

apps/google-analytics-4/frontend/src/components/config-screen/WarningDisplay/constants/warningMessages.ts
```diff
@@ -13,6 +14,13 @@ export const getSlugFieldDeletedMsg = (contentTypeId: string, slugField: string)
...
+
+export const getPropertyDeletedMsg = (propertyId: string): string => {
+  return `The previously configured property (${propertyId}) is no longer available via the configured Google Service Account. Either ensure that the Service Account has access to this property or select a new property and save the configuration.`;
```
jsdalton, line 21: Small style clean up:

> The previously configured property (${propertyId}) is no longer available via the installed Google service account. Either ensure that the service account has access to this property or select a new property and save the configuration.

jsdalton (commented):
I have a few small suggestions and copy ideas but otherwise I think this is looking really great. Let me know what you think!

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertyDropdown.tsx
whitelisab: Makes sense, updated this!

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
whitelisab: This is actually how the content type configuration is working -- if you have an invalid selection (the content type or the slug field that you had previously selected was deleted), then you are not able to save. My initial intention was to encourage the user to immediately address anything invalid, but your point does make sense. There are so many things on this configuration page, so it might not make sense to block a user from saving other parts of the configuration. I'll remove this behavior on the property id dropdown and will also add a note to remove that behavior in the task to clean up some UX with the content type selection. Thanks for pointing this out!

apps/google-analytics-4/frontend/src/components/config-screen/WarningDisplay/constants/warningMessages.ts
whitelisab: This component displays the Note instead of the dropdown when there are no account summaries; it is not aware of the API error state. So I purposely kept the content generic here thinking that the user would be able to look in the API access section for more descriptive info. If I update the content to something more descriptive, then it may not apply to a a general API error as the first screenshot shows. So I think I'd prefer to keep this more generic. What do you think?

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertyDropdown.tsx
```diff
@@ -55,51 +56,47 @@ const MapAccountPropertyDropdown = (props: Props) => {
...
+                  {accountSummary.propertySummaries.map((propertySummary) => (
+                    <Select.Option key={propertySummary.property} value={propertySummary.property}>
+                      {`${propertySummary.displayName} (${getIdOnly(propertySummary.property)})`}
+                    </Select.Option>
+                  ))}
+                </optgroup>
+              ))}
+            </Select>
+          </Stack>
+          <Paragraph marginTop="spacingM" style={{ color: tokens.gray600 }}>
```
jsdalton, line 83: It's super annoying because it's such a tiny thing but I think we should move this to a `MapAccountPropertyDropdown.styles.tsx` file like we do elsewhere (e.g. https://github.com/contentful/apps/blob/master/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentType.styles.ts), just so we avoid adhoc css creeping in.

jsdalton (approved):
Looks great! I just had the one refactor suggestion in the other set of comments but overall is looking good. 👍

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertyDropdown.tsx
whitelisab: No worries, makes sense! Updating now.

PR #2999, opened 2023-03-30
Ux/configuration instructions [INTEG-294]

jsdalton:
## Purpose

Presently our instruction copy is not great for configuring your property and content types. Specifically we need to:

* Help the user understand what choices they are making and why
* Point the user to appropriate instructions copy
* Provide contextual help as simply as possible

## Approach

* Add helpful text on the property dropdown.
* Move contextual "if you don't see a property" text to help text. Lower volume reduces noise and also confusion (since it only applies to some users)
* Add helpful text under content type configuration
* Add tooltips with descriptions + examples. (Examples might sound goofy but I think they make concrete what is otherwise very abstract for normal users)
* Add link to docks. It says "WRONG" because we don't have the real link and this intended to stick out like a sore thumb.

### Before

image

#### API access

image

#### API access

image

#### Google Analytics 4 property

image

#### Content Type configuration

image

image

image

image

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
@@ -146,7 +146,7 @@ export default function SetupServiceAccountCard(props: Props) {
         </FormControl>
       </Card>
       <Paragraph marginBottom="none">
-        To use the Google Analytics app, you will need to provision a{' '}
+        The Google Analytics 4 app requires a{' '}
         <TextLink
           icon={<ExternalLinkTrimmedIcon />}
           alignIcon="end"
@@ -155,12 +155,12 @@ export default function SetupServiceAccountCard(props: Props) {
           rel="noopener noreferrer">
           Google Cloud service account
         </TextLink>{' '}
-        for which you enable "viewer" access to one of your organization's Google Analytics
-        properties.
+        with "viewer" access to one of your organization's Google Analytics 4 properties.
       </Paragraph>
       <Paragraph marginBottom="none">
-        After configuring the service account, you'll download a set of credentials that Contentful
-        will use to access Google Analytics 4 data on this service account's behalf.
+        After configuring the service account, you will create and download a service account key
+        file (JSON) that connects this Contentful space to the Google Analytics 4 property you
+        specify.
       </Paragraph>
     </Stack>
   );
--- a/apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
@@ -155,7 +155,8 @@ export default function SetupServiceAccountCard(props: Props) {
           rel="noopener noreferrer">
           Google Cloud service account
         </TextLink>{' '}
-        for which you enable <i>read access</i> to your organization's Google Analytics data.
+        for which you enable "viewer" access to one of your organization's Google Analytics
+        properties.
       </Paragraph>
       <Paragraph marginBottom="none">
         After configuring the service account, you'll download a set of credentials that Contentful
--- a/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
@@ -165,6 +165,7 @@ const AssignContentTypeSection = (props: Props) => {
         </Paragraph>
         <Paragraph marginBottom="none">
           <TextLink
+            /* TODO: Correct this wrong link. (WRONG below in copy is intentional to ensure we don't forget) */
             href="https://support.google.com/analytics/answer/10759417"
             target="_blank"
             icon={<ExternalLinkIcon />}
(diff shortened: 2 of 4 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
```diff
@@ -148,10 +157,20 @@ const AssignContentTypeSection = (props: Props) => {
...
         <Paragraph marginBottom="none">
-          Select which content types will show the Google Analytics functionality in the sidebar.
-          Specify the slug field that is used for URL generation in your application. Optionally,
-          specify a prefix for the slug.
+          <TextLink
+            href="https://support.google.com/analytics/answer/10759417"
+            target="_blank"
+            icon={<ExternalLinkIcon />}
+            alignIcon="end">
+            WRONG See our help documentation for more details
```
ghepting, line 172: What's going on here with all caps yelling at the user? 😂

apps/google-analytics-4/frontend/src/components/config-screen/map-account-property/MapAccountPropertySection.tsx
```diff
@@ -77,39 +78,47 @@ export default function MapAccountPropertySection(props: Props) {
...
-          <TextLink
-            href="https://support.google.com/analytics/answer/10759417"
-            target="_blank"
-            icon={<ExternalLinkIcon />}
-            alignIcon="end">
-            "Make the switch to Google Analytics 4"
-          </TextLink>{' '}
-          for more information.
+          In the dropdown below, select the Google Analytics 4 property that's connected to the
+          website where content in this Contentful space (environment) appears.
```
ghepting, line 82: Recommend `website where this content appears.` to reduce the complexity here. It's difficult to read and understand. We need to dumb things down

apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
```diff
@@ -155,7 +155,8 @@ export default function SetupServiceAccountCard(props: Props) {
           rel="noopener noreferrer">
           Google Cloud service account
         </TextLink>{' '}
-        for which you enable <i>read access</i> to your organization's Google Analytics data.
+        for which you enable "viewer" access to one of your organization's Google Analytics
+        properties.
```
ghepting, line 159: i recommend:

```
The Google Analytics 4 app requires a
[Google Cloud service account](https://www.contentful.com/help/google-analytics-service-account-setup/)
 with "viewer" access to one or more Google Analytics properties.
```

apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
```diff
@@ -155,7 +155,8 @@ export default function SetupServiceAccountCard(props: Props) {
...
       </Paragraph>
       <Paragraph marginBottom="none">
         After configuring the service account, you'll download a set of credentials that Contentful
```
ghepting, line 162: I would recommend changing this to:

```
With the service account configuration, you will create a new service account key file (JSON) that will enable you to access Google Analytics 4 data from within Contentful.
```

ghepting: might help connect the dots a bit more for folks in the GCP UI I think (specifically saying "service account key file" and "JSON" since they have to navigate through the UI making both of these distinguishing selections)

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
jsdalton: > Add link to docks. It says "WRONG" because we don't have the real link and this intended to stick out like a sore thumb.

Very much intentional because this is the wrong link and it's super easy to overlook forgetting it. Should have it in the next day or so.

apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
jsdalton: I like it 👍

apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
jsdalton: Yes

apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
jsdalton: Going with your suggestion with a few minor modifications:

> The Google Analytics 4 app requires a [Google Cloud service account] with "viewer" access to one of your organization's Google Analytics 4 properties.

apps/google-analytics-4/frontend/src/components/config-screen/api-access/setup/SetupServiceAccountCard.tsx
jsdalton: I think "you" is a bit weird here btw, but leaving it. The more pedantically correct wordings ("this app" or "users of this space environment") are unhelpfully more complicated.

PR #3009, opened 2023-03-30
[Cleanup] Remove process.envs to use the function getEnvironmentVariable instead

ryunsong-contentful:
## Purpose
We want to use `getEnvironmentVariable` to actually access our env vars instead of `process.env` so that we have more control of when the application should error out/succeed when accessing environment variables

Also updates the `getEnvironmentVariable` function to cleanup some of the package.json commands to set some of the env vars.

## Approach
-Make sure all tests run successfully
-Clean up package.json 
-Build should pass successfully

### DIFF
--- a/apps/google-analytics-4/frontend/src/config.ts
+++ b/apps/google-analytics-4/frontend/src/config.ts
@@ -1,6 +1,12 @@
-function getEnvironmentVariable(environmentVariableName: string) {
+type EnvironmentVariable = string | null | number | undefined;
+
+function getEnvironmentVariable(
+  environmentVariableName: string,
+  fallback?: EnvironmentVariable
+): string {
   const environmentVariableValue = process.env[environmentVariableName];
-  if (environmentVariableValue === undefined) {
+  if (!environmentVariableValue) {
+    if (fallback) return fallback.toString();
     throw new Error(`Missing environment variable: '${environmentVariableName}'`);
   }
   return environmentVariableValue;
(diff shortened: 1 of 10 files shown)


ghepting (approved):

apps/google-analytics-4/frontend/src/config.ts
```diff
@@ -1,14 +1,16 @@
-function getEnvironmentVariable(environmentVariableName: string) {
+function getEnvironmentVariable(environmentVariableName: string, fallback?: any) {
```
ghepting, line 1: change `any` to `string | number | null`

apps/google-analytics-4/lambda/src/config.ts
```diff
@@ -1,6 +1,7 @@
-function getEnvironmentVariable(environmentVariableName: string) {
+function getEnvironmentVariable(environmentVariableName: string, fallback?: any) {
```
ghepting, line 1: - [ ] change `any` to `string | number | null`

apps/google-analytics-4/lambda/.env
```diff
@@ -1,6 +1,6 @@
 # Important! These values should only be used when running tests. All environment variables must be
 # specified in serverless.yml or they will not function in the real Lambda environment
-SIGNING_SECRET: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+SIGNING_SECRET: QPFQDCfipitJcFGb4f-xnP0RhZOqnStP2thYUkqg_CcWdTlFWu4KCP6pYrWPto5x
```
matthew-gordon, line 3: was this intentional? Or does this need to revert back to a placeholder?

ryunsong-contentful: it is intentional, it's a random signing secret from my own random app - no security concerns

matthew-gordon (approved):

PR #3037, opened 2023-03-31
Show checks when incomplete or problems with config

lilbitner:
## Purpose
When the configuration steps are incomplete or when there exist issues within the config setup, the checklist is currently hidden and the user has to open it manually, an unnecessary extra step asked of the user. Ex:

image

This checklist should instead render automatically in these cases. 

## Approach
One adjustment in state was needed for the checklist to render appropriately when the checklist is incomplete or there exist error states. 
Screenshot 2023-03-31 at 8 27 40 AM


whitelisab (approved):
LGTM! 🌟

PR #3038, opened 2023-03-31
[INTEG-331 & INTEG-352] Fix content type errors and allow config save

whitelisab:
## Purpose
This PR fixes some issues with the content type configuration:
- When adding a new content type row, do not show an error state before the user has taken any action
- When there are content type errors, the user should still be able to save the configuration page

## Approach
Here's a [quick video](https://www.loom.com/share/8a2221b92342447da3060b5ae89a5867) showing current state and how I've addressed these issues in this PR.
- I removed the error state when adding a new row. If there is an empty row when saving, we are now filtering this row out instead of throwing an error
- I removed the validation on save if there is a content type error (e.g. a previously saved content type that has been deleted)
- I updated relevant tests and also removed a `console.log` from another PR

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
@@ -125,8 +125,20 @@ export default function GoogleAnalyticsConfigPage() {
       return false;
     }
 
+    // Remove empty content type rows
+    let parametersToSave = parameters;
+
+    if (parameters.contentTypes) {
+      const filteredContentTypes = Object.fromEntries(
+        Object.entries(parameters.contentTypes).filter(([key]) => key !== '')
+      );
+
+      parametersToSave = { ...parameters, contentTypes: filteredContentTypes };
+      setParameters(parametersToSave);
+    }
+
     // Assign the app to the sidebar for saved content types
-    const contentTypeIds = Object.keys(parameters.contentTypes ?? {});
+    const contentTypeIds = Object.keys(parametersToSave.contentTypes ?? {});
     const newEditorInterfaceAssignments = generateEditorInterfaceAssignments(
       currentEditorInterface,
       contentTypeIds,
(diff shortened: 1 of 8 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
```diff
@@ -125,8 +125,20 @@ export default function GoogleAnalyticsConfigPage() {
       return false;
     }
 
+    // Remove empty content type rows
+    let parametersToSave = parameters;
+
+    if (parameters.contentTypes) {
+      const filteredContentTypes = Object.fromEntries(
```
jsdalton, line 132: I had to look at this a minute to figure out what you were filtering and then guess why. Maybe we could just make it easier to understand with a more precise label such as `nonEmptyContentTypes`? not sure a comment is needed but could help (maybe something like "// we just want to filter out any empty content types that came from empty rows on the from").

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
```diff
@@ -93,18 +93,7 @@ const AssignContentTypeSection = (props: Props) => {
...
-
-    for (const contentType in newContentTypes) {
-      if (!contentType || !allContentTypeEntries.find((entry) => entry[0] === contentType)) {
-        hasValidContentTypes = false;
-        break;
-      }
-    }
-
-    onIsValidContentTypeAssignment(hasValidContentTypes);
+    onIsValidContentTypeAssignment(true);
```
jsdalton, line 96: Here's another place I might leave a comment because I can see this hardcoded `true` being confusing to someone with fresh eyes on this code base. It'd be great for a line of context on _why_ we're always true here (i.e. because we always want to allow the save)

jsdalton (approved):
One or two small suggestions to help be more clear what we're doing and why otherwise looking 💯 .

apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage.tsx
whitelisab: Yes! I will update this for clarity

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
whitelisab: 👍 added a comment for clarity

PR #3042, opened 2023-03-31
[INTEG-361] Fix sidebar error when there are no content types

whitelisab:
## Purpose
This PR fixes an error in the `Sidebar` component when there are no configured content types. Additionally, I fixed two warnings in the `HyperLink` and `DisplayServiceAccountCard` components.

## Approach
Reproduction steps for `Sidebar` error:
- Install the app but do not configure any content types
- Add the app manually to the sidebar of a content type
- Navigate to an entry of that content type. You should see the error: "Cannot read properties of undefined (reading 'content type id')"

In the `HyperLink` component, there was a warning for "Each child in a list should have a unique 'key' prop"

In the `DisplayServiceAccountCard` component, there was a warning for "div cannot appear as a descendant of p"


lilbitner (approved):
Crushed it 💯

PR #3043, opened 2023-03-31
WIP: Add a next steps section to the config page

lilbitner:
## Purpose
It was brought up in the mob QA that it isn't super clear to the user what to do after everything is configured on the App Config page. Therefore, I have added a new small section to nudge the user to add content types if they have not already, as well as to head to the Content tab to view the app. I have also added a Contact Support link. 

#### [**This**](https://www.loom.com/share/9e458bc2e5d04c0183581a9ab1f3467b) video reviews my thoughts and questions around the implementation. Please watch on 1.5x speed 😅

**Example of section when content type not yet configured:** 
Screenshot 2023-04-03 at 8 16 03 AM

**Example of section when content type configured:** 
Screenshot 2023-04-03 at 8 16 14 AM

### Would appreciate any and all thoughts here, this is very much a WIP PR, as I want to welcome any feedback for functionality/copy/anything/everything 😄 
.


apps/google-analytics-4/frontend/src/components/config-screen/NextStepsSection/NextStepsSection.tsx
```diff
@@ -0,0 +1,45 @@
+import HyperLink from 'components/common/HyperLink/HyperLink';
+import { Paragraph, Subheading } from '@contentful/f36-components';
+import { AppExtensionSDK } from '@contentful/app-sdk';
+import { useSDK } from '@contentful/react-apps-toolkit';
+import { ExternalLinkIcon } from '@contentful/f36-icons';
+
+export const NO_CONTENT_TYPES_MSG =
+  'Be sure to save this configuration. Once you have configured a content type, head over to the Content tab, open an entry associated with a configured content type, and view the app in the sidebar.';
```
whitelisab, line 8: maybe change "...open an entry associated with a configured content type..." to "...open an entry of a configured content type..."? Just a bit less wordy.

apps/google-analytics-4/frontend/src/components/config-screen/NextStepsSection/NextStepsSection.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const NextStepsSection = (props: Props) => {
+  const { isContentTypeConfigured } = props;
+  const sdk = useSDK<AppExtensionSDK>();
+
+  const openEntriesList = () => sdk.navigator.openEntriesList();
+
+  return (
+    <>
+      <Subheading marginBottom="spacingXs">View app</Subheading>
```
whitelisab, line 24: I don't mind "View app" as the section header; maybe another option to consider is "View app on content entry"?

whitelisab (commented):
I like this approach! I think it helps direct the user to where they can verify if they've configured everything correctly. Just adding some minor thoughts re: copy.

apps/google-analytics-4/frontend/src/components/config-screen/NextStepsSection/NextStepsSection.tsx
lilbitner: Nice, I agree, the less wordy the better 💯

apps/google-analytics-4/frontend/src/components/config-screen/NextStepsSection/NextStepsSection.tsx
lilbitner: oh I like that, will adjust, thank you!

PR #3058, opened 2023-04-03
optimizes keyboard navigability in content type configuration section

ghepting:
## Purpose
optimize keyboard navigability and accessibility of the content type configuration section of the GA4 config screen

## Approach
ensure focus is set to newly added content type configuration rows

https://user-images.githubusercontent.com/492573/229544975-255699a1-92b5-4b54-b673-31da2763be90.mov

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeRow.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeRow.tsx
@@ -16,6 +16,7 @@ interface Props {
   onRemoveContentType: (key: string) => void;
   currentEditorInterface: Partial<EditorInterface>;
   originalContentTypes: ContentTypes;
+  focus: boolean;
 }
 
 const AssignContentTypeRow = (props: Props) => {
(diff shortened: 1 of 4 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeRow.tsx
```diff
@@ -126,6 +128,13 @@ const AssignContentTypeRow = (props: Props) => {
     );
   };
 
+  useEffect(() => {
+    if (focus) {
+      const contentTypeSelect = document.getElementById(`contentType-${index}`);
+      if (contentTypeSelect) contentTypeSelect.focus();
```
pkeavenycontentful, line 134: Is there a reason to use an explicit `focus()` in the useEffect instead of `tabindex` on the element itself with incremental integer values? https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

ghepting: this element is dynamically injected into the DOM and _above_ the `Add another content type` button. This handler is responsible for setting the focus to the newly added element for optimal/seamless keyboard navigation

ghepting: also, we're not trying to _alter_ the natural tabindex, just set focus to the new element

whitelisab (approved):
LGTM!

PR #3060, opened 2023-04-03
Fix broken test

lilbitner:
## Purpose
Fix a flaky test that fails intermittently depending on current dates.

## Approach
Adjusted the test to be independent of dates and reference absolute date values. 

## Deployment
This test is failing in deployment to production hence the quick small PR for it.


ghepting (approved):

apps/google-analytics-4/frontend/src/helpers/DateRangeHelpers/DateRangeHelpers.spec.ts
```diff
@@ -1,9 +1,11 @@
 import { DateRangeType } from '@/types';
 import getRangeDates, { RANGE_OPTIONS } from './DateRangeHelpers';
 
+const DAY_IN_MS = 1000 * 3600 * 24;
```
ghepting, line 4: _maybe_ better to have the whole expression like: `1000 * 60 * 60 * 24`? not a blocker :)

pkeavenycontentful (approved):

PR #3061, opened 2023-04-03
[Fix] Date range tests GA4

ryunsong-contentful:
## Purpose
Failures in GA4 tests because we are in a new month. 

## Approach
Update the test to not compare the date diffs, but the actual date number that's extracted from the parser


lilbitner (approved):
👍 🚀 💯

PR #3062, opened 2023-04-03
feat: stores product url instead of only sku id

rusticpenguin:
## Purpose
An experienced developer would want a full URL to fetch their products. This makes the developer experience a lot nicer

## Approach
Store the data as the full URL, and split on `/products/` when we need the sku ID (this avoids the issue of the sku ID containing `/`)

### DIFF
--- a/apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
+++ b/apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
@@ -23,7 +23,7 @@ export async function fetchProductPreviews(
         },
       };
   for (const sku of skus) {
-    skuIds.push(sku.split('/products/').pop()!);
+    skuIds.push(sku.split('/products/').pop() as string);
     const response = await fetch(
       sku +
         '?fields=code,name,summary,price(formattedValue,DEFAULT),images(galleryIndex,FULL),averageRating,stock(DEFAULT),description,availableForPickup,url,numberOfReviews,manufacturer,categories(FULL),priceRange,multidimensional,configuratorType,configurable,tags',
--- a/apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
+++ b/apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
@@ -23,14 +23,9 @@ export async function fetchProductPreviews(
         },
       };
   for (const sku of skus) {
-    const splitSku = sku.split(':');
-    const baseSite = splitSku[0];
-    const skuId = splitSku[1];
-    skuIds.push(skuId);
+    skuIds.push(sku.split('/products/').pop()!);
     const response = await fetch(
-      parameters.installation.apiEndpoint +
-        `/occ/v2/${baseSite}/products/` +
-        skuId +
+      sku +
         '?fields=code,name,summary,price(formattedValue,DEFAULT),images(galleryIndex,FULL),averageRating,stock(DEFAULT),description,availableForPickup,url,numberOfReviews,manufacturer,categories(FULL),priceRange,multidimensional,configuratorType,configurable,tags',
       headers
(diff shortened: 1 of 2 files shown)


apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
```diff
@@ -23,14 +23,9 @@ export async function fetchProductPreviews(
         },
       };
   for (const sku of skus) {
-    const splitSku = sku.split(':');
-    const baseSite = splitSku[0];
-    const skuId = splitSku[1];
-    skuIds.push(skuId);
+    skuIds.push(sku.split('/products/').pop()!);
```
pkeavenycontentful, line 26: Curious what the `!` is for at the end of the function. Is this an assertion? (not blocking)

apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
```diff
@@ -23,14 +23,9 @@ export async function fetchProductPreviews(
...
     const response = await fetch(
-      parameters.installation.apiEndpoint +
-        `/occ/v2/${baseSite}/products/` +
```
pkeavenycontentful, line 32: Is there a reason to remove the `parameters.installation.apiEndpoint + '/occ/etc`? This looks to be causing an issue when the Product Previews are fetched in the Field: 
Screenshot 2023-04-03 at 1 20 54 PM

pkeavenycontentful (changes requested):
Looks like an issue with the fetchProductPreviews() function as we're missing the full api endpoint URL. After that's fixed everything else looks good 👍

apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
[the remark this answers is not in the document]
mgoudy91: oh, it's a type safety thing

mgoudy91: @pkeavenycontentful https://github.com/contentful/apps/pull/3062#discussion_r1156372871 😂 

For what it's worth I think this is pretty tough to read and the fact two of us weren't sure what it did makes me think it's likely to be misunderstood again. Maybe we could rework this slightly to do a quick check on that string before attempting a pop. More lines, but maybe more readable?

apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
rusticpenguin: Convo was over slack, this was due to old data being in the blogpost

apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
rusticpenguin: Yeah this is a non-null assertion that was added in TypeScript 2.0. In this case we can safely assume this will always be defined. 

I can add a check to throw an error instead to keep the same sentiment 

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

pkeavenycontentful (approved):
Tested this using a brand new entry and the fetchProductPreviews() worked, existing entries will not due to the new way data is being stored. 

Confirmed on Publish that the JSON Object now shows absolute URLs instead of the product id. Nice work! 

Screenshot 2023-04-03 at 1 44 16 PM

apps/sap-commerce-cloud/src/api/fetchProductPreviews.ts
rusticpenguin: Changed it to use `as string` to keep with the sentiment of "this will not ever be undefined"

PR #3063, opened 2023-04-03
[INTEG-319] Update content for service account key setup

whitelisab:
## Purpose
When pasting in the service account key for the GA4 app configuration, the validation message `Service account key is valid` was confusing. We are only checking to see if the key is valid JSON, not if the key itself is valid. 

## Approach
Updated the validation message to `Service account key file is valid JSON` and also updated some content in the Note text. Below are examples of the messages presented to users with valid and invalid JSON service account key files.

Screenshot 2023-04-03 at 1 20 14 PM

Screenshot 2023-04-03 at 1 20 30 PM


rusticpenguin (approved):
LGTM!

lilbitner (approved):
I think this looks great! I think this solves the confusion that the key is somehow valid without even testing it. Its just a valid JSON object that is formatted correctly at this stage. Nice! 💯

PR #3065, opened 2023-04-03
chore: update pr template to include breaking changes and testing steps

rusticpenguin:
## Purpose
Two purposes:
1. In a pre-released app we may have breaking changes introduced in PR's. This would cause old data to break the new code.
1. While testing an app, it may be nice for someone not familiar with the tool to be able to run through a happy path first to see if any bugs appear. 

## Approach
Added a field for the dev to input breaking changes and testing steps


mgoudy91 (approved):
😎

PR #3081, opened 2023-04-04
SAP/Remove Application Interface Key

pkeavenycontentful:
## Purpose

We're having some issues with the application interface key in our prod space, SAP blocks access with the header added. In order to unblock prod we're removing this for the time being and have reached out to SAP to help diagnose.

### DIFF
--- a/apps/sap-commerce-cloud/src/index.tsx
+++ b/apps/sap-commerce-cloud/src/index.tsx
@@ -41,8 +41,8 @@ const getApplicationInterfaceKey = async (): Promise<boolean | string> => {
 
 init(async (sdk) => {
   const root = document.getElementById('root');
-  const isTestEnv = config.isTestEnv;
-  const sapApplicationInterfaceKey = !isTestEnv ? await getApplicationInterfaceKey() : '';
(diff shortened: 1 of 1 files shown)


matthew-gordon (approved):

apps/sap-commerce-cloud/src/index.tsx
```diff
@@ -41,8 +41,8 @@ const getApplicationInterfaceKey = async (): Promise<boolean | string> => {
 
 init(async (sdk) => {
   const root = document.getElementById('root');
-  const isTestEnv = config.isTestEnv;
-  const sapApplicationInterfaceKey = !isTestEnv ? await getApplicationInterfaceKey() : '';
+  //const isTestEnv = config.isTestEnv;
```
matthew-gordon, line 44: Can it be removed, or is this coming back into action at a later point? LGTM otherwise nothing blocking

pkeavenycontentful: Coming back into action at some point, we've reached out to SAP to help diagnose as it doesn't look to be an issue on our side.

PR #3084, opened 2023-04-04
[INTEG-356-373] Fix invalid and missing service account error handling

whitelisab:
## Purpose

This PR fixes two issues with error handling related to invalid/missing service account keys for the GA4 app configuration.

## Approach

The first issue is that when there is an invalid service account, the wrong error message was showing. It was showing "unknown error" when it should be showing the service key check as "Service account key is not valid".

Incorrect
image-20230404-144505

Correct
Screenshot 2023-04-04 at 9 48 07 AM

The second issue is when there is a missing service account key error, the API and property checks were showing as "enabled" when they should be showing as "Provide a valid service account key to run this check".

Incorrect
image-20230330-194317

Correct
Screenshot 2023-04-04 at 2 37 15 PM

This PR also adds handling for the "InvalidServiceAccountKey" error and fixes some of the tests in `ServiceAccountChecklist.spec.tsx`.

## Testing steps

To reproduce an invalid service account error from Google: 
- Install a valid service account key
- Remove the key in Google
- Refresh status checks

To reproduce the missing service account key/invalid service account key errors:
- Throw the error directly from the `ApiController`

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ChecklistUtils.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ChecklistUtils.tsx
@@ -291,15 +291,15 @@ export const getServiceKeyChecklistStatus = (
 export const getAdminApiErrorChecklistStatus = (
   isFirstSetup: boolean,
   parameters: KeyValueMap,
-  invalidServiceAccountError: ApiErrorType | undefined,
+  serviceAccountError: ApiErrorType | undefined,
   apiError: ApiErrorType | undefined
 ): ChecklistRow => {
   const checklistUrl = getApiChecklistURLs(parameters, apiError ? true : false).adminApi;
-  if (!invalidServiceAccountError && !apiError)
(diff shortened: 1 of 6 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/api-access/display/ChecklistUtils.tsx
```diff
@@ -291,48 +291,48 @@ export const getServiceKeyChecklistStatus = (
 export const getAdminApiErrorChecklistStatus = (
   isFirstSetup: boolean,
   parameters: KeyValueMap,
-  invalidServiceAccountError: ApiErrorType | undefined,
```
whitelisab, line 294: Renamed this `invalidServiceAccountError` to `serviceAccountError` in all of these functions since it could be an invalid or a missing service account error.

ryunsong-contentful (approved):

lilbitner (approved):
Looks great to me! 💯

PR #3085, opened 2023-04-05
Add waiting period when updating slug field to avoid a network call on every keystroke

lilbitner:
## Purpose

When the user updates the slug field value of an entry while viewing the GA4 app in the sidebar, each keystroke was causing a network request to the `runReports` endpoint. This is because we are using the `useFieldValue` hook from the `react-apps-toolkit` to grab that slug field value, so when the hook updated, all subsequent statetful hooks and components updated. 

## Approach
This was trickier than most classic debounce fixes. We are not debouncing a user typing into an input that is rendered as part of our component tree, for example, but rather trying to debounce a hook's ever changing value and how that value is used. 

I originally began by using the `lodash` `debounce` with the `useSidebarSlug` hook, where we grab the field value, but in any approach that I took to manipulate the hook's value, I encountered one of two problems: 1. A violation of the rules of hooks i.e placing a hook in another hook to conditionally call it or 2. State updates within the `useSidebarSlug` hook or `AnalyticsApp` component would cause re-renders, so the many network requests would continue, just in a delayed fashion. 

Therefore I did some digging and solved it by what seems to be a common approach in these cases: using `setTimeout` to update state (which I know `setTimeout` sometimes seems a little 'meh' to use but actually think it can be pretty powerful). This only updates the state once, after the timeout finishes, therefore only one network call is made. 

I set the timeout to be **700ms** given that felt like it reduced the most amount of network calls without causing a significant delay when I was testing, but happy to adjust lower or higher!

## Testing steps

Make changes to the slug field value of an entry, and observe the network tab. The `runReports` endpoint should not be called on every key stroke.

### DIFF
--- a/apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
+++ b/apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
@@ -19,7 +19,7 @@ export const useSidebarSlug = (slugFieldInfo: ContentTypeValue) => {
   };
 
   useEffect(() => {
-    const timeout = setTimeout(() => setDebouncedSlugFieldValue(slugFieldValue), 500);
+    const timeout = setTimeout(() => setDebouncedSlugFieldValue(slugFieldValue), 700);
 
     return () => clearTimeout(timeout);
   }, [slugFieldValue]);
--- a/apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
+++ b/apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { ContentTypeValue } from 'types';
 import { useSDK } from '@contentful/react-apps-toolkit';
 import { ContentEntitySys, SidebarExtensionSDK } from '@contentful/app-sdk';
@@ -18,17 +18,11 @@ export const useSidebarSlug = (slugFieldInfo: ContentTypeValue) => {
     setIsPublished(Boolean(sys.publishedAt));
   };
 
-  const updateSlugFieldValue = (v: string) => {
-    setDebouncedSlugFieldValue(v);
-  };
-
-  const handleSlugFieldValue = useCallback(updateSlugFieldValue, [updateSlugFieldValue]);
-
   useEffect(() => {
-    const timeout = setTimeout(() => handleSlugFieldValue(slugFieldValue), 500);
+    const timeout = setTimeout(() => setDebouncedSlugFieldValue(slugFieldValue), 500);
 
     return () => clearTimeout(timeout);
-  }, [slugFieldValue, handleSlugFieldValue]);
+  }, [slugFieldValue]);
 
   useEffect(() => {
     // We only want this to update state on component mount.
--- a/apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
+++ b/apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
@@ -31,7 +31,12 @@ export const useSidebarSlug = (slugFieldInfo: ContentTypeValue) => {
   }, [slugFieldValue, handleSlugFieldValue]);
 
   useEffect(() => {
+    // We only want this to update state on component mount.
+    // Otherwise when slugFieldValue changes it is updated
+    // in the above useEffect
+
     setDebouncedSlugFieldValue(slugFieldValue);
+    // eslint-disable-next-line
   }, []);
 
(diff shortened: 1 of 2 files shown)


apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
```diff
@@ -16,11 +18,26 @@ export const useSidebarSlug = (slugFieldInfo: ContentTypeValue) => {
     setIsPublished(Boolean(sys.publishedAt));
   };
 
+  useEffect(() => {
+    const timeout = setTimeout(() => setDebouncedSlugFieldValue(slugFieldValue), 700);
```
jsdalton, line 22: Minor suggestion to pull out the value here (700) as a constant with a meaningful name.

Also, just curious why 700? Seems fairly reasonable, but we might want to try for lower (in general slug fields almost never change, my thinking is we'd probably want to tune this to a fairly low number)

jsdalton (changes requested):
Looks great! One small change, if it makes sense to you.

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
lilbitner: yes I was playing around with the delay here and 500 - 700 felt like a sweet spot of not allowing too many network calls while not making the user wait for the changes to reflect. Happy to pull out into a constant, and will lower down to 500!

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
```diff
@@ -5,22 +5,35 @@ import { ContentEntitySys, SidebarExtensionSDK } from '@contentful/app-sdk';
...
 
+  const [isPublished, setIsPublished] = useState(false);
+  const [debouncedSlugFieldValue, setDebouncedSlugFieldValue] = useState(slugFieldValue);
+
   const handlePublishedStatus = (sys: ContentEntitySys) => {
     setIsPublished(Boolean(sys.publishedAt));
   };
 
+  useEffect(() => {
+    const timeout = setTimeout(
```
ghepting, line 24: ProTip: you can use [lodash debounce](https://lodash.com/docs/#debounce) if you want to avoid roll-your-own :)

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
```diff
@@ -5,22 +5,35 @@ import { ContentEntitySys, SidebarExtensionSDK } from '@contentful/app-sdk';
...
+
 export const useSidebarSlug = (slugFieldInfo: ContentTypeValue) => {
-  const [isPublished, setIsPublished] = useState(false);
   const sdk = useSDK<SidebarExtensionSDK>();
 
   const { slugField, urlPrefix } = slugFieldInfo;
   const slugFieldValue = useGetFieldValue(slugField);
 
+  const [isPublished, setIsPublished] = useState(false);
+  const [debouncedSlugFieldValue, setDebouncedSlugFieldValue] = useState(slugFieldValue);
```
ghepting, line 17: just to confirm my assumption/understanding here –  we need a separate state vs just doing setSlugFieldValue in a debounced fashion because we need a real time state to populate the react component's field value, but a debounced one that is tied to a useEffect elsewhere that updates the chart?

if i'm on the right track here, I actually wonder why we're so eagerly re-rendering the chart in the first place? it could definitely just render with whatever the slug field has when it initializes and then have a refresh button to manually re-render it that's conditionally shown based on a diff state being detected. this could reduce the "magic" here quite a bit i suspect?

jsdalton: I think it's a fair point btw, but for now I'd lean toward shipping this change and seeing if it doesn't make sense later to remove the "reactive" behavior of the chart altogether (or add a button to trigger a refresh etc)

jsdalton (approved):
Let's ship and def observe behavior in production!

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
lilbitner: Yeah you got it, so we are essentially having to "debounce" the state updates caused by the `getFieldValue` hook changing with every key stroke. And you definitely bring up a good point, the app has been consistently updating with any update to the slug field, and we originally thought "oh thats neat" but it certainly caused these unnecessary re-renders. I think this fix might be worth adding for now, as its simple enough imo, but perhaps along the lines of refresh button, we only update the state when the user has 'published' the data, so published data is the single source of truth? If we feel something along those lines is necessary now I am fine implementing that as well 👍

ghepting (approved):

PR #3106, opened 2023-04-05
improve loading status indication onConfigure and onConfigurationCompleted

ghepting:
## Purpose
This addressed a jarring UX from the moment a user clicks "save" or "install" through the post configuration/installation step when we write their secret key to dynamodb.

## Approach
It wraps both the `onConfigure` and `onConfigurationCompleted` callbacks with a boolean loading state so we can eagerly reset from the `isInEditMode` state, but avoid making eager calls to `verifyAdminApi()` and `verifyDataApi()` which are in a useEffect in the DisplayServiceAccountCard.tsx component (see newly added guard clause)

https://user-images.githubusercontent.com/492573/230084523-7fbef44d-bc49-48c7-b852-7871d4b57cb9.mov


lilbitner (approved):
Working really well!

jsdalton (approved):
Looks good! Works for me too. Let's ship it 🚢

PR #3117, opened 2023-04-05
Integ 381 test fixes

mgoudy91:


apps/google-analytics-4/lambda/src/middlewares/setSentryContext.spec.ts
```diff
@@ -17,7 +17,7 @@ describe('setSentryContext', () => {
   app.use(setSentryContext);
 
   app.get('/test', (_req, res) => {
-    res.send('ok');
+    res.sendStatus(200);
```
matthew-gordon, line 20: is there a difference between `status(200)` and `sendStatus(200)`?

mgoudy91: deprecation issue

PR #3121, opened 2023-04-05
Restore original react script version and package lock files

cantstoptheunk:
## Purpose
Regression and unintended downgrade of react script 

## Approach
Restore to original version


jsdalton (approved):
Great!

PR #3124, opened 2023-04-05
Fix package.json react script versions to original version

cantstoptheunk:
## Purpose
Dependabot PR https://github.com/contentful/apps/pull/3047/files downgraded a lot of react-script versions in the package.jsons. 

## Approach
Reverting those downgrades

### DIFF
--- a/apps/google-analytics-4/lambda/tsconfig.json
+++ b/apps/google-analytics-4/lambda/tsconfig.json
@@ -10,6 +10,9 @@
     "inlineSources": true,
     "sourceRoot": "/src"
   },
-  "include": ["src", "test"],
+  "include": [
+    "src",
+    "test",
+  ],
   "exclude": ["node_modules"]
 }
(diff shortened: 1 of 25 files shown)


apps/google-analytics-4/lambda/tsconfig.json
```diff
@@ -10,6 +10,10 @@
     "inlineSources": true,
     "sourceRoot": "/src"
   },
-  "include": ["src", "test"],
+  "include": [
+    "src",
+    "test",
+    ".eslintrc.js"
```
andipaetzold, line 16: Why .eslintrc?

andipaetzold (changes requested):
Upgrade is fine - just one open question

apps/google-analytics-4/lambda/tsconfig.json
ryunsong-contentful: good catch

PR #3138, opened 2023-04-06
tweaks and fixups for ai-content-generator

ghepting:
## Purpose
- fixes issue with poluted state when changing the action after running a different one
- better error handling and presentation to user


jsdalton (approved):
✅

mgoudy91 (approved):
seems good

PR #3141, opened 2023-04-06
adds make go command

ghepting:
`make go`

### DIFF
--- a/apps/google-analytics-4/lambda/Makefile
+++ b/apps/google-analytics-4/lambda/Makefile
@@ -1,8 +1,8 @@
 .PHONY: all
 all: help
 
-## make start-app: Starts a docker image for dynamodb, our serverless lambda, and frontend for local development
-start-app:
+## make go: Starts a docker image for dynamodb, our serverless lambda, and frontend for local development
+go:
 	@docker compose up --build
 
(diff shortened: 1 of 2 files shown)


ryunsong-contentful (approved):

apps/google-analytics-4/lambda/Makefile
```diff
@@ -5,6 +5,10 @@ all: help
 start-app:
 	@docker compose up --build
 
+## make go: Starts a docker image for dynamodb, our serverless lambda, and frontend for local development
```
ryunsong-contentful, line 8: should we remove start-app then? What's the difference?

ghepting: i wanted to make it backward compatible, but i'd like to :)

matthew-gordon (approved):

matthew-gordon:
:sob: I was hoping this day would come :fist:

apps/google-analytics-4/lambda/Makefile
matthew-gordon: 💣 💥 Nuke it!

ghepting: removed `start-app`

PR #3142, opened 2023-04-06
Integ 385 ux use forma36 skeleton instead of custom loader

lilbitner:
## Purpose & Approach

This PR accomplishes two things: 
1. Replaces the custom loading spinner within the GA4 app with the F36 skeleton components. I tried to represent the underlying actual component size with this loader, but am open to having it more generic. 

Screenshot 2023-04-06 at 11 44 42 AM

2. Replaces the old loading state of the Content Types (jumping dots) with the skeleton component used throughout the rest of the config page. I used the same setup as used with the Property Id section to keep things consistent. 

Screenshot 2023-04-06 at 11 34 43 AM

## Testing steps
Load the Config page as well as the app and notice the loading states 😄


apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
```diff
@@ -186,7 +187,11 @@ const AssignContentTypeSection = (props: Props) => {
           )}
         </>
       ) : (
-        <Spinner variant="primary" />
+        <Flex marginTop="none" fullWidth={true}>
```
ghepting, line 190: is the marginTop of a flex container _not_ `none` by default? 🤔

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.styles.ts
mgoudy91, line 1: 🎉

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
```diff
@@ -84,9 +83,11 @@ const AnalyticsApp = (props: Props) => {
   const renderAnalyticContent = () => {
     if (loading || pendingData) {
       return (
-        <Flex justifyContent="center" alignItems="center" className={styles.wrapper}>
-          <div className={styles.loader}></div>
-        </Flex>
+        <Skeleton.Container svgHeight={300}>
+          <Skeleton.Image width={325} height={68} offsetTop={10} />
+          <Skeleton.Image height={160} width={325} offsetTop={93} />
+          <Skeleton.Image height={25} width={200} offsetTop={268} />
```
ghepting, line 89: nit: order width height attributes consistently

ghepting (approved):

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
lilbitner: Good catch - copied this over from where the exact same component was implemented for the Property section since I wanted it to be consistent. Should have noticed that 👍

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
```diff
@@ -186,7 +187,11 @@ const AssignContentTypeSection = (props: Props) => {
...
+          <Skeleton.Container svgHeight={147}>
```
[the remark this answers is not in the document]
lilbitner: Yeah I do agree, I don't like using magic numbers! Though in this case with the loading containers, this is meant to reflect the exact height of the component that eventually renders, so I don't foresee another solution besides this, since SVG height needs to be pretty explicit? Though I will think on it! 

Luckily the container that eventually renders here is 150px so I can really just remove this here, but this pattern can certainly be seen with the other loading states that have been implemented previously on the page.

apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx
lilbitner: nice, will do!

PR #3144, opened 2023-04-06
[INTEG-336] Fix broken tests in watch mode

mgoudy91:
## Purpose
We have a few GA4 lambda tests tests that fail (incorrectly) when run in watch mode, as a result of unneeded hook stubbing.

## Approach
- Add stubbing for `getServiceAccountKeyFile` returning correctly formatted file.
- Remove stubbing for `getServiceAccountKeyFile` that happens every test to provide more granular and accurate stubbing.
- Quick deprecation fix for sendStatus
image

## Testing steps

google-analytics-/lambda > `npm run test:watch`

- Ensure all tests pass first time as well as subsequent runs

### DIFF
--- a/apps/google-analytics-4/lambda/src/controllers/apiController.ts
+++ b/apps/google-analytics-4/lambda/src/controllers/apiController.ts
@@ -44,7 +44,7 @@ const ApiController = {
         serviceAccountKeyFile
       );
 
-      return res.send(200);
+      return res.sendStatus(200);
     } catch (err) {
       next(err);
     }
--- a/apps/google-analytics-4/lambda/src/app.spec.ts
+++ b/apps/google-analytics-4/lambda/src/app.spec.ts
@@ -1,4 +1,6 @@
+import * as NodeAppsToolkit from '@contentful/node-apps-toolkit';
 import { AnalyticsAdminServiceClient } from '@google-analytics/admin';
+import { BetaAnalyticsDataClient } from '@google-analytics/data';
 import chai, { expect } from 'chai';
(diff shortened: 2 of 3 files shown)


apps/google-analytics-4/lambda/src/controllers/apiController.ts
```diff
@@ -44,7 +44,7 @@ const ApiController = {
         serviceAccountKeyFile
       );
 
-      return res.send(200);
+      return res.sendStatus(200);
```
matthew-gordon, line 47: Just curious if we want to be sending the string representation of the HTTP status code, or if we want to maintain the actual code?

sends status code
```typescript
res.status(200);
```

sends string representation of status code
```typescript
res.sendStatus(200);
```

matthew-gordon:
LGTM, small clarifying question but otherwise ✅ ✨

matthew-gordon (approved):

ryunsong-contentful (approved):

apps/google-analytics-4/lambda/src/controllers/apiController.ts
mgoudy91: image
I think it's just a number?

mgoudy91: image

matthew-gordon: interesting, so short hand for the chainable 🤔 👍 neato! Just curious 🙇

PR #3193, opened 2023-04-12
fix: preview and image and marker position[NONE]

vikaskumr:
## Purpose

We are trying to fix the marker and preview image position, previously marker was able to focus even om the blank area.
After this change marker will not be focused apart from image.

Screenshot 2023-04-12 at 18 57 48

Screenshot 2023-04-12 at 18 49 46

## Approach
We allowed max width and height for image to take as much space as needed within parent div without changing its aspect-ratio.


whitelisab (changes requested):
This looks good to me; I confirmed that it addresses the issues presented in the two related support tickets. You'll just need to update the failing snapshot test, then it should be good to go!

colomolo (approved):

whitelisab (approved):

PR #3215, opened 2023-04-13
[INTEG-426] Fix formatting of sidebar date range

whitelisab:
## Purpose

There was a reported issue for a GA4 app user in a different locale where the Sidebar app wasn't displaying data and the 'Open in Google Analytics' link was showing the wrong date range in the Google Analytics report.

## Approach

Initially we were taking the date, converting it to a locale date string, and then reformatting the date string from there. The problem with this is when the locale date format was not in the 'MM/DD/YYYY', our parsing would generate the wrong date string. To fix this, we are now using methods on the JavaScript Date object to get the year, month, and date, rather than formatting to the locale date string first.

## Testing steps

If you change your locale (changing your language and region on your device) and test the app, you should see that the startDate and endDate query parameters are properly formatted (YYYY-MM-DD) and that the linked report in Google Analytics displays the same date range as the chart.


lilbitner (approved):
LGTM :)

PR #3234, opened 2023-04-17
feat: add a note about default limits []

Rowadz:
## Purpose

This might help people to see that the default limit of the `collections` is `100`

https://user-images.githubusercontent.com/38977667/232449691-3c203ec2-8ecb-41cd-b3bd-4ca40f8fdf1f.mp4

### DIFF
--- a/apps/graphql-playground/src/components/GqlPlayground.tsx
+++ b/apps/graphql-playground/src/components/GqlPlayground.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Provider } from 'react-redux';
-import { Playground, store } from 'graphql-playground-react';
+import { Playground, store, getQuery } from 'graphql-playground-react';
 import stripIndent from 'strip-indent';
+import { Note } from '@contentful/forma-36-react-components';
 
 interface Sys {
   id: String;
@@ -33,6 +34,8 @@ function formatQuery(query: string) {
 
 function GqlPlayground(props: GqlPlaygroundProps) {
   const { cpaToken, entry, spaceId, spaceEnvironment, spaceEnvironmentAlias } = props;
+
+  const [hasCollection, setHasCollection] = useState<boolean>();
 
   const tabConfig = {
     endpoint: `https://graphql.contentful.com/content/v1/spaces/${spaceId}/environments/${
@@ -87,10 +90,36 @@ function GqlPlayground(props: GqlPlaygroundProps) {
 
   const settings = { 'editor.theme': 'light' };
 
+  useEffect(() => {
+    const unsubscribe = store.subscribe(() => {
+      const query: string = getQuery(store.getState());
+      const result = query.toLowerCase().includes('collection');
+      if (hasCollection !== result) {
+        setHasCollection(result);
+      }
+    });
+    return unsubscribe;
+  }, [hasCollection]);
+
   return (
-    <Provider store={store}>
-      <Playground tabs={tabs} settings={settings} fixedEndpoint={true} {...tabConfig} />
-    </Provider>
+    <>
+      {hasCollection && (
+        <Note>
+          Did you know that the default limit for a collection is <b>100</b>, lowering this would
+          decrease your query complexity! [
+          <a
+            target="_blank"
+            rel="noreferrer"
+            href="https://www.contentful.com/developers/docs/references/graphql/#/reference/collection-fields/arguments:~:text=default%20is%200-,limit,to%20fetch.%20The%20default%20is%20100%20and%20the%20maximum%20is%201000,-where">
+            source
+          </a>
+          ]
+        </Note>
+      )}
+      <Provider store={store}>
+        <Playground tabs={tabs} settings={settings} fixedEndpoint={true} {...tabConfig} />
+      </Provider>
+    </>
   );
 }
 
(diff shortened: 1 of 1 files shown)


apps/graphql-playground/src/components/GqlPlayground.tsx
```diff
@@ -83,17 +88,41 @@ function GqlPlayground(props: GqlPlaygroundProps) {
         },
       ];
 
-  const settings = { "editor.theme": "light" };
+  const settings = { 'editor.theme': 'light' };
+
+  useEffect(() => {
+    const unsubscribe = store.subscribe(() => {
+      const query: string = getQuery(store.getState());
+      const result = query.toLowerCase().includes('collection');
```
Rowadz, line 96: Unfortunately, that's the only way I managed to pull the data from [graphql-playground-react](https://github.com/graphql/graphql-playground/tree/main/packages/graphql-playground-react)

They don't expose the codemirror instance not provide another way to get the query value.

apps/graphql-playground/src/components/GqlPlayground.tsx
```diff
@@ -83,17 +88,41 @@ function GqlPlayground(props: GqlPlaygroundProps) {
...
+      if (hasCollection !== result) {
+        setHasCollection(result);
+      }
```
Rowadz, line 99: doing this 
```ts
setHasCollection(query.toLowerCase().includes('collection'));
```

will cause an infinite re-render.

apps/graphql-playground/src/components/GqlPlayground.tsx
```diff
@@ -87,10 +90,36 @@ function GqlPlayground(props: GqlPlaygroundProps) {
...
-    </Provider>
+    <>
+      {hasCollection && (
+        <Note>
+          Did you know that the default limit for a collection is <b>100</b>, lowering this would
+          decrease your query complexity! [
+          <a
+            target="_blank"
+            rel="noreferrer"
+            href="https://www.contentful.com/developers/docs/references/graphql/#/reference/collection-fields/arguments:~:text=default%20is%200-,limit,to%20fetch.%20The%20default%20is%20100%20and%20the%20maximum%20is%201000,-where">
```
andipaetzold, line 113: That link should be shortened to `https://www.contentful.com/developers/docs/references/graphql/#/reference/collection-fields/arguments`

apps/graphql-playground/src/components/GqlPlayground.tsx
andipaetzold: Even if you remove `hasCollection` from the dependencies?

apps/graphql-playground/src/components/GqlPlayground.tsx
```diff
@@ -87,10 +90,36 @@ function GqlPlayground(props: GqlPlaygroundProps) {
...
+  useEffect(() => {
+    const unsubscribe = store.subscribe(() => {
+      const query: string = getQuery(store.getState());
+      const result = query.toLowerCase().includes('collection');
+      if (hasCollection !== result) {
+        setHasCollection(result);
+      }
+    });
+    return unsubscribe;
+  }, [hasCollection]);
```
andipaetzold, line 102: `hasCollection` should be removed from the dependencies here. We currently unsubscribe and resubscribe whenever `hasCollection` changes. We actually only want to subscribe on mount and unsubscribe on unmount.

apps/graphql-playground/src/components/GqlPlayground.tsx
```diff
@@ -87,10 +90,36 @@ function GqlPlayground(props: GqlPlaygroundProps) {
...
+    <>
+      {hasCollection && (
+        <Note>
+          Did you know that the default limit for a collection is <b>100</b>, lowering this would
+          decrease your query complexity! [
+          <a
+            target="_blank"
+            rel="noreferrer"
+            href="https://www.contentful.com/developers/docs/references/graphql/#/reference/collection-fields/arguments:~:text=default%20is%200-,limit,to%20fetch.%20The%20default%20is%20100%20and%20the%20maximum%20is%201000,-where">
+            source
```
andipaetzold, line 114: "source" feels quite technical. Maybe "Learn more" without the square brackets?

andipaetzold (changes requested):

apps/graphql-playground/src/components/GqlPlayground.tsx
Rowadz: actually I can just do this 

```ts
  useEffect(() => {
    const unsubscribe = store.subscribe(() => {
      const query: string = getQuery(store.getState());
      setHasCollection(query.toLowerCase().includes('collection'));
    });
    return unsubscribe;
  }, []);
```

maybe I did something else that caused the infinite re-render.

andipaetzold: Yes, that should work

apps/graphql-playground/src/components/GqlPlayground.tsx
```diff
@@ -93,27 +93,23 @@ function GqlPlayground(props: GqlPlaygroundProps) {
...
           Did you know that the default limit for a collection is <b>100</b>, lowering this would
-          decrease your query complexity! [
+          decrease your query complexity!{' '}
           <a
             target="_blank"
             rel="noreferrer"
-            href="https://www.contentful.com/developers/docs/references/graphql/#/reference/collection-fields/arguments:~:text=default%20is%200-,limit,to%20fetch.%20The%20default%20is%20100%20and%20the%20maximum%20is%201000,-where">
-            source
+            href="https://www.contentful.com/developers/docs/references/graphql/#/reference/collection-fields/arguments">
+            learn more
```
andipaetzold, line 111: Maybe capitalize "Learn more". Not sure whether lowercase looks good here.

andipaetzold (approved):

PR #3251, opened 2023-04-18
Adds run in band for CI test

ryunsong-contentful:

### DIFF
--- a/apps/google-analytics-4/frontend/package.json
+++ b/apps/google-analytics-4/frontend/package.json
@@ -28,7 +28,7 @@
     "start": "cross-env BROWSER=none react-scripts start",
     "build": "REACT_APP_RELEASE=$(git rev-parse --short HEAD) react-scripts build",
     "test": "react-scripts test",
-    "test:ci": "CI=true react-scripts test",
+    "test:ci": "CI=true react-scripts test --runInBand",
     "lint": "eslint --max-warnings=0 .",
     "eject": "react-scripts eject",
     "create-app-definition": "contentful-app-scripts create-app-definition",


matthew-gordon (approved):
LGTM ❇️ might also consider setting `maxWorkers` in the future to avoid run up but thats for a different day 💪

LiamStokingerContentful:
> LGTM ❇️ might also consider setting `maxWorkers` in the future to avoid run up but thats for a different day 💪

Interestingly enough, `runInBand` and `maxWorkers=1` are functionally the same thing. So yeah if we want to incrementally increase the number of workers, we can do so.

```
export default function getMaxWorkers(
  argv: Partial<
    Pick<Config.Argv, 'maxWorkers' | 'runInBand' | 'watch' | 'watchAll'>
  >,
  defaultOptions?: Partial<Pick<Config.Argv, 'maxWorkers'>>,
): number {
  if (argv.runInBand) {
    return 1;
  } else if (argv.maxWorkers) {
    return parseWorkers(argv.maxWorkers);
  } else if (defaultOptions && defaultOptions.maxWorkers) {
    return parseWorkers(defaultOptions.maxWorkers);
  } else {
    // In watch mode, Jest should be unobtrusive and not use all available CPUs.
    const numCpus = getNumCpus();
    const isWatchModeEnabled = argv.watch || argv.watchAll;
    return Math.max(
      isWatchModeEnabled ? Math.floor(numCpus / 2) : numCpus - 1,
      1,
    );
  }
}
```

https://github.com/facebook/jest/blob/main/packages/jest-config/src/getMaxWorkers.ts#L21

PR #3256, opened 2023-04-19
chore: update app scripts versions in js and ts templates

martin3walker:
We need the latest version of app scripts available in the ts and js templates in order to automatically create built app actions in the connected app definition.


loweisz:
No lockfile for the TS template?

loweisz (approved):

PR #3258, opened 2023-04-19
Proof of Concept: unit test run without building whole repo

LiamStokingerContentful:
## Purpose

This is a POC which shows that it is possible to hack together a solution where the unit tests for the repo can be run without triggering a build for the entire repo.

## Approach

What I do here is simply build the Smartling frontend package as a prerequisite for running the Smartling lambda tests. This allows the tests to run properly without triggering a build for the entire codebase.

## Deployment

~This PR has stripped actual building of the code, that will need to be added as a separate build step~
This should be now addressed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,9 +48,6 @@ jobs:
           name: Verify project configurations
           command: npm run verify-config
       - run:
-          name: Build
-          command: npm run build
-      - run:
           name: Test
           command: npm run test
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,22 @@ jobs:
           name: Test
           command: npm run test
 
+  build:
+    docker:
+      - image: cimg/base:stable
+    resource_class: medium+
+    steps:
+      - checkout
+      - node/install
+      - run:
+          name: Install dependencies
+          command: |
+            npm ci
+            npm run bootstrap:ci
+      - run:
+          name: Build
+          command: npm run build
+
   deploy-staging:
     docker:
       - image: cimg/base:stable
@@ -170,10 +186,14 @@ workflows:
       - apps-test:
           context:
             - vault
+      - build:
+          context:
+            - vault
       - deploy-staging:
           context:
             - vault
           requires:
+            - build
             - apps-test
           filters:
             branches:
@@ -184,6 +204,7 @@ workflows:
           context:
             - vault
           requires:
+            - build
             - apps-test
           filters:
             branches:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -47,13 +47,28 @@ jobs:
...
           command: npm run test
 
+  build:
+    docker:
+      - image: cimg/base:stable
+    resource_class: medium+
+    steps:
+      - checkout
+      - node/install
+      - vault/get-secrets:
```
LiamStokingerContentful, line 61: Is getting secrets necessary here? I have limited context, so am just curious

apps/smartling/lambda/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "rimraf built && tsc",
     "lint": "tslint --project ./tsconfig.json",
     "test": "jest --watch",
-    "test:ci": "jest",
+    "test:ci": "cd ../frontend && npm run build && cd ../lambda && jest",
```
ryunsong-contentful, line 10: nice (not too ugly and readable tbh)

ryunsong-contentful:
Everything related to the changes itself looks great and love seeing!
Screenshot 2023-04-19 at 2 27 08 PM

Only comment is with one awkwardness that I see with the Pipeline now that the build is separated from the tests.
The `lint` and `prettier` commands only run in the tests when I feel like they should now be separated (lint and prettier together) as well so that it gets applied to the build and tests. It's not for performance gains as they would be minimal, this is just for semantics and organizations so that it's obvious people don't think tests are special in any way as far as formatting goes. Maybe this is out of scope of the PR/task but just food for thought at least.

.circleci/config.yml
jsdalton: Not necessary for this particular step.

apps/smartling/lambda/package.json
jsdalton: Haha not sure about "not too ugly" but the value it adds in decoupling the whole big build for the tests is probably worth it for me.

I mean I might make a small suggestion to make it a bit less hacky by moving this out to a separate task (also to add a bit of clarity) i.e.:

```
"build-frontend": "cd apps/smartling/frontend && npm run build"
"test:ci": "npm run build-frontend && jest"
```

(btw I didn't test to see if you need to cd back in)

jsdalton (commented):
Yeah I love the benefit, unlocks some other improvements as well! 💯 

I'm willing to swallow the hack for now 😄 

I had a small suggestion or two to make that pill a bit easier to swallow though.

PR #3265, opened 2023-04-20
chore: include hosted app action templates

martin3walker:
Hosted app action templates, which we will write into the templates used by cca if a user wants to include a hosted app action.


examples/hosted-app-action-templates/javascript/build-actions.js
```diff
@@ -0,0 +1,28 @@
...
+    const actionsFiles = readdirSync('actions', {
+      encoding: 'utf-8',
+    });
+
+    actionsFiles.forEach((fileName) => {
+      const action = readFileSync(`actions/${fileName}`, {
+        encoding: 'utf-8',
+      }).toString();
+
+      writeFileSync(`build/actions/${fileName}`, action);
```
kdamball, line 18: Can we do this synchronously? It will be faster

martin3walker: Do you mean asynchronously?

examples/hosted-app-action-templates/javascript/build-actions.js
```diff
@@ -0,0 +1,33 @@
...
+
+    await Promise.all(
+      actionsFiles.map((fileName, index) => {
+        return writeFile(`build/actions/${fileName}`, actions[index]);
+      })
+    );
+
+    console.log('App actions successfully built');
+  } catch (e) {
+    console.log(e);
```
kdamball, line 29: You might want to throw an actual error that building failed. Otherwise people will be confused why nothing is showing up

martin3walker: Done 👍

examples/hosted-app-action-templates/javascript/build-actions.js
```diff
@@ -0,0 +1,34 @@
...
+    await Promise.all(
+      actionsFiles.map((fileName, index) => {
+        return writeFile(`build/actions/${fileName}`, actions[index]);
+      })
+    );
+
+    console.log('App actions successfully built');
+  } catch (e) {
+    console.log('Error building app actions');
+    throw Error(e);
```
kdamball, line 30: `throw e` is good enough

kdamball: Ah - then `throw new Error(e)`?

PR #3282, opened 2023-04-21
[INTEG-445] Chore: Update dependencies for Smartling lambda

whitelisab:
## Purpose

Updated packages in Smartling lambda that contained minimist, qs, and simple-git in order to address security vulnerabilities.

## Approach

**Minimist**
This is a dependency of mkdirp, which is a dependency of TSLint. Although TSLint is deprecated, I decided to bump the version of TSLint to the most recent version for the scope of this PR, which resolves Minimist to version 1.2.8. We can address moving from TSLint to ESLint in follow up work (there are other apps that use TSLint, too).

**qs**
This is a dependency in a few packages. I bumped versions of the following packages: qs, serverless-express, body-parser, express, superagent, and supertest. The security vulnerability is resolved in qs >6.10.3 and backported to some versions (https://github.com/advisories/GHSA-hrpp-h998-j3pp).

**simple-git**
This is a dependency of @serverless/dashboard-plugin, which is a dependency of serverless. The fixed version of simple-git is updated in [version 6.1.6](https://github.com/serverless/dashboard-plugin/releases?q=simple-git&expanded=true) of @serverless/dashboard-plugin. However, we would need to upgrade to v3 of serverless to update this dependency. So instead I added an override to the fixed version of @serverless/dashboard-plugin until we can upgrade serverless.

### DIFF
--- a/apps/smartling/lambda/package.json
+++ b/apps/smartling/lambda/package.json
@@ -35,9 +35,12 @@
     "supertest": "^4.0.2",
     "ts-jest": "^27.1.5",
     "ts-node": "^10.9.1",
-    "tslint": "^5.20.1",
+    "tslint": "^6.1.3",
     "tslint-config-prettier": "^1.18.0",
     "tslint-config-standard": "^9.0.0",
     "tslint-no-unused-expression-chai": "^0.1.4"
+  },
+  "overrides": {
+    "@serverless/dashboard-plugin": "6.1.6"
   }
 }
--- a/apps/smartling/lambda/package-lock.json
+++ b/apps/smartling/lambda/package-lock.json
@@ -29,7 +29,7 @@
         "supertest": "^4.0.2",
         "ts-jest": "^27.1.5",
         "ts-node": "^10.9.1",
-        "tslint": "^5.20.1",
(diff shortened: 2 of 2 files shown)


ghepting (approved):

apps/smartling/lambda/package.json
```diff
@@ -35,9 +35,12 @@
...
     "ts-jest": "^27.1.5",
     "ts-node": "^10.9.1",
-    "tslint": "^5.20.1",
+    "tslint": "^6.1.3",
     "tslint-config-prettier": "^1.18.0",
     "tslint-config-standard": "^9.0.0",
     "tslint-no-unused-expression-chai": "^0.1.4"
+  },
+  "overrides": {
+    "@serverless/dashboard-plugin": "6.1.6"
```
ghepting, line 44: as discussed, can we make sure we have a TIMEBOMB backlog ticket to upgrade these lambdas to serverless >= 3 and get rid of this [risky] force override?

whitelisab: Yes, just added it to the backlog!

ryunsong-contentful (approved):

PR #3292, opened 2023-04-24
[INTEG-446] Chore: Update dependencies for Typeform lambda

whitelisab:
## Purpose

Updated packages in Typeform lambda that contained the qs package in order to address security vulnerabilities.

## Approach

The qs package is a dependency in a few packages. I bumped versions of the following packages: qs, body-parser, and express. The security vulnerability is resolved in qs >6.10.3 and backported to some versions (https://github.com/advisories/GHSA-hrpp-h998-j3pp).

### DIFF
--- a/apps/typeform/lambda/package-lock.json
+++ b/apps/typeform/lambda/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "aws-serverless-express": "3.3.6",
-        "express": "4.17.1",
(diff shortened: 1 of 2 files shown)


mgoudy91 (approved):
Works for me, though I am confused we aren't also bumping express to like "^4.x.y" where 4.x.y is whatever version will pull in the right deps (appears to be 4.18.2)

whitelisab:
@mgoudy91 yup, I can update that!

PR #3304, opened 2023-04-25
DRAFT: Update Mux dependencies

lilbitner:


whitelisab (approved):
I also tested everything locally and this looks good to me! It's strange that it does accept that CSS variable according to the documentation, but this looks like a good resolution to me! ⭐

PR #3318, opened 2023-04-26
fix: Dependabot/npm and yarn/contentful/field editor test utils 0.17.3 wistia downgrade

lilbitner:
## Purpose

This PR resolves two issues found [this](dependabot/npm_and_yarn/contentful/field-editor-test-utils-0.17.3) dependabot PR: 
1. This fixes the test error that occurs within the `graphql-playground` app. 
2. This reverses the dependabot incurred downgrade found in the `wistia` `package.json` with the `field-editor-test-utils` package. 

**To clarify, this PR includes the changes from the original dependabot PR, but with the appropriate changes made.**
 
## Approach

1. In order to resolve the test error found in the PR, I installed the latest version of the `field-editor-test-utils` package within the `graphql-playground` `package.json`. This update resolved the peer dependency issues. I would prefer to test that everything is as it should be locally with the `graphql-playground` app, however the local development version is fairly broken, with maintainability work that is out of scope of this update. (this has been communicated to Extensibility and Integration PMs). I am trusting that updating to the latest version of the package should cause little to no change, but am open to holding off if others feel differently. 

2. I also manually bumped the version of the `wistia` package of `field-editor-test-utils` to the latest version. Can confirm all is working is working with that app.

I limited the scope of changes to those two adjustments ^, so as not to make any extraneous adjustments for this PR.

### DIFF
--- a/apps/graphql-playground/package-lock.json
+++ b/apps/graphql-playground/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.2.11",
       "dependencies": {
         "@contentful/field-editor-single-line": "^0.16.0",
-        "@contentful/field-editor-test-utils": "^0.8.2",
+        "@contentful/field-editor-test-utils": "^0.17.3",
         "@contentful/forma-36-fcss": "0.3.5",
         "@contentful/forma-36-react-components": "^3.100.7",
         "@contentful/forma-36-tokens": "^0.11.2",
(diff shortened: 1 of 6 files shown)


matthew-gordon (approved):
LGTM 🌮 ✨ Thank you for the extra detail around `graphql-playground` as well 🙏

PR #3329, opened 2023-04-27
refactored to unified sentry release job

ghepting:
## Purpose
We don't want each app responsible for the shared sentry project release, so this moves it from the ad hoc post-deploy phase that lerna runs to a single job in the deploy-prod workflow


mgoudy91 (approved):
🎉

PR #3336, opened 2023-04-28
Fix Mux specific @typescript-eslint Dependabot updates 

lilbitner:
## Purpose

The purpose of this PR is to remedy a build error found in these following `dependabot` initiated PRs: 

1 . https://github.com/contentful/apps/pull/3301 : @typescript-eslint/eslint-plugin update (changes incorporated into this PR).
2. https://github.com/contentful/apps/pull/3302 : @typescript-eslint/parser update (build error will be fixed within this open PR).

## Approach

The build error occurring in both PRs, specifically when building the Mux app, is the following error: 

`Failed to load config "react-app" to extend from.`

After investigating the issue, I believe it is due to an incompatibility of the `eslint-config-react-app` package, which is found as part of the outdated `react-scripts` package within the Mux app. Therefore, my first attempt at fixing this issue was simply to update the `react-scripts` package, however this seemed to bring up another incompatibility error found between the `react-scripts` package and the `mux-uploader-react` package, hence why this is a very Mux specific issue. This latest incompatibility causes another build error. Therefore I decided to place efforts in remedying the original build error.  

The most popular solution for the original error, as per others who have encountered the same error, is to simply remove the `eslintConfig` within the package.json. 

Therefore, I did decide to do this, but for the following reasons, with the intention of preserving `eslint` standards: 
1. To separate the `eslint` configuration into a separate file (notice the `.eslintrc.js` file addition), which is a pattern we follow in several other apps. I believe this to be a good abstraction of configs. 
2. To begin implementing `eslint` configurations for this Mux app. It seems there is no significant implementation of `eslint` within the app aside from the original extending of` react-app`, which is not an optimal pattern. A potentially simpler solution to this build error would have been to remove the typescript `eslint` dependencies all together within this app, but again, I think that might be a slight regression. 

You might notice that there is no extending of `eslint `rules within the new config file. I opted to remove those, as adding them revealed many `eslint` violations in the Mux app, which seems out of scope in this PR. I would like to add the following standard config to the app in a follow-up task, accompanied by remedied `eslint` errors: 

`extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended']`

## Testing steps

Pull down this PR and test that the Mux app still work as expected within the Mux (development) space of the Apps - Development and Testing organization. 

(I have done this myself and all is well on my end!)

[Standard eslint configuration](https://typescript-eslint.io/getting-started)

### DIFF
--- a/apps/mux/.eslintrc.js
+++ b/apps/mux/.eslintrc.js
@@ -1,0 +1,8 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
+  plugins: ['@typescript-eslint'],
+  root: true,
+  env: {
+    node: true,
+  },
+};
--- a/apps/ecommerce/lambda/package-lock.json
+++ b/apps/ecommerce/lambda/package-lock.json
@@ -2641,15 +2641,15 @@
       }
     },
     "node_modules/@typescript-eslint/eslint-plugin": {
-      "version": "5.59.0",
-      "resolved": "https://registry.npmjs.org/@typescript-eslint/eslint-plugin/-/eslint-plugin-5.59.0.tgz",
-      "integrity": "sha512-p0QgrEyrxAWBecR56gyn3wkG15TJdI//eetInP3zYRewDh0XS+DhB3VUAd3QqvziFsfaQIoIuZMxZRB7vXYaYw==",
+      "version": "5.59.1",
+      "resolved": "https://registry.npmjs.org/@typescript-eslint/eslint-plugin/-/eslint-plugin-5.59.1.tgz",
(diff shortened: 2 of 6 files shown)


apps/mux/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
rusticpenguin, line 1: love moving this into it's own file!

rusticpenguin (approved):
LGTM!

ghepting (approved):

ghepting:
> You might notice that there is no extending of eslint rules within the new config file. I opted to remove those, as adding them revealed many eslint violations in the Mux app, which seems out of scope in this PR. I would like to add the following standard config to the app in a follow-up task, accompanied by remedied eslint errors:
> extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended']

💯

PR #3352, opened 2023-05-01
Feature INTEG-498: Added static response JSON to new /product endpoint

pkeavenycontentful:
## Purpose

For the purposes of scaffolding, we'll need an API endpoint that returns a static JSON to unblock the frontend. 

## Approach

I just put a `/product` endpoint with an `:id` for future considerations but the id does nothing at this point.


rusticpenguin (approved):
Perfect!

apps/ecommerce/lambda/src/routers/apiRouter.ts
```diff
@@ -3,5 +3,6 @@ import { ApiController } from '../controllers';
 const app = Router();
 
 app.get('/ping', ApiController.ping);
+app.get('/product/:id', ApiController.product);
```
ghepting, line 6: again here, make this `resource` or something generic.

PR #3371, opened 2023-05-04
Integ 500 proxy provider

pkeavenycontentful:
## Purpose

For the purposes of scaffolding, we're adding the ability to call a provider with mock data to save us some time when we hook up the real provider logic down the road. 

## Approach

- Per Gary's advice, it makes sense to have an http call to a provider for the moment, since we will likely have to make an http call anyway (say to SAP's public product URLs)

## Testing steps

I set up a Postman collection for testing purposes: https://grey-spaceship-425061.postman.co/workspace/New-Team-Workspace~4ca0371f-88c4-46de-b128-da67c31e4460/collection/26682020-0fc91c3a-a49c-4aaa-a155-824d6f87f6b1?action=share&creator=26682020

Feel free to use my collection for testing or copy into your own collection to test use cases 👍 

- Axios for http calls (seems like Fetch API is available, but haven't had any luck getting it to work: https://nodejs.org/dist/latest-v19.x/docs/api/globals.html#fetch ) 
- Nock for mocking http calls in test suites


apps/ecommerce/lambda/src/app.ts
```diff
@@ -33,6 +33,13 @@ app.use(express.static('public'));
 
 app.use('/api', ApiRouter);
 
+app.use('/shopify', ShopifyRouter);
+
+// for Postman calls in development. This bypasses request signing if using Postman to call localhost
+if (config.environment === 'development' || config.environment === 'test') {
+  app.use('/', ApiRouter);
+}
```
ghepting, line 41: don't commit this

apps/ecommerce/lambda/src/controllers/apiController.ts
```diff
@@ -1,25 +1,33 @@
...
+    res: Response<HydratedResourceData | { error: any; message: string }>
+  ) => {
+    const body: ResourceLink = req.body;
+    const BASE_URL = config.baseUrl;
+    try {
+      const url = `${BASE_URL}/${body.sys.provider.toLowerCase()}/resource/${body.sys.linkType}/${
+        req.params.id
+      }`;
+
+      const response: AxiosResponse<HydratedResourceData> = await axios.post(url, body);
```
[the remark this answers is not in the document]
ghepting: @pkeavenycontentful i think we can more of less hard code the URL to the proxy to avoid the VULN introduced here that CodeQL is flagging. Something like a mapping of providers to proxy URL 


```
const PROVIDERS = {
  shopify: `${config.baseUrl}/shopify/resource`
}
```

and then `axios.post(PROVIDERS[req.body.sys.provider.toLowerCase(), req.body)` perhaps? we can toy around with this if we want to, just spit balling here

apps/ecommerce/lambda/src/controllers/shopifyController.ts
```diff
@@ -0,0 +1,21 @@
...
+  resource: (
+    req: Request<ResourceLink>,
+    res: Response<CombinedResource>
+  ): Response<CombinedResource> => {
+    return res
+      .send({
+        sys: req.body.sys,
+        ...mockResourceData,
+      })
+      .status(200);
```
ghepting, line 17: this part is default actually! can drop the explicit `status(200)` here :)

apps/ecommerce/lambda/src/routers/apiRouter.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        .request(app)
+        .post('/resource/product/1234')
+        .send({ sys: newResourceLink })
+        .end((error, res) => {
+          expect(res).to.have.status(500);
+          done();
+        });
+    });
+  });
+});
```
ghepting, line 63: 💪 nice

apps/ecommerce/lambda/src/routers/apiRouter.ts
```diff
@@ -1,8 +1,8 @@
 import { Router } from 'express';
-import { ApiController } from '../controllers';
+import { ApiController, ShopifyController } from '../controllers';
 const app = Router();
 
 app.get('/ping', ApiController.ping);
-app.get('/product/:id', ApiController.product);
+app.post('/resource/:resourceType/:id', ApiController.resource);
```
ghepting, line 6: my thoughts here: we should just drop these path params since all of the data we need is in the post body. it feels too redundant

apps/ecommerce/lambda/src/config.ts
```diff
@@ -26,4 +26,5 @@ export const config = {
   sentryDSN: getEnvironmentVariable('SENTRY_DSN'),
   environment: getEnvironmentVariable('NODE_ENV'),
   release: getEnvironmentVariable('CIRCLE_SHA1', 'no-release-hash-set'),
+  baseUrl: getEnvironmentVariable('BASE_URL'),
```
jsdalton, line 29: For now this work but it brings up that we eventually need to retrieve this data from the provider installation itself. That is, if you installed Magento app it should find a different provider base URL than Shopify, i.e. to point to those different provider instances.

apps/ecommerce/lambda/src/controllers/apiController.ts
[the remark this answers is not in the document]
ghepting: - [ ] refactor to avoid SSRF vuln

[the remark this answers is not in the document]
jsdalton: eventually we can move this to a client wrapper which handles the params etc. so i'd support whatever easy fix allows us to avoid the vulneratiblity.

apps/ecommerce/lambda/src/routers/apiRouter.ts
jsdalton: true, i agree. to my thinking since we really need to take a step back and consider the API design (endpoints, methods and schemas together) we could punt on this until later.

jsdalton (approved):
Looks good to me for now, aside from fixing the vulnerability thing. As a scaffold it already exposes some questions and considerations, which is a good thing for me. But I suggest we address most of my questions in a subsequent PR.

apps/ecommerce/lambda/src/controllers/apiController.ts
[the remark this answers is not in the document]
ghepting: yep absolutely

apps/ecommerce/lambda/src/config.ts
ghepting: yep, absolutely

apps/ecommerce/lambda/src/app.ts
```diff
@@ -26,13 +26,16 @@ app.use(Sentry.Handlers.requestHandler() as express.RequestHandler);
 app.use(apiRouteConstraint, cors(corsConfig));
 
 // verify signed requests on /api/* routes
-app.use(apiRouteConstraint, Middleware.verifiySignedRequests);
+// to test endpoints with Postman, comment out below
+// app.use(apiRouteConstraint, Middleware.verifiySignedRequests);
```
ghepting, line 30: - [ ] uncomment this before merging

PR #3379, opened 2023-05-04
add error boundary with retry button

rusticpenguin:
## Purpose

This will catch an errors in the component try and display a debug editor for directly manipulating the JSON

## Approach

Currently the only way to properly handle these errors in React is through using class components. This was written in a way so children components can still be functional components use hooks. 

In a future where it's _finally_ added, we can swap this class' logic out with a functional component


apps/ecommerce/frontend/src/helpers/fakeFetch.ts
```diff
@@ -0,0 +1,28 @@
+import { HydratedResourceData, ResourceLink } from '../types';
```
rusticpenguin, line 1: Accidentally added this file to the commit. Whoops! But I will be adding this later. It's just abstracting out that fake data call

apps/ecommerce/frontend/src/components/FieldFallback.tsx
```diff
@@ -0,0 +1,28 @@
+import { Box, Button, Note } from '@contentful/f36-components';
+import { Field as DefaultField } from '@contentful/default-field-editors';
+import { useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
+import { FieldAppSDK } from '@contentful/app-sdk';
+import { ErrorComponentProps } from '../types';
+import tokens from '@contentful/f36-tokens';
+
+const FieldFallback = ({ error, resetErrorHandler }: ErrorComponentProps) => {
+  const sdk = useSDK<FieldAppSDK>();
+  useAutoResizer();
```
rusticpenguin, line 10: This is a fix for if the field loads and instantly errors. This will make sure the field fully resizes to the fallback's height

ghepting (approved):

apps/ecommerce/frontend/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,39 @@
+import { Component, ErrorInfo } from 'react';
+import type { ErrorBoundaryState, ErrorBoundaryProps } from '../types';
+
+class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
+  constructor(props: ErrorBoundaryProps) {
+    super(props);
+  }
```
ghepting, line 7: accordingly to eslint, i think this whole constructor can go away 😄

apps/ecommerce/frontend/src/types.ts
```diff
@@ -55,3 +56,18 @@ export interface ResourceCardProps {
   onMoveToTop?: Function;
   onMoveToBottom?: Function;
 }
+
+type ErrorBoundaryErrored = { hasError: true; error: Error; info: ErrorInfo };
+type ErrorBoundaryStandby = { hasError: false; error: null; info: null };
```
pkeavenycontentful, line 61: Since these are types, is there a reason to set explicit values like `true/false/null` instead of `boolean` or `unknown`?

pkeavenycontentful (approved):

apps/ecommerce/frontend/src/types.ts
rusticpenguin: Think of these types more as "states" of the component

Give a scenario where we are in an errored state, we can assume that `hasError` is true, this let's typescript know that error and info will also be defined. 

Since there's only two possible states this component could be in, this removes the need to if check both error/info or explicitly cast it as defined `eg. error!`

PR #3398, opened 2023-05-05
Extend linting to mux app

lilbitner:
## Purpose

The purpose of this PR is to follow-up the addition of an .`eslintrc.js` file to the Mux app code (see [this](https://github.com/contentful/apps/pull/3336) PR for context there.), with an extension of `eslint` itself. Therefore enforcing recommended `eslint` rules. 

## Approach

I wanted to be as minimally disruptive to active code as possible, while also providing basic `eslint` guidelines and enforcment (and not ignoring too many rules).

Therefore in files and on specific lines you will notice I disabled `eslint` rules where I felt that changing the code may cause unintended consequences.

I also ignored rules that I felt were rampant throughout the code, and could not be safely adjusted without unforeseen consequences. 

Open to any and all suggestions/opinions on the calls I made here! 

## Testing steps

Test that the Mux app still works locally with this branch. It worked locally for me. (The Mux (development) environment within the testing org is a good place for this).

### DIFF
--- a/apps/mux/src/components/playercode.tsx
+++ b/apps/mux/src/components/playercode.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Textarea, CopyButton, TextLink, Box } from '@contentful/f36-components';
 
 interface GetPlayerCodeProps {
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
   params: any;
 }
 
@@ -46,8 +47,7 @@ class PlayerCode extends React.Component<GetPlayerCodeProps, {}> {
           <TextLink
             href="https://docs.mux.com/guides/video/mux-player"
             target="_blank"
-            rel="noopener noreferrer"
-          >
+            rel="noopener noreferrer">
             Read more about Mux Player
           </TextLink>
         </p>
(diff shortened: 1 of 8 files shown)


apps/mux/src/components/playercode.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Textarea, CopyButton, TextLink, Box } from '@contentful/f36-components';
 
 interface GetPlayerCodeProps {
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
```
matthew-gordon, line 5: instead of `any` would it be better to cast this to something like

```typescript
interface GetPlayerCodeProps {
  params: { [key: string]: string | boolean }
}
```
understand if this isn't warranted

matthew-gordon (approved):
LGTM ✅ just questions around potential type casting instead of many ignore comments or explicit `any`

apps/mux/src/components/playercode.tsx
lilbitner: Yeah I agree that any types are not ideal. For the scope of this PR, I didn't want to adjust too many things, hence why I made the call to just ignore this line (the rest of the app is enforced with 'no any types'). I played around with the types of this `params` props itself, and I noticed that it is similar to what you suggest as an array of objects, but am worried about any variation not accounted for - unsure if original developer placed the any type to account for variation or not. 🤷‍♀️

PR #3413, opened 2023-05-08
Feature INTEG-516: Add separate, shareable SAP app with AIR header added

pkeavenycontentful:
## Purpose

As part of the partner program with SAP, Contentful needs to provide a header for requests made to SAP to verify that we are part of the partner program. This header contains an ID that identifies Contentful as a partner. 

This is a copy of the original SAP connector with the header added. 

## Approach

Using the header in a real environment causes CORs errors, which our team and their team have been unable to find the root problem for. SAP is fine with having a separate app with the header added for verification purposes. 

## Testing steps

1. Login to Contentful, access the following Space: https://app.contentful.com/spaces/rp6w4ccbjqs5
2. Select or add an entry, hit the Select Products button
3. In the Network tab, verify that the header `Application-interface-key: {keyID}` is added to any calls to `.commerce.ondemand.com` 
4. NOTE: the call will fail due to CORs errors. We're unable to figure out why this is happening. 

## Deployment

- I added the new app definition id based on the app I created in the testing space. Let me know if this isn't the right approach and I can adjust


matthew-gordon (approved):
LGTM ❇️ ✨ 💪

ryunsong-contentful (approved):

ghepting (approved):
rubber stamping this since it's a massive copy/paste

PR #3434, opened 2023-05-09
Feat/support magento

ghepting:
## Purpose
Adds support for Magento and faux Magento backend proxy service

### DIFF
--- a/apps/ecommerce/frontend/src/components/MultipleResources.tsx
+++ b/apps/ecommerce/frontend/src/components/MultipleResources.tsx
@@ -1,12 +1,13 @@
 import { useEffect, useState } from 'react';
 import { FieldAppSDK } from '@contentful/app-sdk';
 import { /* useCMA, */ useSDK } from '@contentful/react-apps-toolkit';
-import { ExternalResourceLink } from '../types';
+import { ExternalResourceLink, ProviderLabel } from '../types';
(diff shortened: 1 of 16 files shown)


apps/ecommerce/frontend/src/components/MultipleResources.tsx
```diff
@@ -40,10 +41,10 @@ const MultipleResources = () => {
 
   const mockValue: ExternalResourceLink = {
     sys: {
-      urn: 'gid://products/8191006998814',
+      urn: crypto.randomUUID(),
```
jsdalton, line 44: I hesitate to care about this in the context of this PR, but I don't like the pattern of envisioning a UUID or other opaque string here. It's too early to formalize it but let's at least start off in the mock value with something that mimics what we eventually expect to sea, even e.g

```
`http://api.somemagento.com/products/${crypto.randomUUID()}`
```

jsdalton (approved):
Just one small comment which I feel medium strong about, but not enough to hold up the PR at this scaffolding stage.

apps/ecommerce/frontend/src/components/MultipleResources.tsx
ghepting: yeah this is 100% just for generalization of the mock data purposes currently. definitely not wanting to keep this around either!

PR #3436, opened 2023-05-09
[Fix] Deploy staging and prod works now

ryunsong-contentful:
## Purpose
Apps don't deploy to staging and production because the --since option doesn't work once the PR is merged. So a separate command is needed at the root level for the apps to deploy


mgoudy91 (approved):
Just noting here that we should probably fine a better solution long term but this looks good to unblock deploys for now

PR #3451, opened 2023-05-10
refactors proxy base url to come from app def instance params

ghepting:
## Purpose
refactors backend to receive proxy base URL from app def instance parameters instead of being hard coded. this change is complimentary to the app definition configuration changed made to shopify and magento app def instance parameter configurations

Screenshot 2023-05-10 at 9 21 19 AM

Screenshot 2023-05-10 at 9 21 13 AM


apps/ecommerce/lambda/src/controllers/apiController.ts
```diff
@@ -21,20 +15,34 @@ const ApiController = {
...
+      const proxyResourceUrl = new URL(proxyBaseUrl);
+      proxyResourceUrl.pathname += `/resource`;
+
+      // prevent SSRF exploits
+      // TODO: refactor this to get the proxy resource url from the app definition
+      // configuration parameters once this is an internal Contentful service
+      if (proxyResourceUrl.hostname !== config.domainName) {
+        // temporary: only support data provider proxying on the same domain to avoid SSRF exploits
+        throw new Error(`Invalid data provider base URL. Hostname must be ${config.domainName}`);
+      }
```
ghepting, line 40: @jsdalton PTAL at the temporary SSRF exploit remediation strategy. Open to better ideas, but I think we have to make this an internal Contentful service

PR #3452, opened 2023-05-10
adds pre-push lint git hook

ghepting:
## Purpose
adds git pre-push hook to lint changed files

NOTE: developers MUST configure git local config to execute the git hooks!

```
git config --local core.hooksPath .githooks/
```

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+# run lint on changed files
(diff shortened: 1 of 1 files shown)


jsdalton (changes requested):
We already have husky + lint-staged set up in this repo. Let's lean on those -- or discuss removing them entirely -- instead of setting up a parallel set of git hooks.

ghepting:
> 

Hmm okay let's chat. Maybe i'm missing something!

PR #3466, opened 2023-05-11
refactors to align with CEP for ResourceLink schema spec

ghepting:
## Purpose
refactors to align with ResourceLink specification in [CEP draft](https://contentful.atlassian.net/wiki/spaces/ENG/pages/4221043197/DRAFT+CEP-0163+Extending+ResourceLink+to+reference+external+resources)

### DIFF
--- a/apps/ecommerce/frontend/src/components/ResourceCard.tsx
+++ b/apps/ecommerce/frontend/src/components/ResourceCard.tsx
@@ -1,5 +1,5 @@
 import { Badge, Box, Card, Flex, Grid, Text } from '@contentful/f36-components';
-import { HydratedResourceData, ResourceCardProps, ExternalResourceLink } from '../types';
+import { ExternalResource, ResourceCardProps, ExternalResourceLink } from '../types';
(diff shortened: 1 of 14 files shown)


apps/ecommerce/frontend/src/components/ResourceCard.tsx
```diff
@@ -79,10 +80,7 @@ const ResourceCard = (props: ResourceCardProps) => {
     );
   }
 
-  const provider = props.value.sys?.provider || 'External Resource';
-
-  let resourceType = props.value.sys?.linkType?.toString();
-  if (resourceType) resourceType = resourceType?.split(':')[1];
+  const [resourceProvider, resourceType] = props.value.sys?.linkType?.split(':');
```
jsdalton, line 83: Seeing this makes me realize we'll need to eventually add more rigid runtime checking to this value, e.g. if someone uses an extra `:` in the provider or type, which would pass the type check but could produce unexpected results at run time.

jsdalton (approved):
👍

PR #3467, opened 2023-05-11
Feat/refactor fields

rusticpenguin:
## Purpose
Having each field be it's own maintained component makes it harder keep both components consistent

## Approach

Moved the mockData into it's own function, and pulled out all the rendering logic into a core component `ResourceList`. Both Single/Multiple resources wrap this list and pass down their specific features and logic

### DIFF
--- a/apps/ecommerce/frontend/src/components/AddContentButton.tsx
+++ b/apps/ecommerce/frontend/src/components/AddContentButton.tsx
@@ -17,9 +17,11 @@ const container = css({
 });
 
 export const AddContentButton = (props: AddContentButtonProps) => {
+  const { onClick } = props;
+
   return (
     <div className={container}>
-      <Button variant="secondary" startIcon={<PlusIcon />} size="small" onClick={props.onClick}>
+      <Button variant="secondary" startIcon={<PlusIcon />} size="small" onClick={onClick}>
         <Text fontWeight="fontWeightDemiBold">Add content</Text>
       </Button>
     </div>
--- a/apps/ecommerce/frontend/src/components/MultipleResources.tsx
+++ b/apps/ecommerce/frontend/src/components/MultipleResources.tsx
@@ -18,6 +18,7 @@ const MultipleResources = () => {
   }, [sdk.field, setValue]);
 
   const addContent = () => {
+    // TODO: Update this function to add a new resource(s) to the list
     const newValue = mockValue(sdk);
 
     if (value) {
@@ -27,18 +28,18 @@ const MultipleResources = () => {
     }
   };
 
-  const onRemove = (index: number) => {
+  const handleRemove = (index: number) => {
     const newValue = value.filter((obj, i) => i !== index);
     sdk.field.setValue(newValue);
   };
 
-  const onMoveToTop = (index: number) => {
+  const handleMoveToTop = (index: number) => {
     const newValue = [...value];
     newValue.unshift(newValue.splice(index, 1)[0]);
     sdk.field.setValue(newValue);
   };
 
-  const onMoveToBottom = (index: number) => {
+  const handleMoveToBottom = (index: number) => {
     const newValue = [...value];
     newValue.push(newValue.splice(index, 1)[0]);
     sdk.field.setValue(newValue);
@@ -48,9 +49,9 @@ const MultipleResources = () => {
     <ResourceList
       addContent={addContent}
       isMultiple={true}
-      onMoveToBottom={onMoveToBottom}
-      onMoveToTop={onMoveToTop}
-      onRemove={onRemove}
+      onMoveToBottom={handleMoveToBottom}
+      onMoveToTop={handleMoveToTop}
+      onRemove={handleRemove}
       total={total}
       value={value}
     />
--- a/apps/ecommerce/frontend/src/components/MultipleResources.tsx
+++ b/apps/ecommerce/frontend/src/components/MultipleResources.tsx
@@ -2,17 +2,13 @@ import { useEffect, useState } from 'react';
 import { FieldAppSDK } from '@contentful/app-sdk';
 import { /* useCMA, */ useSDK } from '@contentful/react-apps-toolkit';
 import { ExternalResourceLink } from '../types';
-import ResourceCard from './ResourceCard';
-import { Field } from '@contentful/default-field-editors';
-import { Collapse, Grid, TextLink } from '@contentful/f36-components';
-import { AddContentButton } from './AddContentButton';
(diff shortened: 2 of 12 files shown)


apps/ecommerce/frontend/src/helpers/getResourceProviderAndType.ts
```diff
@@ -0,0 +1,12 @@
+import { ExternalResourceLink } from '../types';
+
+const getResourceProviderAndType = (resource?: ExternalResourceLink) => {
+  const [resourceProvider, resourceType] = resource?.sys?.linkType?.split(':') || [];
```
rusticpenguin, line 4: incase the linktype is undefined, the output will be [undefined, undefined]. We could also have these values be an error state too like `No Provider` and `No Resource Type`

pkeavenycontentful (approved):

apps/ecommerce/frontend/src/components/AddContentButton.tsx
```diff
@@ -17,9 +17,11 @@ const container = css({
 });
 
 export const AddContentButton = (props: AddContentButtonProps) => {
+  const { onClick } = props;
```
ryunsong-contentful, line 20: 🤝

apps/ecommerce/frontend/src/components/MultipleResources.tsx
```diff
@@ -21,6 +17,16 @@ const MultipleResources = () => {
     });
   }, [sdk.field, setValue]);
 
+  const addContent = () => {
+    const newValue = mockValue(sdk);
```
ryunsong-contentful, line 21: :nit Can we add a TODO here?

whitelisab (approved):
LGTM 🥇

apps/ecommerce/frontend/src/components/MultipleResources.tsx
```diff
@@ -38,62 +44,16 @@ const MultipleResources = () => {
...
-        </TextLink>
-        <Collapse isExpanded={isExpanded}>
-          <Field sdk={sdk} />
-        </Collapse>
-      </Grid.Item>
-    </Grid>
+    <ResourceList
+      addContent={addContent}
+      isMultiple={true}
+      onMoveToBottom={onMoveToBottom}
```
ryunsong-contentful, line 51: Annoying nit: One convention I think we should follow is that upon creating the function the name should be 'handle'. When it's passed as a prop we name it with the `on` prefix

apps/ecommerce/frontend/src/components/ResourceCard.tsx
```diff
@@ -3,61 +3,74 @@ import { useState } from 'react';
...
+import { ExternalResource, ExternalResourceLink } from '../types';
+import { RenderDragFn } from '@contentful/field-editor-reference/dist/types';
+import getResourceProviderAndType from '../helpers/getResourceProviderAndType';
 
-const ResourceCard = (props: ResourceCardProps) => {
-  const debouncedValue = useDebounce(props.value, 300);
-  const { hydratedResourceData, isLoading, error, errorMessage, errorStatus } =
+interface Props {
+  value: ExternalResourceLink;
+  data?: ExternalResource;
```
ryunsong-contentful, line 15: this isn't used anywhere - or is this intentional?

apps/ecommerce/frontend/src/components/ResourceList.tsx
```diff
@@ -0,0 +1,73 @@
...
+          items={value}
+          axis="y"
+          useDragHandle={isMultiple}
+          isInitiallyDisabled={true}
+          isDisabled={false}
+          hasCardEditActions={false}
+          sdk={sdk}
+          viewType={'card'}
+          parameters={{ instance: {} }}>
+          {({ item, isDisabled, DragHandle, index }) => {
```
ryunsong-contentful, line 40: nit: isDisabled is never used here

apps/ecommerce/frontend/src/helpers/getResourceProviderAndType.ts
```diff
@@ -0,0 +1,12 @@
+import { ExternalResourceLink } from '../types';
```
ryunsong-contentful, line 1: I think the file name should be change to either `utils` or `ResourceProviderUtils`. Something like that

apps/ecommerce/frontend/src/helpers/getResourceProviderAndType.ts
ryunsong-contentful: I think you don't need to make the parameter optional. Personally, I think utils functions should be pure functions without all this error handling.

Basically it should be 

```
const getResourceProviderAndType = (resource: ExternalResourceLink) => {
  const [resourceProvider, resourceType] = resource.sys.linkType.split(':') 

  return {
    resourceProvider,
    resourceType,
  };
};
```

and you should do some resource error checking in the caller who will have context on how to handle those situations. That way these functions are cut and dry and what they do, and each component can handle their errors as they need to

ryunsong-contentful (changes requested):
Mainly cleanup and some design considerations I'd like to talk to you about in regards to these Resource Cards

apps/ecommerce/frontend/src/components/ResourceCard.tsx
```diff
@@ -3,61 +3,74 @@ import { useState } from 'react';
...
             {resourceProvider} {resourceType}
           </Text>
           <Flex alignItems="center" isInline={true}>
-            {hydratedResourceData.status && (
-              <Badge variant="featured">{hydratedResourceData.status}</Badge>
-            )}
+            {externalResource.status && <Badge variant="featured">{externalResource.status}</Badge>}
             <ResourceCardMenu
-              onRemove={() => props.onRemove(props.index)}
+              onRemove={() => onRemove(index)}
```
ryunsong-contentful, line 67: Note/Re-iteration since we found a bug: Being more declarative on how we set state rather than relying on some toggle method would improve code quality/bug prevention

apps/ecommerce/frontend/src/components/MultipleResources.tsx
rusticpenguin: All for this!

ryunsong-contentful (approved):
I'm approving in advance as I am on leave and don't want to be a blocker.

I'm also assuming the changes requested will be addressed

apps/ecommerce/frontend/src/components/ResourceCard.tsx
rusticpenguin: Yeah this is later on when we are actually populating cards with a bulk endpoint

PR #3478, opened 2023-05-12
fix: build script for actions in typescript template[EXT-4509]

vikaskumr:
## Purpose

Introduced a build script to bundle the transpiled code in ESM.

## Approach

Every action will be bundled  in ESM code for the TypeScript template with the help of build script.

### DIFF
--- a/examples/hosted-app-action-templates/typescript/build-action.js
+++ b/examples/hosted-app-action-templates/typescript/build-action.js
@@ -1,0 +1,35 @@
+const esbuild = require('esbuild');
+const { resolve } = require('path');
+const yargs = require('yargs/yargs');
+const { hideBin } = require('yargs/helpers');
+
+const manifest = require('./contentful-app-manifest.json');
+
+const argv = yargs(hideBin(process.argv)).argv;
+
+const getEntryPoints = () =>
+  manifest.actions.reduce((result, action) => {
+    result[action.id] = resolve(__dirname, action.path);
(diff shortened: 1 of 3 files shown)


examples/hosted-app-action-templates/typescript/build-action.js
```diff
@@ -0,0 +1,35 @@
+const esbuild = require('esbuild');
+const { resolve } = require('path');
+const yargs = require('yargs/yargs');
```
kdamball, line 3: Is this package already available?

examples/hosted-app-action-templates/typescript/build-action.js
```diff
@@ -0,0 +1,35 @@
...
+const { hideBin } = require('yargs/helpers');
```
kdamball, line 4: Same here with the package

kdamball (commented):
minor stuff missing but good start

examples/hosted-app-action-templates/typescript/build-action.js
vikaskumr: Yes, when I tested it didn't gave any error regarding this package to add it as a dev dependency

examples/hosted-app-action-templates/typescript/build-action.js
vikaskumr: Yes

PR #3481, opened 2023-05-12
[INTEG-479] Delete Bynder app

ryunsong-contentful:
## Purpose
Remove deploy scripts before deleting the entire app. We'll delete the entire app once we're 99.99999% confident the deployment was successful

## Approach
Make sure Bynder works


matthew-gordon (approved):
LGTM when we are ready to 💣 💥 ✅

PR #3482, opened 2023-05-12
Feature INTEG-540: Connect to and retrieve products from Shopify

pkeavenycontentful:
## Purpose

For the e-comm stack work, it's necessary to connect our lambda to a real Provider to emulate customer needs. 

This PR connects to our Partner Shopify account for testing against our test shop data. 

## Approach

- Was trying to use `shopify-api-js`, but that ended up being way too much overhead and requiring user-based authentication with OAuth, we might revisit this down the road, but it will depend on the UX for space admins adding Shopify as an integration (i.e., do we want them to authenticate the same way we do with GA4, through a dialog, or just with access token/domain)
- Switched to https://shopify.dev/docs/custom-storefronts/additional-sdks/js-buy since it's lighter weight and requires less for authentication

## Testing steps

- Can test with the following Postman collection if you comment out L30 in `app.ts`: https://grey-spaceship-425061.postman.co/workspace/New-Team-Workspace~4ca0371f-88c4-46de-b128-da67c31e4460/request/26682020-fe3c70cb-367b-4538-8f38-4a685fdac4bc

- JS Buy SDK: https://shopify.dev/docs/custom-storefronts/additional-sdks/js-buy

### DIFF
--- a/apps/ecommerce/lambda/src/controllers/apiController.ts
+++ b/apps/ecommerce/lambda/src/controllers/apiController.ts
@@ -21,8 +21,9 @@ const ApiController = {
       if (!proxyBaseUrl) {
         return res.status(404).send({
           status: 'error',
-          message: `Base URL configuration for provider${resourceProvider ? `: "${resourceProvider}"` : ''
-            } not found`,
+          message: `Base URL configuration for provider${
+            resourceProvider ? `: "${resourceProvider}"` : ''
+          } not found`,
         });
       }
 
--- a/apps/ecommerce/lambda/src/controllers/apiController.ts
+++ b/apps/ecommerce/lambda/src/controllers/apiController.ts
@@ -26,9 +26,8 @@ const ApiController = {
       if (!proxyUrl) {
         return res.status(404).send({
           status: 'error',
-          message: `Provider${
-            resourceLink.sys.provider ? `: ${resourceLink.sys.provider}` : ''
-          } not found`,
+          message: `Provider${resourceLink.sys.provider ? `: ${resourceLink.sys.provider}` : ''
+            } not found`,
         });
       }
 
@@ -44,7 +43,6 @@ const ApiController = {
             .send(JSON.parse(JSON.stringify((response as AxiosResponse).data)));
         }
       } catch (error) {
-        console.log('error', error);
         res.status(500).send({
           status: 'error',
           message: 'Error fetching resource',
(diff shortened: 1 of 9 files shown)


apps/ecommerce/lambda/src/routers/shopifyRouter.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+const shopifyClientStub = {
+    product: {
+        fetch: () => Sinon.promise((resolve) => {
+            return resolve({
+                title: 'Some Product title',
+                description: 'Some product description',
+                images: [{ url: 'some-product-url' }]
+            } as Product)
+        }),
+        fetchAll: () => Sinon.promise(Sinon.spy()),
```
pkeavenycontentful, line 22: Seems like a lot of overhead for a Sinon stub, but I kept running into errors unless I implemented spies for every method contained in the SDK. Open to any feedback on how to do this better!

mgoudy91: yeah that's weird but I think what you've got here makes sense given that's the case!

apps/ecommerce/lambda/src/routers/shopifyRouter.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+        throw new Error('Function not implemented.');
+    }
+}
+
+describe('Shopify Router', () => {
+    beforeEach((done) => {
+        sandbox.stub(NodeAppsToolkit, 'verifyRequest').get(() => {
+            return () => true;
+        });
+        done();
```
mgoudy91: I really don't want to poke the hornet's nest that is our async testing woes, but do you need `done` for these methods that aren't async?

mgoudy91 (approved):
I think this looks great, nice work!

apps/ecommerce/lambda/src/routers/shopifyRouter.spec.ts
pkeavenycontentful: Good point. I did try to remove these `done()`'s, and it caused timeouts when testing. I also tried removing `done()`'s elsewhere in the file and ran into timeout errors. It seems like the nature of the sinon sandbox is a lot of async calls, so these are needed to avoid timeouts. Good suggestion though 👍

matthew-gordon (approved):

apps/ecommerce/lambda/src/controllers/apiController.ts
```diff
@@ -51,7 +51,6 @@ const ApiController = {
             .send(JSON.parse(JSON.stringify((response as AxiosResponse).data)));
         }
       } catch (error) {
-        console.log('error', error);
```
ghepting, line 54: why was this removed?

PR #3498, opened 2023-05-15
[INTEG-536-550] App config page and store credentials

whitelisab:
## Purpose

A space admin should be able to enter the required provider credentials in the configuration page when installing the app. 

## Approach

The configuration page will render differently based on the e-commerce provider. The configuration can be accessed from a new `/config` endpoint. The space admin can then enter the credentials for the provider and these are persisted in the app installation parameters. I also added a `Check Credentials` button for testing that calls the `/healthcheck` endpoint in the Shopify controller. This PR also adds support for absolute imports on the frontend and organizes the frontend components.

Note that this is just the first chunk of work for this config page, there will be some follow up work to improve things like error handling.

Also the Magento configuration is just a placeholder for now.

Screenshot 2023-05-15 at 3 50 10 PM

Screenshot 2023-05-15 at 4 14 30 PM

## Testing

In our e-commerce testing environment, you should see different configuration screens for the Shopify 2.0 and Magento apps. You should enter the credentials from the Shopify (production) space in our Apps Testing org. The `Check Credentials` button will only work for Shopify (not Magento), and you should see a confirmation notification that the credentials are connected successfully after clicking. You should also be able to enter save credentials in the configuration screen.


ryunsong-contentful (approved):

ryunsong-contentful (approved):
That `ConfigPage.tsx` was some 🔥 code. 

Also great job on cleaning up other files with imports and such

apps/ecommerce/frontend/src/components/Config/ConfigPage.tsx
```diff
@@ -111,26 +112,27 @@ const ConfigPage = () => {
...
+        <Skeleton.DisplayText />
+        <Skeleton.BodyText offsetTop={40} numberOfLines={4} />
       </Skeleton.Container>
     );
   };
 
   return (
     <>
-      <Box className={styles.background} testId="configPageContainer" />
+      <Box className={styles.background(providerConfig.color ?? '')} testId="configPageContainer" />
```
ryunsong-contentful, line 126: There's one smell I have here which is related to checking isLoading twice.

This container shouldn't be just some element at the top which you extend and manipulate the css to create the background. It should contain all of its children subcomponents.

More simply, this Box component should wrap everything below here. You should base your height off that and use a minHeight in the css. Once you have this element as an actual container, you can then just do your skeleton loading check once for both the logo and the field contents

apps/ecommerce/frontend/src/components/Config/ConfigPage.tsx
```diff
@@ -164,6 +166,9 @@ const ConfigPage = () => {
           </>
         )}
       </Box>
+      <Box className={styles.icon}>
+        {isLoading ? null : <img src={providerConfig.logo} alt="App logo" />}
```
ryunsong-contentful, line 170: Shouldn't have to check twice once you have a true Box container

ryunsong-contentful (approved):
Approved with contingency/don't want to be a blocker

rusticpenguin (approved):

PR #3510, opened 2023-05-16
Fix/build script for action [EXT-4548]

vikaskumr:
## Purpose

Fixing the build script used in app actions hosted, target for the build.
Name of the file for script was wrong, and because of that it was in the actions folder instead of being in root of the app, after correcting the file name all the problems related to build script are solved.


examples/hosted-app-action-templates/typescript/build-actions.js
```diff
@@ -14,21 +14,28 @@ const getEntryPoints = () =>
...
-  const config = {
-    entryPoints: getEntryPoints(),
-    minify: true,
-    bundle: true,
-    platform: 'node',
-    outdir: 'build/actions',
-    logLevel: 'info',
-    format: 'esm',
-  };
+  try {
```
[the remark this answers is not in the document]
vikaskumr: Yeah, I thought may be some user experience if we show the error, but removed it.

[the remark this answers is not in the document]
kdamball: I'm not opposed to it but wondering if there's more we can do to that

ivo-toby:
I tried the steps CCA takes manually, that works!

I do have some questions;

- the manifest is configured to build a `js` file, while it's a `ts`-file. From esm I can understand, but this is typescript. I'd expect it to refer to a ts-file in the path param of an action. 
- build process does not check if the id of an action is unique, so if 2 actions have the same id, it will overwrite the first action with the same id. Question is; should actions have a unique id? 
- the build process does not keep subfolders in /actions/ while they are defined in the manifest, the output in build does not have folders. @kdamball will this become an issue when lambda wants to load the action from `appActionPath` ?

vikaskumr:
> How do we know the JS build actions works also?

JS template has it's own script and it just copy and paste the code and we already converted the JS code to ESM
Am I missing something here 🤔 ?

ivo-toby:
> How do we know the JS build actions works also?

if the copy-only step has not been changed, it won't. It needs bundling, we have a separate ticket for that. Not sure if we need to do that if the Javascript template is not used much tho

kdamball:
As long as we upload the correct path from the manifest file as it's referring to the build folder, we're ok. 

> build process does not check if the id of an action is unique, so if 2 actions have the same id, it will overwrite the first action with the same id. Question is; should actions have a unique id? if not the build-process should warn if they are not unique

This is a bug. Needs fixing

vikaskumr:
> 
This is correct, right?
If we had the same unique id it will override the first action, for this case we need to throw error right?
@ivo-toby

examples/hosted-app-action-templates/javascript/build-actions.js
```diff
@@ -1,34 +1,75 @@
...
+    if (uniqueValues.has(action.path)) {
+      throw new Error(`Duplicate action path: '${action.path}'`);
+    }
+    if (uniqueValues.has(action.entryFile)) {
+      throw new Error(`Duplicate entryFile path: '${action.entryFile}'`);
+    }
+
+    uniqueValues.add(action.entryFile);
+    uniqueValues.add(action.path);
+    uniqueValues.add(action.id);
```
ivo-toby, line 33: Very nit; if you're referring to props of an object more than once, I'd destruct them at the top 
( no need to change it though, only if you are really inclined to do so :D )

examples/hosted-app-action-templates/typescript/contentful-app-manifest.json
```diff
@@ -1,6 +1,7 @@
 {
   "actions": [
     {
+      "id": "example",
       "name": "Example App Action",
       "category": "Custom",
       "description": "This is an example to help you learn how hosted app actions work.",
```
ivo-toby, line 7: small thing; can you add
```allowNetworks: [] ``` to the manifest at the example-action (no validation needed here)?

ivo-toby (approved):
Small very nit thingy and a request to add something to the manifest. But besides that 👍🏻

examples/hosted-app-action-templates/typescript/contentful-app-manifest.json
vikaskumr: we are already doing validation for allowedNetworks in CCA and attaching the `[]` so I think no need here.

ivo-toby: No, I meant it to add it to this as example.. We have all other props here but not allowNetworks, so why not add it here

ivo-toby: for improved DX :)

vikaskumr: Yeah, I was also thinking about DX and you are right, I will raise another PR for that

PR #3511, opened 2023-05-16
Feature INTEG-549: Content editors can see the ecommerce product linked to a given content entry 

pkeavenycontentful:
## Purpose

As part of the e-comm stack, Content Editors will need to see ecommerce product data in a field attached to an entry. 

## Approach

Building off what was done in: https://github.com/contentful/apps/pull/3498

## Testing steps

- Configure Shopify app in your space with our partner credentials (See Shopify (Production) space in Apps Development org), hit Save
- Navigate to an entry, click "Add Product" (will randomize between a real product id or a fake one)
- Validate that product details show up, or Not Found is shown for missing product


apps/ecommerce/frontend/src/hooks/useExternalResource.tsx
```diff
@@ -19,7 +20,7 @@ const useExternalResource = (resource?: ExternalResourceLink) => {
 
   const hydrateExternalResource = useCallback(
     async (resource: ExternalResourceLink) => {
-      const url = new URL(`${config.backendApiUrl}/api/resource`);
+      const url = new URL(`${config.backendApiUrl}/shopify/resource`);
```
whitelisab, line 23: Should we be using the `baseUrl` from the instance parameters here instead of pulling from this config? The `baseUrl` is what I used in the `ConfigPage` component (e.g. https://github.com/contentful/apps/blob/master/apps/ecommerce/frontend/src/components/Config/ConfigPage.tsx#L75). This way, you don't have to hardcode `shopify` in the url

apps/ecommerce/frontend/src/hooks/useExternalResource.tsx
```diff
@@ -30,13 +31,15 @@ const useExternalResource = (resource?: ExternalResourceLink) => {
         {
           'x-contentful-data-provider': resourceProvider.toLowerCase(),
           'X-Contentful-Data-Provider-BaseURL': sdk.parameters.instance.baseUrl,
+          'x-contentful-shopify-domain': apiEndpoint,
```
whitelisab, line 34: Probably a question for the future and not for this PR, but I am curious how we are going to handle different providers using this approach. I guess maybe use information from the config to construct the headers rather than hardcoding them?

whitelisab (commented):
I left some questions for you, and I also have a question about the `ResourceCard`, should we be expecting to see a description and an image from Shopify at this point? I didn't see this, maybe it was just the specific product I had, but wanted to ask!

^update to the above, that specific product doesn't have a description, but I was able to validate that the description shows with another product that has one. For the image though, I think you'll have to change the `shopifyAdapter`. I was able to get the image if I used the `src` property, but that property is deprecated according to the types in `shopify-buy` 🤔

apps/ecommerce/frontend/src/hooks/useExternalResource.tsx
pkeavenycontentful: Oh cool, good call! Will switch it to `baseUrl` 👌

apps/ecommerce/frontend/src/hooks/useExternalResource.tsx
pkeavenycontentful: True, we are aiming for the universal frontend being agnostic of the provider, I can try generalizing this more.

apps/ecommerce/frontend/src/components/Config/ConfigPage.tsx
```diff
@@ -83,16 +83,11 @@ const ConfigPage = () => {
...
-      cma,
-      'POST',
-      {
-        'X-Contentful-Data-Provider': sdk.parameters.instance.provider,
-      },
-      { domain: parameters.apiEndpoint, storefrontAccessToken: parameters.storefrontAccessToken }
-    )
+    fetchWithSignedRequest(url, sdk.ids.app, cma, 'POST', {
+      'X-Contentful-Data-Provider': sdk.parameters.instance.provider,
+      'x-contentful-shopify-domain': parameters.apiEndpoint,
```
ghepting, line 88: we should *not* be passing this through as a header. it should be coming from the [provider configs](https://github.com/contentful/apps/blob/master/apps/ecommerce/lambda/src/mocks/providerConfigs.ts) that @whitelisab added to avoid introducing man-in-the-middle/SSRF attacks

pkeavenycontentful: Not sure I follow, how do the values get passed to the endpoint via the provider configs? Should `apiEndpoint/storefrontAccessToken` be body params?

ghepting: to be clear, i'm referring to the apiEndpoint – URL things shouldn't be coming from the frontend because someone could make a request with a bad actor endpoint like `iscrapecredentials.com` and then get the access tokens of our customers.

let's start a thread in the eng chat to discuss further

apps/ecommerce/frontend/src/hooks/useExternalResource.tsx
pkeavenycontentful: Poked around at this and seems like it might be more work out of scope for this ticket. Would it be okay to address as a follow-up?

apps/ecommerce/frontend/src/hooks/useExternalResource.tsx
pkeavenycontentful: Updated to use `baseURL` in latest commit.

apps/ecommerce/frontend/src/hooks/useExternalResource.tsx
whitelisab: Yes, that makes sense!

pkeavenycontentful:
There's an odd issue with the Shopify Buy SDK, it says the `src` is deprecated, but none of the other values available on `ImageResource` have valid URL strings. Not sure when they're planning on deprecating it but as of the current version it's the only property that has a valid image URL. Just FYI for anyone who sees this 👌

whitelisab (approved):
Thank you for addressing those changes! LGTM ⭐

matthew-gordon (approved):

pkeavenycontentful:
Ticket created to address SSRF issues in follow-up work.

PR #3521, opened 2023-05-17
update API gateway integration to disclude filepath so it works for relative assets (duh)

ghepting:
## Purpose
fix api gateway configuration for ecommerce frontend


whitelisab (approved):
LGTM 🎊

PR #3550, opened 2023-05-22
Refactor/sprint work

rusticpenguin:
## Purpose

Prop drilling was starting to become unwieldy. And as we grow out the app we will want to keep the code more organized

## Approach

Utilizing context and providers to get rid of prop drilling. Added hooks to simplify the component's code. Moved data around so only the components that "care" about that data will have access to it

## Testing steps

Run the shopify app, verify all functionality remains unchanged


apps/ecommerce/frontend/src/components/Field/ResourceCard/ResourceCard.tsx
```diff
@@ -32,17 +31,11 @@ const ResourceCard = (props: Props) => {
   if (error) {
     return (
       <MissingResourceCard
-        index={index}
-        total={total}
-        onRemove={onRemove}
-        onMoveToBottom={onMoveToBottom}
-        onMoveToTop={onMoveToTop}
-        isLoading={isLoading}
+        {...props}
```
pkeavenycontentful, line 34: 👍

pkeavenycontentful (approved):

apps/ecommerce/frontend/src/components/Field/ResourceCard/useResourceMenuItems.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (typeof index !== 'undefined' && total && total > 1) {
+      if (index > 0 && onMoveToTop) {
+        newMenuItems.push(
+          <Menu.Item key="moveToTop" onClick={() => onMoveToTop()}>
+            Move to top
+          </Menu.Item>
+        );
+      }
+
+      if (index + 1 !== total && onMoveToBottom) {
```
ryunsong-contentful, line 32: Can't you just do ```index + 1 < total```?

Also why do we even keep a total track? Isn't the total just the menuItems array length?

apps/ecommerce/frontend/src/components/Field/Wrappers/MultipleResources.tsx
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      sdk.field.setValue([newValue]);
+    }
+  };
+
+  const handleRemove = (index: number) => {
+    const resourceArray = sdk.field.getValue();
+
+    const newValue = resourceArray.filter((obj: ExternalResourceLink, i: number) => i !== index);
+    sdk.field.setValue(newValue);
```
ryunsong-contentful, line 27: Slight inconsistency here with the single value. For a single, the remove makes this value undefined. But here, on the last item in the array, it'll just be an empty array.

Do we care about making this distinction?

ryunsong-contentful:
Overall lgtm - overall cleaner

ryunsong-contentful (approved):

PR #3574, opened 2023-05-24
Refactor to use app access token to retrieve secret params

ghepting:
## Purpose
Refactors the ecom stack to utilize app access tokens to retrieve sensitive app installation parameters via the CMA instead of passing them front the frontend as headers to avoid SSRF/other consequences, depending on the data and how it's used in the backend.


apps/ecommerce/frontend/src/components/Config/ConfigPage.tsx
```diff
@@ -81,23 +83,21 @@ const ConfigPage = () => {
...
-    fetchWithSignedRequest(url, sdk.ids.app, cma, 'POST', {
-      'X-Contentful-Data-Provider': sdk.parameters.instance.provider,
-      'x-contentful-shopify-domain': parameters.apiEndpoint,
-      'x-contentful-shopify-token': parameters.storefrontAccessToken,
+  const checkCredentials = async () => {
+    const url = new URL(`${config.backendApiUrl}/api/credentials`);
+    fetchWithSignedRequest(url, sdk.ids.app, cma, sdk, 'GET', {
+      ...contentfulContextHeaders(sdk),
+      // we only pass the installation parameters to the backend via headers here because they are not yet persisted
+      'x-data-provider-parameters': JSON.stringify(parameters),
```
ghepting, line 91: this is feeling really icky 👎

apps/ecommerce/frontend/src/components/Config/ConfigPage.tsx
```diff
@@ -71,22 +70,20 @@ const ConfigPage = () => {
...
+    fetchWithSignedRequest(url, sdk.ids.app, cma, sdk, 'GET', {
+      // we only pass the installation parameters to the backend via headers here because they are not yet persisted
+      'x-data-provider-parameters': JSON.stringify(parameters),
     })
-      .then((res) => {
-        if (res.status !== 200) {
-          const error = `Error: ${res.status} - ${res.statusText}`;
-          throw new Error(error);
+      .then((res) => res.json())
+      .then((data) => {
```
jsdalton, line 80: Style thought: I see you're just being consistent with what's already here, and we don't really have a style rule for this or anything, but it seems like we prefer await/async these days vs. then promise chains.

apps/ecommerce/lambda/src/app.ts
```diff
@@ -22,12 +22,10 @@
...
 
-// enable CORS on /api/* routes
-app.use(apiRouteConstraint, cors(corsConfig));
-
-// verify signed requests on /api/* routes
-// to test endpoints with Postman, comment out below
-app.use(apiRouteConstraint, Middleware.verifiySignedRequests);
+app.use(apiRouteConstraint, cors(corsConfig)); // enable CORS on /api/* routes
+app.use(apiRouteConstraint, Middleware.loadAppConfig); // load app config on /api/* routes
+app.use(apiRouteConstraint, Middleware.verifiySignedRequests); // verify signed requests on /api/* routes
```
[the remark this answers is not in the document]
jsdalton: ☝️ Should we create a JIRA ticket to add rate limiting? (We can definitely defer)

apps/ecommerce/lambda/src/middlewares/getAppInstallationParameters.ts
```diff
@@ -0,0 +1,86 @@
...
+      const environmentId = req.header('x-contentful-environment-id');
+
+      if (!appId || !spaceId || !environmentId) {
+        console.log('headers', req.headers);
+        throw new Error('Missing required headers!');
+      }
+
+      if (!req.appConfig) throw new Error('App config not found');
+
+      const appAccessToken = await getManagementToken(req.appConfig.privateKey, {
```
jsdalton, line 34: It could just be, but mixing await with then here really does my head in, and also leads us to awkward things like `then((token) => token)` This could all be refactored to:

```js
let appAccessToken
try {
  appAccessToken = await getManagementToken(req.appConfig.privateKey, {
        appInstallationId: appId,
        spaceId,
        environmentId,
      })
} catch(e) {
  throw new UnableToGetAppAccessToken(e.message, {
    cause: e,
  });
}
```

jsdalton (approved):
Looks good overall. I ended up with a handful of nitpicky comments none of which are critical. Two high level things that stood out:

* We seem to be mixing and matching then promise chains with async/await. I think in GA4 we just avoided promise chains entirely. Not sure if there's a benefit here to promise chains I'm missing. 
* A lot of console.error messages. Are these intentional? They can clutter up the test output. We might consider stubbing for tests if we want to use that.

PR #3575, opened 2023-05-24
scripts package

lnsflive:
## Purpose

The Vue template is missing @contentfu/app-scripts package and NPM scripts.

## Approach

This might have been possibly overlooked as React seems to be the favorable approach to creating apps with Contentful.
I didn't find anything relating to the missing @contentful/app-scripts needed to properly upload via CLI in the documentation in https://www.contentful.com/developers/docs/extensibility/app-framework/create-contentful-app/

## Testing steps

None

## Breaking Changes

No

This was my original issue, I didn't see any documentation on where or how to add @contentful/app-scripts as a package when creating a new app using a template or other than React in https://www.contentful.com/developers/docs/extensibility/app-framework/create-contentful-app/

## Deployment

No

### DIFF
--- a/examples/vue/package.json
+++ b/examples/vue/package.json
@@ -7,13 +7,17 @@
     "preview": "vite preview --port 5050",
     "test:unit": "vitest --environment jsdom",
     "typecheck": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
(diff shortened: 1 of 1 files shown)


lnsflive:
Thanks for the feedback. I was also having issues when uploading. All the asset src files were wrong and vite might need base: "/" to fix this. Still testing and will more likely make a new PR for that too

PR #3588, opened 2023-05-25
Feature/integ 492 serverless upgrade

pkeavenycontentful:
## Purpose

Serverless v3 is here, so it's necessary to upgrade our serverless v2s before they become too deprecated. 

## Approach

- used the `Serverless CLI` to test these, but stopped before the deploy step, so any errors that occur during deploy are currently untested 

## Breaking Changes

Serverless v3 might break production builds, need to test in staging/test environment first. 

## Deployment

- Test the apps in this PR in a staging deploy (or equivalent) before deploying to production


apps/smartling/lambda/package.json
whitelisab, line 1: You'll also want to remove the `"overrides": {
    "@serverless/dashboard-plugin": "6.1.6"
  }` in this file, since that was added a temporary fix until we could update to serverless v3. Smartling is the only app that has this override in the `package.json` file.

apps/ai-image-tagging/lambda/package.json
```diff
@@ -7,15 +7,16 @@
   "scripts": {
     "lint": "eslint .",
     "test": "jest",
-    "test:ci": "jest",
     "serve": "node serve.js",
     "deploy": "sls deploy --stage $STAGE",
-    "deploy:test": "npm run deploy"
+    "deploy:test": "npm run deploy",
+    "test:ci": "npm run test:ci:build-frontend && jest",
+    "test:ci:build-frontend": "cd ../frontend && npm run build && cd ../lambda"
```
mgoudy91, line 14: did we end up needing this change after all?

mgoudy91 (approved):
lgtm!

apps/ai-image-tagging/lambda/package.json
pkeavenycontentful: I don't think we need to at this point. Nice catch👌

pkeavenycontentful:
Most of the merge work was done on the following branch: `feature/INTEG-492_serverless_increment `. After that, the only remaining work on this PR was removing an unneeded script and a minor aws-sdk version update. Merging and closing.

PR #3598, opened 2023-05-26
Integ 287 slack app does not emit messages in environments other than the configured one

lilbitner:
## Purpose

The purpose of this PR is to adjust how the `AuthToken`, held within DynamoDB, is being queried/referenced. Originally it was being referenced with a unique id that was compiled of various attributes, one of them being the `environmentId` that is associated with the current app installation environment. This reference was causing particular issues for environments that are aliases - the configuration page would throw an error due to a lack of credentials accessed via DynamoDB.

## Approach

Given the problem described above, one solution is to instead have an environment agnostic reference to the `AuthToken`, so that all environments within the space can properly reference the appropriate credentials. 

Therefore I implemented the use of an `installationUuid` attribute that is now being used as the reference. This `installationUuid` is added to the DynamoDB table as well as the app installation parameters for persistence. 

Workflow:
1. User installs the Slack app - when connecting to slack, a window opens that takes them through the auth flow. This renders the `useConnect` component, which creates the initial `installationUuid` and eagerly POSTs to the Lamdba, which then creates the appropriate DynamoDb row of data. (this follows the same workflow as before, just with an added `installationUuid` column). 
2. Once the user saves the installation configuration, the `installationUuid` is saved as an app installation parameter. Therefore the `installationUuid` is persisted in the params as well as the DynamoDB. 
3. Once this `installationUuid` exists as a parameter, the backend can pull that information using the CMA, and can then use it to grab the associated `AuthToken` within DynamoDB. 

### Know limitations to this approach: 
1. The user must save the config page to persist the `installationUuid`. If they do not, the auth credentials saved within DynamoDB cannot be accessed, and they will instead have to reinitiate the auth flow (which is fairly easy for the user). The main drawback here is just adding new rows to DynamoDB each time this occurs, so a proposed solution moving forward would be to remind the user to save the page on first install if they try and navigate away or to adjust the flow more significantly to not eagerly create the data in DynamoDB until the config has been saved. (follow-up work, most likely - ticket to come)
2. There is a bug within the CMA/appInstallation workflow that does not correctly provide installation parameters when environments are aliased multiple times. This causes an error to be thrown with the Slack app installation process, but this is very edge case and the plan is to eventually escalate and tackle on DX team with App Framework adjustments. 

## Testing steps

Pull down this code in conjunction with [this](https://github.com/contentful/apps/pull/3634) PR. Within the Slack (development) space created as part of our testing organization, confirm the proper installation and use of the Slack app within environments, that are or are not aliases. In order to test the event triggering functionality (sending slack notifications when publishing content for example), refer to the internal Slack app wiki (would link here if I could - can I?) - ping me with questions!


apps/slack/lambda/lib/routes/events/controller.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
lilbitner, line 1: You will notice eslint changes throughout the PR, for some reason these lint errors were being thrown as part of the PR build, but these files were untouched. Added these disable lines to fix the issue for now, and to not extend work too far out of scope.

apps/slack/lambda/lib/middlewares/corsConfig.ts
```diff
@@ -0,0 +1,34 @@
...
+    'X-Contentful-Entry',
+    'X-Contentful-Environment',
+    'X-Contentful-EnvironmentAlias',
+    'X-Contentful-Field',
+    'X-Contentful-Location',
+    'X-Contentful-Organization',
+    'X-Contentful-Space',
+    'X-Contentful-User',
+  ],
+};
```
lilbitner, line 34: took this cors config logic from GA4, as it works well to define the headers as well as adjust any of the headers (as per the new uuid header)

apps/slack/lambda/lib/routes/auth-token/repository.ts
```diff
@@ -47,24 +49,43 @@ export class AuthTokenRepository {
...
+    );
+    const installationUuid = parameters.installationUuid;
+    const spaceEnvUuid = AuthTokenRepository.uuid(
+      context.spaceId,
+      context.environmentId,
+      workspaceId
     );
 
+    // installationUuid is preferred, spaceEnvUuid is for backwards compatibility
+    const possibleUuids = installationUuid || spaceEnvUuid;
```
lilbitner, line 86: backwards compatibility flow here

jsdalton: Super small nit but it confused me at first: Maybe rename to `possibleUuid` (singular) or even `uuidToUse` or something. Your current name made me at first think this was an array / collection of uuids.

jsdalton (approved):
I'm not as close to this as the other and didn't test, but code wise everything looks sensible and correct. Nice solution! 💯

apps/slack/lambda/lib/routes/auth-token/repository.ts
lilbitner: ah yes good point! Will adjust 👍

PR #3599, opened 2023-05-26
[Provider] Adds Shopify skeleton provider integration

ryunsong-contentful:
## Purpose
Barebones Shopify provider - purpose is to have it launched offline/locally successfully and can be run in parallel with the integration proxy lambda locally. 

No deployment, no testing - just the barebones to get the service up and running. Deployment, testing, api endpoints will be developed in the next iterative PR.

## Approach
Make sure Integration Proxy, Provider, and Client can all be run in parallel locally

## Testing steps
Test locally in the approach section

## Breaking Changes
None 

None


apps/ecommerce/providers/shopify/serverless.yml
```diff
@@ -0,0 +1,49 @@
+service: sls-apps-ecommerce
```
whitelisab, line 1: Should this line (and line 34) be changed to something provider-specific?

apps/ecommerce/providers/shopify/tsconfig.json
```diff
@@ -0,0 +1,103 @@
+{
+  "include": ["src", "test"],
+  "exclude": ["node_modules"],
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
```
whitelisab, line 5: Do we want to keep all these comments in here?

whitelisab (approved):
Confirmed that the Integration Proxy, Provider, and Client can all be run in parallel locally. Looks good to me as a basic skeleton for the Shopify provider lambda 👍

apps/ecommerce/providers/shopify/tsconfig.json
ryunsong-contentful: Don't have a strong opinion here. Preferentially, I personally like having all these options. Also I placed all the configuration we are using to the top of the file

PR #3649, opened 2023-06-01
[INTEG-617] Update shopify backend to support fetching of product list

ryunsong-contentful:
## Purpose
Unlock picker frontend work by enabling the backend to fetch a list of products

## Approach
Add new endpoint to existing router

## Testing steps
Unit test, and test in app

## Breaking Changes
None

None


whitelisab (commented):
These changes look good! Should we also update `apiRouter.ts` and `apiController.ts` with the `/resources` endpoint? I think that's the last piece that's needed in order for this to be usable by the frontend.

whitelisab (approved):

PR #3654, opened 2023-06-02
[INTEG-618] Add Resource Selection Dialog to display products

whitelisab:
## Purpose

This PR adds in the functionality to allow a content editor to open a dialog where they can view their products from their configured e-commerce vendor. 

## Approach

The scope of this PR is to create a dialog where users can see their product catalog. The ability to select, save, sort, filter, search, etc. will be addressed in future stories. For the time being, I kept the `Add Content` button which saves a product to the field and then I added a new `Add a Product` button which opens the dialog. We will combine this into one button in future work. Also to be addressed later is loading/error states for the dialog.

Screenshot 2023-06-02 at 8 49 03 AM


apps/ecommerce/frontend/src/components/Dialog/Dialog.styles.ts
```diff
@@ -0,0 +1,47 @@
+import { css } from 'emotion';
+import tokens from '@contentful/f36-tokens';
+
+const HEADER_HEIGHT = 114;
```
whitelisab, line 4: I took much of this css from the existing ecom base app, I imagine we will update this css as we continue to build out the search/filter/selection functionality in the dialog.

apps/ecommerce/frontend/src/components/Dialog/ResourceSelectionDialog.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { contentfulContextHeaders } from 'helpers/contentfulContext';
+import { config } from 'config';
+import DialogHeader from './DialogHeader';
+import DialogBody from './DialogBody';
+import { ExternalResource } from 'types';
+import { KeyValueMap } from 'contentful-management';
+import { getResourceProviderAndType } from 'helpers/resourceProviderUtils';
+
+interface DialogInvocationParameters extends KeyValueMap {
+  linkType: string;
```
ryunsong-contentful, line 14: should the `fieldType` attribute be here as well?

apps/ecommerce/frontend/src/components/Dialog/ResourceList.tsx
```diff
@@ -0,0 +1,31 @@
...
+  resourceProvider: string;
+  resourceType: string;
+}
+
+const ResourceList = (props: Props) => {
+  const { externalResources, resourceProvider, resourceType } = props;
+
+  return (
+    <Flex className={styles.productList}>
+      {externalResources?.map((item, index) => {
```
ryunsong-contentful, line 17: No need for the nullable `?` operator here as externalResources has to have a value.

apps/ecommerce/frontend/src/components/Dialog/ResourceSelectionDialog.tsx
```diff
@@ -0,0 +1,66 @@
...
+        setExternalResources(data);
+      } catch (error: any) {
+        console.error(error.message);
+      }
+    })();
+  }, [sdk, cma]);
+
+  return (
+    <>
+      <DialogHeader
```
ryunsong-contentful, line 51: I think we should add the total up here in the header

apps/ecommerce/frontend/src/components/Field/Wrappers/SingleResource.tsx
```diff
@@ -11,6 +11,19 @@ const SingleResource = () => {
     sdk.field.setValue(undefined);
   };
 
+  const handleAddResource = async (): Promise<any[]> => {
```
ryunsong-contentful, line 14: This returns an array right? Not a promise (part 2)

apps/ecommerce/frontend/src/context/ResourceFieldContext.tsx
```diff
@@ -2,6 +2,7 @@ import { createContext } from 'react';
 
 export type ResourceFieldContextType = {
   isMultiple: boolean;
+  handleAddResource: () => Promise<any[]>;
```
ryunsong-contentful, line 5: Same comment here about the return type being a promise instead of an array

ryunsong-contentful (approved):

ryunsong-contentful:
Approved since we chatted in person + the fixes are coming in fast

apps/ecommerce/frontend/src/components/Dialog/ResourceCard/ResourceCard.tsx
```diff
@@ -0,0 +1,57 @@
+import { Badge, Box, Card, Flex, Grid, Text } from '@contentful/f36-components';
+import { styles } from './ResourceCard.styles';
+import { ExternalResource } from 'types';
+
+export interface ResourceCardProps {
+  resource: ExternalResource;
+  total: number;
```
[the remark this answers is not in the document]
whitelisab: Good catch, we don't need it here!

PR #3658, opened 2023-06-02
Segment consent aware client

mgoudy91:
## Purpose

We need to use appropriate user consent when tracking user actions.

## Approach

- Consumes changes to sdk to collect user's given consent record.
- Slight rework of segment implementation to use a singleton segment client that is aware of consent information


package.json
```diff
@@ -12,15 +12,15 @@
     "prettier": "2.8.8"
   },
   "scripts": {
-    "bootstrap": "lerna bootstrap --no-ci --since master",
-    "bootstrap:ci": "lerna bootstrap --ci --concurrency=1 --since master",
+    "bootstrap": "lerna bootstrap --no-ci --since master --include-dependencies",
```
mgoudy91, line 15: @pkeavenycontentful and I identified this flag as being necessary in his serverless upgrade work

It ended up being necessary in this as well, so I've added it here.

https://lerna.js.org/docs/api-reference/commands#--include-dependencies

matthew-gordon (approved):
LGTM ✨✅

PR #3666, opened 2023-06-05
[Provider] Adds Magento skeleton provider integration

ryunsong-contentful:
## Purpose
Barebones Magento provider - purpose is to have it launched offline/locally successfully and can be run in parallel with the integration proxy lambda locally. 

No deployment, no testing - just the barebones to get the service up and running. Deployment, testing, api endpoints will be developed in the next iterative PR.

## Approach
Make sure Integration Proxy, Provider, and Client can all be run in parallel locally

## Testing steps
Test locally in the approach section

## Breaking Changes
None 

None

### DIFF
--- a/apps/ecommerce/providers/magento/.eslintignore
+++ b/apps/ecommerce/providers/magento/.eslintignore
@@ -1,0 +1,3 @@
+build
+node_modules
+.eslintrc.js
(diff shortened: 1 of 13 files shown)


whitelisab (approved):
LGTM

PR #3678, opened 2023-06-06
[INTEG-645] Build out Shopify api skeleton

ryunsong-contentful:
## Purpose
Get all the skeleton APIs in place before moving the business logic

## Approach
Makes everything safer, easier to review, and easier to test and see what went wrong

## Testing steps
Check APIs work

## Breaking Changes
Can't get absolute imports to work for the app.ts file

None


apps/ecommerce/providers/shopify/src/app.ts
```diff
@@ -10,7 +12,10 @@ app.use(
   })
 );
 
-app.get('/api', async (req, res) => {
+app.use('/', MetadataController);
+app.use('/resourceTypes', ResourceTypeController);
```
whitelisab, line 16: nit: according to our technical brief, I believe that this should be `/resourcesTypes`

whitelisab (approved):
One small comment, otherwise good to go! 🌟

PR #3679, opened 2023-06-06
[INTEG-646] Shopify provider business logic

ryunsong-contentful:
## Purpose
Fill out the logic to hit the Shopify API for a single products, and a list of products.

DOES NOT HANDLE PAGINATION, QUERYING, NOR PAGINATION

## Approach
This backend logic is not yet hooked into the proxy. This will be done in a separate PR where that PR will delete all the business Shopify API logic in the proxy and instead forward those requests to this Provider backend

## Testing steps
Use postman to get data back (link to postman collection: https://elements.getpostman.com/redirect?entityId=27394115-97ca61c3-8f3b-4b81-869f-0432b6914a13&entityType=collection)

SINGLE
Screenshot 2023-06-06 at 4 09 48 PM

MULTIPLE
Screenshot 2023-06-06 at 4 10 08 PM

## Breaking Changes
None

None

### DIFF
--- a/apps/ecommerce/providers/shopify/src/app.ts
+++ b/apps/ecommerce/providers/shopify/src/app.ts
@@ -12,7 +12,7 @@ app.use(
   })
 );
 
-app.use('/', MetadataController);
+app.use('/metadata', MetadataController);
 app.use('/resourcesTypes', ResourceTypeController);
 
 app.get('/ping', async (req, res) => {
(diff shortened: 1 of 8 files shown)


whitelisab (approved):
Approving so that you can move forward, we discussed that the `/metadata` endpoint will need to be updated later!

PR #3693, opened 2023-06-07
Select a resource in the product picker

whitelisab:
## Purpose

Content editors needs to be able to select a product from the picker and save that selection to an entry.

## Approach

In the product picker I added:
- Hover state for the displayed resources
- Selection state for when a resource is clicked
- Header updates to reflect the number of items selected
- On save, the dialog closes and the field is then updated with the selected item

This works for the single reference field as well as a multiple reference field. For multiple reference, it works similar to the native reference field where items selected are just appended to the field.

For now, there is still the "Add content" button as well as the "Add a product" button until we have our design sync and decide on what we want for that.

Single reference field showing hover on a card:
Screenshot 2023-06-07 at 1 49 30 PM

Multiple reference field showing items selected:
Screenshot 2023-06-07 at 1 48 54 PM

### DIFF
--- a/apps/ecommerce/frontend/src/components/Dialog/DialogBody.tsx
+++ b/apps/ecommerce/frontend/src/components/Dialog/DialogBody.tsx
@@ -7,6 +7,8 @@ interface Props {
   externalResources: ExternalResource[];
   resourceProvider: string;
   resourceType: string;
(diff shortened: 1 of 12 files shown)


lilbitner (approved):
Everything looks great to me! ⭐

PR #3710, opened 2023-06-08
Feature/integ 622: Add AWS Build button to Sidebar Location

pkeavenycontentful:
## Purpose

Content editors will need a way to trigger AWS Amplify builds from the sidebar. This PR adds the Build button to the sidebar location. 

## Approach
For testing purposes and to get around the CORS errors thrown from Amplify, I used [cors-anywhere](https://www.npmjs.com/package/cors-anywhere) to proxy the request from localhost to a CORS enabled server. This is meant only for development purposes. 

## Testing steps

- Create an AWS amplify app and point it to your localhost
- Go to any Content screen, verify Build button shows up in the sidebar
- Execute Build button, ensure appropriate error messaging occurs
- To test success messaging, follow instructions in the Readme to enable `cors-anywhere`

PR #3721, opened 2023-06-09
[INTEG-621] Add amplify base app

matthew-gordon:
## Purpose
We want to allow space admins to provide Contentful with an AWS Amplify webhook URL that will kick off a build.

closes: 
https://contentful.atlassian.net/browse/INTEG-621
https://contentful.atlassian.net/browse/INTEG-624

## Approach
- adds base AWS amplify project shell
- adds initial configuration screen with input for adding/saving an amplify webhook url

## Testing steps

- Install AWS Amplify custom app to space/environment
- On configuration screen navigation, paste a valid webhook url in the `AWS Amplify Webhook URL` (webhookUrl) input
- Click Install and save configuration

## Considerations

This should be merged BEFORE https://github.com/contentful/apps/pull/3710


apps/aws-amplify/src/components/common/AmplifyIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
...5+vfv6IQZMvTBaHbPkcoqUQDvMRv5Gz0PzAaqxx+PhfIlIxEaL53AXikC2LQ03eodY46GmMlrr9ur3kcjWs+mnVd7SI8eEW8GglrInEFSgJE2J67JgOuv61crAAITaalgiXOIpyx+Ueh6Yt6QoiZOvCJEZLHUmzevnHinsCN3kAVh1N3hWHnN7myZgQc8Si4JFOFhKitu2pdDRWNDZAtIK6mjw+bRV+IBxmDilR/u0wImRLwWjxvcnrTF+HYliFtgnRXiDsGo/fq2NC9AiHeXsrCwmxSxlBecnTwnpouXVM8vTREg9LU3zOkdmUyaTDjNFKEGscpla6e5yEaIlTYXsWOJgQHSpkS02vhAXQUg8ivOQKozg1ef92sY9kyBxq8vYhRMiWmYXaVlroMP9z5WDFS4iSeEHp3Nk5TIg64TCPPjZJC7GalvoLgWNbrhob4jRtGbcKx4czHJVDTd4jFgX/2spCbSuT2nztaKzK2BA7IRbmEQ4zh4TIx91EhWxhQlQO6Nv23vs5JKrB80P/HxK9BT3CuwAPSZgQbcJh3tJMQ0Ls4n3a3snffhBS1fZpK3ySCNnikBCZdDkRsqVjQkz22Y/+/r+EOKxWv0mMQuXFDjOH5gdrNhGypaQwIaprhGztGnnJRMXA2yLWFSw0ZOxsefOokK1sFpm2sqN2rfaSfy6QqGLgbZlGr3iNg8uhJq9e3EOEbLEsdLbks/t6yQCsdr3ACO4Wjg/3mTqHmjzlMB9kZbNIk+9im/zfK+jKE3R8hbhCHLIVcESItMN8g4wJke00vNpLpisOLzByT2wtdHaXmRxq8jbhMO8npTFaFYd5zLflsGa1W4jE5pAJPoeEmLAOFI8Px0mYEFNjx1R7yXTV/5BYvxCdh3YV3mxY7mFCNJeLHeY9rk2CR2vVGod7Kl5yrIWoFsTXgw7MyZgQcxPEQnwh6vHAbKptu010nYl4KGzJQyIuRNPtFV+BMjAhyncNI0K2GCZE/vcrveR/+x6rryv1+bTPV9e0D/5c0z4or1XIt6XFfvxIgg/mdwsXbpv5Pj+USohQn59F8Q6xDbieIfVdPV8pXillZXF7YseL//4f/4nVf6xd/7W6oj/pW/+9Vv3P6or+pG+9W7veelmctf/s02vWleJ6en3x0drx0itfWLM2FNeVm/4vT14upjkIA7kAAAAASUVORK5CYII="
```
matthew-gordon, line 21: this is, just, ew. Will update this with actual asset(s) when applicable, possibly follow up PR for deploy checklist items

apps/aws-amplify/src/components/common/Splitter.tsx
```diff
@@ -0,0 +1,16 @@
+import tokens from '@contentful/f36-tokens';
```
matthew-gordon, line 1: actually not sure this is necessary anymore, will get this 🧹

apps/aws-amplify/src/lib/isVaildUrl.ts
```diff
@@ -0,0 +1,11 @@
+export const isValidUrl = (urlString: string) => {
```
matthew-gordon, line 1: Initially just using this method to verify with builtin `URL` constructor, can migrate to a more finite `REGEX` pattern if we think its necessary, these webhook url's get pretty dicy though so open to opinions here

pkeavenycontentful: All webhooks should follow the pattern of `https://webhooks.amplify.{amazon region}.amazonaws.com/prod/webhooks{some id and token}` so we could potentially use a regex pattern on `webhooks.amplify` or some variation on that.

apps/aws-amplify/src/locations/ConfigScreen.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+      await act(async () => {
+        const res = await saveAppInstallation();
+
+        expect(res).toEqual(false);
+      });
+    });
+
+    it('allows the app to be installed with a valid webhook url', async () => {
+      const user = userEvent.setup();
```
pkeavenycontentful, line 49: 💪

pkeavenycontentful (approved):
LGTM

apps/aws-amplify/src/lib/isVaildUrl.ts
matthew-gordon: OK so thinking of security,we'llneed to restrict what can be in amplifyWebhookUrl (similar to what cors-anywhere calls originWhitelist), there is a risk of abuse to our service with arbitrary Urls. SO this might need to live in a seperate PR depending on the Lambda Function URL implementation 🤔 I say we add this as is and then in a follow up rework this

apps/aws-amplify/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "react-scripts": "5.0.1"
+  },
+  "scripts": {
+    "start": "cross-env BROWSER=none react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "create-app-definition": "contentful-app-scripts create-app-definition",
+    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEFINITIONS_ORG_ID} --definition-id YlAVHsxexkcs5PG3INCmc --token ${CONTENTFUL_CMA_TOKEN}",
+    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${TEST_ORG_ID} --definition-id ${TEST_APP_ID} --token ${CONTENTFUL_CMA_TOKEN}"
```
matthew-gordon, line 24: This is the PROD Amplify app definition ID

apps/aws-amplify/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "react-dom": "17.0.2",
+    "react-scripts": "5.0.1"
+  },
+  "scripts": {
+    "start": "cross-env BROWSER=none react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "create-app-definition": "contentful-app-scripts create-app-definition",
+    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEFINITIONS_ORG_ID} --definition-id YlAVHsxexkcs5PG3INCmc --token ${CONTENTFUL_CMA_TOKEN}",
```
matthew-gordon, line 23: This is the PROD Amplify app definition ID

apps/aws-amplify/README.md
```diff
@@ -0,0 +1,81 @@
+This project was bootstrapped with [Create Contentful App](https://github.com/contentful/create-contentful-app).
+
+## How to use
+
+Execute create-contentful-app with npm, npx or yarn to bootstrap the example:
```
ryunsong-contentful, line 5: I don't think this section is relevant for this app

apps/aws-amplify/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### `npm run build`
+
+Builds the app for production to the `build` folder.
+It correctly bundles React in production mode and optimizes the build for the best performance.
+
+The build is minified and the filenames include the hashes.
+Your app is ready to be deployed!
+
+#### `npm run upload`
```
ryunsong-contentful, line 38: This command doesn't even exist in the package.json

apps/aws-amplify/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Uploads the build folder to contentful and creates a bundle that is automatically activated.
+The command guides you through the deployment process and asks for all required arguments.
+Read [here](https://www.contentful.com/developers/docs/extensibility/app-framework/create-contentful-app/#deploy-with-contentful) for more information about the deployment process.
+
+#### `npm run upload-ci`
```
ryunsong-contentful, line 44: Same comment as above, command doesn't exist - instead this should be `npm run deploy`

apps/aws-amplify/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+For this command to work, the following environment variables must be set:
+
+- `CONTENTFUL_ORG_ID` - The ID of your organization
+- `CONTENTFUL_APP_DEF_ID` - The ID of the app to which to add the bundle
+- `CONTENTFUL_ACCESS_TOKEN` - A personal [access token](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/personal-access-tokens)
+
+## Libraries to use
+
+To make your app look and feel like Contentful use the following libraries:
```
ryunsong-contentful, line 58: I don't think this is relevant for the app here as well

apps/aws-amplify/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+- [Forma 36](https://f36.contentful.com/) – Contentful's design system
+- [Contentful Field Editors](https://www.contentful.com/developers/docs/extensibility/field-editors/) – Contentful's field editor React components
+
+## Using the `contentful-management` SDK
```
ryunsong-contentful, line 63: Same comment about relevancy

apps/aws-amplify/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+```js
+// Use the client
+cma.locale.getMany({}).then((locales) => console.log(locales));
+```
+
+Visit the [`contentful-management` documentation](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#using-the-contentful-management-library)
+to find out more.
+
+## Learn More
```
ryunsong-contentful, line 77: Same comment again about relevancy

apps/aws-amplify/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "aws-amplify",
```
ryunsong-contentful, line 2: this should be `@contentful/aws-amplify`

apps/aws-amplify/package.json
ryunsong-contentful, line 23: Did a sanity check here and it matches and looks good! Just create a ticket to make sure we handle the Samson deployment and take away it's `localhost:3000` reference in the app defintion

apps/aws-amplify/src/components/common/AmplifyIcon.tsx
ryunsong-contentful: Yes please!

apps/aws-amplify/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,156 @@
...
+import tokens from '@contentful/f36-tokens';
+import { ExternalLinkIcon } from '@contentful/f36-icons';
+import AmplifyIcon, { AWSAmplifyBrand } from '../components/common/AmplifyIcon';
+import { isValidUrl } from '../lib/isVaildUrl';
+
+export interface AppInstallationParameters {
+  amplifyWebhookUrl?: string;
+}
+
+export const styles = {
```
ryunsong-contentful, line 25: Annoying nit: One thing @lilbitner has been pushing for and a pattern we've kind of taken on as a team on the frontend is to separate the styles and components as two separate files unless the styles is dead simple and short

apps/aws-amplify/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,156 @@
...
+    };
+  }, [parameters, sdk, amplifyWebhookUrl]);
+
+  useEffect(() => {
+    sdk.app.onConfigure(() => onConfigure());
+  }, [sdk, onConfigure]);
+
+  useEffect(() => {
+    (async () => {
+      const currentParameters: AppInstallationParameters | null = await sdk.app.getParameters();
```
ryunsong-contentful, line 88: Pretty sure you can just not type this at all (basically remove `AppInstallationParameters | null` and this will all work just fine

apps/aws-amplify/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+      if (currentParameters) {
+        setParameters(currentParameters);
+      }
+
+      if (currentParameters?.amplifyWebhookUrl) {
```
ryunsong-contentful, line 94: I don't think it's necessary to do this check again, you can just set your amplifyWebhookUrl on line 91 since you're already checking for `currentParameters`.

Basically get rid of this if block and in the above just have
```
      if (currentParameters) {
        setParameters(currentParameters);
        setAmplifyWebhookUrl(currentParameters.amplifyWebhookUrl);
      }
```

apps/aws-amplify/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,156 @@
...
+  };
+
+  return (
+    <>
+      <Box className={styles.background} />
+      <Box className={styles.body}>
+        <Box>
+          <Heading>Set Up AWS Amplify</Heading>
+          <Paragraph>
+            Amplify is a set of purpose-built tools and features that enables frontend web and
```
ryunsong-contentful, line 113: I'm not sure about this paragraph and how useful this is to users. Not that I have anything better to suggest but I think before we deploy this to customer, double check with Gavin on what exactly we want to say to the users.

My suspicion here is that we can find something more impactful and informative to users.

apps/aws-amplify/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
ryunsong-contentful, line 1: Delete?

ryunsong-contentful (changes requested):
Overall solid technical work and looks good overall. Just a few things where most of my comments were concentrated in.

1. We should write a new README specific to this app we're launching (GA4 is probably a decent example). Although, I have my doubts on how many eyeballs reach this repo and even more so these README's, this is a public repo and I think it's important we put our best foot forward even on easy to forget things like the README. AKA NO COPY PASTAS HERE :)
2. The wording on the Config Page I think is a low-hanging fruit that we can definitely make progress on with Gavin's help or just by thinking about what would really help out customers.

apps/aws-amplify/src/react-app-env.d.ts
matthew-gordon: Im inclined to leave it, this file has a triple-slash directive which is just referring to a file with some types definitions. You can inspect it or open it directly from the node_module: `node_modules/react-scripts/lib/react-app.d.ts`. I think generally it's better to not remove or change since it might get updated by any possible changes in create-react-app

apps/aws-amplify/src/locations/ConfigScreen.tsx
matthew-gordon: No worries, I usually like to keep things flat and co-located as much as tolerable for readability/maintainability but I can definitely support separate the styles into another file

apps/aws-amplify/src/locations/ConfigScreen.tsx
matthew-gordon: this came from the boilerplate and while it implicitly assumes the correct types because of the return value, I don't think it hurts to have the visible definition for readability. I can definitely remove it if we feel strongly about it though!

apps/aws-amplify/src/locations/ConfigScreen.tsx
matthew-gordon: Yeah I'll connect with design on this for better copy, I just took the direct amplify description for base app shell

apps/aws-amplify/src/locations/ConfigScreen.tsx
matthew-gordon: the TS compiler doesn't like the call though until explicitly checking the `AppInstallationParameters` field of `amplifyWebhookUrl` as a string, at that point it is still treating it as `SetStateAction<string>`

```typescript
Argument of type 'string | undefined' is not assignable to parameter of type 'SetStateAction<string>'.
  Type 'undefined' is not assignable to type 'SetStateAction<string>'
```

Can remove the default empty string from the `useState` call but then need to type that `as string` to overload the `string | undefined` which could potentially be missing on initial render. I feel like leaving it for now if we want to have a fast follow for cleanup of this?

apps/aws-amplify/src/lib/isVaildUrl.ts
matthew-gordon: Going to add a ticket for more robust validation on this when we actually have the business logic connected for the button trigger in a follow up PR

apps/aws-amplify/src/locations/ConfigScreen.tsx
matthew-gordon: I went a head and removed it actually 👍 will rely on the implict types

apps/aws-amplify/README.md
```diff
@@ -0,0 +1 @@
+# AWS Amplify
```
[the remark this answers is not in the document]
ryunsong-contentful: Okay, so you're saying until we get the end to end working we can hold off on the README? As long as we have a ticket to track that work, I'm good with that

apps/aws-amplify/src/locations/ConfigScreen.tsx
ryunsong-contentful: Wait, it complains? It works for me

```
  useEffect(() => {
    (async () => {
      const currentParameters = await sdk.app.getParameters();

      if (currentParameters) {
        setParameters(currentParameters);
        setAmplifyWebhookUrl(currentParameters.amplifyWebhookUrl);
      }

      sdk.app.setReady();
    })();
  }, [sdk]);
```

ryunsong-contentful (approved):

PR #3735, opened 2023-06-12
Integ 662 add tests to existing components

lilbitner:
## Purpose

The purpose of this PR is to add the first round of supplemental unit tests to the existing frontend components. 

## Approach

A spec file was added for each component that was without one - and each spec file should have at least one basic test. 
Some test files are more extensive than others, as certain components are going to change in the near future, so found there to be diminishing returns on spending too much time on, for example, the `ResourceCard` components. 

Additionally, there will be more tests added for the hooks and a few other components, in another PR. I wanted to keep momentum going by opening this PR for basic tests right now.


apps/ecommerce/frontend/src/components/App/Dialog/ResourceCard/ResourceCard.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+      />
+    );
+
+    const productName = externalResource.name!;
+    await userEvent.click(getByText(productName));
+
+    const card = getByTestId('cf-ui-card');
+
+    expect(mockOnSelect).toHaveBeenCalled();
+    expect(card).toBeEnabled();
```
whitelisab, line 44: Is this the right assertion here? I am wondering if this will return true, regardless of whether the card has been selected?

whitelisab (approved):
This looks great! 🔥 Agreed that we can fill in the gaps after we refactor some more of these components, but this is a great foundation.

apps/ecommerce/frontend/src/components/App/Dialog/ResourceCard/ResourceCard.spec.tsx
lilbitner: Yeah good point! It is always enabled, looks like the card itself doesn't have an easy attribute to assert being active vs not active. I have removed for now, especially with the future changes of this card 👍

PR #3736, opened 2023-06-12
[INTEG-654] Update Add Products button in ResourceField

whitelisab:
## Purpose

We had two buttons (Add Content and Add Products) in the Field location up until now. Now that we have come to a design decision, this PR reflects the new Add Products button states.

## Approach

I removed the old Add Content button and handlers. This new button has the provider logo and its text changes depending on whether it is a single or multiple field. For a single field, the text changes from "Add" to "Edit" when a resource is selected. Below are the various button states:

Multiple reference field:
Screenshot 2023-06-12 at 2 55 34 PM

Single reference field (item selected):
Screenshot 2023-06-12 at 2 55 55 PM

Single reference field (item not selected): 
Screenshot 2023-06-12 at 2 55 39 PM


apps/ecommerce/frontend/src/components/App/Field/ResourceField/ResourceField/ResourceField.tsx
```diff
@@ -8,26 +7,29 @@ import SortableResourceList from '../SortableResourceList/SortableResourceList';
...
   const { value } = useResourceValue(isMultiple);
   const sdk = useSDK<FieldAppSDK>();
   const { linkType } = sdk.parameters.instance;
   const { resourceType } = getResourceProviderAndType(linkType);
-  const buttonText = isMultiple ? `Add ${resourceType}s` : `Add a ${resourceType}`;
+  const buttonText = isMultiple
+    ? `Add ${resourceType}s`
+    : value.length
+    ? `Edit ${resourceType}`
+    : `Add ${resourceType}`;
```
lilbitner, line 22: Really like these changes you added in this file! A little nitpicky, but or this logic, I tend to avoid using nested ternaries for readability and complexity reasons, and fallback to a function with an if statement such as 

const getButtonText = () => {
  if (isMultiple) return `Add ${resourceType}s`;
  return value.length ? `Edit ${resourceType}` : `Add ${resourceType}`
}

or something along those lines!

apps/ecommerce/frontend/src/components/App/Field/ResourceField/ResourceField/ResourceField.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+  <div>Mock Sortable Resource List</div>
+));
+
+describe('ResourceField component', () => {
+  it('mounts', () => {
+    render(<ResourceField />);
+
+    expect(screen.getByTestId('cf-ui-button')).toBeInTheDocument();
+  });
+});
```
lilbitner, line 33: fabulous testing file ⭐

apps/ecommerce/frontend/src/components/App/Field/SingleResource/SingleResource.tsx
```diff
@@ -1,11 +1,36 @@
...
+        }
+
+        const data = await res.json();
+
+        setLogoUrl(data.logoUrl);
+      } catch (error: any) {
+        console.error(error.message);
+      }
+    })();
+  }, [sdk, cma]);
```
lilbitner, line 33: more of a larger thing to discuss, nonblocking, but it seems kind of funky to me that we have this logic into two places, for Single and Multiple resources - is there potentially a way to put this at an even higher level?

lilbitner (approved):
Looking awesome 💯 a few non-blocking comments!

apps/ecommerce/frontend/src/components/App/Field/ResourceField/ResourceField/ResourceField.tsx
whitelisab: Good call!

PR #3765, opened 2023-06-14
[INTEG-677] Remove deploy command from the ecomm lambda to have builds running again

ryunsong-contentful:
## Purpose
Builds won't deploy and work because of some limit lambdas imposes on header sizes (5120bytes)

## Approach
Remove deploy command until we have a real solution for this. Real solution will bring back this deploy command.

## Testing steps
Builds pass 

## Breaking Changes
None

None

## Deployment
Builds should start working again

### DIFF
--- a/apps/ecommerce/frontend/src/components/App/Field/SingleResource/SingleResource.spec.tsx
+++ b/apps/ecommerce/frontend/src/components/App/Field/SingleResource/SingleResource.spec.tsx
@@ -58,7 +58,7 @@ describe('SingleResource component', () => {
     expect(resourceField).toBeVisible();
   });
 
-  it('handles adding a resource ', async () => {
+  it.skip('handles adding a resource ', async () => {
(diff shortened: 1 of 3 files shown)


apps/ecommerce/frontend/src/components/App/Field/SingleResource/SingleResource.spec.tsx
```diff
@@ -59,19 +58,16 @@ describe('SingleResource component', () => {
   });
 
   it('handles adding a resource ', async () => {
```
lilbitner, line 60: is it easier to just skip this test instead of comment it out?

lilbitner: and perhaps also skip the MultipleResource.spec until figure out flakiness ?

lilbitner (approved):

PR #3781, opened 2023-06-15
[NONE] Update app id for aws-amplify

ryunsong-contentful:
## Purpose
Was pointing at the app id in the wrong org. Now we have the correct app id in the correct org

## Approach
Build should upload

## Testing step
Deploys to this app defn https://app.contentful.com/account/organizations/5EJGHo8tYJcjnEhYWDxivp/apps/definitions/1mVogDvuE0GuW4qp4dk4zQ/general


matthew-gordon (approved):
💪 ✅

PR #3784, opened 2023-06-15
Feat/integ 673 675

lilbitner:


apps/ecommerce/frontend/src/types.ts
```diff
@@ -98,3 +98,12 @@ export interface DialogInvocationParameters extends KeyValueMap {
...
 }
+
+// the variant states of ProductCard, for now, labeled by location
+export type ProductCardType = 'field' | 'dialog';
+
+export type ExternalResourceError = {
+  error: string;
+  errorMessage: string;
+  errorStatus: number;
+};
```
lilbitner, line 109: conflicted about this, I find the current setup of `useExternalResource` hook, with `error`, `errorMessage`, and `errorStatus` being returned as a bit confusing because I expect error to be the `error` object that encapsulates all of this. Any thoughts here appreciated!

apps/ecommerce/frontend/src/types.ts
```diff
@@ -98,3 +98,12 @@ export interface DialogInvocationParameters extends KeyValueMap {
   linkType: string;
   fieldType: FieldType;
 }
+
+// the variant states of ProductCard, for now, labeled by location
+export type ProductCardType = 'field' | 'dialog';
```
lilbitner, line 103: this for now is a simpler way to designate card variants, this will most likely me adjusted as we continue to figure out the product card mapping on the frontend.

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCard/ProductCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+      isHovered={isHovered}>
+      <ProductCardHeader
+        headerTitle={cardHeader}
+        status={resource.status}
+        handleRemove={handleRemove}
+        cardIndex={cardIndex}
+        totalCards={totalCards}
+        showJson={showJson}
+        handleShowJson={setShowJson}
+        showExternalResourceLinkDetails={!externalResourceError?.error && Boolean(externalResourceLink)} // do we want to show this icon button when there is an error, or the showData option?
```
lilbitner, line 78: question about showing the external link icon - as well as conditionally rendering the 'Show data' menu item if there exists an error!

lilbitner:
I notice there is some prettier failures, but I am a bit confused by the log from the build 🤔

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBody/ProductCardBody.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Grid.Item>
+        <Text>{productDescription}</Text>
+      </Grid.Item>
+      <Grid.Item>
+        <Text fontColor="gray600">{productId}</Text>
+      </Grid.Item>
+    </Grid>
+  );
+
+  // TODO: Handle missing image
```
lilbitner, line 52: perhaps a quick low hanging follow-up task to do this? scope creep was starting with this PR 😅

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBadge/ProductCardBadge.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const ProductCardBadge = (props: ProductCardBadgeProps) => {
+  const { showHeaderMenu, externalResourceError, resource = {} } = props;
+  const { error, errorMessage, errorStatus } = externalResourceError || {};
+
+  // TO DO: Ensure the status/badge state is consistent based on data provided
+  const getDisplayStatus = () => {
+    if (resource.status) return resource.status;
+    return resource.availableForSale ? 'Available' : 'Not Available'
+  }
```
lilbitner, line 25: perhaps this is just still in flux, but would love to align on the 'status' piece of a resource right now and how that renders via the badge.

whitelisab (approved):
Awesome work here! ⭐ Nothing truly blocking, just some comments/questions for you. You can decide whether we fix up the badge in this PR or in a follow up.

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBadge/ProductCardBadge.tsx
lilbitner: Perfect that sounds great!

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBody/ProductCardBody.tsx
```diff
@@ -0,0 +1,64 @@
...
+          lineHeight="lineHeightL"
+          isWordBreak={true}>
+          {productName}
+        </Text>
+      </Grid.Item>
+      <Grid.Item>
+        <Text>{productDescription}</Text>
+      </Grid.Item>
+      <Grid.Item>
+        <Text fontColor="gray600">{productId}</Text>
```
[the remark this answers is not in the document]
lilbitner: ah great eye will add! 🦅

apps/ecommerce/frontend/src/hooks/field/useResourceMenuItems.tsx
```diff
@@ -1,10 +1,10 @@
 import { Menu } from '@contentful/f36-components';
 import { useEffect, useMemo, useState } from 'react';
-import type { ResourceCardMenuProps } from 'components/App/Field/ResourceCard/ResourceCardMenu/ResourceCardMenu';
+import type { ProductCardMenuProps } from 'components/Common/ProductCard/ProductCardMenu/ProductCardMenu'; // adjust?????
```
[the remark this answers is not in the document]
lilbitner: ah whoops thanks for catching, nothing relevant now, will remove!

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCard/ProductCard.tsx
lilbitner: Yeah totally agree - sorry my wording was a bit confusing, I meant do we want this to show if an error exists? The default state is that the icon shows, but the conditional here in the code is that it only shows if no error exists

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCard/ProductCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+  cardHeader: string;
+  productCardType?: ProductCardType;
+  handleRemove?: (index?: number) => void;
+  onSelect?: (resource: ExternalResource) => void;
+  isSelected?: boolean;
+  dragHandleRender?: RenderDragFn;
+  isLoading?: boolean;
+  cardIndex?: number;
+  totalCards?: number;
+  externalResourceLink?: any;
```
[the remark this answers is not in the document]
lilbitner: ah yes 100% good catch!

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCard/ProductCardStories/mocks.ts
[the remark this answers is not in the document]
lilbitner: yeah totally - I did think about this and felt it was best to allow there to be one test file and one storybook file in case of customization, but perhaps that is putting the cart before the horse and over optimizing. I will go ahead and just pull from the test mocks for now and remove this file 👍 perhaps as we move forward, we can remove the mocks from just the test directory to their own directory if they will be used in stories as well!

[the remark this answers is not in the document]
lilbitner: oops okay now that I tried to consolidate, it looks like the test directory is at the same level as the storybook directory so I am running into errors if I pull from the already existing test mocks. I have decided to remove the local `ProductCard` stories mocks and opt for a mocks directory under the storybook directory, so they can at least me reusable. I think a technical task in the future is to consolidate all testing and storybook mocks (maybe move the testing mocks directory into the src directory?).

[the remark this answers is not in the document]
whitelisab: Yeah, totally fine for us to address this later and just keep the mocks as is!

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBadge/ProductCardBadge.tsx
```diff
@@ -0,0 +1,37 @@
+import { Badge, Box, EntityStatus } from '@contentful/f36-components';
+import { ExternalResource, ExternalResourceError } from 'types';
+import { styles } from './ProductCardBadge.styles';
+
+interface ProductCardBadgeProps {
+  showHeaderMenu?: boolean;
```
ryunsong-contentful, line 6: I think it's somewhat of an anti-pattern to have these simple badge components handle all this optional/nullable logic. There should be instead be multiple types of ProductCardBadges, one for the good state, and the other for missing states and let the parent handle which one to render. That way these simpler badge components are more re-usable

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBody/ProductCardBody.tsx
```diff
@@ -0,0 +1,64 @@
+import { Box, Flex, Grid, Text } from '@contentful/f36-components';
+import { ExternalResourceError } from 'types';
+
+interface ProductCardProps {
+  title?: string;
```
ryunsong-contentful, line 5: Similar comment here - it seems weird that all these props are optional when the ProductCard itself conceptually definitely should have props that must be required, like a title and description for example.

However, an image being optional makes sense to me.

Also some of these props if undefined will just print nothing at all - is this what we want? I suspect there is a more wholeistic design here if you make title and description required and then go from there.

apps/ecommerce/frontend/src/providers/ResourceFieldProvider.tsx
```diff
@@ -4,9 +4,9 @@ import { ReactNode } from 'react';
 export type ResourceFieldProviderProps = {
   isMultiple: boolean;
   handleAddResource: () => Promise<any[]>;
-  handleRemove: (index: number) => void;
-  handleMoveToTop?: (index: number) => void;
-  handleMoveToBottom?: (index: number) => void;
+  handleRemove: (index?: number) => void;
```
ryunsong-contentful, line 7: Repeat comment for completeness sake: Is the index now optional so that it works for both multiple and single?

apps/ecommerce/frontend/src/types.ts
ryunsong-contentful: What you did here I think is a great placeholder for now

ryunsong-contentful (commented):
Nice! Two main themes in the review
1. Algo improvement opportunity
2. Design level of base components and Product Card components

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBadge/ProductCardBadge.tsx
```diff
@@ -0,0 +1,37 @@
...
+  externalResourceError?: ExternalResourceError;
```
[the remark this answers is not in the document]
lilbitner: it is being used on line 19 to grab the error info!

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBody/ProductCardBody.tsx
lilbitner: yeah this is a good point. Tbh these types all extend from the original `ExternalResource` type that we have throughout the application thus far. I somewhat assumed given how in flux everything is, that these were all designated as optional and as part of the iterative process will be refined as required. I am happy to make that change as part of this effort and require a title and description because I think that logically makes sense. My understanding since joining the project is that this is more about "when", not an "if" we would make this change. Happy to adjust 👍

lilbitner: also within this component specifically, there might exist an error, and therefore the title and description might not be provided.

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBadge/ProductCardBadge.tsx
lilbitner: yeah this is something I certainly debated and is a good callout. The reason I decided to move the logic to this component itself was simply because the atom it is creating is already existent within F36, and any customization added is a very specific implementation of the component (limiting usability). Otherwise would certainly opt for just moving the logic to the parent component. Its kind of a fine line imo, congesting the parent with many different conditional renders, or pulling those out into a hook or some helper, or just adding simple logic to the child component if it is a very specific implementation, which imo this is a very specific implementation of the Badge and I somewhat foresee us adding more conditionals given the future differing states of a ProductCard. In general I think parents should handle conditional rendering until you reach several different variations, which is what I am noticing here, so I treat this more as an adjacent component handling logic, instead of a lower level atom. But that was just my original rationale and totally agree with your point 👍

apps/ecommerce/frontend/src/components/App/Dialog/ResourceList/ResourceList.tsx
```diff
@@ -17,13 +17,16 @@ const ResourceList = (props: Props) => {
   return (
     <Flex className={styles.productList}>
       {externalResources.map((item, index) => {
+        const isSelectedResource = selectedResources.find((selectedItem) => {
```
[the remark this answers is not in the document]
lilbitner: awesome idea, love it! Will implement 💯 🔥

apps/ecommerce/frontend/src/components/App/Field/MultipleResources/MultipleResources.tsx
```diff
@@ -66,23 +66,26 @@ const MultipleResources = () => {
     return Array.isArray(resources) ? resources : [];
   };
 
-  const handleRemove = (index: number) => {
+  const handleRemove = (index?: number) => {
+    if (!index) return;
```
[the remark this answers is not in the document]
lilbitner: The reason I had to make the "index" optional here is because this `ProductCard` may or may not exist within a larger context of a list at some point. And the `cardIndex` that is provided as a prop is optional, given that there is no need to provide a specific cardIndex outside of the context of the `SortableList` component. Admittedly it felt weird doing this, and I wen't back and forth a few times. Because I agree with you that the actions that handle this should require an index, but the actions themselves are so coupled to the `SortableList` component which is then using the reusable `ProductCard` so it felt justified doing this. Though I can add a default index somehow, either when from where the event is called or within these functions. What do you think?

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCard/ProductCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+    dragHandleRender,
+    externalResourceLink,
+    cardMovementCallbacks,
+    isHovered,
+    externalResourceError,
+  } = props;
+
+  const { handleMoveToBottom, handleMoveToTop } = cardMovementCallbacks || {};
+  const fieldProductCardType = productCardType === 'field';
+  const renderRawData = !externalResourceError?.error && fieldProductCardType && externalResourceLink && showJson;
```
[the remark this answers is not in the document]
lilbitner: ah yes this makes a ton of sense! Will adjust this conditional for now 👍

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBody/ProductCardBody.tsx
ryunsong-contentful: Hmm okay, given the news we got yesterday in standup - to me it's just a matter of time/effort/value evaluation since this work will be handed off. I think you'll have the best judgement here on whether or not we should or should not clean this up in this PR.

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBadge/ProductCardBadge.tsx
ryunsong-contentful: Basically you would pass in the Badge from the parent, and then the HOC component would return the modified Badge with whatever wrapper components around it.

Honestly, I don't think this optimization matters - I think there's a very good chance that the Platform team will re-design/build on top and change a lot of this. So I'll let them decide what to do here!

ryunsong-contentful (approved):

ryunsong-contentful (approved):
Approving because I don't think the last bit of design optimizations is a great use of your time and there's a good chance even if we optimized HOC badge component, it would be changed and refactored

apps/ecommerce/frontend/src/components/Common/ProductCard/ProductCardBadge/ProductCardBadge.tsx
lilbitner: ah okay I see, I like the idea, but agreed that things will be in flux with this!

PR #3802, opened 2023-06-19
EAP sign up - Configuration page (Commercetools & Shopify) [MONET-1253]

cemreyuksel:
## Purpose

[Jira Ticket ](https://contentful.atlassian.net/browse/MONET-1253)

## Approach

This PR adds a new prop to the `AppConfig` component in order to determine whether the orchestration EAP Note should be rendered.

There will be follow-up PRs for enabling the prop on Shopify and E-commerce market app repositories.

## Testing steps

Unit tests.

## Breaking Changes

How it looks: 
Screenshot 2023-06-19 at 14 57 58

View after note is dismissed: 
Screenshot 2023-06-19 at 15 00 58

### DIFF
--- a/packages/ecommerce-app-base/src/AppConfig/AppConfig.spec.tsx
+++ b/packages/ecommerce-app-base/src/AppConfig/AppConfig.spec.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, cleanup, fireEvent, waitFor } from '@testing-library/react';
+import { render, cleanup, fireEvent, waitFor, screen } from '@testing-library/react';
 
 import { AppExtensionSDK } from '@contentful/app-sdk';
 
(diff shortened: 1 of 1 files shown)


j-pea (approved):
Texted update, thanks

packages/ecommerce-app-base/src/interfaces.ts
```diff
@@ -164,6 +164,7 @@ export type OpenDialogFn = (
 export type DisabledPredicateFn = (currentValue: string[], config: Config) => boolean;
 
 export interface Integration {
+  isInOrchestrationEAP: boolean | undefined;
```
kdamball, line 167: nit: can also just be `isInOrchestrationEAP?: boolean`

packages/ecommerce-app-base/src/AppConfig/AppConfig.spec.tsx
```diff
@@ -62,6 +62,7 @@ const renderComponent = (sdk: unknown) => {
       logo="some-logo.svg"
       color="red"
       description="App description"
+      isInOrchestrationEAP={isInOrchestrationEAP ?? true}
```
ryunsong-contentful, line 65: Why is the default true? Isn't Orchestration opt-in which would mean the default being false would be a more accurate representation?

cemreyuksel: 👍 I will apply this change in the [commercetools PR](https://github.com/contentful/apps/pull/3855).

PR #3803, opened 2023-06-19
feat: add orchestration eap prop as true [MONET-1253]

cemreyuksel:
## Purpose

[Jira Ticket](https://contentful.atlassian.net/browse/MONET-1253)

This PR is a follow-up PR to add `isInOrchestrationEAP` prop to commercetools app so that we can render the EAP note.

Related PR: #3802


cemreyuksel:
Build will pass after #3802 is merged.

kdamball (approved):

PR #3859, opened 2023-06-22
Fix MuxUploader double drop area

dylanjha:
By default `MuxUploader` will render a drop, since we made our own drop with `MuxUploaderDrop` use the `noDrop` property to prevent `MuxUploader` from rendering the default one

## Purpose

Current bug in Mux Uploader shows 2 drop areas

Screenshot 2023-06-21 at 11 35 55 AM

## Approach

There was a slight change between `beta.8` and `beta.9` that included `MuxUploader` rendering a drop area by default.

See docs: https://docs.mux.com/guides/video/mux-uploader

## Breaking Changes

None

### DIFF
--- a/apps/mux/src/index.tsx
+++ b/apps/mux/src/index.tsx
@@ -1074,6 +1074,7 @@ export class App extends React.Component<AppProps, AppState> {
                 type="bar"
                 onSuccess={this.onUploadSuccess}
                 endpoint={this.getUploadUrl}
+                noDrop
                 //onError={this.onUploadError}
                 style={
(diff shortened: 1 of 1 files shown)


ryunsong-contentful (commented):
There is a build failure but happy to approve once the build failure is fixed!

whitelisab (approved):
I tested this locally and it looks great. Thank you for this contribution @dylanjha! Apologies for the delay in getting this approved, we were working through some issues with our CI checks for external contributions and just got that resolved.

PR #3876, opened 2023-06-23
Fix/revert revert

jsdalton:
## Purpose

The previous revert from #3875 needs to be reverted, so we can restore the functionality that was disabled. This is possible now that the build is unbroken and the fix published.

## Approach

* Revert the revert
* Pin to correct ecomm package


matthew-gordon (approved):
🧹 ✨ LGTM

PR #3890, opened 2023-06-26
[INTEG-762] Adds signed request verification logic for SAP AIR

ryunsong-contentful:
## Purpose
To get around cors we built a backend but that means we need to do some verification signed request logic to ensure property security practices.

Next step will be to port over api logic and hook up the frontend to backend

## Testing steps
Able to make frontend request from the backend from a Contentful Context - otherwise rejects the request

## Breaking Changes
None

None

## Deployment
Standard


apps/sap-commerce-cloud-with-air/lambda/src/middlewares/verifySignedRequests.ts
```diff
@@ -0,0 +1,58 @@
...
+  const headers = { ...req.headers };
+
+  // coerce all headers to strings
+  Object.keys(headers).forEach((key) => {
+    headers[key] = headers[key]?.toString();
+  });
+
+  // express.json() makes body always an object ({}), even if there is no body
+  // so we need to check if the body is empty and set it to undefined
+  const bodyOrUndefined = Object.keys(req.body).length === 0 ? undefined : JSON.stringify(req.body);
```
lilbitner, line 51: `isEmpty` from lodash is super useful for this check in the future!

apps/sap-commerce-cloud-with-air/lambda/src/middlewares/getAppInstallationParameters.ts
```diff
@@ -0,0 +1,75 @@
...
+    const appId = JSON.parse(JSON.stringify(req.header('x-contentful-app')));
+    const spaceId = JSON.parse(JSON.stringify(req.header('x-contentful-space-id')));
+    const environmentId = JSON.parse(JSON.stringify(req.header('x-contentful-environment-id')));
+    req.params.appId = appId;
+    req.params.spaceId = spaceId;
+    req.params.environmentId = environmentId;
+
+    if (!appId || !spaceId || !environmentId) {
+      throw new Error('Missing required headers!');
+    }
```
lilbitner, line 25: just for my understanding of the SAP workflow, why are these required? I see why they were used in ecommerce project, this is to check credentials for GET requests prior to them saving the config (so then we can subsequently access via CMA?)?

apps/sap-commerce-cloud-with-air/lambda/src/config.ts
```diff
@@ -0,0 +1,23 @@
...
+  const environmentVariableValue = process.env[environmentVariableName];
+  if (!environmentVariableValue) {
+    if (fallback) return fallback.toString();
+    throw new Error(`Missing environment variable: '${environmentVariableName}'`);
+  }
+  return environmentVariableValue;
+}
+
+export const config = {
+  stage: getEnvironmentVariable('STAGE'),
```
lilbitner, line 16: what does "stage" mean in this context? (again this is curiosity question). Based on the code context, assuming this is just any environment besides production?

lilbitner (commented):
Added a few comments, mostly questions for my own understanding! Looking great :)

apps/sap-commerce-cloud-with-air/lambda/src/middlewares/corsConfig.ts
```diff
@@ -0,0 +1,29 @@
...
+    // Contentful SDK Headers
+    'X-Contentful-Timestamp',
+    'X-Contentful-Signed-Headers',
+    'X-Contentful-Signature',
+    'X-Contentful-User-ID',
+    'X-Contentful-Space-ID',
+    'X-Contentful-Environment-ID',
+    'X-Contentful-App-ID',
+    // ContentfulServiceAccountHeaders
+    'X-Contentful-ServiceAccountKeyId',
```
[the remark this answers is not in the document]
ryunsong-contentful: Lazy copy pasta from Ecommerce lambda

apps/sap-commerce-cloud-with-air/lambda/src/config.ts
ryunsong-contentful: Yeah there's some serverless stuff that defines the env vars of what stage we are in (and all of these env vars really) and then in the application code we can flexibly write if-else logic based on the stages we are in. Not a thing that will be necessarily used, but more of a future-proof type of useful to have type of deal

apps/sap-commerce-cloud-with-air/lambda/src/middlewares/getAppInstallationParameters.ts
ryunsong-contentful: Excellent question. If you don't mind instead of writing this all out - I can just tuple with you and explain in person!

lilbitner (approved):
Looking good 🔥

PR #3938, opened 2023-06-29
config page done

rusticpenguin:


rusticpenguin:
Merged in with the AICG feature branch

PR #3956, opened 2023-06-30
[INTEG-720] - App Actions Backend Proxy

matthew-gordon:
## Purpose

This adds MVP functionality for triggering amplify builds from a user supplied webhook using hosted app actions.

view the action [here](https://github.com/contentful/apps/blob/mjg/feat/add-hosted-actions/apps/aws-amplify/actions/amplify.ts)

## How to test

With access:

1.) visit the amplify staging space [here](https://app.contentful.com/spaces/b7sxrca414e5/environments/master)

2.) make sure the Amplify app is installed (staging)

3.) Navigate to a piece of content, and click `Build now` button to initiate a build

4.) Observe build status in the Amplify console


apps/aws-amplify/src/lib/localStorage.ts
```diff
@@ -0,0 +1,19 @@
+export const build_status_key = '__contentful_build_timestamp__';
```
matthew-gordon, line 1: this is temporary until we add a persistence layer

apps/aws-amplify/actions/amplify.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const handler = async (payload: AppActionCallParameters, context: AppActionCallContext) => {
+  const { amplifyWebhookUrl } = payload;
+  try {
+    await fetch(amplifyWebhookUrl, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
```
ghepting, line 15: post w/out body intentional/how it works yeah?

ghepting: I would have expected that `AppActionCallParameters` or *something* was passed through

matthew-gordon: Yep it is intentional, could pass an empty `{}` but wasn't necessary, absolutely 0 info passed/recieved from triggering the build though

apps/aws-amplify/src/locations/Sidebar.tsx
```diff
@@ -50,9 +52,10 @@ const Sidebar = () => {
         isFullWidth={true}
         isLoading={isLoading}
         isDisabled={isLoading}
-        onClick={buildAmplifyApp}>
-        {isLoading ? 'Building' : 'Build website'}
+        onClick={handleBuildAppActionCall}>
+        {isLoading ? 'Initiating build...' : 'Build now'}
       </Button>
+      <span>Last build started: {lastBuildInitiated}</span>
```
whitelisab, line 58: It seems odd to me to have this `Button` and `span` rendering within a `Paragraph` component. I would suggest using the `Paragraph` component for the `Last build started` text and then move the button outside.

apps/aws-amplify/src/locations/__tests__/ConfigScreen.spec.tsx
whitelisab, line 1: This is something we should document somewhere, but we've started a convention in many of our apps of having the test files and css files within the same directory as their component. Currently the test and css for the ConfigScreen component are located in different directories. 

We also had the location components (`ConfigScreen` and `Sidebar`) hold little logic and just render other components that are located in the components directory.

apps/aws-amplify/src/locations/Sidebar.tsx
whitelisab, line 58: Also, there could probably be some more padding between the button at the `Last build started` text.

apps/aws-amplify/src/locations/Sidebar.tsx
```diff
@@ -50,9 +52,10 @@ const Sidebar = () => {
         isFullWidth={true}
         isLoading={isLoading}
         isDisabled={isLoading}
-        onClick={buildAmplifyApp}>
-        {isLoading ? 'Building' : 'Build website'}
+        onClick={handleBuildAppActionCall}>
+        {isLoading ? 'Initiating build...' : 'Build now'}
```
whitelisab, line 56: Eventually, we should add some error handling here and useful messages to the user. I am getting a `404 AppAction does not exist` and the button is stuck in the `Initiating build` state.

whitelisab (approved):
Great job on this hosted app actions solution! I am approving since none of the comments I added are blocking. I also had two general questions:

1. When I run `npm install` I get a diff on my `package-lock.json` file
2. I am not able to get it working locally, I am probably missing a step? When I click the button, I am seeing an error `AppAction does not exist`

apps/aws-amplify/src/locations/Sidebar.tsx
matthew-gordon: This is tricky with the hosted app actions currently there is no documentation and they dont execute locally 🤔 this will definitely need to be handled at some point beyond the limited MVP we have flushed out here

apps/aws-amplify/src/locations/Sidebar.tsx
matthew-gordon: This was left over from another Dev's branch, but agreed elements shouldn't be nested under a paragraph tag

PR #3959, opened 2023-06-30
Add logging for help debug production sap-air 404 issues with lambda

ryunsong-contentful:
## Purpose
Getting 404's and can't tell what URL they are hitting so need logging

## Testing steps
None as they are just console.logs

## Breaking Changes
None

None

## Deployment
Standrard


rusticpenguin (approved):
Looks good!

PR #3987, opened 2023-07-04
feat: upgrade templates to React 18

andipaetzold:
Update all projects in `examples` to React 18. Projects using dam/ecommerce-app-base could not be updated as these libraries don't support React 18 (yet).

This sometimes required updating F36 or Testing Library to a compatible version.


colomolo (approved):
:shipit:

PR #3994, opened 2023-07-05
Adjust configs

lilbitner:
## Purpose

This PR removes extraneous code from other app config templates. It also consolidates private-key and signing secret logic to config files.


whitelisab (approved):
A++

PR #3995, opened 2023-07-06
Fix: Filter out deleted files in prettier check

whitelisab:
## Purpose

Currently, any changed files (added, modified, or deleted) are being included in the prettier check. Adding a filter to exclude deleted files from the prettier check to avoid getting an error of `No files matching the pattern were found` on this check. Also removed the `HEAD` from this line since we are already getting that in the line above for `MERGE_BASE`.

Fun fact, using the lowercase `d` in the filter excludes that status whereas an uppercase `D` includes that status in the filter (https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB82308203).


lilbitner (approved):
Woo!!

PR #4010, opened 2023-07-06
[INTEG-786] Fix: Remove Headers from SAP

whitelisab:
## Purpose

Remove old code from the SAP app where we were beginning to test out the AIR header.


lilbitner (approved):
Looks good to me! Not sure I have all the context, but just assuming the `applicationInterfaceKey` logic was added here originally before pivoting to creating a new app 🙃 ?

whitelisab:
> 

@lilbitner exactly, this was just cleaning up leftover code from the first attempt before a new app was created!

PR #4116, opened 2023-07-12
adds hooks for config handling, imports react-query

rusticpenguin:
## Purpose

Utilizing hooks and reducers we can push code to the edge, rather than having it live within components. This gives us the added benefit of predictable code and simplified testing since the components can be tested on predictable hook/reducer outputs

Other benefits for reducers:  
- Reduces developer error as dispatches require an enum for the specific dispatch
- Each dispatch event is strictly typed
- Less cognitive load/easier time debugging as the developer only has to remember [parameter, dispatch] as opposed to each individual useState
- Dispatch-driven state allows for an easy entry point for validation


apps/ai-content-generator/src/components/config/ConfigForm.tsx
```diff
@@ -0,0 +1,10 @@
+interface AppInstallationParameters {
+  model: string;
+  apiKey: string;
+  profile: string;
+}
+
+const ConfigForm = () => {};
```
matthew-gordon, line 7: noop for a follow up here?

apps/ai-content-generator/src/hooks/config/useInitializeParameters.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    dispatch({
+      type: ParameterAction.CONTENTFUL_PARAMETERS,
+      value: parameters,
+    });
+
+    sdk.app.setReady();
+  };
+
+  useQuery(['appParameters'], getCurrentParameters(sdk), {
```
matthew-gordon, line 23: does this value need to be returned from the hook? Not all the way familiar with React Query

matthew-gordon (approved):
Just some clarifying question but nothing blocking, super clean! 👍 ❇️

apps/ai-content-generator/src/components/config/ConfigForm.tsx
rusticpenguin: Yup! This will be added to next PR

apps/ai-content-generator/src/hooks/config/useInitializeParameters.ts
rusticpenguin: All good! So when the query resolves (the query in this case is getCurrentParameters) it'll call `dispatchParameters` which will update the state of the app installation parameters the parent.

This is the code from the next PR with the hook being used:

```
const ConfigForm = () => {
  const [parameters, dispatchParameters] = useReducer(parameterReducer, initialParameters);

  useSaveConfigHandler(parameters);
  useInitializeParameters(dispatchParameters);
  
...
```

So the OOO is:
1. Component Loads
2. `useInitializeParameters` runs
3. `getCurrentParemeters` resolves with the InstallationParameters
4. The InstallationParameters are dispatched
5. `parameters` is now updated with the current InstallationParameters from Contentful

matthew-gordon: makes sense!

ryunsong-contentful (changes requested):
Awesome stuff. Loving the initial structure and form you set out here. Good work here

The comments I have are really about cleaning things up and making sure we avoid over-abstracting things unnecessarily.

apps/ai-content-generator/src/components/config/parameterReducer.ts
```diff
@@ -0,0 +1,38 @@
...
+  MODEL = 'model',
+  APIKEY = 'apiKey',
+  PROFILE = 'profile',
+  CONTENTFUL_PARAMETERS = 'contentfulParameters',
+}
+
+const { MODEL, APIKEY, PROFILE, CONTENTFUL_PARAMETERS } = ParameterAction;
+
+type ParameterStringActions = {
+  type: typeof MODEL | typeof APIKEY | typeof PROFILE;
```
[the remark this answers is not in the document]
rusticpenguin: So there's two ways we can do a list of enum values here:

First way is referencing the enum directly, so we have the proper typing
```js
type ParameterStringActions = {
  type: ParameterAction.MODEL | ParameterAction.APIKEY | ParameterAction.PROFILE;
...
```

Second way is if we want to destructure ParameterAction, like we are here, we have to use typeOf. Since we are using the enum typing here, not the actual value. 

Eg. if we don't use typeOf, the destructured enum's type will be unresolved/any: 
image

In the case that we use typeOf, the typeOf resolves to `ParameterAction.<EnumName>`
image

apps/ai-content-generator/src/components/config/parameterReducer.ts
```diff
@@ -0,0 +1,38 @@
+import { AppInstallationParameters } from './ConfigForm';
+
+export enum ParameterAction {
```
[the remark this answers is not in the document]
rusticpenguin: That would work as well, but I feel an enum would be more descriptive in this case. As it's a predefined set of constants
https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

apps/ai-content-generator/src/react-query/queries/getCurrentParameters.ts
```diff
@@ -0,0 +1,11 @@
+import { AppInstallationParameters } from '@/components/config/ConfigForm';
```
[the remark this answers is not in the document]
rusticpenguin: Agreed, let's get this out of react-query

apps/ai-content-generator/src/components/config/parameterReducer.ts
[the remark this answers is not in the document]
rusticpenguin: I found a way to make this look cleaner than destructuring and `or`-ing the keys: we can instead use `Exclude` in TS

```ts
type ParameterStringActions = {
  type: Exclude<ParameterAction, ParameterAction.CONTENTFUL_PARAMETERS>;
  value: string;
};
type ParameterObjectActions = {
  type: ParameterAction.CONTENTFUL_PARAMETERS;
  value: AppInstallationParameters;
};
```

PR #4118, opened 2023-07-12
refactors config page

rusticpenguin:
## Purpose

Chunking the config page into small UI pieces alleviates the cognitive load on the developer. As well as leaning into hooks and reducers leads to the component's code being simple and descriptive


apps/ai-content-generator/src/index.tsx
```diff
@@ -6,7 +6,7 @@ import { SDKProvider } from '@contentful/react-apps-toolkit';
 import App from './App';
 import LocalhostWarning from '@components/common/LocalhostWarning';
 import { QueryClientProvider } from 'react-query';
-import queryClient from '@/react-query';
+import queryClient from '@react-query/index';
```
matthew-gordon, line 9: is this meant to pull from the local module or the installed package?

matthew-gordon (approved):
🏆 ✅

apps/ai-content-generator/src/index.tsx
rusticpenguin: I was having an issue with the alias directly resolving to the module. Adding an alias that points directly to it, I am able to fix this.

I'll add this to the next PR

PR #4130, opened 2023-07-13
Thank you message

gavinmatt:
## Purpose

Adding a thank you message to capture the git history for the original authors of the app!


lilbitner (approved):
Looks good to me 👍

PR #4148, opened 2023-07-14
Feature [INTEG-810]: Add new custom commercetools app with different API

whitelisab:
## Purpose

This PR introduces a new app based off of our existing commercetools app. This shows an example of how you can customize an app for a specific use case. In this example, if a customer doesn't have indexing activated in their commercetools project, then our marketplace app doesn't work properly since the search API relies on indexing being active. This app changes the API endpoint used, but then does not fully support full-text search. It does support searching for a full product name in the picker.

## Approach

This is pretty much all a copy of the existing commercetools app, the changes are in the `api/products.ts` file. Here you can see the the [Product Projections API endpoint](https://docs.commercetools.com/api/projects/productProjections#query-productprojections) is used instead of the [Product Projections Search API endpoint](https://docs.commercetools.com/api/projects/products-search).

I purposely did not add a deploy script in the `package.json`, as this app is not listed in the marketplace.

## Testing steps

We have spaces in our testing org for local development and production. You should be able to add products/categories to an entry as usual. The search functionality in the product picker will be different. You can only search for full names of products.


apps/commercetools-without-search/src/api/products.ts
```diff
@@ -0,0 +1,91 @@
...
+  config: ConfigurationParameters
+): Promise<Product[]> {
+  if (skus.length === 0) {
+    return [];
+  }
+
+  const searchQuery = skus
+    ? {
+        queryArgs: {
+          where: `masterVariant(sku in (${skus.map((sku) => `"${sku}"`).join(',')}))`,
```
ryunsong-contentful, line 34: Strange but trust I shall!

ryunsong-contentful (approved):
I mainly focused on the changes you highlighted and didn't focus on the other files too much. But it seems like the overall code is pretty good and clean.

Well done!

PR #4149, opened 2023-07-14
Improve error logging in Slack app and fix bug

lilbitner:
## Purpose

This PR adjusts the following: 

1. Adds more meaningful messages and identifiers to errors thrown within the Slack app lambda. 
2. Fixes [this](https://contentful.atlassian.net/browse/INTEG-652) issue. 

Also as you will notice, lots of trailing commas are added as per the prettier commit hook. I am not sure if this is standard, but I have trouble reverting the changes locally given the commit hook does them, though I could via github if they are an issue. Please let me know your thoughts!

## Approach

1. This PR adjusts the `Exception` class within the API to handle the addition of identifying information like `environmentId` and `spaceId`. Additionally, more meaningful error messages are provided when the errors are manually thrown. 
2. This was a small issue where a snackbar was rendering and indicating to the user that an error occurred, despite successful installation. This snackbar is being initialized from the `useConnect` component which is the popup that renders when connecting to Slack via the auth flow. The error being thrown that is causing this snackbar is simply a race condition of the API grabbing installation parameters via the `CMA` just a little before the app is actually labeled as "installed". I believe the only reason this is occurring is because the `EventsService` class, that fetches these installation parameters on subsequent requests, is being instantiated when the auth token is created (`POST` call to `/api/tokens`) in the `useConnect` popup auth flow. The error is not meaningful and would require a bit of an overhaul to completely prevent it from being thrown (as far as I understand, but please let me know if you believe otherwise!), so for now to improve user experience, this PR just renders an error snackbar if the error is anything other than this specific race condition error. 

## Testing steps

Uninstall and install the Slack app in your local environment, and confirm that all works correctly and no extraneous error messages appear.

### DIFF
--- a/apps/slack/frontend/src/useConnect.ts
+++ b/apps/slack/frontend/src/useConnect.ts
@@ -38,7 +38,8 @@ export const useConnect = () => {
 
           // this error is a race condition that occurs when instantiating the eventsService within the lambda.
           // the error is not valid to show to the user.
-          if (JSON.parse(error?.message)?.details !== 'AppInstallation does not exist.') {
+          const errorMessage = JSON.parse(error?.message);
+          if (errorMessage && errorMessage.details !== 'AppInstallation does not exist.') {
             console.error(e);
             sdk.notifier.error(genericErrorMessage);
           }
--- a/apps/slack/frontend/src/useConnect.ts
+++ b/apps/slack/frontend/src/useConnect.ts
@@ -34,7 +34,7 @@ export const useConnect = () => {
 
           await apiClient.createAuthToken(sdk, cma, token, installationUuid);
         } catch (e) {
-          const error = e as any;
+          const error = e as Error;
 
           // this 'AppInstallation does not exist error' is a race condition that occurs when instantiating the eventsService within the lambda.
           // it is not a valid error to show to the user.
--- a/apps/slack/frontend/src/useConnect.ts
+++ b/apps/slack/frontend/src/useConnect.ts
@@ -46,7 +46,7 @@ export const useConnect = () => {
         }
       }
     },
-    [cma, sdk, temporaryRefreshToken],
+    [cma, sdk, temporaryRefreshToken]
   );
 
   const { addConnectedWorkspace, setWorkspaceState, setNotificationsLoading, setChannels } =
@@ -81,7 +81,7 @@ export const useConnect = () => {
         sdk.notifier.success('Connected to the Slack workspace successfully.');
       } catch (e) {
         sdk.notifier.error(
-          'Something went wrong while authenticating with Slack. Please try again.',
+          'Something went wrong while authenticating with Slack. Please try again.'
         );
       } finally {
         setNotificationsLoading(false);
--- a/apps/slack/frontend/src/useConnect.ts
+++ b/apps/slack/frontend/src/useConnect.ts
@@ -34,12 +34,18 @@ export const useConnect = () => {
 
           await apiClient.createAuthToken(sdk, cma, token, installationUuid);
         } catch (e) {
-          console.error(e);
-          sdk.notifier.error(genericErrorMessage);
+          const error = e as any;
+
+          // this error is a race condition that occurs when instantiating the eventsService within the lambda.
+          // the error is not valid to show to the user.
+          if (JSON.parse(error?.message)?.details !== 'AppInstallation does not exist.') {
+            console.error(e);
+            sdk.notifier.error(genericErrorMessage);
+          }
         }
       }
     },
-    [cma, sdk, temporaryRefreshToken]
+    [cma, sdk, temporaryRefreshToken],
   );
 
   const { addConnectedWorkspace, setWorkspaceState, setNotificationsLoading, setChannels } =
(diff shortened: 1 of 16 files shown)


apps/slack/frontend/src/useConnect.ts
```diff
@@ -34,12 +34,19 @@ export const useConnect = () => {
...
-          sdk.notifier.error(genericErrorMessage);
+          const error = e as Error;
+
+          // this 'AppInstallation does not exist error' is a race condition that occurs when instantiating the eventsService within the lambda.
+          // it is not a valid error to show to the user.
+          const errorMessage = JSON.parse(error?.message);
+          if (errorMessage && errorMessage.details !== 'AppInstallation does not exist.') {
+            console.error(e);
+            sdk.notifier.error(genericErrorMessage);
+          }
```
lilbitner, line 45: this is where, as per the PR description, we are allowing any other error to be handled appropriately but selecting out the one ignorable error.

apps/slack/lambda/lib/routes/auth-token/repository.ts
```diff
@@ -64,41 +75,36 @@ export class AuthTokenRepository {
...
-    );
+    const spaceEnvUuid = AuthTokenRepository.uuid(spaceId, environmentId, workspaceId);
 
     // installationUuid is preferred, spaceEnvUuid is for backwards compatibility
     const possibleUuids = installationUuid || spaceEnvUuid;
     let authToken = await this.singleTableClient.get(Entity.AuthToken, possibleUuids);
 
     if (!authToken) {
-      throw new NotFoundException('Entity could not be found');
+      throw new NotFoundException({ errMessage: 'Auth token not found', environmentId, spaceId });
```
whitelisab, line 96: Yes! These new message instead of `Entity could not be found` will be helpful!

apps/slack/lambda/lib/errors/exception.ts
```diff
@@ -1,10 +1,13 @@
+import { ErrorObject } from 'ajv/dist/types';
```
whitelisab, line 1: I will have to look into avj, I haven't used it before!

apps/slack/lambda/lib/errors/exception.ts
```diff
@@ -31,3 +34,10 @@ export class Exception<T> extends Error {
 
 export const isException = (e: unknown): e is Exception<unknown> =>
   typeof e == 'object' && !!e && 'isException' in e;
+
+export type ExceptionDetails = {
```
whitelisab, line 38: Nice, this is great to pass in these additional details when possible

apps/slack/lambda/lib/errors/not-found.ts
```diff
@@ -1,7 +1,7 @@
-import { Exception } from './exception';
+import { Exception, ExceptionDetails } from './exception';
 
 export class NotFoundException extends Exception<string> {
```
whitelisab, line 3: Should this be updated to `export class NotFoundException extends Exception<ExceptionDetails> {` since we are no longer passing in a string? Same thing in a few other files.

whitelisab (commented):
This is great, thanks for fixing the error message bug and for adding for useful messages within the lambda! 🎊 

I think ideally it would be nice to not have the prettier fixes in this PR, if possible. Have you tried `git commit` with the `--no-verify` option to skip the pre-commit hook?

apps/slack/frontend/src/useConnect.ts
```diff
@@ -34,8 +34,15 @@ export const useConnect = () => {
 
           await apiClient.createAuthToken(sdk, cma, token, installationUuid);
         } catch (e) {
-          console.error(e);
-          sdk.notifier.error(genericErrorMessage);
+          const error = e as Error;
+
+          // this 'AppInstallation does not exist error' is a race condition that occurs when instantiating the eventsService within the lambda.
+          // it is not a valid error to show to the user.
+          const errorMessage = JSON.parse(error?.message);
```
ryunsong-contentful, line 41: Why would error ever be undefined or null in this catch block?

apps/slack/frontend/src/useConnect.ts
ryunsong-contentful: I'm assuming you checked but don't want to review based off assumptions. Is there an error code or something more unique we can attach to the check rather than a string check or is this our best bet?

apps/slack/lambda/lib/errors/exception.ts
lilbitner: yes! also new to me!

apps/slack/lambda/lib/errors/not-found.ts
lilbitner: good eye! adjusted!!

ryunsong-contentful (approved):
Overall great stuff! Just minor comments and double checking if there's a better stronger way to identify the one error that's really not an error

apps/slack/frontend/src/useConnect.ts
lilbitner: ah yeah I think I originally had escape character on `error?.message?.details` to handle if one of those objects does not exist, though I think you are right we don't need the first escape character since that is just handling and undefined error. Will adjust! Thank you!

ryunsong-contentful (approved):

PR #4156, opened 2023-07-14
create interstitial modal, remove react-query dependency

rusticpenguin:
## Purpose

Creates an interstitial modal to display AI features


apps/ai-content-generator/src/hooks/dialog/useEntryAndContentType.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useState } from 'react';
+import { ContentTypeProps, EntryProps } from 'contentful-management';
+import { DialogAppSDK } from '@contentful/app-sdk';
+import { useSDK } from '@contentful/react-apps-toolkit';
+
+/**
+ * This hook is used to get the entry and content type of the entry.
```
matthew-gordon, line 7: 🎊

matthew-gordon (approved):
This looks great, coming together amazing 🙇 🙌

apps/ai-content-generator/src/hooks/dialog/useAIFeature.tsx
```diff
@@ -0,0 +1,17 @@
+import featureConfig, { AIFeature, AIFeatureType } from '@configs/features/featureConfig';
+import { FeatureComponentProps } from '@configs/features/featureTypes';
+import { FC } from 'react';
+
+const useAIFeature = (aiFeature?: AIFeature): FC<FeatureComponentProps> | null => {
```
ryunsong-contentful, line 5: This does not need to be a hook and it doesn't follow the hook conventions - Hooks are mainly used to abstracting out useEffects/useMemos/other native hooks. This is just a regular old switch case function that can just be part of the base dialog

apps/ai-content-generator/src/configs/features/featureConfig.tsx
```diff
@@ -12,31 +13,41 @@ enum AIFeature {
   SEO_KEYWORDS = 'seoKeywords',
 }
 
+export enum AIFeatureType {
```
ryunsong-contentful, line 16: I dont understand why we need this. All you need is the key such as AIFeature.TITLE/AIFeature.CONTENT that is consumed by your Dialog component. Your base Dialog component should contain the switch-case logic and return the component to render based on that.

That base Dialog component should contain the mapping, you don't need to define any types here.

PR #4157, opened 2023-07-14
Fix [INTEG-529]: Display Sidebar Note when GA4 installation params are missing

whitelisab:
## Purpose

When the GA4 app is programmatically installed without the necessary app installation parameters, the Sidebar app was throwing errors for a missing `serviceAccountKeyId` and/or a missing `propertyId`. 

## Approach

This PR handles checking whether these installation parameters exist, and then displaying a helpful message if they don't exist rather than throwing an error and not rendering anything in the Sidebar. The Sentry errors we are seeing should be resolved by this.

Screenshot 2023-07-13 at 9 10 41 PM

## Testing steps

Change the values of the app installation parameters in the Sidebar component to undefined to see the Note in the Sidebar.

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
+++ b/apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
@@ -1,3 +1,5 @@
+import { ServiceAccountKeyId } from 'types';
+
 export const DEFAULT_ERR_MSG =
   'An unknown error has occurred. Please try again or contact support.';
 export const EMPTY_DATA_MSG =
@@ -14,3 +16,20 @@ export const getContentTypeSpecificMsg = (contentTypeName: string) => ({
   noSlugContentMsg: `This ${contentTypeName} entry does not have a valid slug field. Please add a field of type short text to this entry and configure it on the app configuration page.`,
   notPublishedMsg: `This ${contentTypeName} entry has not yet been published. Please publish the entry to view this Google Analytics 4 sidebar app.`,
 });
+
+export const getAppInstallationParamsMsg = (
+  serviceAccountKeyId: ServiceAccountKeyId | undefined,
+  propertyId: string | undefined
+) => {
+  const serviceAccountMsg = 'Service Account Key';
+  const propertyMsg = 'Google Analytics 4 property';
+  let missingParams = '';
+
+  if (!serviceAccountKeyId && !propertyId)
+    missingParams = `${serviceAccountMsg} and ${propertyMsg}`;
+  else {
+    !serviceAccountKeyId ? (missingParams = serviceAccountMsg) : (missingParams = propertyMsg);
+  }
+
+  return `No ${missingParams} provided or found in app installation parameters. Please update your app configuration page. `;
+};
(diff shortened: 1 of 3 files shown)


apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
```diff
@@ -14,3 +16,20 @@ export const getContentTypeSpecificMsg = (contentTypeName: string) => ({
   noSlugContentMsg: `This ${contentTypeName} entry does not have a valid slug field. Please add a field of type short text to this entry and configure it on the app configuration page.`,
   notPublishedMsg: `This ${contentTypeName} entry has not yet been published. Please publish the entry to view this Google Analytics 4 sidebar app.`,
 });
+
+export const getAppInstallationParamsMsg = (
+  serviceAccountKeyId: ServiceAccountKeyId | undefined,
```
ryunsong-contentful, line 21: Do these two properties need to be undefined?

apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
```diff
@@ -14,3 +16,20 @@ export const getContentTypeSpecificMsg = (contentTypeName: string) => ({
...
+  propertyId: string | undefined
+) => {
+  const serviceAccountMsg = 'Service Account Key';
+  const propertyMsg = 'Google Analytics 4 property';
+  let missingParams = '';
+
+  if (!serviceAccountKeyId && !propertyId)
+    missingParams = `${serviceAccountMsg} and ${propertyMsg}`;
+  else {
```
ryunsong-contentful, line 30: I'm assuming what you're trying to do here is rendering the appropriate message when both the `serviceAccountId` and `propertyId` is missing or if one of them is missing.

But your else logic here assumes that at least one of the two are always missing. For example, what if there is a `propertyId`? Then it would render `(missingParams = propertyMsg);` still.

So it's one of two things
1. We always assume that at least one of the two properties are missing, but it's not exactly clear from the code that this is the assumption.
2. This is a bug in the code in the case where there is a propertyId

ryunsong-contentful (approved):
Approving as this code works and I can't really think of any quick solutions off the top of my head for the design improvements. However I challenge you (in an encouraging fun puzzle way) to potentially figure something cleaner here. Don't spend too much time on it and if you can't figure it out - no worries, I'd say just move on

apps/google-analytics-4/frontend/src/components/main-app/constants/noteMessages.ts
ryunsong-contentful: Actually I think this is best without investing too much time. Just do if checks forget using elses and then just add a comment that one of the two params are guaranteed to be null. Also I would adjust the function name to be more in tune with the code logic.

```
  if (!serviceAccountKeyId && !propertyId) missingParams = `${serviceAccountMsg} and ${propertyMsg}`;
  if (!serviceAccountKeyId) missingParams = `${serviceAccountMsg};
  if (!propertyId) missingParams = `${propertyMsg};
```

whitelisab: Thanks for this feedback! I made these adjustments.

PR #4158, opened 2023-07-14
Feat/integ 815

rusticpenguin:
## Purpose

This will allow us to easily communicate with chatGPT


apps/ai-content-generator/src/utils/AI/index.ts
rusticpenguin, line 1: Decided to move the core AI logic into a class. It would have been this way if we were able to use the openAI library. This keeps the hook more focused on getting and returning the data. 

We will move this logic into a lambda at some point because of secret storage

apps/ai-content-generator/src/hooks/dialog/useAI.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        setOutput((prev) => prev + streamOutput);
+        completeMessage += streamOutput;
+      }
+    } catch (error: any) {
+      console.error(error);
+    }
+
+    setStream(null);
```
ryunsong-contentful, line 65: You probably want this in a finally block instead just in case some runtime error happens, the finally block will always execute and put your stream state in a good place.

rusticpenguin: Love this, thank you

ryunsong-contentful (approved):

PR #4251, opened 2023-07-20
Add styling fixes

lilbitner:
## Purpose

✨ 🖌️ 🎨 

This PR implements the following styling changes to the AI Content Generator app v2 app: 

1. Removes the resizing indicator of the text area within the Dialog.
2. Removes default 'Help text' as the tooltip shown when hovering over the buttons in the sidebar. I would like to confirm what these tooltips should actually be with Emily next week as a follow-up task. 
3. Adds borders and adjusts the layout of the dialog slightly. 
4. Fixes the copies of the title shown on when the dialog is rendered. 
5. Fixes the copies of the buttons within the sidebar. 
6.  Imports reusable components from the[ Integration Component Library ](https://github.com/contentful/integration-component-library)yay !
7. Adds the disclaimer text shown below the sidebar buttons. For now the hyperlink links to the openAI but will confirm this. 
8. Render a different copy on the left-most panel when generating a title (From prompt vs From new text). Emily is in process of confirming this.
9. Refactor the `dialogConfig` into its own config, to support defining "magic numbers" and logical decoupling within the `FeatureButtons` component. 

## Notes

1. **The one update I had difficulty with was adding the 'sparkle' icon next to the title of the Dialog. The title type accepted by the app framework logic is pretty strict to only accept a string. Any creative ideas for how to get around this?** 
Screenshot 2023-07-21 at 11 18 51 AM

2. I didn't go any farther than these adjustments as per Emily's visual explorations, I think the layout and design will change fairly soon. 
3. "Current" designs are here for comparison: https://www.figma.com/file/TQsQcvECzI6R2Jy2Utu6xP/AI-Content-Generator-Revision-2?type=design&node-id=118-59938&mode=design&t=HARwTZBPMOdBYQ5K-0


apps/ai-content-generator/src/configs/dialog/dialogConfig.ts
```diff
@@ -0,0 +1,19 @@
...
+  title: string,
+): openDialogOptions => {
+  return {
+    position: 'center',
+    width: 'fullWidth',
+    minHeight: DIALOG_MIN_HEIGHT,
+    title,
+    parameters,
+  };
+};
```
rusticpenguin, line 19: I love this!

apps/ai-content-generator/src/configs/features/featureTypes.tsx
```diff
@@ -16,6 +16,7 @@ export type FeatureComponentProps = {
 
 type FeatureItem = {
   title: string;
+  buttonTitle: string;
```
rusticpenguin, line 19: 👍

apps/ai-content-generator/src/locations/Dialog.tsx
```diff
@@ -26,7 +26,7 @@ const Dialog = () => {
   };
 
   return (
-    <GeneratorProvider entryId={entryId} prompt={featureConfig[feature].prompt}>
+    <GeneratorProvider entryId={entryId} feature={feature} prompt={featureConfig[feature].prompt}>
```
rusticpenguin, line 29: Since we're only using the `feature` in the provider to check for a bool, we could follow a similar pattern to the `isTranslate` prop on CommonGenerator

So instead of getting feature from context, we can have a flag like:
```
case : AIFeature.Title:
  return <CommonGenerator fromPrompt />
```
and then you can pass down that `FromPrompt` into `NewOrExistingText`

whitelisab (commented):
This looks great and a nice balance of some quick wins while not investing too much as we rework the designs. 🌟 The only thing I noticed on here is that the sidebar is now scrollable, maybe we want the buttons and disclaimer text to take the full height instead? 

Screenshot 2023-07-21 at 12 54 09 PM

apps/ai-content-generator/src/locations/Dialog.tsx
lilbitner: this is a great point! I felt like what I was doing was a little bit more of a heavy lift than necessary. I ended up implementing you suggestion, with a small change in the prop name to be `isTitle` to keep consistency with the `isTranslate` pattern.

apps/ai-content-generator/src/configs/features/featureConfig.tsx
```diff
@@ -15,28 +15,33 @@ enum AIFeature {
...
+    title: 'Generate a title',
+    buttonTitle: 'Title',
+    helpText: '',
     prompt: titlePrompt,
   },
   [AIFeature.CONTENT]: {
-    title: 'Content',
+    title: 'Generate content',
+    buttonTitle: 'Content',
     helpText: 'Generate content based off of your data',
```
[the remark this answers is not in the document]
lilbitner: removed this!

lilbitner:
> This looks great and a nice balance of some quick wins while not investing too much as we rework the designs. 🌟 The only thing I noticed on here is that the sidebar is now scrollable, maybe we want the buttons and disclaimer text to take the full height instead?
> 
> Screenshot 2023-07-21 at 12 54 09 PM

Yes good catch! I also noticed the overflow while developing and am having a difficult time adjusting it, because totally agree that ideally the scroll would not appear. I notice in a few other apps (GA4 included) that the scroll isn't present, but haven't had success in finding the difference. Has to be something small so will keep digging. Would love any suggestions if you have them! :)

whitelisab:
@lilbitner I think I found it -- you can import `useAutoResizer` in the Sidebar component and that should fix it! Here's an example in GA4: https://github.com/contentful/apps/blob/master/apps/google-analytics-4/frontend/src/components/main-app/AnalyticsApp/AnalyticsApp.tsx#L29

apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
```diff
@@ -41,11 +41,13 @@ const CommonGenerator = (props: FeatureComponentProps) => {
   useEffect(updateProviderData, [parameters.targetLocale, parameters.locale, dispatch]);
 
   return (
-    <Flex>
-      <NewOrExistingText isNewText={parameters.isNewText} dispatch={dispatch} />
-      <FieldSelector parameters={parameters} isTranslate={isTranslate} fieldTypes={TextFields} />
+    <div className={styles.root}>
```
rusticpenguin, line 44: Nonblocking Q: Do we want to have these styles be applied inside of the components their meant for instead of at the top level?

whitelisab (approved):
Looks great! 🎊

PR #4262, opened 2023-07-21
[INTEG-875] Point apps deploy:staging to actual apps in the dev & testing orgs

ryunsong-contentful:
## Purpose
Have our `staging:deploy` script actually do something and point and deploy to an actual app in our Dev & Testing org.
Also completed one part of 3 movements in improving our infrastructure https://contentful.atlassian.net/browse/INTEG-873.

## Approach
Create staging apps in the Apps - Dev & Testing org.
Update scripts to point to those app ids

## Testing steps
Check app ids from package.json versus the [Apps marketplace place](https://app.contentful.com/account/organizations/6xdLsz6lCsk0yPOccSsDK7/apps)
Deployment should succeed

## Breaking Changes
Could break if there are mismatched app ids

## Deployment
Standard

### DIFF
--- a/apps/dropbox/package.json
+++ b/apps/dropbox/package.json
@@ -16,7 +16,7 @@
     "start": "cross-env BROWSER=none react-scripts start",
     "build": "react-scripts build",
     "deploy": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEFINITIONS_ORG_ID} --definition-id 6YdAwxoPHopeTeuwh43UJu --token ${CONTENTFUL_CMA_TOKEN}",
-    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${TEST_ORG_ID} --definition-id ${TEST_APP_ID} --token ${CONTENTFUL_CMA_TOKEN}"
+    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEV_TESTING_ORG_ID} --definition-id 4OLkT2GLsM5HxBFGlnsR5q --token ${TEST_CMA_TOKEN}"
   },
   "eslintConfig": {
     "extends": "react-app"
--- a/apps/ecommerce/frontend/src/components/App/Dialog/ResourceSelectionDialog/ResourceSelectionDialog.spec.tsx
+++ b/apps/ecommerce/frontend/src/components/App/Dialog/ResourceSelectionDialog/ResourceSelectionDialog.spec.tsx
@@ -25,7 +25,7 @@ const mockFetchWithSignedRequest = fetchWithSignedRequest as jest.MockedFunction
 
 const { getByText, getByTestId, findByText } = screen;
 
-describe('ResourceSelectionDialog component', () => {
+describe.skip('ResourceSelectionDialog component', () => {
   beforeEach(() => {
     const mockApiResponse = {
       ok: true,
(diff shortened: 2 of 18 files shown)


apps/dropbox/package.json
```diff
@@ -16,7 +16,7 @@
     "start": "cross-env BROWSER=none react-scripts start",
     "build": "react-scripts build",
     "deploy": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEFINITIONS_ORG_ID} --definition-id 6YdAwxoPHopeTeuwh43UJu --token ${CONTENTFUL_CMA_TOKEN}",
-    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${TEST_ORG_ID} --definition-id ${TEST_APP_ID} --token ${CONTENTFUL_CMA_TOKEN}"
```
rusticpenguin, line 19: Should these definitions be ENV variables?

apps/dropbox/package.json
```diff
@@ -16,7 +16,7 @@
...
+    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEV_TESTING_ORG_ID} --definition-id 4OLkT2GLsM5HxBFGlnsR5q --token ${TEST_CMA_TOKEN}"
```
rusticpenguin, line 19: Should these definitions be ENV variables?

apps/ecommerce/frontend/src/components/App/Dialog/ResourceSelectionDialog/ResourceSelectionDialog.spec.tsx
```diff
@@ -25,7 +25,7 @@ const mockFetchWithSignedRequest = fetchWithSignedRequest as jest.MockedFunction
 
 const { getByText, getByTestId, findByText } = screen;
 
-describe('ResourceSelectionDialog component', () => {
+describe.skip('ResourceSelectionDialog component', () => {
```
lilbitner, line 28: is this (and the other skips in this PR) because the test is flaky?

lilbitner (commented):
thank you for taking the initiative to do this!!!! ⭐ do you mind elaborating on what the next logical steps are here after this change? Just for clarity on my end, I understand that this will now support deploy to both staging and prod environments at the same time (emphasis on staging now!), but curious about the next step - assuming more extensive work with the pipeline to only deploy to staging on merge, and then add commands to approve prod deployment?

apps/ecommerce/frontend/src/components/App/Dialog/ResourceSelectionDialog/ResourceSelectionDialog.spec.tsx
ryunsong-contentful: Yes, i skipped them because it's ecommerce which is no longer really part of our project and flakey tests

apps/dropbox/package.json
ryunsong-contentful: It's fine for the app id to be public - there's no security concerns

apps/ecommerce/frontend/src/components/App/Dialog/ResourceSelectionDialog/ResourceSelectionDialog.spec.tsx
lilbitner: oh word, good point 💯

ryunsong-contentful:
So the logical next step would be to separate the deploy of Prod and Staging. However, I don't think that will be coming in the short term as we should decide as a team how we should do that. In the MPA repo we have two separate branches called (Main -> Staging) and (Production -> Production) but we may not want to do that here for the apps repo.

lilbitner (approved):

PR #4263, opened 2023-07-21
[INTEG-880] add testing utils and tests for config hooks

rusticpenguin:
This adds unit testing to the Configs hooks. I also created some testing utils to better help with testing with the app sdk.

The MockSDK class is designed to a spot for a lot of repeat code revolving the mockSDK. A great example here is the method `MockSDK.reset` used in beforeEach.

Create a few static App installation params, as well as a generator function for randomly generating test data

To verify: 
- Test should pass


lilbitner (approved):
This is an awesome start to testing things in the` ai-content-generator`! Love the utils and mocks that you made to make testing way simpler in the app 💯  Way to go!!! 🚀 just seeing what seems like a flaky build issue, but other than that, looks fantastic to me!

PR #4279, opened 2023-07-24
[INTEG-871]: Update filtering for source and output field selectors

whitelisab:
## Purpose

This PR updates the logic for which fields will be shown in the source and output fields. 

## Approach

- The source field should show: Short text, long text, and rich text fields that have content in the field
- The output field should show: All Short text, long text, and rich text fields 

Additionally, I also updated these field selectors so that field id is saved rather than field name because field name is not unique.

Note that there are some quirks regarding whether a field is localized that will be addressed in another PR.

## Testing steps

Tested with fields of supported types and ensured that fields of different types were not listed as options.


rusticpenguin (approved):
Really good choice on var names and good swap to using the ID. This looks amazing

PR #4297, opened 2023-07-25
Feat/integ 869

lilbitner:


apps/ai-content-generator/src/components/app/sidebar/feature-button/FeatureButton.tsx
```diff
@@ -19,7 +19,7 @@ const FeatureButton = (props: Props) => {
   const handleOnClick = () => {
     const entryId = sdk.entry.getSys().id;
 
-    const dialogConfig = makeDialogConfig({ feature, entryId }, dialogTitle);
+    const dialogConfig = makeDialogConfig({ feature, entryId });
```
rusticpenguin, line 21: Since we're just using the feature we no longer need the dialogTitle prop in this component

apps/ai-content-generator/src/locations/Dialog.tsx
```diff
@@ -29,7 +29,7 @@ const Dialog = () => {
   };
 
   return (
-    <GeneratorProvider entryId={entryId} prompt={featureConfig[feature].prompt}>
+    <GeneratorProvider feature={feature} entryId={entryId} prompt={featureConfig[feature].prompt}>
```
rusticpenguin, line 32: I feel since we're giving full access to the feature here, we can remove the `prompt` property

apps/ai-content-generator/src/configs/features/featureConfig.tsx
```diff
@@ -15,33 +15,33 @@ enum AIFeature {
...
-    helpText: '',
+    description: 'generate SEO Keywords',
     prompt: seoKeywordsPrompt,
   },
   [AIFeature.SEO_DESCRIPTION]: {
-    title: 'Generate an SEO description',
+    dialogTitle: 'Generate an SEO description',
     buttonTitle: 'SEO description',
-    helpText: '',
+    description: 'generate an SEO description',
```
lilbitner, line 44: `dialogTitle` is explicit now, and added the name of the other field as a `description` field. What do you think? I really am having trouble thinking of another name for it 😅

apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
```diff
@@ -17,11 +18,12 @@ const initialParameters: GeneratorParameters = {
   sourceField: '',
   outputField: '',
   originalText: '',
+  canGenerateTextFromField: false,
 };
 
 const CommonGenerator = (props: FeatureComponentProps) => {
   const { isTranslate, isTitle } = props;
-  const { setProviderData } = useContext(GeneratorContext);
+  const { setProviderData, feature } = useContext(GeneratorContext);
```
rusticpenguin, line 26: Since feature lives in context I think it'd be cleaner to just call useContext in the `NoFieldsSelectedMessage`

apps/ai-content-generator/src/components/app/sidebar/SidebarButtons.tsx
```diff
@@ -3,18 +3,8 @@ import featureConfig, { AIFeature } from '@configs/features/featureConfig';
...
-      <FeatureButton
-        key={featureId}
-        text={buttonTitle}
-        dialogTitle={title}
-        helpText={helpText}
-        feature={featureId as AIFeature}
-      />
-    );
+  const featureList = Object.entries(featureConfig).map(([featureId]) => {
+    return <FeatureButton key={featureId} feature={featureId as AIFeature} />;
```
rusticpenguin, line 7: So much easier to look at now!

You can change it from `Object.entries` to just `Object.keys` since we only need the key now!
```suggestion
  const featureList = Object.keys(featureConfig).map((featureId: AIFeature) => {
    return <FeatureButton key={featureId} feature={featureId} />;
```

apps/ai-content-generator/src/configs/features/featureConfig.tsx
rusticpenguin: What do you think about `invalidFormSuffix`? (or if I forgot about a place where it wasn't just used for an invalid spot, just `messageSuffix`

rusticpenguin (approved):
This look great!! Lots of good code design decisions here

apps/ai-content-generator/src/components/app/sidebar/SidebarButtons.tsx
rusticpenguin: makes complete sense! I was wondering if key would accept an enum as I was typing it

PR #4301, opened 2023-07-25
chore: add deploy commands to package

rusticpenguin:
## Purpose
We need to deploy AICG to prod and staging but there were no scripts to do so in the package.json.
We also need to deploy two different versions of the build to staging/prod respectively

## Approach
- Add the correct deploy commands as well as populate the definition ID's
- Add NODE_ENV to our CircleCI pipelines so we can differentiate which environment we are building for

**Read:** This branch is only to deploy to Staging at the moment. We will deploy to prod later today at 4pm MT


lilbitner (commented):
This all makes sense to me! Just noticing a build failure here that might be an issue?

rusticpenguin:
> This all makes sense to me! Just noticing a build failure here that might be an issue?

agreed! trying to see what's up with the build

matthew-gordon (approved):

PR #4323, opened 2023-07-27
[INTEG-893] Properly handle languages for localized fields

whitelisab:
## Purpose

There are a couple of issues currently in the modal for localized fields:
- If you try to apply output to a field that doesn't have the localized language or when where there is not content already, you get an error and cannot apply the output text
- We have decided that we want to remove the source/target language fields and instead display all of the fields and their language options within the field selectors

This is what the new field selector options look like:
Screenshot 2023-07-31 at 8 59 55 AM

## Approach

Within the modal we have access to the entry from `cma.entry.get()` and the contentType from `cma.contentType.get()`. However, the fields array on the entry only contains fields that have content, and the fields array on the contentType only shows whether a field is localized. Therefore, we currently do not have a way of seeing all of the language options for a field within the modal. The Dialog location does not support `sdk.entry.fields` unfortunately. So in order for us to have access to all the language options for all the fields, I am getting this information from the Sidebar SDK and passing it via invocation parameters to the Dialog. I am definitely open to other approaches if anyone has any ideas!

I also wanted to have the order of the fields in the dropdown be in the same order as they appear on on the entry, which is the default locale first, and then the rest alphabetically by the full language name.

## Testing steps

Ensure you have localization turned on and have some fields that have several localization language options. You should now see all of the fields and their locale options in the field selectors. These should be ordered the same way that the fields are ordered on the entry. You should also be able to apply the output to fields consistently without errors (check when a field doesn't have content and when it's not the default locale). Test each of the actions and they should work as expected.


apps/ai-content-generator/src/components/app/dialog/common-generator/output/output-text-panels/OutputTextPanels.tsx
```diff
@@ -22,10 +22,7 @@ const OutputTextPanels = (props: Props) => {
   const sdk = useSDK<DialogAppSDK>();
 
   const handleEntryApply = async () => {
-    const fieldKey = outputField.toLocaleLowerCase();
```
whitelisab, line 25: This was making the `outputField` lowercase and therefore causing errors where the field wasn't recognized, so I removed it.

apps/ai-content-generator/src/hooks/dialog/useEntryAndContentType.ts
```diff
@@ -39,13 +39,21 @@ const useEntryAndContentType = (entryId: string) => {
...
 
-      entry.fields[fieldKey][fieldLocale] = isRichText
+      const formattedText = isRichText
         ? richTextModel(updatedText)
         : updatedText
             .replace(/^\s+|\s+$/g, '')
             .replace(/^['"]+|['"]+$/g, '')
             .trim();
 
+      if (!entry.fields[fieldKey]) {
```
whitelisab, line 49: This fixes the issue where we couldn't apply the output to a field if it didn't already have text in it.

apps/ai-content-generator/src/utils/dialog/common-generator/field-selector/fieldSelectorHelpers.ts
```diff
@@ -6,16 +6,24 @@ import { Field } from '@hooks/dialog/useSupportedFields';
...
  */
-const getFieldData = (fieldId: string, fields: Field[]) => {
-  const field = fields.find((field) => field.id === fieldId);
-  return field
-    ? { id: field.id, name: field.name, data: field.data || '' }
-    : { id: '', name: '', data: '' };
+const getFieldData = (fieldKey: string, fields: Field[]) => {
+  const field = fields.find((field) => field.key === fieldKey);
+  return (
+    field ?? {
```
rusticpenguin, line 17: so clean

apps/ai-content-generator/src/utils/dialog/supported-fields/supportedFieldsHelpers.ts
```diff
@@ -28,28 +43,55 @@ const formatField = (field: ContentFields, entry: EntryProps, targetLocale: stri
...
+        const formattedField = formatField(field, entry, locale, localeNames, defaultLocale);
+
+        if (hasContent) {
+          fieldsWithContent.push(formattedField);
+        }
+        allFields.push(formattedField);
+      });
+
+      // Sort the locales so that default is first, then alphabetically by language name
+      const sortedFieldsWithContent = orderBy(
```
rusticpenguin, line 83: I feel this can be done with the vanilla sort, as opposed to bringing in lodash here:

```js
// another helper function
const sortLanguages =  (a, b) => {
  if (a.isDefaultLocale) {
    // send default to top
    return -1;
  }

  if (a.language > b.language) {
    return 1
  }

  if (a.language < b.language) {
    return -1
  }
  
  return 0;
}

// inside isSupported
...

const sortedFieldsWithContext = fieldsWithContext.sort(sortLanguages);
const sortedAllFields = allFields.sort(sortLanguages)
```

rusticpenguin:
This looks great! All I have is a comment on using vanilla over lodash

apps/ai-content-generator/src/utils/dialog/supported-fields/supportedFieldsHelpers.ts
whitelisab: I removed lodash and made this change; had to update the logic a little bit from what you proposed here to get the sorting working right, but it's now implemented with vanilla sort!

rusticpenguin: Had to pull this down to mess around with it a bit more. I riffed on your implementation a bit to cover the edge case of if b is the default locale. Also found `localeCompare` which simplifies this a bit more!

```js
      const sortedAllFields = allFields.sort((a: Field, b: Field) => {
        // Default should be listed first
        if (a.isDefaultLocale) {
          return -1;
        }

        // Default should be listed first
        if (b.isDefaultLocale) {
          return 1;
        }

        return a.name.localeCompare(b.name);
      });
```

PR #4325, opened 2023-07-27
fix: content generation now uses gavin's new prompt

rusticpenguin:
## Purpose

The old content prompt would output weird extra data

## Approach

Update the prompt with a bit more constraints (courtesy of Gavin)

## Testing steps

Try generating some content from a title or an existing blog post. It should be in a normal text style now


whitelisab (approved):
Confirmed that it removes the headings from the generated content! 🌟

PR #4326, opened 2023-07-27
chore: add aicg deploy script

rusticpenguin:
This will deploy the merged AICG changes to prod.

Outcome:
Visibly nothing should change on prod, it should remain using V1


matthew-gordon (approved):
🖖

PR #4344, opened 2023-07-31
feat: implement launchdarkly

rusticpenguin:
## Purpose

We're using environment variables right now to display V1 or v2 and I decided to just try out implementing launch darkly over launch. 

Using the following flag, we are displaying V2 in both the staging and development spaces in contentful:
~~https://app.launchdarkly.com/default/production/features/integrations-ai-content-generator-v-1/targeting~~
https://app.launchdarkly.com/default/production/features/integrations-ai-content-generator-v-2/insights


apps/ai-content-generator/package.json
```diff
@@ -23,7 +24,7 @@
     "build": "tsc && vite build --mode ${NODE_ENV:='production'}",
     "test": "vitest",
     "create-app-definition": "contentful-app-scripts create-app-definition",
-    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEFINITIONS_ORG_ID} --definition-id 3iszK8Gl7aaxLvxGyCOhgA --token ${CONTENTFUL_CMA_TOKEN}",
+    "doNotDeployToProd": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEFINITIONS_ORG_ID} --definition-id 3iszK8Gl7aaxLvxGyCOhgA --token ${CONTENTFUL_CMA_TOKEN}",
```
jsdalton, line 27: so i take it you're disabling the prod deploy here to verify that everything with the LD ff works, and then you'll re-enable it later?

apps/ai-content-generator/src/App.tsx
```diff
@@ -11,29 +11,42 @@ import Page from '@locations/Page';
...
+  [locations.LOCATION_ENTRY_SIDEBAR]: isV2 ? Sidebar : SidebarV1,
   [locations.LOCATION_PAGE]: Page,
   [locations.LOCATION_HOME]: Home,
-};
+});
 
 const App = () => {
   const sdk = useSDK();
+  const { integrationsAiContentGeneratorV2: isV2 } = useFlags<ldConfigType>();
+  const ldClient = useLDClient();
```
jsdalton, line 30: Eventually it'd be nice to extract this setup into its own hook that we could put in our `marketplace-apps-toolkit` and just do something `useLaunchDarkly(sdk)` in future iterations.

apps/ai-content-generator/src/configs/launch-darkly/ldConfig.ts
```diff
@@ -0,0 +1,16 @@
+const ldConfig = {
+  clientSideID: `${import.meta.env.VITE_LAUNCHDARKLY_CLIENT_ID}`,
+  deferInitialization: false,
+  flags: {
+    'integrations-ai-content-generator-v-2': true,
```
jsdalton, line 5: Do we have a naming convention for these? I lost my launch darkly access so i actually can't check. 🤷

apps/ai-content-generator/src/configs/launch-darkly/ldConfig.ts
```diff
@@ -0,0 +1,16 @@
...
+  },
+  options: {
+    bootstrap: localStorage,
+  },
+};
+
+type ldConfigType = { integrationsAiContentGeneratorV2: boolean };
```
jsdalton, line 12: Okay this is like a magic thing that the LD client does at compile time where it automatcally converts the feature flag `integrations-ai-content-generator-v-2` to that camelCase type? Interesting.

jsdalton (approved):
A few questions and comments but overall I'm good with proceeding.

apps/ai-content-generator/package.json
rusticpenguin: Yup! Exactly that

apps/ai-content-generator/src/configs/launch-darkly/ldConfig.ts
rusticpenguin: Yeah they magically convert it for us. We can disable that in the config with a flag, and it'll spit out `integerations-ai-...`

apps/ai-content-generator/src/App.tsx
rusticpenguin: Agreed, I think it'd be nice to move this ldClient stuff out of the component

PR #4353, opened 2023-08-01
fix: save content before launching ai content generator

rusticpenguin:
## Purpose

When a user edits a field and launches the ai content generator, if it has not auto-saved yet we are using old data

## Approach

We decided to go with the route of least resistance here and force a save before launching the modal.

Edge cases:
- We are handling if a user spams the button
- We are handling if a user clicks one button and a different one before the save completes
- We are handling if the save fails for any reason

## Testing steps

1. Edit the entry
2. Click a button
3. Verify the spinner appears and then the modal launches
___
1. Don't edit anything
2. Click a button
3. Verify it opens instantaneously 
___
1. Edit the entry
2. Spam click the button
3. Verify nothing bad happens
___
1. Edit the entry
2. Click two different features
3. Verify nothing bad happens

### DIFF
--- a/apps/ai-content-generator/src/components/app/sidebar/feature-button/FeatureButton.tsx
+++ b/apps/ai-content-generator/src/components/app/sidebar/feature-button/FeatureButton.tsx
@@ -9,7 +9,7 @@ import { useState } from 'react';
 interface Props {
   feature: AIFeature;
   isSaving: boolean;
-  setIsSaving: (isSaving: boolean) => void;
+  onSaving: (isSaving: boolean) => void;
 }
 
 interface FieldLocales {
@@ -17,7 +17,7 @@ interface FieldLocales {
 }
 
 const FeatureButton = (props: Props) => {
-  const { feature, isSaving, setIsSaving } = props;
+  const { feature, isSaving, onSaving } = props;
   const buttonCopy = featureConfig[feature].buttonTitle;
   const sdk = useSDK<SidebarAppSDK>();
 
@@ -30,8 +30,8 @@ const FeatureButton = (props: Props) => {
     return acc;
   }, {} as FieldLocales);
 
-  const toggleSaving = (toggleTo: boolean) => {
-    setIsSaving(toggleTo);
+  const updateSaveState = (toggleTo: boolean) => {
+    onSaving(toggleTo);
     setIsOpeningDialog(toggleTo);
   };
 
@@ -41,12 +41,12 @@ const FeatureButton = (props: Props) => {
     }
 
     try {
-      toggleSaving(true);
+      updateSaveState(true);
       await sdk.entry.save();
     } catch (error) {
       console.error(error);
     } finally {
-      toggleSaving(false);
+      updateSaveState(false);
     }
 
     const entryId = sdk.entry.getSys().id;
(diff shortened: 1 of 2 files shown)


apps/ai-content-generator/src/components/app/sidebar/feature-button/FeatureButton.tsx
```diff
@@ -1,31 +1,54 @@
...
+import { Button, Spinner, Tooltip } from '@contentful/f36-components';
 import featureConfig, { AIFeature } from '@configs/features/featureConfig';
 import { styles } from './FeatureButton.styles';
 import { makeDialogConfig } from '@configs/dialog/dialogConfig';
+import { useState } from 'react';
 
 interface Props {
   feature: AIFeature;
+  isSaving: boolean;
+  setIsSaving: (isSaving: boolean) => void;
```
ryunsong-contentful, line 12: This is a general anti-pattern for statement management functions. It's better to use the handleFunctionName and onFunctionName convention` and separating your state setters and your functions and use the functions to be the objects you pass as props (aka callback functions).

apps/ai-content-generator/src/components/app/sidebar/feature-button/FeatureButton.tsx
```diff
@@ -1,31 +1,54 @@
...
   const fields = sdk.entry.fields;
   const fieldLocales = Object.entries(fields).reduce((acc, entryField) => {
     const [fieldId, field] = entryField;
     acc[fieldId] = field.locales;
     return acc;
   }, {} as FieldLocales);
 
-  const handleOnClick = () => {
+  const toggleSaving = (toggleTo: boolean) => {
+    setIsSaving(toggleTo);
```
ryunsong-contentful, line 34: What point is there in having these two states when they effectively represent the same thing and can be used in the same fashion?

apps/ai-content-generator/src/components/app/sidebar/feature-button/FeatureButton.tsx
```diff
@@ -1,31 +1,54 @@
...
+    setIsOpeningDialog(toggleTo);
+  };
+
+  const handleOnClick = async () => {
+    if (isSaving) {
+      return;
+    }
+
+    try {
+      toggleSaving(true);
```
ryunsong-contentful, line 44: toggle isn't representative of what you're doing as you're correctly setting the actual boolean. Maybe `updateSaveStates` is a better name?

ryunsong-contentful (changes requested):
Nice PR - The code/logic works but there's one anti-pattern section and then some smaller comments just for readability

apps/ai-content-generator/src/components/app/sidebar/feature-button/FeatureButton.tsx
rusticpenguin: Thank you, I was moving too fast on this. I appreciate the catch!

apps/ai-content-generator/src/components/app/sidebar/feature-button/FeatureButton.tsx
rusticpenguin: isOpeningDialog is to show the spinner, which would be local to the button that was pressed. isSaving is to block any button from firing a launch modal command, and subsequently another save

ryunsong-contentful: Cool makes sense to me

ryunsong-contentful (approved):

ryunsong-contentful (approved):

PR #4356, opened 2023-08-01
feat: add character count and validation to aicg

rusticpenguin:
## Purpose

A user may or may not know that that have exceeded or are under the character limits. When a user applies this may put their entry in an unpublishable state.

## Approach

Forma36's counter requires a limit on the text field. This forces the user to backspace until they get below the limit. This UX seems unintuitive for our case.

This custom counter allows the user to exceed the limit (but disable the publishing), allowing them to edit the text as they see fit until they get below the maximum limit.

We also are showing an error for when the user is below the minimum limit.

## Testing steps

Test using a field with no limit
- verify it only shows the current character count.
- verify the apply is never disabled

Test using a field with an upper and/or lower limit
- verify that if you are over the limit you cannot apply
- verify that if you are below the limit you cannot apply
- verify that if you are within the constrains you can apply

### DIFF
--- a/apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
+++ b/apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
@@ -12,9 +12,12 @@ import NoFieldsSelectedMessage from './output/output-text-panels/no-fields-selec
 const initialParameters: GeneratorParameters = {
   isNewText: false,
   sourceField: '',
-  outputField: '',
-  outputFieldId: '',
-  outputFieldLocale: '',
+  output: {
+    field: '',
+    id: '',
+    locale: '',
+    validation: null,
+  },
   originalText: '',
   canGenerateTextFromField: false,
 };
@@ -41,8 +44,9 @@ const CommonGenerator = (props: FeatureComponentProps) => {
       </div>
       {parameters.canGenerateTextFromField ? (
         <Output
-          outputFieldId={parameters.outputFieldId}
-          outputFieldLocale={parameters.outputFieldLocale}
+          outputFieldId={parameters.output.field}
+          outputFieldLocale={parameters.output.locale}
+          outputFieldValidation={parameters.output.validation}
           inputText={parameters.originalText}
         />
       ) : (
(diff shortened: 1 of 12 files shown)


apps/ai-content-generator/src/components/common/text-counter/TextCounter.styles.ts
```diff
@@ -0,0 +1,13 @@
+import { css } from '@emotion/react';
+
+export const styles = {
+  // How do I access these colors from forma36?
```
rusticpenguin, line 4: Is this a thing we can do?

whitelisab: Yes, you can access the forma36 colors from tokens, here is an example: https://github.com/contentful/apps/blob/master/apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.styles.tsx#L7

apps/ai-content-generator/package-lock.json
whitelisab, line 1: FYI: I had quite a large diff on the `package-lock.json` when I ran `npm install` locally on your branch.

whitelisab (changes requested):
Looks really great, nicely done! Overall I made some comments about cleaning up the `package.json` and `package-lock.json` files and had some minor questions.

apps/ai-content-generator/src/components/common/text-counter/TextCounter.tsx
```diff
@@ -0,0 +1,52 @@
...
+          {text.length} characters
+        </Paragraph>
+        {maxLength && (
+          <Paragraph marginTop="spacing2Xs" css={styles.validCount}>
+            Maximum {maxLength} characters
+          </Paragraph>
+        )}
+      </Flex>
+      <Flex
+        css={{ visibility: isValid ? 'hidden' : 'visible' }}
```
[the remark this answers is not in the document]
rusticpenguin: I chose do this so we can keep that space reserved, this keeps the ui from popping when the it appears/disappears

apps/ai-content-generator/package-lock.json
rusticpenguin: I'll run the npm install again. I was flopping between branches so maybe I lost it. Great catch!

whitelisab (approved):
Looks great! ⭐

PR #4357, opened 2023-08-01
Disable staging deploy to prevent race condition in hosted app action deploys

jsdalton:
## Purpose

There's a bug in extensibility-api right now where if the same app action id is included for two different org ids / app definition ids, the most recent deploy overwrites the app action it does not own instead of creating a new one.

## Approach

* Disable "staging" deploy so that we have only one app action deploy active


matthew-gordon (approved):
👍

PR #4377, opened 2023-08-02
[INTEG-863] Fix sidebar and app installation issues for AICG for empty api keys

ryunsong-contentful:
## Purpose
Massive amounts of CMA calls which add up in terms of cost, resources, and time on other support teams.

Screenshot 2023-08-02 at 9 20 20 AM
Screenshot 2023-08-02 at 9 25 29 AM

## Approach
Same as GA4, just check if there's a key and terminate early with a note component if it's an empty key

## Testing steps
Smoke test in development environment

## Breaking Changes
None 

None

## Deployment
Standard

### DIFF
--- a/apps/ai-content-generator/src/locations/ConfigScreenV1.tsx
+++ b/apps/ai-content-generator/src/locations/ConfigScreenV1.tsx
@@ -25,6 +25,12 @@ const ConfigScreen = () => {
     const currentState = await sdk.app.getCurrentState();
     setApiKey('');
     setEditing(false);
+
+    if (parameters.key === '') {
+      sdk.notifier.error('A valid OpenAI API key is required');
+      return false;
+    }
+
     return {
       parameters,
       targetState: currentState,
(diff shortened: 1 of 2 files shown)


apps/ai-content-generator/src/locations/ConfigScreenV1.tsx
```diff
@@ -25,6 +25,12 @@ const ConfigScreen = () => {
     const currentState = await sdk.app.getCurrentState();
     setApiKey('');
     setEditing(false);
+
+    if (parameters.key === '') {
+      sdk.notifier.error('A non-empty api key is required');
```
jsdalton, line 30: Small copy suggestion. You can just say:

> A valid OpenAI API key is required

jsdalton (approved):
Small copy suggestions, otherwise looks good.

apps/ai-content-generator/src/locations/ConfigScreenV1.tsx
ryunsong-contentful: Hmm - I avoided that because it only checks for empty string - it doesn't actually fully check if an API key is required but I'll make the changes regardless

PR #4378, opened 2023-08-02
Fix(deps): Dependabot/npm and yarn/contentful/default field editors 1.4.10

lilbitner:
## Purpose

This PR fixes an [error](https://app.circleci.com/pipelines/github/contentful/apps/22390/workflows/5bd63496-c0aa-44bd-a542-8326c0764b5c/jobs/63503) thrown with the accompanying dependabot update within ecomm. The new update caused a type error, and frankly because of the deprecation of ecomm, I made adjustments that passed the build, but nothing beyond that. The app still renders locally. 

This fixes this PR: https://github.com/contentful/apps/pull/4300


whitelisab (approved):
LGTM! Agree with your approach on this.

PR #4380, opened 2023-08-02
[INTEG-964] Update AICG to global modal design

whitelisab:
## Purpose

This PR refactors the AICG dialog to match the new designs. Here's a screenshot of how it looks now:
Screenshot 2023-08-02 at 3 38 47 PM

## Approach

These are the changes made on this PR:
- Changed size/orientation of modal to match vertical orientation
- Created a custom header to include the Sparkle icon
- Moved content source from buttons to select component
- Ensured correct field selectors showed based on content source
- Renamed and move Source and Result tabs to top right side of text area
- Added placeholder and helper text as variables
- Ensured that source text area and generate button are disabled appropriately

There are a few things that will be addressed in follow up PRs:
- Character count warning
- Fixing the behavior when you switch between content source options
- Adding test coverage
- Reviewing and updating all of the copy

A note about how I approached styling in this PR. If the Forma36 component had a prop, I used it. If not, I used Emotion's css prop. I only created a separate styles file if there was a lot of custom css, otherwise I just kept it in the context of the component.

## Testing steps

- Ensure that all of the actions still work as expected
- The "Generate" button should be disabled and a helper text should be shown next to the button if the source/output is not selected
- When Field is the Content Source, the textarea will be disabled until an input field is selected


apps/ai-content-generator/src/components/app/dialog/common-generator/field-selector/content-source/ContentSource.tsx
```diff
@@ -0,0 +1,26 @@
+import { Box, FormControl, Select } from '@contentful/f36-components';
+
+interface Props {
+  title: string;
+  isNewText: boolean;
+  onChange: () => void;
+  selectFieldWidth: number;
+}
+
+function EntryFieldList(props: Props) {
```
rusticpenguin, line 10: Looks like this could be renamed to `ContentSource`

whitelisab: Good catch, thank you!

apps/ai-content-generator/src/components/app/dialog/common-generator/field-selector/content-source/ContentSource.tsx
```diff
@@ -0,0 +1,26 @@
...
+  isNewText: boolean;
+  onChange: () => void;
+  selectFieldWidth: number;
+}
+
+function ContentSource(props: Props) {
+  const { title, isNewText, onChange, selectFieldWidth } = props;
+
+  return (
+    <Box css={{ width: selectFieldWidth }}>
```
rusticpenguin, line 14: What do you think about moving the html into a generic component? That way we're not duplicating html between this and the Entry field list

rusticpenguin (approved):

PR #4399, opened 2023-08-03
[WIP] update app action name

matthew-gordon:
## Purpose

- update app action name to redeploy and make use of new index compound key


lilbitner (approved):
👍

PR #4402, opened 2023-08-04
[INTEG-965-973] Updates to Field and Content Selectors

whitelisab:
## Purpose

This PR addresses some follow up items for the new AICG dialog design:

1. Updates the behavior when changing between content sources
2. Creates a generic Selector component that can be used for the content source, source field, and output field selectors
3. Changes the default content source to "Field"
4. Changes the sparkle icon to the new one with a black fill

## Approach

Notes on how the behavior when changing between content sources is changing:
Previously, if you updated the content source then it would change the input/output field selections. Also, the input text was not retained when switching between sources. The new behavior is that when you switch between content sources, the field selections don't change and the input text you had for that content source is retained.

## Testing steps

Open the AICG dialog and toggle between "Field" and "Prompt" for the content source. You should see that the field selections don't change and that any input you have in the text area is retained when switching between them.


apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
```diff
@@ -33,6 +33,10 @@ const CommonGenerator = () => {
 
   useEffect(updateProviderData, [dispatch]);
 
+  const inputText = parameters.isNewText
+    ? parameters.originalText.prompt
+    : parameters.originalText.field;
```
rusticpenguin, line 38: This is a clean change!

rusticpenguin (approved):
Really beautiful code here! This brings so many great improvements, good job!

PR #4404, opened 2023-08-04
Fix (dependabot): Remove deprecated sentry tracing package

lilbitner:
## Purpose

This PR resolves a dependabot build issue found in the GA4 app, in [this](https://github.com/contentful/apps/pull/4350) PR. The dependabot update was updating `@sentry/tracy` package, which caused a type error in the build. This package is actually deprecated and no longer necessary, as its functionality is encapsulated by the Sentry react sdk. 

## Approach

I have updated the file where the type error is being thrown by adjusting the import. Additionally have removed the `@sentry/tracing` package entirely as its antiquated now. 

Confirmed that the app still works as before. 

Sentry React package: https://www.npmjs.com/package/@sentry/react 
Sentry Tracing package deprecation info: https://www.npmjs.com/package/@sentry/tracing

### DIFF
--- a/apps/google-analytics-4/frontend/package-lock.json
+++ b/apps/google-analytics-4/frontend/package-lock.json
@@ -14,7 +14,6 @@
         "@contentful/react-apps-toolkit": "1.2.16",
         "@segment/analytics-next": "^1.51.3",
         "@sentry/react": "^7.44.2",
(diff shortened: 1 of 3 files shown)


matthew-gordon (approved):
👍 🧹 ✨

PR #4419, opened 2023-08-07
[INTEG-1027] Launching a basic hosted app action for AIIG

ryunsong-contentful:
## Purpose
Quickly test app actions connection with experience-package changes as quickly as possible

## Approach
It is the quickest way possible

## Testing steps
Check if the AWS resources are deployed

## Breaking Changes
None

None

## Deployment
Standard


matthew-gordon (commented):
This looks good, I think you are missing a `build-actions.js` at the root, you can copy from this example:

https://github.com/contentful/apps/blob/master/apps/aws-amplify/build-actions.js

We will also need to add `esbuild` as a dev dependency

Lastly, we need to add a new `build actions` script, and update the original build command with both:

https://github.com/contentful/apps/blob/master/apps/aws-amplify/package.json#L25
https://github.com/contentful/apps/blob/master/apps/aws-amplify/package.json#L19

apps/ai-image-generator/frontend/actions/aiig.ts
```diff
@@ -0,0 +1,21 @@
+import { AppActionCallContext } from '@contentful/node-apps-toolkit';
+
+interface AppActionCallParameters {
+  amplifyWebhookUrl: string;
+}
+
+export const handler = async (payload: AppActionCallParameters, context: AppActionCallContext) => {
+  const { amplifyWebhookUrl } = payload;
+  try {
+    await fetch(amplifyWebhookUrl, {
```
matthew-gordon, line 10: I know this is copy/pasta and temporary but we should probably rename this

matthew-gordon (approved):
Awesome 💪 too legit to quit 🏁 ✨

PR #4430, opened 2023-08-08
Resolve Salesforce app TypeError

lilbitner:
## Purpose

This PR fixes a TypeError that a customer encountered while using the Salesforce app. See the thread [here](https://discord.com/channels/1030176999471321129/1111352040413724712/1138094064558952478). 

## Approach

I had difficulty replicating this issue locally, and I believe it is because of variance in images being provided as part of a product object. Locally these images are provided, but I am making the assumption, in this first attempt at resolving the customer's issue, that images are not always guaranteed. This can also be further supported by the way the code was setup to conditionally render the images if they exist. Unfortunately the code did not originally account for an undefined image object, so I have added several safeguards here. 

I can also add optional chaining to the `product` objects, but felt this change would be a good first step since the `image` object itself is the one throwing the error as `undefined`.  

Salesforce app works as before, but will hopefully not throw this blocking error for the user now.

### DIFF
--- a/apps/salesforce-commerce-cloud/src/components/dialog/ProductPreview.tsx
+++ b/apps/salesforce-commerce-cloud/src/components/dialog/ProductPreview.tsx
@@ -14,11 +14,11 @@ const ProductPreview = (props: { product: any }) => {
         : product.shortDescription.default.markup;
   }
 
+  const imageUrl = product.image?.absUrl;
+
   return (
     <>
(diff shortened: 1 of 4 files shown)


matthew-gordon (approved):
Awesome, noice and clean 🥂 ✅

whitelisab (approved):
Looks good to me! Confirmed the app is working as expected

PR #4431, opened 2023-08-08
[INTEG-941] Revert aiig from using vite to create react app

ryunsong-contentful:
## Purpose
Vite is incompatible with HAA so we have to use CRA for now

## Approach
HAA will not work with Vite

## Testing steps
HAA attached the app action successfully

## Breaking Changes
None

CRA instead of Vite

## Deployment
Standard


matthew-gordon (approved):
LGTM, maybe there is some more discovery around getting this to work with Vite 🤔 ✅

PR #4432, opened 2023-08-08
[INTEG-1034] Update AICG config screen copy and styling

whitelisab:
## Purpose

This PR updates the configuration screen for the AI Content Generator app to match the new designs. Here is the updated configuration screen:

Screenshot 2023-08-08 at 4 33 08 PM

## Approach

I created a new component structure. `ConfigPage` is the parent component, and then each section is its own component. This will make it easier to implement the rest of the Brand Profile section as well as the Content Type selection in follow up work.

## Testing steps

Install the app and change the configuration page fields. It should save the configuration information as app installation parameters.


rusticpenguin (approved):
This looks good!

PR #4481, opened 2023-08-14
[INTEG-1064] HAA adds 3 more hardcoded urls and return it as an array

ryunsong-contentful:
## Purpose
Experience package PR relies on this change https://github.com/contentful/experience-packages/pull/2234. This is so that we can deal with an array of images and return them so we can continue developing the modal

## Approach
Hardcode urls

## Testing steps
Tested it in the user_interface override - aka local development

## Breaking Changes
None

None

## Deployment
Standard

### DIFF
--- a/apps/ai-image-generator/frontend/build-actions.js
+++ b/apps/ai-image-generator/frontend/build-actions.js
@@ -58,7 +58,7 @@ const main = async (watch = false) => {
       platform: 'node',
       outdir: 'build',
       logLevel: 'info',
-      format: 'cjs',
+      format: 'esm',
       target: 'es6',
       external: ['node:*'],
     };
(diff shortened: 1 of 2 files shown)


apps/ai-image-generator/frontend/build-actions.js
```diff
@@ -58,7 +58,7 @@ const main = async (watch = false) => {
       platform: 'node',
       outdir: 'build',
       logLevel: 'info',
-      format: 'cjs',
+      format: 'esm',
```
lilbitner, line 61: definitely want to keep an eye on this when we build for production, but if that helps it work locally then 👍

lilbitner (approved):
✅ !

PR #4482, opened 2023-08-14
feat: introduce integration context [MONET-1338]

marcolink:
## Purpose
Prepare `ecommerce-app-base` for future changes

## Approach

Introduce an Integration (React) context to avoid excessive prop drilling through the code in the future.  
Allow react hooks usage by migrating the Field component to a functional component.  

This is exclusively a cleanup up task for future work on the codebase. No changes to the public interface have been intorduced.

## Testing steps
For now this is only tested with realtive-deps on the shopify app

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ tags
 **/.coverage/**
 lerna-debug.log
 .eslintcache
+.tool-versions
--- a/packages/ecommerce-app-base/package-lock.json
+++ b/packages/ecommerce-app-base/package-lock.json
@@ -13,6 +13,7 @@
         "@contentful/f36-components": "^4.0.42",
         "@contentful/f36-icons": "^4.25.0",
         "@contentful/f36-tokens": "^4.0.0",
+        "@contentful/react-apps-toolkit": "^1.2.16",
         "array-move": "^3.0.0",
         "contentful-management": "^10.0.0",
         "emotion": "^10.0.0",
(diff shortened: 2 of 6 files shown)


lilbitner (commented):
The code looks great. Based on the code changes here, it seems like there would be little disruption to the output of the `ecommerce-app-base` package, but am wondering if moving forward the intention is to change the resulting package extensively for the current apps that utilize that package (I believe `Saleor`, and `commerce-tools`, `commercetools-without-search`), and if so, how might that be tested and managed?

marcolink:
> The code looks great. Just curious around a few testing details. Based on the code changes here, it seems like there would be little disruption to the output of the `ecommerce-app-base` package, but am wondering if moving forward the intention is to change the resulting package extensively for the current apps that utilize that package (I believe `Saleor`, and `commerce-tools`, `commercetools-without-search`), and if so, how might that be tested and managed?

@lilbitner valid point 🙏  - what I think we should do:
- Add more tests (ins a separate PR) to ensure everything works as expected BEFORE we start ~changing~/adding functionality
- Add more functionality (the plan is to NOT introduce a breaking change at this point, to avoid any manual migration on the downstream consumers). Said that, the new features should be (where possible) opt-in and configurable through the integration props. 

Do you have anything else in mind?

lilbitner:
> > The code looks great. Just curious around a few testing details. Based on the code changes here, it seems like there would be little disruption to the output of the `ecommerce-app-base` package, but am wondering if moving forward the intention is to change the resulting package extensively for the current apps that utilize that package (I believe `Saleor`, and `commerce-tools`, `commercetools-without-search`), and if so, how might that be tested and managed?
> 
> @lilbitner valid point 🙏 - what I think we should do:
> 
> * Add more tests (ins a separate PR) to ensure everything works as expected BEFORE we start ~changing~/adding functionality
> * Add more functionality (the plan is to NOT introduce a breaking change at this point, to avoid any manual migration on the downstream consumers). Said that, the new features should be (where possible) opt-in and configurable through the integration props.
> 
> Do you have anything else in mind?

I think that is a great approach for future changes here @marcolink. Thank you for clarifying!

lilbitner (approved):

PR #4483, opened 2023-08-14
refactor: separate backend from frontend

jsdalton:
## Purpose

Coupling the frontend and the backend creates numerous downsides for us, including:

* Can't use Vite, must use CRA
* Backend dependencies end up in frontend
* Weird programmatic build instead of standard build

Separating the backend and frontend into isolated projects works better.

## Approach

* All backend "actions" code lives in `backend` folder. It's a separate NPM project, with a separate build and separate installs.
* Frontend continues to live in `frontend` but all code and dependencies related to actions is removed.
* A top level project is created which solely exists to build the subprojects, host the contentful-app-upload actions, and supervise a shared build folder


apps/ai-image-generator/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "ai-image-generator",
+  "version": "0.1.0",
```
ryunsong-contentful, line 3: do we care about matching these versions to start initially? Idk how much we care about these things since we don't even bother to update them 🤷

apps/ai-image-generator/contentful-app-manifest.json
```diff
@@ -6,8 +6,7 @@
       "category": "Custom",
       "description": "This action is used to launch hosted app builds for AIIG.",
       "type": "function",
-      "path": "actions/aiig.js",
-      "entryFile": "./actions/aiig.ts",
+      "path": "backend/actions.js",
```
ryunsong-contentful, line 9: You or Matt will have to explain to me tomorrow why this works without the entry file that points to `actions.ts`

apps/ai-image-generator/backend/scripts/validate-actions.js
```diff
@@ -0,0 +1,33 @@
+const manifest = require('../../contentful-app-manifest.json');
+
+const validateActions = () => {
+  const requiredProperties = ['id', 'path'];
```
ryunsong-contentful, line 4: I lack understanding on why we don't have to check for the entryFile

apps/ai-image-generator/backend/scripts/validate-actions.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    uniqueValues.add(path);
+    uniqueValues.add(id);
+  });
+};
+
+const main = async () => {
+  console.log('Building app actions');
+  validateActions();
```
ryunsong-contentful, line 30: Also would love to know more about why we don't care to look for esBuild configs

apps/ai-image-generator/backend/scripts/validate-actions.js
jsdalton: It's a great question. From what i can tell this is not needed. It's only need for the previous `build-actions.js` script to work (since the entry file value is provided to esbuild).

apps/ai-image-generator/backend/scripts/validate-actions.js
jsdalton: Yes! We ditch esbuild entirely and just use `tsc` (same way most other typescript projects are built today. We could use esbuild if we wanted or any build. But it's separate from this script now. We just build in our project.

apps/ai-image-generator/contentful-app-manifest.json
jsdalton: But BTW i wouldn't be surprised if I"m wrong here, maybe it wont' work. 🤷 

The main reason I have a hunch I'm right: Our previous entryFile is a typescript file `./actions/aiig.ts` and there are no typescript files in our build. So that's why I believe this is only a compile time thing vs. a runtime thing.

jsdalton: As (somewhat) expected `entryFile` is just a compile thing that was needed by the previous `build-actions.js` but not actually required.

apps/ai-image-generator/backend/src/actions.ts
ryunsong-contentful, line 1: Should we consider naming this to be GenerateImageAction? Since we know we'll have one action per function - naming everything actions.ts (even the plurality) makes less sense

ryunsong-contentful (approved):
Approved - just not the one small comment about file naming. 

This is tangential and not the point of this PR, but I wonder if separating the builds are enough? I think for now it is definitely 100% okay having the backend and frontend deploys bundled together as there's no clear workaround anyways.

I'll file a ticket to keep track of separating the frontend and backend deploys

apps/ai-image-generator/backend/src/actions.ts
jsdalton: Yes. We should. I propose we just rename and do a bit of refactoring when we do the actual implementation of this (vs. trying to guess how it will look right now).

PR #4484, opened 2023-08-14
[INTEG-1031] Fix: AICG modal height

whitelisab:
## Purpose

The new vertical modal design was not very useable on small screens. This PR changes the output textarea to be responsive to screen size. There is also a fix in here to reduce re-renders in the Dialog that are being caused by an addition to the dependency array of a `useEffect` for the generator components.

Examples of the modal on a small browser window:
Screenshot 2023-08-14 at 3 50 30 PM
Screenshot 2023-08-14 at 3 51 14 PM

## Approach

In order to fix the modal size, I removed the `minHeight` prop from the `dialogConfig` and added the `useAutoResizer` hook to the `Dialog` component. I then updated the `Output` component to calculate its height based on the heights of the other elements in the modal. Note that the Rewrite action has a different layout than the other actions, so I accounted for this by passing in the height of of the other components as a prop to `Output`.

## Testing steps

Resize your browser and try opening all of the different actions. You should be able to see all of the content in the modal (including the generate button) on a variety of sizes. I did add a minimum modal height of 620 pixels, as anything smaller on the rewrite action made the textarea too small.

### DIFF
--- a/apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
+++ b/apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
@@ -1,5 +1,5 @@
-import { useContext, useEffect, useReducer } from 'react';
-import { Flex } from '@contentful/f36-components';
+import { useContext, useEffect, useReducer, useRef, useState } from 'react';
+import { Box, Flex } from '@contentful/f36-components';
 import { GeneratorContext } from '@providers/generatorProvider';
 import SourceAndFieldSelectors from '@components/app/dialog/common-generator/field-selector/SourceAndFieldSelectors';
 import Header from '@components/app/dialog/common-generator/header/Header';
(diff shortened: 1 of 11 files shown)


apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
```diff
@@ -26,14 +26,23 @@ const CommonGenerator = () => {
...
+  const headerRef = useRef<HTMLElement>(null);
 
   const updateProviderData = () => {
     setProviderData({
       dispatch,
     });
   };
 
-  useEffect(updateProviderData, [dispatch, setProviderData]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
matthew-gordon, line 38: we may want to just add `setProviderData` to the dependency array here, this could lead to buggy code potentially

matthew-gordon (approved):
LGTM, working as expected small comment (non blocker) ✅

apps/ai-content-generator/src/components/app/dialog/common-generator/CommonGenerator.tsx
whitelisab: Currently having `setProviderData` in the dependency array is causing lots of re-renders for the `CommonGenerator` component. I'm removing it and adding this comment for now and we can come back to it to refactor.

PR #4497, opened 2023-08-15
chore: add storybook [MONET-1338]

marcolink:
## Purpose

Use storybook for faster development cycles of visual components

## Approach

Add storybook and add basic stories for existing components

Screenshot 2023-08-15 at 22 21 38


whitelisab (approved):
Definitely makes sense to add Storybook here to aid in the development of these new components. LGTM!

PR #4498, opened 2023-08-15
Add new components to config page and add storybook for development for AI Image Generator

lilbitner:
## Purpose

This PR adds the basic config page functionality for the AI Image Generator ✨ 

Additionally it adds storybook for development 🖌️ , as well as absolute import configurations 🧹 

Ultimately, since these components are almost identical to the AI Content Generator, it would be ideal to add these to the Integration Component Library. I will do so in a follow-up task after we reach the steel thread (it will require resolving some issues at that library as well).

## Approach

The API key section has been added here. Installing the app with the key will save to the app installation params. Additionally there exists error handling for an empty API key i.e we will prevent install if the field is empty. 

Screenshot 2023-08-15 at 4 28 46 PM

## Testing

Go ahead and pull this down and use the AI Image Generator custom app to view this config page locally. Right now you can enter an string of characters for the API key and install it. You should not be able to install if the API key is empty.


ryunsong-contentful (approved):

apps/ai-image-generator/frontend/src/configs/launch-darkly/launchDarklyConfig.ts
```diff
@@ -1,5 +1,6 @@
 const launchDarklyConfig = {
-  clientSideID: `${process.env.REACT_LAUNCHDARKLY_CLIENT_ID}`,
+  // TO DO: Add as env variable
+  clientSideID: '588a047044b03e0b3211298e',
```
ryunsong-contentful, line 3: we probably want to undo this for the PR - Here's the ticket to fix it https://contentful.atlassian.net/browse/INTEG-1070

lilbitner: good call! just updated

PR #4508, opened 2023-08-16
fix(deps): slack fronend lock file [NONE]

marcolink:
## Purpose
Last pushed changed (adding linting) on `master`.  

## Approach

The steps executed on master are different from those on pull requests. Turned out, some lock files where out of sync with their package.json file. Additionally, the slack/frontend was causing issues with the current set of deps version.
This has been fixed by switching out the depricated `@contentful/eslint-config-extension` package. 

With proper linting in place, some errors had to be suppressed in order to pass the build steps. 
Ideally, we manage to remove those linting comments. (I explicitly didn't try to fix those, to not intorduce any side-effects with this PR).

## Testing steps

To test the changes against the steps that broke before, we temporarily changed to circleci config to run one of the install commands the same way as on master 
Screenshot 2023-08-16 at 17 05 59

This is reverted now. The successful run can be found [here](https://app.circleci.com/pipelines/github/contentful/apps/23011/workflows/ef13f00d-22d6-4024-844a-06926e349b52/jobs/66314).


apps/ai-image-tagging/frontend/package.json
```diff
@@ -6,7 +6,8 @@
   "devDependencies": {
     "@testing-library/react": "11.2.5",
     "cross-env": "7.0.3",
-    "fetch-mock": "9.11.0"
+    "fetch-mock": "9.11.0",
+    "typescript": "^4.9.5"
```
andipaetzold, line 10: Pin this

apps/slack/frontend/package.json
```diff
@@ -38,8 +38,8 @@
     ]
   },
   "devDependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
```
andipaetzold, line 41: Why is this needed? Where is it used?

andipaetzold (approved):

PR #4510, opened 2023-08-16
chore: fix package publish step [NONE]

marcolink:
## Purpose

Latest changes broke the publish step for packages (see circleci build [here](https://app.circleci.com/pipelines/github/contentful/apps/23017/workflows/80d00eeb-0ff6-4622-b019-0beca72b8319/jobs/66343)). After some googling, it looks like `prettier@3` is not compatible with `lerna@5` (we should definitely update to `lerna@6` eventually, to get the benefits from `nx`). 

## Approach
Downgrading prettier to the latest 2.x version, and hope it finally publishes again 🤞 
(running a different major version from prettier after the install caused again formatting changes) 

## Testing steps
AFAIK there is no other way than merging this, and see if the publish step works.


matthew-gordon (approved):
I see mostly formatting, and downgrade to `prettier@2.8.8` LGTM

I will look into the build failures 👀

jsdalton (approved):
👍

PR #4512, opened 2023-08-16
Fix package naming, version and metadata

jsdalton:
## Purpose

Install was occasionally missed during CI build of ai-image-generator-backend.

## Approach

* Fix the issue for now by moving the main build steps back into frontend
* Hard code the npm ci requirement for backend when called from frontend (to ensure everythign is installed)
* Need to revisit ot find a better way!


ryunsong-contentful (approved):

ryunsong-contentful (commented):
Still having issues with HAA from the user_interface side - however, this may fix the /apps build for now so we can push this forward and then fix the issue below
```
{
    "body": "{\"errorMessage\":\"exports is not defined\",\"errorType\":\"ReferenceError\"}",
    "statusCode": 200
}
```

PR #4513, opened 2023-08-16
[INTEG-1068] AICG config screen brand profile section

whitelisab:
## Purpose

This PR adds the brand profile fields on the config screen for AI Content Generator. 

Screenshot 2023-08-16 at 11 58 35 AM

## Approach

Added new fields on the config screen according to the designs. I also updated the `useSaveConfigHandler` hook to reduce the calls to `sdk.app.getCurrentState()`. I also updated some tests, more will be added later 😄 This PR does not sync up the new brand profile info to the prompts yet, that will be addressed in a follow up PR.

## Testing steps

Navigate to the AI Content Generator app config page to see the new fields. Add some text, save the config, and the inputs should be persisted in the app installation parameters.


apps/ai-content-generator/src/components/config/configText.ts
```diff
@@ -6,10 +6,66 @@ const ModelText = {
...
-  title: 'Describe your brand or product.',
-  textAreaPlaceholder: 'Example: Contentful is a headless content management system.',
-};
+const BrandProfileFields = [
+  {
+    id: 'profile',
+    title: 'Describe your brand or product.',
+    textAreaPlaceholder: 'Example: Contentful is a headless content management system.',
+    isRequired: true,
+    type: 'textarea',
```
lilbitner, line 15: I think this a great way to organize lots of copy! nit: I am wondering if the 'type' attribute should also be 'typed' in the sense that there exist only 'textarea' | 'textinput' etc. Just to keep things strict here? Unless this is sort of a one-off detail for this particular implementation. And perhaps type could be labeled as `fieldType`?

apps/ai-content-generator/src/components/config/profile/Profile.tsx
```diff
@@ -1,31 +1,48 @@
...
-import { ProfileText } from '../configText';
+import { BrandProfileFields } from '../configText';
 
 interface Props {
-  profile: string;
+  profile: { [key: string]: string };
   dispatch: Dispatch<ParameterReducer>;
 }
 
+const TEXTAREA_ROWS = 5;
```
lilbitner, line 11: love pulling this out into a constant. No magic numbers 😅 🪄

apps/ai-content-generator/src/locations/ConfigScreen.spec.tsx
```diff
@@ -12,9 +13,8 @@ vi.mock('@contentful/react-apps-toolkit', () => ({
 }));
 
 describe('Config Screen component', () => {
-  it('Component text exists', async () => {
+  it('Config header exists', async () => {
     const { getByText } = render(<ConfigScreen />);
-    // simulate the user clicking the install button
-    expect(getByText('OpenAI API key')).toBeTruthy();
+    expect(getByText(Sections.pageHeading)).toBeTruthy();
```
lilbitner, line 18: love constants in tests 🙌  single source of truth 👌

lilbitner (approved):
Looks awesome! Just had a few small comments, non blocking. I pulled this down to look at it but might not be privy to what environment it works in/if I need to adjust the flag? I was only able to see the old config page in AI Content Generator (development) (must be something small/silly I am doing!)

apps/ai-content-generator/src/components/config/configText.ts
whitelisab: Great suggestions, I will update to `fieldType` and added an enum for `FieldTypes`!

PR #4514, opened 2023-08-16
Refactor/builds

jsdalton:
## Purpose

One final attempt to restore sanity to the setup of this project. The current way had weird cross depedencies.

## Approach

* top level: a simple npm project that knows how to build the frontend and backend to a common build folder and then deploy that folder
* frontend: a basic react app that doesn't know anything about the backend build steps or care. has an npm build script that builds as normal
* backend: a basic set of function that doesn't know anything about the frontend build steps or care. has an npm build script that builds as normal.

The main downside here is a _bit_ of extra work to do `npm ci` before build at the top level just in case the full npm install has not already run.

Also adding back the regular build means we ensure those builds work (in case the top level build doesn't run)

### DIFF
--- a/apps/ai-image-generator/package.json
+++ b/apps/ai-image-generator/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "create-app-definition": "contentful-app-scripts create-app-definition",
     "build": "rimraf ./build && npm run build-frontend && npm run build-backend",
-    "build-backend": "cd frontend && npm ci --include=dev && npm run build",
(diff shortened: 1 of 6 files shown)


apps/ai-image-generator/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@contentful/ai-image-generator",
+  "version": "0.1.0",
+  "private": true,
```
ryunsong-contentful, line 4: is this needed anymore?

ryunsong-contentful (approved):

PR #4543, opened 2023-08-18
[INTEG-1069] AICG Utilize brand fields in prompts

whitelisab:
## Purpose

This PR hooks up the new brand profile fields on the config page with the prompts. I also made a few small changes to the rewrite prompt (we didn't need to pass in locale), and removed the brand/copy link from the config page (this is something that we will not support now as it doesn't work reliably with ChatGPT).

## Approach

The brand profile information is being passed in as system context. Since most of these fields are optional, I constructed the system context prompt based on whether the user has added input in those fields.

## Testing steps

When you add content to the brand profile fields in your config page, the output from ChatGPT should reflect those changes. Try updating the fields on the config page and see how the output changes.


apps/ai-content-generator/src/configs/prompts/baseSystemPrompt.ts
```diff
@@ -1,13 +1,64 @@
 import { ChatCompletionRequestMessage } from 'openai';
+import { ProfileFields } from '@components/config/configText';
+import { ProfileType } from '@locations/ConfigScreen';
 
-const baseSystemPrompt = (profile: string, locale: string): ChatCompletionRequestMessage[] => [
+const generateBrandProfile = (profile: ProfileType) => {
```
matthew-gordon, line 5: This is great! How would you feel about maybe using an array to push prompt pieces, and then join them at the end? Also I think we can destructure typescript enums, Just thinking of readability 🤔 not opposed to leaving it as is, but could look something like this:

```typescript
const generateBrandProfile = (profile: ProfileType) => {
  const {
    PROFILE,
    ADDITIONAL,
    VALUES,
    TONE,
    AUDIENCE,
    EXCLUDE,
    INCLUDE
  } = ProfileFields;

  const profilePromptParts = [];

  if (profile[PROFILE]) {
    profilePromptParts.push(`You are working for a company with the following profile: ${profile[PROFILE]}.`);
  }

  if (profile[ADDITIONAL]) {
    profilePromptParts.push(`It is important that you also know the following details about the company: ${profile[ADDITIONAL]}.`);
  }

  if (profile[VALUES]) {
    profilePromptParts.push(`Your company's brand has the following values and attributes: ${profile[VALUES]}.`);
  }

  if (profile[TONE]) {
    profilePromptParts.push(`Your company's voice can be described as ${profile[TONE]}.`);
  }

  if (profile[AUDIENCE]) {
    profilePromptParts.push(`Your company's target audience is: ${profile[AUDIENCE]}.`);
  }

  if (profile[EXCLUDE]) {
    profilePromptParts.push(`Your company's brand states that you should never use the following words: ${profile[EXCLUDE]}.`);
  }

  if (profile[INCLUDE]) {
    profilePromptParts.push(`Your company's brand commonly uses these words: ${profile[INCLUDE]}.`);
  }

  const profilePrompt = profilePromptParts.join(' ');

  return profilePrompt;
};
```

apps/ai-content-generator/src/configs/prompts/baseSystemPrompt.ts
```diff
@@ -1,13 +1,64 @@
...
+      profile[ProfileFields.EXCLUDE]
+    } . `;
+  }
+
+  if (profile[ProfileFields.INCLUDE]) {
+    profilePrompt += `Your company's brand commonly uses these words: ${
+      profile[ProfileFields.INCLUDE]
+    } . `;
+  }
+  console.log(profilePrompt);
```
matthew-gordon, line 45: Is this for debugging, or can we 🧹 💣 💥 ?

matthew-gordon (approved):
LGTM!! 🌮 👾 ✨ Small comments nothing blocking, feel free to ignore my suggestion if it doesn't make sense!

apps/ai-content-generator/src/configs/prompts/baseSystemPrompt.ts
whitelisab: 🧹 !

apps/ai-content-generator/src/configs/prompts/baseSystemPrompt.ts
whitelisab: This suggestion looks great, thank you! Making the changes.

PR #4562, opened 2023-08-21
[INTEG-1092-1093] AICG validation for installation parameters

whitelisab:
## Purpose

This PR addresses a few things:

1. If there are no app installation parameters (e.g. if the app is installed programmatically) then it renders a warning in the Sidebar to have the user update their app configuration
2. On the configuration screen, if any of the 3 required fields (API Key, model, and brand profile) are not filled out, then the user is not able to save the config and a toast message lets them know what field to complete
3. I added some common components for Note and HyperLink

## Approach

1. Sidebar warnings for missing app installation parameters
Below is the sidebar warning for missing an API Key or Model. In this case, the warning takes up the whole Sidebar since missing either of these will mean that the generate operation in the modal will fail.
Screenshot 2023-08-21 at 9 32 58 AM

Below is the sidebar warning for missing the brand profile. In this cases, I am still showing the buttons since the generate operation will run, but the results will be improved if they complete their profile.
Screenshot 2023-08-21 at 9 33 24 AM

2. Below is an example of the toast errors if you are missing required installation parameters. I updated the `useSaveConfigHandler` hook to accept a function for validating parameters. If this function returns errors, then the toasts appear and the config is not saved.
Screenshot 2023-08-21 at 9 34 31 AM

3. The `Note` and `HyperLink` components are the same as the ones in the `integration-component-library`, and we should use the components from the library once we get that resolved.

## Testing steps

On the config screen, leave some of the required fields (API Key, model, brand profile) blank to see the error toast on save. For the sidebar, have some of the parameters missing to see the warning messages.


rusticpenguin (approved):
Looks great!

PR #4563, opened 2023-08-21
Fix build to match correct runtime

jsdalton:
## Purpose

Since the AIIG hosted app action is an internal org, we recently realized this will run with the Node18 runtime instead of Deno. We also fixed our development organization to be marked as "internal" to better align with this.

Doing that broke our current build settings, so we need to make some tweaks.

## Approach

* Set .nvrmc explicitly to the LTS version of Node 18 used by AWS (`lts/hydrogen`)
* Set target to `cjs` since that's what the Node18 runtime uses
* Align the `tsconfig` settings

* https://aws.amazon.com/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/

### DIFF
--- a/apps/ai-image-generator/backend/.nvmrc
+++ b/apps/ai-image-generator/backend/.nvmrc
@@ -1,1 +1,1 @@
-v16.18
+lts/hydrogen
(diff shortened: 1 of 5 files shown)


ryunsong-contentful (approved):
New bundle and it still works

Screenshot 2023-08-21 at 12 38 23 PM
entful/apps/assets/124832189/d72605c2-d411-425d-8d8d-a1721660e613)

PR #4564, opened 2023-08-21
Update responses to agreed upon types

jsdalton:
## Purpose

We need to have a fairly standard yet lightweight way to differentiate between successful and unsuccessful app action calls.

## Approach

* Implement a "result type" to differentiate between error and success
* Update the generate image action to use this format

### DIFF
--- a/apps/ai-image-generator/backend/src/actions/aiig-generate-image.ts
+++ b/apps/ai-image-generator/backend/src/actions/aiig-generate-image.ts
@@ -6,11 +6,29 @@ interface AppActionCallParameters {
   prompt: string;
 }
 
-interface AppActionResponse {
-  status: number;
-  prompt: string;
-  images: string[];
+interface Image {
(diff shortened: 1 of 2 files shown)


apps/ai-image-generator/backend/src/actions/aiig-generate-image.ts
```diff
@@ -6,12 +6,30 @@ interface AppActionCallParameters {
...
+    images: Image[];
+  };
+}
+
+export interface AppActionCallResponseError {
+  ok: false;
+  error: {
+    type: string;
+    message: string;
+    details: Record<string, any>;
```
ryunsong-contentful, line 27: This opens up a lot of interesting options. Would be curious to see what your intention is down the line when we start formalizing and tackling the error side of things in a few weeks.

ryunsong-contentful (approved):

PR #4591, opened 2023-08-23
fix: add hosted action deps []

kdamball:
## Purpose

We had them in typescript template but they got deleted. Readding them to the correct folder this time


examples/hosted-app-action-templates/javascript/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "devDependencies": {
+    "@contentful/node-apps-toolkit": "2.2.0",
```
andipaetzold, line 3: We don't use the dependency here. It's only used as type import for the typescript-one

examples/hosted-app-action-templates/typescript/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "devDependencies": {
+    "@contentful/node-apps-toolkit": "^2.2.0",
```
andipaetzold, line 3: Pin?

Also, why a dev dependency if it's used in runtime code?

examples/hosted-app-action-templates/javascript/package.json
kdamball: It's only used for types :)

andipaetzold: But it can be used to import methods 🤔   So users would need to move it over.

kdamball: I meant to comment on TS file. but which methods do you refer to?

andipaetzold (approved):

examples/hosted-app-action-templates/javascript/package.json
kdamball: That's why we have hosted app actions in the first place 😛 We don't need any of that stuff anymore. Nor getting app tokens

PR #4592, opened 2023-08-23
Add image type to backend result

jsdalton:
## Purpose

We don't want to hardcode an assumption in the front end that image types will always be a certain type (ie png) so instead we want the backend to specify the image type explicitly.

## Approach

* Add an `imageType` attribute
* Hard code to `png` for OpenAI
* Idea is it's the second half of the mime type https://www.iana.org/assignments/media-types/media-types.xhtml#image

### DIFF
--- a/apps/ai-image-generator/backend/src/actions/aiig-generate-image.ts
+++ b/apps/ai-image-generator/backend/src/actions/aiig-generate-image.ts
@@ -1,6 +1,7 @@
 import { AppActionCallContext } from '@contentful/node-apps-toolkit';
 import { PlainClientAPI } from 'contentful-management';
 import { OpenAiApiService } from '../services/openaiApiService';
+import OpenAI from 'openai';
 
 interface AppActionCallParameters {
   prompt: string;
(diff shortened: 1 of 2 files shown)


apps/ai-image-generator/backend/src/actions/aiig-generate-image.ts
```diff
@@ -57,7 +58,9 @@ export const handler = async (
     numImages: 4,
     size: '1024x1024',
   });
-  const images = openAiImages.filter((image): image is Image => !!image.url);
+  const images = openAiImages
+    .map((image) => ({ url: image.url, imageType: 'png' }))
```
ryunsong-contentful, line 62: Wait - small optimization here. Why can't we filter first and then map. Seems more efficient right? I guess it doesn't really matter with 4 images

jsdalton: Honest answer: Harder to deal with types to filter first so I swapped it. The OpenAI type is super annoying to work with since `url` is actually optional.

Some ideas here on how we could just create an interim type to work around it better https://stackoverflow.com/questions/40370214/how-to-override-a-property-to-be-non-nullable-in-typescript

PR #4593, opened 2023-08-23
Handle errrors with our action response type

jsdalton:
## Purpose

We'd like to add minimal error handling so we at least handle errors with a common format.

## Approach

* Use our Result type and treat any error as `ok: false`
* Change error to plural (assume possible to have multiple errors at some point)


lilbitner (approved):
Looks good to me! 👍

PR #4605, opened 2023-08-24
feat: update example to make use of new interface [MONET-1335]

BasKiers:
## Purpose

The interface for Delivery Functions has changed with [this PR](https://github.com/contentful/node-apps-toolkit/pull/327)

This PR updates the `hosted-delivery-function-templates` to confirm to the new interface

In addition, the example has been expanded to include a minimal setup to call a GraphQL API to better show developers how they could implement custom delivery functions for their own GraphQL API

Note that the example does call a publicly hosted Starwars GraphQL API. If this is not an appropriate API we can easily change it with a different one.

### DIFF
--- a/examples/hosted-delivery-function-templates/typescript/example.ts
+++ b/examples/hosted-delivery-function-templates/typescript/example.ts
@@ -1,24 +1,47 @@
-import { DeliveryFunctionEventHandlers, DeliveryFunctionRequestEventType } from '@contentful/node-apps-toolkit';
+import {
+  DeliveryFunctionEventHandler,
+  DeliveryFunctionRequestEventType,
+  GraphQLQueryResponse,
+} from '@contentful/node-apps-toolkit';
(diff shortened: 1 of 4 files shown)


examples/hosted-delivery-function-templates/typescript/example.ts
```diff
@@ -1,24 +1,47 @@
...
+
+export const handler: DeliveryFunctionEventHandler = (event, context) => {
+  if (event.type === DeliveryFunctionRequestEventType.GRAPHQL_FIELD_MAPPING) {
+    return fieldMappingHandler(event, context);
+  }
+  if (event.type === DeliveryFunctionRequestEventType.GRAPHQL_QUERY) {
+    return queryHandler(event, context);
   }
-}
+  throw new Error('Unknown Event');
```
kdamball, line 46: Curious, is will this ever be a possibility?

PR #4619, opened 2023-08-25
[INTEG-1044] Remove hidden fields from field selectors

whitelisab:
## Purpose

This PR filters out fields that are marked as `Hidden when editing` so that they are not available as an option in the field selectors in the AICG modal.

## Approach

The field property is called `field.disabled`, so I filtered out fields set to true.

## Testing steps

Add a text field on a content type and mark it as `Hidden when editing`. You should not see it as an option in the modal for either source or output field.


rusticpenguin (approved):
Wooo! This looks great!

PR #4638, opened 2023-08-28
[INTEG-1110] AICG add content type selection to config screen

whitelisab:
## Purpose

Space admins want an easy way to assign the AI Content Generator app to the Sidebar location for multiple content types. This PR adds this ability on the configuration page. See screenshots for what this looks like on the config page and what it looks like when there are no content types in the space:

Screenshot 2023-08-28 at 12 01 36 PM

Screenshot 2023-08-22 at 8 08 27 AM

## Approach

This follows a similar pattern for what we have in other apps where you get all content types from the CMA and get which content types have the app in the sidebar via the editor interface. On saving the config, the selected content types will be added to the editor interface. I added a new hook `useGetContentTypes` and a new reducer `contentTypeReducer` to facilitate this.

## Testing steps

Navigate to the app config page for the AI Content Generator. Here are a few things to test:
- If there are no content types in the space, you should see the warning note and a link to go create a content type
- Content types should be unchecked when initially installing the app
- The select all/deselect all option should work and it should display the total number of content types
- On saving the config, the newly selected content types should have the app assigned to the sidebar
- If the app is manually removed from the sidebar, it should be reflected on the config page
- If a content type is added or deleted, this should be reflected on the config page

### DIFF
--- a/apps/ai-content-generator/src/components/config/content-type-selection/ContentTypeSelection.tsx
+++ b/apps/ai-content-generator/src/components/config/content-type-selection/ContentTypeSelection.tsx
@@ -17,16 +17,8 @@ const ContentTypeSelection = (props: Props) => {
 
   useEffect(() => {
     const areAllItemsChecked = allContentTypes.length === selectedContentTypes.size;
-    const areAllButOneItemsChecked = allContentTypes.length - selectedContentTypes.size === 1;
-
-    if (!isSelectAllChecked && areAllItemsChecked) {
-      setIsSelectAllChecked(true);
-    }
-
-    if (isSelectAllChecked && areAllButOneItemsChecked) {
-      setIsSelectAllChecked(false);
-    }
-  }, [allContentTypes.length, isSelectAllChecked, selectedContentTypes.size]);
+    setIsSelectAllChecked(areAllItemsChecked);
+  }, [allContentTypes.length, selectedContentTypes.size]);
 
   const handleSelectAll = () => {
     if (isSelectAllChecked) {
@@ -35,8 +27,6 @@ const ContentTypeSelection = (props: Props) => {
       const allContentTypeIds = allContentTypes.map((ct) => ct.sys.id);
       dispatch({ type: ContentTypeAction.ADD_ALL, value: allContentTypeIds });
     }
-
-    setIsSelectAllChecked(!isSelectAllChecked);
   };
 
   const handleCheckboxChange = (contentType: ContentTypeProps, isChecked: boolean) => {


apps/ai-content-generator/src/hooks/config/useSaveConfigHandler.ts
```diff
@@ -26,11 +28,30 @@ const useSaveConfigHandler = (
 
     const currentState = await sdk.app.getCurrentState();
 
+    let currentEditorInterface = {};
+
+    if (currentState !== null) {
+      const { EditorInterface } = currentState;
+      currentEditorInterface = EditorInterface;
+    }
```
rusticpenguin, line 36: I feel this could read faster
```suggestion
    const currentEditorInterface = currentState?.EditorInterface || {};
```

apps/ai-content-generator/src/components/config/content-type-selection/ContentTypeSelection.tsx
```diff
@@ -0,0 +1,74 @@
...
+    setIsSelectAllChecked(!isSelectAllChecked);
+  };
+
+  const handleCheckboxChange = (contentType: ContentTypeProps, isChecked: boolean) => {
+    const action = isChecked ? ContentTypeAction.REMOVE : ContentTypeAction.ADD;
+    dispatch({ type: action, value: contentType.sys.id });
+  };
+
+  const isChecked = (contentTypeId: string) => {
+    return selectedContentTypes.includes(contentTypeId);
```
rusticpenguin, line 36: We may want to consider storing the selected content types in a `Set` vs an array. This is creating an O(n^2) on every render.

This would be a pretty easy rewrite, all you'd need to do is to use the `.has()` method here! 

Anywhere you'd actually need the array (like for saving) you would need to just update it to do `Array.from(selectedContentTypes)` to access it as an array. 

And in places where you're adding/deleting there are methods on Set `.delete()` and `.add()`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

apps/ai-content-generator/src/components/config/content-type-selection/ContentTypeSelection.tsx
```diff
@@ -0,0 +1,74 @@
...
+  allContentTypes: ContentTypeProps[];
+  selectedContentTypes: string[];
+  dispatch: Dispatch<ContentTypeReducer>;
+}
+
+const ContentTypeSelection = (props: Props) => {
+  const { allContentTypes, selectedContentTypes, dispatch } = props;
+  const [isSelectAllChecked, setIsSelectAllChecked] = useState(false);
+
+  const allContentTypeIds = allContentTypes.map((ct) => ct.sys.id);
```
rusticpenguin, line 18: Could we just move this map inside of the code block on line 24? That way this doesn't get called on every rerender

apps/ai-content-generator/src/components/config/contentTypeReducer.ts
```diff
@@ -0,0 +1,44 @@
...
+  | ContentTypeRemoveAllAction;
+
+const { ADD, REMOVE, ADD_ALL, REMOVE_ALL } = ContentTypeAction;
+
+const contentTypeReducer = (state: string[], action: ContentTypeReducer) => {
+  switch (action.type) {
+    case ADD:
+      return [...state, action.value];
+    case REMOVE:
+      return state.filter((ct) => ct !== action.value);
```
rusticpenguin, line 34: if we're using a set vs an array, we can do `state.delete(action.value)` here!

apps/ai-content-generator/src/components/config/contentTypeReducer.ts
```diff
@@ -0,0 +1,44 @@
...
+  | ContentTypeActions
+  | ContentTypeAddAllAction
+  | ContentTypeRemoveAllAction;
+
+const { ADD, REMOVE, ADD_ALL, REMOVE_ALL } = ContentTypeAction;
+
+const contentTypeReducer = (state: string[], action: ContentTypeReducer) => {
+  switch (action.type) {
+    case ADD:
+      return [...state, action.value];
```
rusticpenguin, line 32: if using a set, we can do `state.add(action.value)`

apps/ai-content-generator/src/components/config/contentTypeReducer.ts
```diff
@@ -0,0 +1,44 @@
...
+    case REMOVE:
+      return state.filter((ct) => ct !== action.value);
+    case ADD_ALL:
+      return action.value;
+    case REMOVE_ALL:
+      return [];
```
rusticpenguin, line 38: if using a set, we can call `state.clear()` here

apps/ai-content-generator/src/hooks/config/useSaveConfigHandler.ts
```diff
@@ -26,11 +28,30 @@ const useSaveConfigHandler = (
...
+
+    // Assign the app to the sidebar for saved content types
+    const newEditorInterfaceAssignments = generateEditorInterfaceAssignments(
+      currentEditorInterface,
+      contentTypes,
```
rusticpenguin, line 41: here is where you would use `Array.from(contentTypes)` to get access to that array

apps/ai-content-generator/src/components/config/content-type-selection/ContentTypeSelection.tsx
```diff
@@ -0,0 +1,74 @@
...
+      dispatch({ type: ContentTypeAction.ADD_ALL, value: allContentTypeIds });
+    }
+
+    setIsSelectAllChecked(!isSelectAllChecked);
+  };
+
+  const handleCheckboxChange = (contentType: ContentTypeProps, isChecked: boolean) => {
+    const action = isChecked ? ContentTypeAction.REMOVE : ContentTypeAction.ADD;
+    dispatch({ type: action, value: contentType.sys.id });
+  };
```
rusticpenguin, line 33: If we deselect a checkbox when isSelectAllChecked is true, should we set isSelectAllChecked to false?

And vice versa if we select all of the fields by hand, should isSelectedAllChecked become true

rusticpenguin (commented):
This overall looks great! I have a comment on an optimization and a readability thing!

apps/ai-content-generator/src/components/config/content-type-selection/ContentTypeSelection.tsx
whitelisab: Good point, added handling for this!

apps/ai-content-generator/src/components/config/contentTypeReducer.ts
```diff
@@ -0,0 +1,48 @@
...
+  switch (action.type) {
+    case ADD:
+      state.add(action.value);
+      return new Set(state);
+    case REMOVE:
+      state.delete(action.value);
+      return new Set(state);
+    case ADD_ALL:
+      action.value.forEach((item) => state.add(item));
+      return new Set(state);
```
rusticpenguin, line 39: `new Set(action.value)` should work here!

apps/ai-content-generator/src/components/config/content-type-selection/ContentTypeSelection.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const areAllButOneItemsChecked = allContentTypes.length - selectedContentTypes.size === 1;
+
+    if (!isSelectAllChecked && areAllItemsChecked) {
+      setIsSelectAllChecked(true);
+    }
+
+    if (isSelectAllChecked && areAllButOneItemsChecked) {
+      setIsSelectAllChecked(false);
+    }
+  }, [allContentTypes.length, isSelectAllChecked, selectedContentTypes.size]);
```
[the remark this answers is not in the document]
rusticpenguin: and if we want this to not call the set whenever the selectedContentTypes are updated we can do instead
```
if (isSelectAllChecked !== areAllItemsChecked) {
  setIsSelectAllChecked(areAllItemsChecked);
}
```

apps/ai-content-generator/src/components/config/content-type-selection/ContentTypeSelection.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const handleSelectAll = () => {
+    if (isSelectAllChecked) {
+      dispatch({ type: ContentTypeAction.REMOVE_ALL });
+    } else {
+      const allContentTypeIds = allContentTypes.map((ct) => ct.sys.id);
+      dispatch({ type: ContentTypeAction.ADD_ALL, value: allContentTypeIds });
+    }
+
+    setIsSelectAllChecked(!isSelectAllChecked);
```
rusticpenguin, line 39: With the changes above we can just remove this!
```suggestion
```

rusticpenguin (approved):
Looks awesome!!! Great work

PR #4639, opened 2023-08-28
[INTEG-1147] Add default max length to char validation

whitelisab:
## Purpose

For AI Content Generator, we noticed an issue with applying the content to fields. We were handling fields where there was custom size validation, but if there was no custom size validation, then we were not respecting the [default field character length](https://www.contentful.com/developers/docs/technical-limits/). For example, Short Text fields have a default max length of 256 characters. If you had a Short Text field and wanted to output content there that was 500 characters, the apply button was still enabled, you could click it, but then it wouldn't apply and a useful error was not being shown to the user.

## Approach

To fix this, I leaned into our existing handling for character count validation. I added a method when formatting the field response to add in a max length if one doesn't exist. If there was a custom max length, I defaulted to that instead. I added handling for the 3 field types that we support: Short Text, Long Text, and Rich Text.

## Testing steps

Create a Short Text field with no custom validation. Try to add content from the modal that's length is more than 256 characters. It should now prevent you from doing so.

### DIFF
--- a/apps/ai-content-generator/src/utils/dialog/supported-fields/supportedFieldsHelpers.ts
+++ b/apps/ai-content-generator/src/utils/dialog/supported-fields/supportedFieldsHelpers.ts
@@ -7,6 +7,39 @@ import {
 import { ContentFields, EntryProps } from 'contentful-management';
 import { FieldLocales } from '@locations/Dialog';
 import { LocaleNames } from '@providers/generatorProvider';
+import { ContentTypeFieldValidation } from 'contentful-management/types';
+
+/**
+ * This creates the size validation for the field
+ * If no custom size validation exists, use the default character length for the field type
+ * as listed in our technical limits: https://www.contentful.com/developers/docs/technical-limits/.
+ * If a custom max length exists, use that instead of the default.
+ * @param fieldValidations
+ * @param fieldType
+ * @returns ContentTypeFieldValidation
+ */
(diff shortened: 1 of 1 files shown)


apps/ai-content-generator/src/utils/dialog/supported-fields/supportedFieldsHelpers.ts
```diff
@@ -7,6 +7,39 @@ import {
...
+  const defaultMax = DEFAULT_CHAR_LENGTH[fieldType as SupportedFieldTypes];
+
+  const customSizeValidation = fieldValidations?.find((validation) => validation?.size) || {
+    size: { max: defaultMax },
+  };
+
+  if (customSizeValidation.size) {
+    const newMax = customSizeValidation.size?.max ? customSizeValidation.size.max : defaultMax;
+    customSizeValidation.size.max = newMax;
+  }
```
rusticpenguin, line 39: I feel this would read a bit faster:
```suggestion
  if (!customSizeValidation.size?.max) {
    customSizeValidation.size.max = defaultMax;
  }
```

whitelisab: Thank you for this suggestion! I made this update and I also had to add a check to see if `customSizeValidation.size` exists to make TypeScript happy 😄

PR #4655, opened 2023-08-29
feat: add X-Contentful-CRN to cors config

andipaetzold:
Adds the new X-Contentful-CRN header to the cors config.

Why are we so picky about the headers at all? 🤷


rusticpenguin (approved):
This looks good! Gonna talk to the team about making this more general instead of forcing specific headers

PR #4667, opened 2023-08-30
Fix file name length and adjust action name

jsdalton:
## Purpose

Old action name too long

## Approach

* Shorten name
* FIx action id


ryunsong-contentful (approved):
I think we landed on Select-and-Edit but I think we should get this pushed through so that we can test other things and I'll clean up the file naming later on this/next week

PR #4668, opened 2023-08-30
Transform source image and mask image to correct dimensions

jsdalton:
## Purpose

To do an edit of an image, OpenAI requires the source image and mask image to be in a specific format:

* Both must be PNG
* Both must have matching sizes
* Both must be square
* Both must be <4MB

## Approach

* Add an ImageTransformer class to manage the specifics


apps/ai-image-generator/backend/src/transform-images.ts
```diff
@@ -0,0 +1,158 @@
+import * as nodeFetch from 'node-fetch';
+import { default as sharp } from 'sharp';
+import { areEqualColors, toRGBA, toSharp } from './utils';
+
+interface TransformedImages {
+  image: Buffer;
+  mask: Buffer;
+}
+
+export const ERASE_COLOR: sharp.RGBA = { r: 174, g: 193, b: 204 };
```
ryunsong-contentful, line 10: @lilbitner Can you confirm this is the exact values we want

apps/ai-image-generator/backend/src/transform-images.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    return {
+      image: await sharpImage.toBuffer(),
+      dimensions: { width: side, height: side },
+    };
+  }
+
+  private constrainDimensions(dimensions: Dimensions): {
+    side: number;
```
ryunsong-contentful, line 71: Would look cleaner with a defined return type

apps/ai-image-generator/backend/src/transform-images.ts
```diff
@@ -0,0 +1,158 @@
...
+    changed: boolean;
+  } {
+    const { width, height } = dimensions;
+
+    let side: number;
```
ryunsong-contentful, line 76: I think you can clean up this entire logic by doing

```

  private constrainDimensions(dimensions: Dimensions): {
    side: number;
    changed: boolean;
  } {
    const { width, height } = dimensions;
    let side = Math.min(width, height, MAX_SIDE);
    return {
      side,
      changed: side !== width || side !== height,
    };
  }
```

apps/ai-image-generator/backend/src/transform-images.ts
lilbitner: yeah so at this very moment in master this is the correct color (gray400)! I think we might adjust this to `gray300` with some styling updates on the frontend, but makes sense to me to keep this aligned with master right now to test functionality.

apps/ai-image-generator/backend/src/actions/aiig-select-fill.ts
```diff
@@ -32,20 +34,25 @@ export const handler = async (
...
+    const { mask: maskBuffer, image: imageBuffer } = await transformImages({
+      sourceImageResponse,
+      maskImageResponse,
+    });
+
     const openAiImages = await openAiApiService.editImage({
       prompt,
-      image,
-      mask,
+      image: await toFile(imageBuffer),
```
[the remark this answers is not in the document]
jsdalton: Yeah! In general I'm hopeful that it "just works" 🙏

apps/ai-image-generator/backend/src/transform-images.ts
jsdalton: 👍 Yep, and we can evolve this pretty easily as time goes on. (We coud even make it a param to the app action if we wanted to further customability)

apps/ai-image-generator/backend/src/transform-images.ts
```diff
@@ -0,0 +1,158 @@
...
+      mask,
+    };
+  }
+
+  private async normalizeSourceImage(): Promise<{
+    image: Buffer;
+    dimensions: Dimensions;
+  }> {
+    const sharpImage = toSharp(this.sourceImageResponse.body);
+    const initialMetadata = await sharpImage.metadata();
```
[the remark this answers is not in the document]
jsdalton: There's a subtle reason why I separated this. I will add a comment since it's non obvious.

The reason is that for certain formats it's possible _we will not have a width and height from the initial metadata_.

After we force the file to a PNG in the following line, we are _guaranteed_ to have a width and height since it's part of the PNG file spec. So that's why we do the check later, when we konw the values will be present. (And we error if not).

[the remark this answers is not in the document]
jsdalton: Note that I will still streamline this and remove the unneeded interim variable.

apps/ai-image-generator/backend/src/transform-images.ts
```diff
@@ -0,0 +1,158 @@
...
+    let side: number;
+    if (width !== height) {
+      side = Math.min(width, height);
+    } else {
+      side = width;
+    }
+
+    if (side > MAX_SIDE) {
+      side = MAX_SIDE;
+    }
```
rusticpenguin, line 85: nit: This could be abstracted into a clamp function for readability:

```
const clamp = (value, min, max) => Math.min(Math.max(value, min), max);
```

```suggestion
    isNotSquare = width !== height;
    const side = isNotSquare ? clamp(width, height, MAX_SIDE) || width; 
```

apps/ai-image-generator/backend/src/transform-images.ts
jsdalton: Great suggestion. ❤️

apps/ai-image-generator/backend/src/transform-images.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    if (!hasAlpha) return sharpImage;
+
+    const { data: imagePixels, info } = await sharpImage
+      .raw()
+      .toBuffer({ resolveWithObject: true });
+
+    // image "pixels" are sets of four integers in a row, ie [red, blue, green, alpha]. so we
+    // iterate over the image pixels in batches of 4
+    for (let i = 0; i < imagePixels.length; i += 4) {
```
rusticpenguin, line 123: Will this make any pixel that's this color transparent?

apps/ai-image-generator/backend/src/transform-images.ts
rusticpenguin: Oh just saw @ryunsong-contentful had the exact same review!

apps/ai-image-generator/backend/src/transform-images.ts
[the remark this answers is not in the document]
ryunsong-contentful: Okay got it, makes sense now. 👍

apps/ai-image-generator/bundle-sharp.js
```diff
@@ -1,5 +1,7 @@
 const fs = require('fs/promises');
 
+console.log('Copying sharp files from node_modules to build/node_modules');
```
ryunsong-contentful, line 3: This looks intentional to me but just want to double check

ryunsong-contentful (approved):
You will have to teach me what `"install-linux-sharp"` does but that can be done later.
LGTM and I'm going to post this PR https://github.com/contentful/app-upload-to-delivery/pull/48 here so we can keep track

PR #4669, opened 2023-08-31
Chore/use crn

rusticpenguin:
## Purpose

We would like to get access tokens that are location aware on the backend.

## Approach

Reading the `x-contentful-crn` we created helper functions to map the crn to the correct baseUrl. Updating any location that uses the CMA client or calls the api directly

### DIFF
--- a/apps/ecommerce/lambda/src/middlewares/getAppInstallationParameters.ts
+++ b/apps/ecommerce/lambda/src/middlewares/getAppInstallationParameters.ts
@@ -67,8 +67,8 @@ export const getAppInstallationParametersMiddleware: RequestHandler = async (
         Object.assign(installationParameters, appInstallation.parameters);
 
         req.installationParameters = installationParameters;
-      } catch (e: any) {
-        console.error(e.message);
+      } catch (e: unknown) {
+        console.error((e as { message: string }).message);
         throw new UnableToGetAppInstallationParameters(
           `Unable to get app installation parameters: cause: ${e}`
         );
--- a/apps/ecommerce/lambda/src/middlewares/getAppInstallationParameters.ts
+++ b/apps/ecommerce/lambda/src/middlewares/getAppInstallationParameters.ts
@@ -4,6 +4,7 @@ import axios from 'axios';
 import { UnableToGetAppInstallationParameters } from '../errors/unableToGetAppInstallationParameters';
 import { AppInstallationParameters } from '../types';
 import { UnableToGetAppAccessToken } from '../errors/unableToGetAppAccessToken';
+import { getHost } from '../helpers/getHost';
 
 export const getAppInstallationParametersMiddleware: RequestHandler = async (
   req: Request,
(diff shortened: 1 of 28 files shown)


apps/sap-commerce-cloud-with-air/lambda/src/middlewares/getAppInstallationParameters.ts
```diff
@@ -46,16 +49,13 @@
...
-            'Content-Type': 'application/json',
-          },
-        }
-      )
+      .get(`${host}/spaces/${spaceId}/environments/${environmentId}/app_installations/${appId}`, {
+        headers: {
+          Authorization: `Bearer ${appAccessToken}`,
+          'Content-Type': 'application/json',
+        },
+      })
```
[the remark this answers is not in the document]
rusticpenguin: ok

apps/ecommerce/lambda/src/middlewares/getAppInstallationParameters.ts
```diff
@@ -34,6 +36,7 @@ export const getAppInstallationParametersMiddleware: RequestHandler = async (
         appInstallationId: appId,
         spaceId,
         environmentId,
+        host,
```
andipaetzold, line 39: When passing the `host` to `getManagementToken`, you have to add the protocol. There is already a ticket to change that as its not intuitive.

```suggestion
        host: `https://${host}`,
```

apps/ecommerce/lambda/src/middlewares/getAppInstallationParameters.ts
```diff
@@ -43,33 +46,34 @@ export const getAppInstallationParametersMiddleware: RequestHandler = async (
...
+        );
+
+        const appInstallation = await cmaClient.appInstallation.get({
+          appDefinitionId: appId,
         });
-    }
+        Object.assign(installationParameters, appInstallation.parameters);
 
-    req.installationParameters = installationParameters;
+        req.installationParameters = installationParameters;
```
andipaetzold, line 69: This line must be run outside of the if-condition that starts in line 21. Otherwise we don't set `req.installationParameters` when `parametersHeader` is provided

apps/slack/lambda/lib/clients/cma.ts
```diff
@@ -21,12 +22,13 @@ export const makeSpaceEnvClient = async (
     appInstallationId: APP_ID as string,
     spaceId,
     environmentId,
+    host,
```
andipaetzold, line 25:
```suggestion
    host: `https://${host}`,
```

apps/slack/lambda/lib/helpers/getWorkspaceId.ts
```diff
@@ -4,13 +4,18 @@ import { NotFoundException } from '../errors';
 export async function getWorkspaceId(
   spaceId: string,
   environmentId: string,
+  host: string,
   workspaceIdFromParameters?: string
```
andipaetzold, line 8: nit: Shall we pass an object here. We now have 4 string arguments that are easy to mix up

andipaetzold (changes requested):

andipaetzold (approved):

PR #4670, opened 2023-08-31
update all urls to use the sdk hostnames

rusticpenguin:
## Purpose

We want to keep ourselves from hardcoding the hostnames for contentful. The SDK recently added a `hostnames` property. This updates all user-facing urls with that sdk property

## Approach

Updated the App SDK, sometimes had to update React for really old apps. We also needed to use legacy-peer-deps on gatsby because the gatsby sdk we're using has not been updated in forever and only supports react 16 (and we need to use react 17)


apps/gatsby/package.json
```diff
@@ -9,15 +9,16 @@
...
     "@contentful/forma-36-fcss": "0.3.5",
     "@contentful/forma-36-react-components": "3.100.7",
     "@contentful/forma-36-tokens": "0.11.2",
     "@gatsby-cloud-pkg/gatsby-cms-extension-base": "0.0.50",
     "emotion": "10.0.27",
     "prop-types": "15.8.1",
-    "react": "16.8.6",
-    "react-dom": "16.8.6",
+    "react": "^17.0.0",
+    "react-dom": "^17.0.0",
```
andipaetzold, line 21: We should pin the versions (remove the `^`)

apps/gatsby/package.json
```diff
@@ -9,15 +9,16 @@
     "cross-env": "7.0.3"
   },
   "dependencies": {
-    "@contentful/app-sdk": "4.9.0",
+    "@contentful/app-sdk": "4.23.0",
+    "@contentful/react-apps-toolkit": "1.2.16",
```
andipaetzold, line 13: If we haven't used react apps toolkit here before, we need to wrap the whole app in `SDKProvider`. How does this work without that?

rusticpenguin: Thank you for this catch!

andipaetzold:
Do we want to extract the package changes to a different PR?

andipaetzold (approved):

PR #4679, opened 2023-08-31
fix: remove curly braces on description fallback

anho:
## Purpose

Remove curly braces, were accidentally added.


rusticpenguin (approved):
Good catch!

PR #4680, opened 2023-08-31
[INTEG-1184]: Update character limit validation warning

whitelisab:
## Purpose

Updates the character limit validation for AI Content Generator to match new designs. Example:
Screenshot 2023-08-31 at 11 21 53 AM

## Approach

Specifically, this removes the error message, moves the error icon next to the character count, and makes both the character count and limits red when the count doesn't meet the limit. Updated tests for the component accordingly.

## Testing steps

Select an output field with character limits and then generate content that is out of bounds for that limit.


rusticpenguin (approved):
This looks so good

PR #4682, opened 2023-08-31
[INTEG-1187] Add cost and disclaimer to config and modal

whitelisab:
## Purpose

For AI Content Generator, we want to add cost and disclaimer information in the app to inform users that using the app utilizes ChatGPT tokens.

## Approach

**Config Page:**
Added new Cost and Disclaimer sections with appropriate links to OpenAI. I also fixed the spacing at the bottom of each section to be consistent.

Screenshot 2023-08-31 at 3 37 41 PM

**Modal:**
Added a message next to the generate button warning that ChatGPT uses tokens

Screenshot 2023-08-31 at 3 29 46 PM

## Testing steps

View the app config to see the new sections. For the modal, you'll see the help text next to the generate button until the button is enabled, then you will see the new warning message. You should see this for all actions and for both field and prompt selections.

### DIFF
--- a/apps/ai-content-generator/src/configs/features/featureConfig.tsx
+++ b/apps/ai-content-generator/src/configs/features/featureConfig.tsx
@@ -18,15 +18,6 @@ enum AIFeature {
 const FIELD_HELP_TEXT_BASE = 'Select a source field and output field';
 const FIELD_PLACEHOLDER = 'Your source and results will appear here...';
 const PROMPT_HELP_TEXT_BASE = 'Select an output field and';
-const TOKEN_WARNING = 'Generating content uses Chat GPT tokens.';
-const TOKEN_SUBSTRING = 'Chat GPT tokens.';
-const TOKEN_LINK = 'https://openai.com/pricing';
-
-const tokenWarning = {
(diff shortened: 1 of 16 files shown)


apps/ai-content-generator/src/configs/features/featureConfig.tsx
```diff
@@ -30,6 +39,7 @@ const featureConfig: FeatureConfig = {
       promptHelpText: `${PROMPT_HELP_TEXT_BASE} enter a prompt to generate a title`,
       promptPlaceholder:
         "Write a prompt, for example...\nAn article about quantum computing in simple terms\nA blog post about creative ideas for a 10 year old's birthday\nA tutorial about making an HTTP request in JavaScript",
+      tokenWarning,
```
rusticpenguin, line 42: Are there any places where the tokenWarning would/could be different? 

If not I think this could live in it's own config and get imported in the component that we're passing it into currently

whitelisab: Great suggestion, that will definitely streamline this! Made the change.

rusticpenguin (approved):
Looks great!

PR #4694, opened 2023-09-01
fixes GA4 tests and cma usage

rusticpenguin:
## Purpose

GA4 broke when I merged, this will fix it.

The tests were failing and the typing of the CMA changed when the SDK was updated


matthew-gordon (approved):
🖖 ✅

PR #4811, opened 2023-09-13
Feat/integ 1133 - Backwards compatibility with V1 config

rusticpenguin:
## Purpose
We were clearing the V1 configs on update to v2

## Approach

There were two approaches, I think this one will be more implicit and lead to less developer error long term. The original versioning config plan had too many potholes and added some in-repo knowledge that a new dev wouldn't necessarily know


whitelisab (approved):
Looks great! 🚀

PR #4822, opened 2023-09-14
Feat/integ 1120: Add segment to AICG

rusticpenguin:
## Purpose

We would like to get insights on a user's usage of our Ai content generator app

## Approach

We chose to initialize a segment client inside of a provider, that way we can reference it anywhere within the react app. For MVP we are cool with having the types declared manually, but would like to see us using typewriter in the near future.

All we need to do is call `trackEvent` anywhere in the app with the correct segment action enum and it will log the event.

This also respects a user's cookie-based consent

The following events are added with this PR:
`config_saved`
`sidebar_rendered`
`dialog_opened`


whitelisab (approved):
Looks very clean to me! I really like how the `trackEvent` is very simple to implement.

PR #4834, opened 2023-09-15
feat: add trailing slash configuration option for GA4 [INTEG-1287]

whitelisab:
## Purpose

We received feedback that some customers have all of their URLs automatically redirect to a trailing slash. Therefore, when they want to see data in Google Analytics, they would want to see the analytics for `https://mywebsite.com/products/cool-shirt/` instead of `https://mywebsite.com/products/cool-shirt`, as these are tracked as different pages in Google Analytics. Our current GA4 app does not allow for the user to easily configure the app to include a trailing slash.

## Approach

This PR adds a checkbox to the configuration page where users can select if they would like all their URLs to have a trailing slash appended to the end. If this checkbox is selected and the configuration page saved, then in the Sidebar app the user can see that a trailing slash is now added to the end of their page path. This is done via a new installation parameter, `useTrailingSlash`.

**Updated section on configuration screen:**
Screenshot 2023-09-15 at 11 56 55 AM

**Updated page path on sidebar app:**
Screenshot 2023-09-15 at 11 52 02 AM

**Updated page path filter in Google Analytics report:**
Screenshot 2023-09-15 at 11 52 17 AM

## Testing steps

The new checkbox will default to false, so for customers with the app installed, there should be no change.

Check the box on the config page and save the config. The navigate to an entry where the app is enabled in the sidebar. The page path should appear as: `/${urlPrefix}/${slug}/`. When you click on `Open in Google Analytics`, the filter in the report page should reflect this trailing slash as well.

## Breaking Changes

This change should not impact current app installations.

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
@@ -41,7 +41,7 @@ const AssignContentTypeSection = (props: Props) => {
     originalContentTypes,
   } = props;
 
-  const [useTrailingSlash, setUseTrailingSlash] = useState<boolean>(false);
+  const [forceTrailingSlash, setForceTrailingSlash] = useState<boolean>(false);
 
   // Content type state
   const [contentTypes, setContentTypes] = useState<ContentTypes>({} as ContentTypes);
@@ -62,10 +62,11 @@ const AssignContentTypeSection = (props: Props) => {
   const sdk = useSDK<AppExtensionSDK>();
 
   useEffect(() => {
-    if (parameters.useTrailingSlash) setUseTrailingSlash(parameters.useTrailingSlash);
+    setLoadingContentTypes(true);
+    if (parameters.forceTrailingSlash) setForceTrailingSlash(parameters.forceTrailingSlash);
     if (parameters.contentTypes) setContentTypes(parameters.contentTypes);
     setLoadingContentTypes(false);
-  }, [parameters.contentTypes, parameters.useTrailingSlash]);
+  }, [parameters.contentTypes, parameters.forceTrailingSlash]);
 
   useEffect(() => {
     setHasContentTypes(Object.keys(contentTypes).length ? true : false);
@@ -95,8 +96,8 @@ const AssignContentTypeSection = (props: Props) => {
   }, [sdk]);
 
   const trailingSlashHandler = () => {
-    setUseTrailingSlash(!useTrailingSlash);
-    mergeSdkParameters({ useTrailingSlash: !useTrailingSlash });
+    setForceTrailingSlash(!forceTrailingSlash);
+    mergeSdkParameters({ forceTrailingSlash: !forceTrailingSlash });
     onIsValidContentTypeAssignment(true);
   };
 
@@ -178,7 +179,7 @@ const AssignContentTypeSection = (props: Props) => {
         <Checkbox
           name="use-trailing-slash"
           id="use-trailing-slash"
-          isChecked={useTrailingSlash}
+          isChecked={forceTrailingSlash}
           onChange={trailingSlashHandler}>
           Use trailing slash for all page paths
         </Checkbox>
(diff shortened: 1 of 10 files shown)


apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
```diff
@@ -59,9 +62,10 @@ const AssignContentTypeSection = (props: Props) => {
   const sdk = useSDK<AppExtensionSDK>();
 
   useEffect(() => {
+    if (parameters.useTrailingSlash) setUseTrailingSlash(parameters.useTrailingSlash);
```
ryunsong-contentful, line 65: might be worth putting `setLoadingCntentTypes(true)` here as there are useEffects in place that may re-trigger this logic.

ryunsong-contentful (approved):
LGTM

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
```diff
@@ -7,7 +7,7 @@ import useGetFieldValue from '../useGetFieldValue';
 
 const SLUG_FIELD_INPUT_DELAY = 500;
 
-export const useSidebarSlug = (slugFieldInfo: ContentTypeValue) => {
+export const useSidebarSlug = (slugFieldInfo: ContentTypeValue, useTrailingSlash: boolean) => {
   const sdk = useSDK<SidebarExtensionSDK>();
```
rusticpenguin, line 11: Instead of passing down useTrailingSlash down to components/into here, what do you think about getting it from the sdk in this hook?

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
```diff
@@ -33,7 +33,9 @@ export const useSidebarSlug = (slugFieldInfo: ContentTypeValue) => {
     sdk.entry.onSysChanged((sys) => handlePublishedStatus(sys));
   }, [sdk.entry]);
 
-  const reportSlug = `/${pathJoin(urlPrefix || '', debouncedSlugFieldValue || '')}`;
+  const reportSlug = `/${pathJoin(urlPrefix || '', debouncedSlugFieldValue || '')}${
+    useTrailingSlash ? '/' : ''
+  }`;
```
rusticpenguin, line 38: Would this create a double / if the field value already ends in a /?

If so, you can take the regex from here:
https://github.com/contentful/marketplace-partner-apps/blob/main/apps/shopify/src/dataTransformer.js#L22

and do something like:
```js
const path = pathJoin(urlPrefix || '', debouncedSlugFieldValue || '';

const trailingSlash = /\/$/;
const reportSlug = useTrailingSlash ? path.replace(trailingSlash, '') + '/' : path;
```

apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentTypeSection.tsx
```diff
@@ -59,9 +62,10 @@ const AssignContentTypeSection = (props: Props) => {
...
     if (parameters.contentTypes) setContentTypes(parameters.contentTypes);
     setLoadingContentTypes(false);
-  }, [parameters.contentTypes]);
+  }, [parameters.contentTypes, parameters.useTrailingSlash]);
```
rusticpenguin, line 68: a nit: I feel useTrailingSlash in code looks more like a react hook. Maybe we could name it `forceTrailingSlash` or something like that

whitelisab: Good point, totally agree! Updated the name to `forceTrailingSlash`

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
whitelisab: Yes, good call!

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.tsx
whitelisab: Good callout! I confirmed that this line takes care of it already for us: https://github.com/contentful/apps/blob/master/apps/google-analytics-4/frontend/src/utils/pathJoin.ts#L3 and have verified that a slug name with a trailing slash will not have a second one added on.

PR #4866, opened 2023-09-19
[INTEG-1253] Changes select and fill naming to select and edit for naming consistency

ryunsong-contentful:
## Purpose
Naming consistency 

## Breaking Changes
Yes once this gets merged in, there is an experience packages PR that will need to be merged in and deployed in user interface as well so that it can find the correct app action id

User_interface and Experience Packages

## Deployment
Standard


lilbitner (approved):
This looks right to me! Assuming the new manifest will just overwrite the current `aaigSelectFill` app action?

ryunsong-contentful:
> This looks right to me! Assuming the new manifest will just overwrite the current `aaigSelectFill` app action?

I manually overwrote it just in case i believe

PR #4878, opened 2023-09-20
[INTEG-1295] - add getting started section for AIIG config page

matthew-gordon:
## Purpose

This adds a getting started section, as well as a clear CTA for creating a new asset for users from the config page

## Approach

Config page aligns with generated comps for getting started section. Aside form comps, the decision to add a CTA for users to create and open a new asset has been added to demystify asset creation location

## New Look

Screenshot 2023-09-21 at 7 14 59 AM

## Testing steps

- Install AI Image Generator to a space
- Instructions for getting started should match comps [here](https://www.figma.com/file/MyzNyHqDo1vjyAVGCMD852/AI-Image-Generator?type=design&node-id=64-40894&mode=design&t=mbiCAS64Tyf9C2Xr-0)
- `Try AI content generator` button should not be visible if app not installed
- `Try AI content generator` button should be visible when app is installed


apps/ai-image-generator/frontend/src/components/config/GettingStartedSection/GettingStartedSection.tsx
```diff
@@ -0,0 +1,89 @@
+import { useEffect, useState } from 'react';
+import { Card, Flex, HelpText, Text } from '@contentful/f36-components';
+import { ConfigAppSDK } from '@contentful/app-sdk';
+import { styles } from './GettingStartedSection.styles';
+
+import configPageCopies from 'constants/configPageCopies';
+
+import getstarted1 from '../../../assets/get-started-1.jpg';
```
ryunsong-contentful, line 8: nit: name these `getStartedImageGeneration` and `getStartedSelectAndEdit`

ryunsong-contentful (approved):
I think some of the spacing is off - will ask Emily, and also the Read the docs link is not yet implemented so we'll create tickets to track that work. Otherwise 🔥 

Also FYI it doesn't make sense to have the button above the Generate an Image section - it should be under.
Screenshot 2023-09-21 at 9 58 50 AM

ryunsong-contentful:
Interesting the header and sub heading are the same size but we can polish that up next week after user testing because I think there are several tweaks we should review as a group ✅

PR #4897, opened 2023-09-20
Fix pollution of package and package lock during sharp build

jsdalton:
## Purpose

We started seeing build failures during the `publish-package` step of the deploy-prod workflow. Example: https://app.circleci.com/pipelines/github/contentful/apps/24206/workflows/72689d12-eadd-4067-aaaa-81f2abe09c02/jobs/71745

Further investigation uncovered the problem was in the `build-sharp` script of the `ai-image-generator` app. This step (used to build an version of sharp that can run in an AWS lambda) was using `npm install`, which was polluting the git working space. This was causing a failure down the road when lerna tried to do it's git magic during `lerna version --conventional-commits --no-private --force-git-tag --create-release github --yes && lerna publish from-git --yes`

## Approach

* Use `npm ci` instead of `npm install`: https://docs.npmjs.com/cli/v7/commands/npm-ci . "It will never write to package.json or any of the package-locks: installs are essentially frozen."


ryunsong-contentful (approved):
nasty little bugger this

PR #4905, opened 2023-09-21
feat: add AICG sidebar error handling [INTEG-1182]

whitelisab:
## Purpose

We want to improve the error handling in the Sidebar for AI Content Generator.

## Approach

This PR adds several things:
- Checks to see if the API key is valid by hitting the OpenAI models endpoint
- Handles when model is undefined by defaulting to a model in the AI class
- Displays appropriate warning messages and disables the sidebar buttons when there is an OpenAI error

I added different warning components and created a hook for the logic to check the sidebar parameters.

**Example of a missing brand profile (this is a dismissible warning and buttons are still enabled):**
Screenshot 2023-09-21 at 4 01 56 PM

**Example of a missing or invalid API Key:**
Screenshot 2023-09-21 at 4 08 36 PM

**Example of Chat GPT not available :**
Screenshot 2023-09-21 at 4 51 57 PM

**Example of a different OpenAI API error, showing the error message from OpenAI:**
Screenshot 2023-09-21 at 4 07 56 PM

**Example of default error message (when there is no message from OpenAI):** 
Screenshot 2023-09-21 at 4 48 54 PM

## Testing steps

- If key, model, or profile are undefined, the Sidebar app should still load
- If model is undefined, the default one should be used
- If profile is missing, warning should display. User should be able to dismiss the warning and the buttons should still be enabled
- An invalid API key should display the proper warning with buttons disabled

## Breaking Changes

None

Common OpenAI errors: https://platform.openai.com/docs/guides/error-codes/api-errors

### DIFF
--- a/apps/ai-content-generator/src/components/app/sidebar/BrandProfileMissingWarning.spec.tsx
+++ b/apps/ai-content-generator/src/components/app/sidebar/BrandProfileMissingWarning.spec.tsx
@@ -1,0 +1,25 @@
+import { render } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { mockCma, MockSdk } from '../../../../test/mocks';
+import BrandProfileMissingWarning from './BrandProfileMissingWarning';
+
+const mockSdk = new MockSdk();
+const sdk = mockSdk.sdk;
+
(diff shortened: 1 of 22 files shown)


rusticpenguin (approved):
Looks good!!

PR #4920, opened 2023-09-22
[MONET-1473] feat: adds hosted delivery function example for PotterDB

thomasspiesser:
## Purpose

Ticket https://contentful.atlassian.net/browse/MONET-1473

This PR adds an example that includes a hosted delivery function. The functions forwards GraphQL queries to the [PotterDB](https://docs.potterdb.com/apis/graphql) so that we can query Contentfuls GraphQL API for data from the potter universe.

image

### DIFF
--- a/examples/hosted-delivery-function-potterdb/contentful-app-manifest.json
+++ b/examples/hosted-delivery-function-potterdb/contentful-app-manifest.json
@@ -1,0 +1,12 @@
+{
+  "deliveryFunctions": [
+    {
+      "id": "PotterDB",
+      "name": "Example Delivery Function querying Potter-DB GraphQL API",
+      "description": "This is an example including hosted delivery functions.",
+      "path": "delivery-functions/index.js",
+      "entryFile": "delivery-functions/index.ts",
+      "allowNetworks": ["api.potterdb.com"]
+    }
+  ]
+}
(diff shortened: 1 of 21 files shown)


examples/hosted-delivery-function-potterdb/contentful-app-manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+  "deliveryFunctions": [
+    {
+      "id": "PotterDB",
+      "name": "Example Delivery Function querying Potter-DB GraphQL API",
```
kdamball, line 5: nit: You could move this to description and just call it "PotterDB Delivery Function"

examples/hosted-delivery-function-potterdb/delivery-functions/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  })
+
+  return {
+    namespace: 'PotterDB',
+    fields,
+  }
+}
+
+const queryHandler: EventHandler<EventType.GRAPHQL_QUERY> = async (event, context) => {
+  const response = await fetch('https://api.potterdb.com/graphql', {
```
kdamball, line 21: I would add a comment here about `isIntrospectionQuery` and how you're forwarding both to the api. Introspection query is the first thing that breaks usually 😅

kdamball (approved):
Only minor comments to make life easier for someone new to this

examples/hosted-delivery-function-potterdb/delivery-functions/index.ts
thomasspiesser: Added a comment. Lmk if it addresses what you had in mind.

kdamball (approved):

PR #4921, opened 2023-09-22
feat: add character limit to brand fields and the prompt field in aicg

rusticpenguin:
## Purpose

ChatGPT 3 has a character limit of ~16k. Also the ai gets significantly worse when user adds too much data for it to consider

## Approach

Create a `Validator` type that takes an existing type and recursively adds `{ value: originalValue, isValid: bool }` to every field.

Then reorganized the code so the validation happens during the dispatch

## Testing steps

Verify:
- The config page still has original functionality
- You can't submit the form when exceeding the character count
- Prompt/From Source shows a character count when using gpt3 and not in gpt4

### DIFF
--- a/apps/ai-content-generator/src/components/config/config-section/ConfigSection.tsx
+++ b/apps/ai-content-generator/src/components/config/config-section/ConfigSection.tsx
@@ -10,27 +10,17 @@ interface Props {
   model: string;
   dispatch: Dispatch<ParameterReducer>;
   isApiKeyValid: boolean;
-  localApiKey: string;
-  onApiKeyChange: (key: string) => void;
-  validateApiKey: (key: string) => Promise<void>;
 }
 
 const ConfigSection = (props: Props) => {
-  const { apiKey, model, dispatch, isApiKeyValid, localApiKey, onApiKeyChange, validateApiKey } =
-    props;
+  const { apiKey, model, dispatch, isApiKeyValid } = props;
 
   return (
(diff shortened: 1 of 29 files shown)


apps/ai-content-generator/src/components/config/config-section/ConfigSection.tsx
```diff
@@ -10,27 +10,17 @@ interface Props {
...
       <Box>
         <Form>
-          <APIKey
-            apiKey={apiKey}
-            isInvalid={!isApiKeyValid}
-            localApiKey={localApiKey}
-            onApiKeyChange={onApiKeyChange}
-            validateApiKey={validateApiKey}
-          />
+          <APIKey apiKey={apiKey} isInvalid={!isApiKeyValid} dispatch={dispatch} />
```
matthew-gordon, line 23: `!isApiKeyValid` will negate the boolean value returned by `isApiKeyValid` right? `isApiKeyValid` returns true, then `!isApiKeyValid` will be false, and if `isApiKeyValid` returns false, then `!isApiKeyValid` will be true, 🤔 does that work as intended?

rusticpenguin: I'm gonna rename this for readability since we're always flipping it

whitelisab (approved):
This looks great and works well on the config page and in the prompt section of the dialog!

PR #4922, opened 2023-09-22
Support EU region in upload URLs

jsdalton:
## Purpose

Right now, we upload all images to the US region (hard coded). We also report back in our `url` field that the image was uploaded to the US bucket.

Going forward, images might be uploaded to the EU region, depending on how the cma client was instantiated. We need to ensure that the URL points to the correct bucket, no matter where the image is uploaded.

## Approach

* We have no way of knowing which host or upload host the cmaClient was initialized with. However we can use the `cmaHost` value to deduce which region was used and build a correct url computation.
* We don't have support for flinkly or other environments. This current implementation is pretty brittle as a result. Better to build the upload url at the API level then hardcode this stuff in apps though.

## Testing steps

* Go to the EU region, to a space where this app is installed
* Do a select and fill
* See that the result images display correctly from the eu bucket

### DIFF
--- a/apps/ai-image-generator/backend/src/actions/aiig-select-edit.ts
+++ b/apps/ai-image-generator/backend/src/actions/aiig-select-edit.ts
@@ -25,7 +25,7 @@ export const handler = async (
 ): Promise<AppActionCallResponse<ImageEditResult>> => {
   const {
     cma,
-    appActionCallContext: { appInstallationId, spaceId, cmaHost },
+    appActionCallContext: { appInstallationId, spaceId, uploadHost },
   } = context;
 
   let images: ImageWithUpload[];
@@ -72,7 +72,7 @@ export const handler = async (
       imagesWithStreams: processedImages,
       cmaClient: cma,
       spaceId,
-      cmaHost,
+      uploadHost,
     });
   } catch (e) {
     if (!(e instanceof Error)) {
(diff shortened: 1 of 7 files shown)


apps/ai-image-generator/backend/src/actions/aiig-select-edit.ts
```diff
@@ -25,7 +25,7 @@ export const handler = async (
 ): Promise<AppActionCallResponse<ImageEditResult>> => {
   const {
     cma,
-    appActionCallContext: { appInstallationId, spaceId },
+    appActionCallContext: { appInstallationId, spaceId, cmaHost },
```
ghepting, line 28: add uploadHost?

apps/ai-image-generator/backend/src/helpers/upload-images.ts
```diff
@@ -35,20 +38,23 @@ export class UploadImages {
...
 
-  // TODO Handle eu!
   private urlFromUpload(upload: UploadProps): string {
     const uploadId = upload.sys.id;
     const uploadPath = `${this.spaceId}!upload!${uploadId}`;
-    return [UPLOAD_DOMAIN.us.toString(), uploadPath].join('/');
+    const uploadDomain = UPLOAD_DOMAIN[this.cmaHost];
+    if (!uploadDomain)
+      throw new Error(`Invalid cmaHost '${this.cmaHost}' -- could not find upload bucket`);
+    return [uploadDomain, uploadPath].join('/');
```
ghepting, line 47: this can all go away if you use uploadHost now

ryunsong-contentful (approved):

PR #4936, opened 2023-09-22
feat: support short text list fields in GA4 [INTEG-1344]

whitelisab:
## Purpose

We received customer feedback that they want to use short text list fields as the "slug" field when configuring the GA4 app. Currently, the app only supports short text fields.

## Approach

In order to support short text list fields, I made the following changes:

1. Config page: Updated the help text displayed when selecting a slug field and updated the filtering so that both short text and short text list fields are displayed as options.
2. Sidebar app: Updated how we are creating the page path to handle an array of strings. The elements of the array will be joined by a forward slash

**Updated config page:**
Screenshot 2023-09-22 at 2 35 39 PM

**Example field on an entry and what appears in the sidebar:**
Screenshot 2023-09-22 at 2 37 21 PM

## Testing steps

1. Create a short text list field on a content type
2. Select that field as the slug field on the app config
4. Navigate to an entry and enter values in the field. See the page path update in the sidebar app

## Breaking Changes

This change is not breaking, it will just display additional fields to select on the app config page.

### DIFF
--- a/apps/google-analytics-4/frontend/src/helpers/contentTypeHelpers/contentTypeHelpers.ts
+++ b/apps/google-analytics-4/frontend/src/helpers/contentTypeHelpers/contentTypeHelpers.ts
@@ -1,7 +1,11 @@
-import { EditorInterface } from '@contentful/app-sdk';
+import { ContentTypeField, EditorInterface } from '@contentful/app-sdk';
 import { ContentTypeProps } from 'contentful-management';
 import { AllContentTypes, EditorInterfaceAssignment } from '../../types';
 import sortBy from 'lodash/sortBy';
+
+interface FieldItem {
+  type: string;
+}
 
 export const generateEditorInterfaceAssignments = (
   currentEditorInterface: Partial<EditorInterface>,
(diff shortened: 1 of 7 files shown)


apps/google-analytics-4/frontend/src/helpers/contentTypeHelpers/contentTypeHelpers.ts
```diff
@@ -37,17 +41,19 @@ export const generateEditorInterfaceAssignments = (
   return { ...newAssignments, ...assignmentsToAdd };
 };
 
+// only include short text and short text list fields in the slug field dropdown
+const isCompatibleFieldType = (field: ContentTypeField): boolean => {
+  const isArray = field.type === 'Array';
+  return field.type === 'Symbol' || (isArray && (field.items as FieldItem).type === 'Symbol');
```
matthew-gordon, line 47: Does this need to be `FieldItem[]` referencing the list?

ryunsong-contentful: What should the behavior be if one of the items isn't the type symbol? Should we first filter it out?

ryunsong-contentful (approved):
One small question/comment

matthew-gordon (approved):

apps/google-analytics-4/frontend/src/helpers/contentTypeHelpers/contentTypeHelpers.ts
whitelisab: If you look at the `Array Fields` section of [this page](https://www.contentful.com/developers/docs/concepts/data-model/) in the docs, you'll see that the items property contains one type per field. We handle the inclusion of short text list fields this way in the ecommerce base app: https://github.com/contentful/apps/blob/master/packages/ecommerce-app-base/src/AppConfig/fields.ts#L39-L42

Open to other suggestions @matthew-gordon and @ryunsong-contentful if there's a better way to approach this!

PR #4954, opened 2023-09-25
fix: event types [MONET-1476]

kdamball:
## Purpose

Hardcode types into every generated template code.

## Approach

Until we update node-apps-toolkit to use esm, we will hardcode the types.


examples/hosted-delivery-function-templates/typescript/example.ts
```diff
@@ -1,11 +1,12 @@
...
 
-enum EventType {
-  GRAPHQL_FIELD_MAPPING = 'graphql.field.mapping',
-  GRAPHQL_QUERY = 'graphql.query',
-}
+const GRAPHQL_FIELD_MAPPING_EVENT = 'graphql.field.mapping';
+const GRAPHQL_QUERY_EVENT = 'graphql.query';
 
-const fieldMappingHandler: EventHandler<EventType.GRAPHQL_FIELD_MAPPING> = (event, context) => {
+const fieldMappingHandler: EventHandler<typeof GRAPHQL_FIELD_MAPPING_EVENT> = (event, context) => {
```
anho, line 9:
```suggestion
const fieldMappingHandler: EventHandler<'graphql.field.mapping'> = (event, context) => {
```

examples/hosted-delivery-function-templates/typescript/example.ts
```diff
@@ -22,18 +23,18 @@ const fieldMappingHandler: EventHandler<EventType.GRAPHQL_FIELD_MAPPING> = (even
   };
 };
 
-const queryHandler: EventHandler<EventType.GRAPHQL_QUERY> = (event, context) => {
+const queryHandler: EventHandler<typeof GRAPHQL_QUERY_EVENT> = (event, context) => {
```
anho, line 26:
```suggestion
const queryHandler: EventHandler<'graphql.query'> = (event, context) => {
```

anho:
Let's also fix the files in [hosted-delivery-function-potterdb](https://github.com/contentful/apps/tree/master/examples/hosted-delivery-function-potterdb).

examples/hosted-delivery-function-potterdb/delivery-functions/index.ts
```diff
@@ -37,11 +43,11 @@ const queryHandler: EventHandler<EventType.GRAPHQL_QUERY> = async (event, contex
   return response.json()
 }
 
-export const handler: EventHandler = (event, context) => {
-  if (event.type === EventType.GRAPHQL_FIELD_MAPPING) {
+export const handler: EventHandler<EventType> = (event, context) => {
```
BasKiers, line 46: The generic argument should default to all types automatcially
```suggestion
export const handler: EventHandler = (event, context) => {
```

BasKiers (approved):

examples/hosted-delivery-function-templates/typescript/example.ts
```diff
@@ -22,18 +20,18 @@ const fieldMappingHandler: EventHandler<EventType.GRAPHQL_FIELD_MAPPING> = (even
...
+const queryHandler: EventHandler<'graphql.query'> = (event, context) => {
   return {
     data: {},
     errors: [],
   };
 };
 
 export const handler: EventHandler = (event, context) => {
-  if (event.type === EventType.GRAPHQL_FIELD_MAPPING) {
+  if (event.type === GRAPHQL_FIELD_MAPPING_EVENT) {
```
anho, line 31: You could replace this (and other occasions) with `graphql.field.mapping` as well.

kdamball: We have it as a `const` at the top - why against that?

anho: I don't see the point of introducing the constant in the first place. The event types are under our control and properly exposed via node-apps-toolkit, so there is no additional type safety to gain here. But feel free to move on.

examples/typescript/tsconfig.json
```diff
@@ -4,5 +4,6 @@
     "target": "es5",
     "lib": ["dom", "dom.iterable", "esnext"]
   },
-  "include": ["src"]
+  "include": ["src"],
+  "exclude": ["actions", "delivery-functions"]
```
BasKiers, line 8: Which tsconfig will be used during development? Will we add this to the template?

kdamball: `build-delivery-function.js` has a config :)

kdamball: Happy to see include an extra one if we find settings we see the need

ghepting (approved):

PR #4955, opened 2023-09-25
feat: add error handling to modal

rusticpenguin:
## Purpose

There was no error handling outside of some try/catches in the original modal.

## Approach

Have stateful errors in the modal, as well as launch toasts as needed

### DIFF
--- a/apps/ai-content-generator/src/components/app/dialog/common-generator/output/output-text-panels/OutputTextPanels.tsx
+++ b/apps/ai-content-generator/src/components/app/dialog/common-generator/output/output-text-panels/OutputTextPanels.tsx
@@ -1,6 +1,4 @@
 import { useContext } from 'react';
-import { useSDK } from '@contentful/react-apps-toolkit';
-import { DialogAppSDK } from '@contentful/app-sdk';
 import { GeneratorContext } from '@providers/generatorProvider';
 import useEntryAndContentType from '@hooks/dialog/useEntryAndContentType';
 import useAI, { GenerateMessage } from '@hooks/dialog/useAI';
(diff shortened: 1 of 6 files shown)


apps/ai-content-generator/src/components/app/dialog/common-generator/output/output-text-panels/OutputTextPanels.tsx
```diff
@@ -38,7 +38,10 @@ const OutputTextPanels = (props: Props) => {
     const successfullyUpdated = await updateEntry(outputFieldId, outputFieldLocale, ai.output);
 
     if (successfullyUpdated) {
+      sdk.notifier.success('Content applied successfully.');
```
whitelisab, line 41: These notification toasts had been previously added here: https://github.com/contentful/apps/blob/master/apps/ai-content-generator/src/hooks/dialog/useEntryAndContentType.ts#L66-L71, so we should just have them in one place.

whitelisab (commented):
Just one comment about the toasts, otherwise looks good!

apps/ai-content-generator/src/components/app/dialog/common-generator/output/output-text-panels/OutputTextPanels.tsx
rusticpenguin: Ahh that's a good callout. I must have missed that! 

What are your thoughts on making this updateEntry return void and we just call it here?

whitelisab: Yeah I'm fine with moving them here in the `OutputTextPanels` component, and then cleaning up what had been added in the `updateEntry` function!

PR #4987, opened 2023-09-27
Capture more error details in logs

jsdalton:
## Purpose

We still don't have enough detail to figure out why some users are seeing a 403 in the workspaces Typeform endpoint

## Approach

* Console log more details


whitelisab (approved):
Looking good! 👍

PR #4988, opened 2023-09-27
fix: handling for no model saved on config

whitelisab:
## Purpose

AICG config page V2 was not rendering if you had not saved a model in V1.


rusticpenguin (approved):
Wooo!

PR #5001, opened 2023-09-28
Add last-modified header to static frontend

jsdalton:
## Purpose

We have not been providing a `Last-Modified` header to our express static middleware used to host the Typeform frontend; as a result, the value of `Last-Modified` sent by AWS gateway has been 1980.

This has caused aggressive browser local caching, which often leads to "broken" deploys since the browser doesn't try to fetch the most recent value.

## Approach

* Capture the "current time" at deploy; pass through to as env var to serverless

### DIFF
--- a/apps/typeform/lambda/app.js
+++ b/apps/typeform/lambda/app.js
@@ -15,6 +15,18 @@ const app = express();
 
 const FRONTEND = path.dirname(require.resolve('@contentful/typeform-frontend'));
 
+const computeLastModifiedTime = () => {
+  const deployTime = process.env.DEPLOY_TIME_UNIX;
+  if (typeof deployTime === 'undefined') {
+    throw new Error('Missing DEPLOY_TIME_UNIX env var');
+  }
+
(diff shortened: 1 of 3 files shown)


apps/typeform/lambda/app.js
```diff
@@ -15,6 +15,18 @@ const app = express();
 
 const FRONTEND = path.dirname(require.resolve('@contentful/typeform-frontend'));
 
+const computeLastModifiedTime = () => {
+  const deployTime = process.env.DEPLOY_TIME_UNIX;
+  if (typeof deployTime === 'undefined') {
+    throw new Error('Missing DEPLOY_TIME_UNIX env var');
```
ghepting, line 21: you want to break things or just do "10 minutes ago" as a fallback?

ghepting (approved):

apps/typeform/lambda/app.js
jsdalton: this will fail hard and loud when the server starts, which is always my personal prefernce for env vars 🤷

PR #5005, opened 2023-09-28
fix serverless env var

jsdalton:
## Purpose

Forgot to rename the env var in serverless.yaml

### DIFF
--- a/apps/typeform/lambda/serverless.yml
+++ b/apps/typeform/lambda/serverless.yml
@@ -33,7 +33,7 @@ functions:
     environment:
       CLIENT_ID: ${env:TYPEFORM_CLIENT_ID}
       CLIENT_SECRET: ${env:TYPEFORM_CLIENT_SECRET}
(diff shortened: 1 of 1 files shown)


ghepting (approved):

apps/typeform/lambda/serverless.yml
```diff
@@ -33,7 +33,7 @@ functions:
     environment:
       CLIENT_ID: ${env:TYPEFORM_CLIENT_ID}
       CLIENT_SECRET: ${env:TYPEFORM_CLIENT_SECRET}
-      DEPLOY_TIME: ${env:DEPLOY_TIME}
+      DEPLOY_TIME_UNIX: ${env:DEPLOY_TIME_UNIX}
```
jsdalton, line 36: Needs to match https://github.com/contentful/apps/blob/62a6405bd0c00c8d98d43ddb293619043d6e6d64/apps/typeform/lambda/package.json#L12

ghepting: that might help :)

PR #5012, opened 2023-09-29
feat: improved handling for rate limits [INTEG-1381]

whitelisab:
## Purpose

We wanted to improve the feedback to AICG users about rate limits for Chat GPT. This PR adds more helpful messaging to the config page and the modal.

## Approach

Added new copy to the config page:
Screenshot 2023-09-29 at 7 50 01 AM

For the modal, refactored the API call to the completions endpoint so that we could display different messages based on the response status. If a 429 response is returned, we have a custom message, otherwise, the default will show.

Rate limit error message:
Screenshot 2023-09-28 at 9 09 40 PM

Default modal error message:
Screenshot 2023-09-28 at 9 10 43 PM

## Testing steps

- Ensure that all actions still work as expected for the happy path
- Trigger an error (e.g. make an invalid API key) and the error message should appear in the modal after clicking "Generate"

## Breaking Changes

None

Links go to this page: https://platform.openai.com/docs/guides/rate-limits/overview
OpenAI error codes:https://platform.openai.com/docs/guides/error-codes

### DIFF
--- a/apps/ai-content-generator/src/utils/aiApi/index.ts
+++ b/apps/ai-content-generator/src/utils/aiApi/index.ts
@@ -41,19 +41,19 @@ class AI {
       stream: true,
     });
 
-    const stream = (
-      await fetch(this.baseUrl, {
-        method: 'POST',
-        headers,
-        body,
-      })
-    ).body?.getReader();
+    const streamResponse = await fetch(this.baseUrl, {
+      method: 'POST',
+      headers,
+      body,
+    });
 
-    if (!stream) {
-      throw new Error('Unable to create stream');
+    if (streamResponse.status === 200) {
+      const stream = streamResponse.body?.getReader();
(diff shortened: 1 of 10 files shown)


apps/ai-content-generator/src/utils/aiApi/index.ts
```diff
@@ -41,19 +41,19 @@ class AI {
...
+      body,
+    });
 
-    return stream;
+    if (streamResponse.status === 200) {
+      const stream = streamResponse.body?.getReader();
+      return stream;
+    } else {
+      const streamJson = await streamResponse.json();
+      validateResponseStatus(streamResponse, streamJson);
```
rusticpenguin, line 55: I like thsi

rusticpenguin (approved):
I like this! Very clean

PR #5020, opened 2023-10-02
feat: update handling for non-image assets [INTEG-1404]

whitelisab:
## Purpose

This PR updates the `dam-app-base` package to better handle assets that are not images in the Field Editor location. For example, in the Bynder app, when you add an asset that does not have a `webimage` property, it is saved to the entry but in the Field Editor the asset displays as `Asset not available`.

Original state: First item is an asset with the thumbnail that's hard to discern what the image is. Third image is a file type that doesn't have a `webimage` property, so the default `Asset not available` is displayed.
Screenshot 2023-10-02 at 11 02 58 AM

## Approach

The `dam-app-base` accepts a `makeThumbnail` function that returns a tuple containing the asset url and a text representation of that asset. When there is no url, the asset was being rendered as `Asset not available` in the UI, even though the asset is saved in the JSON object in the field. To make this less confusing for users, we are adding a different default state for when there is no url for an asset -- display a file icon and the text representation of that asset. Only if an asset has no url or text representation will the `Asset not available` state appear. 

In addition, it is hard to tell the name of the asset from the Field Editor since all you can see is the thumbnail image. This PR adds a title attribute with the text representation so that the user can see the asset name on hover. This is also helpful for an image with a long name, you can then see the full name on hover.

Updated Field Editor component showing various states:
- File name displays on hover
- When there is no url, a default file icon and the file name now displays
- A long file name will be truncated, but the full name is displayed on hover
- The `Asset not available` will only display if the asset doesn't have a url and doesn't have a text representation
- If the asset has a broken url, the text representation will display

Screenshot 2023-10-02 at 11 03 13 AM

## Testing steps

Run the Bynder app locally and use `npm link` to test the changes to the `dam-app-base`. I also tested these changes with other apps that utilize this package, including Dropbox and Brandfolder.

Note: The Bynder app currently uses the asset Id as the text representation. Another PR will update that so that the asset name appears, which will make it easier for the user.

## Breaking Changes

This is not a breaking change as it does not affect how the asset is stored in Contentful, it just changes the display in the Field Editor for when there is no url or text representation for an asset.


matthew-gordon (approved):
Honestly this is great! Looks good and is a massive improvement 👍 🙌 🙇 ✅

PR #5054, opened 2023-10-04
fix: set localstorage in correct place

jsdalton:
## Purpose

In browsers where cross-site tracking and/or third party storage access partitioning features were enabled, token and expire time values received from Jira during the Oauth flow were not being saved to local storage in a way such that they were accessible from the main Jira Contentful app.

Instead, what was observed was that those values were not present in local storage from the main app at all.

Basically same as https://github.com/contentful/apps/pull/5000

## Approach

* Send data via postMessage


lilbitner (approved):
makes sense to me 👍

PR #5055, opened 2023-10-04
Add SDK provider

jsdalton:
## Purpose

There's a bug that's masking error messaging:

```
react-dom.production.min.js:216 Error: SDKContext not found. Make sure this hook is used inside the SDKProvider
    at t.useSDK (useSDK.js:14:15)
    at y (ContentTypeStep.tsx:28:15)
    at ca (react-dom.production.min.js:157:137)
    at Yc (react-dom.production.min.js:267:460)
    at Cs (react-dom.production.min.js:250:347)
    at Ts (react-dom.production.min.js:250:278)
    at Is (react-dom.production.min.js:250:138)
    at bs (react-dom.production.min.js:243:163)
    at react-dom.production.min.js:123:115
    at t.unstable_runWithPriority (scheduler.production.min.js:18:343)
 ```

## Approach

* Add the missing `SDKProvider`

### DIFF
--- a/apps/jira/jira-app/src/index.tsx
+++ b/apps/jira/jira-app/src/index.tsx
@@ -9,6 +9,7 @@ import '@contentful/forma-36-react-components/dist/styles.css';
 import './index.scss';
 import JiraClient from './jiraClient';
 import { InstallationParameters } from './interfaces';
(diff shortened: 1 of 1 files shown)


lilbitner (approved):
LGTM 👍

PR #5068, opened 2023-10-05
Add component from new component library to AWS Amplify

lilbitner:
## Purpose

The purpose of this PR is to introduce the component library import into a low-risk (not high profile like the AI apps right now) app, to ensure that the production build works! I will then have a follow-up PR to add more components to the AIIG app to establish an even more foundational POC for this component library. 

## Approach

I added the LocalHostWarning component from the component library to replace the current version in the AWS Amplify app.

I was able to test that the component rendered locally within the app, the build worked correctly, and the tests ran ✅ 

[
Integration-component-library](https://github.com/contentful/integration-component-library)


matthew-gordon (approved):
🤩 🚀 ITS HERE!

PR #5102, opened 2023-10-10
feat: adds hosted delivery function example for PotterDB REST API

anho:
## Purpose

Another example for delivery functions using the PotterDB REST API (despite it having a GraphQL counterpart) to showcase capabilities.


examples/hosted-delivery-function-potterdb-rest-api/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+This project was bootstrapped with [Create Contentful App](https://github.com/contentful/create-contentful-app).
+
+## How to use
+
+Execute create-contentful-app with npm, npx or yarn to bootstrap the example:
+
+```bash
+# npx
+npx create-contentful-app --typescript
```
BasKiers, line 14: Do we want to make these example specific, e.g.
```suggestion
npx create-contentful-app --example hosted-delivery-function-potterdb-rest-api
```

examples/hosted-delivery-function-potterdb-rest-api/delivery-functions/index.ts
```diff
@@ -0,0 +1,94 @@
+import type { DeliveryFunctionEventHandler as EventHandler } from '@contentful/node-apps-toolkit'
+import { createSchema, createYoga } from 'graphql-yoga'
+import { typeDefs } from './schema'
+import { NodeRequest } from '@whatwg-node/server/typings/utils'
+import { GraphQLError } from 'graphql'
+
+const schema = createSchema({
```
BasKiers, line 7: Personally, to more easily understand what is going on for GraphQL novices, I would put the schema definition and resolvers in the same file (e.g. move the `createSchema` call to `schema.ts`

BasKiers (approved):
Like commented I would move the resolvers to be located next to the schema. Apart from that the example would benefit from having inline comments describing what is going on but I'm fine with doing that on a follow up task.

Implementation looks good, pretty simple to get working!

examples/hosted-delivery-function-potterdb-rest-api/README.md
anho: Then we (or @contentful/team-developer-experience) should probably look a bit more holistically at those files.

PR #5114, opened 2023-10-11
chore: upgrade Lerna to v6 [INTEG-1422]

whitelisab:
## Purpose

We are currently using Lerna v5 in this repo, upgrading to v6 will allow us to take advantage of some useful improvements. Additionally, it should resolve the alerts noted here: https://github.com/contentful/apps/pull/4071

## Approach

Upgrading from v5 to v6 is non-breaking. I upgraded lerna and then ran `npx lerna add-caching` to generate a `nx.json` file to enable caching (https://lerna.js.org/docs/features/cache-tasks).

Upgrading to v7 is a breaking change that will require us to replace our usage of `lerna bootstrap` (https://lerna.js.org/docs/legacy-package-management). This can be addressed in a follow up task. 

## Testing steps

I ran several lerna commands locally (lint, test, bootstrap, build) and verified that it is still working as anticipated.

## Breaking Changes

Upgrading to v6 should not be a breaking change according to this: https://dev.to/nx/lerna-reborn-whats-new-in-v6-245j

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,0 +1,10 @@
+{
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "nx/tasks-runners/default",
+      "options": {
+        "cacheableOperations": ["build", "test", "lint"]
+      }
+    }
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "name": "@contentful/apps",
       "dependencies": {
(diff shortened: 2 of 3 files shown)


ryunsong-contentful (approved):

lilbitner (approved):
Awesome!! This caching feature will be awesome 🙌 A good thing to keep in mind if we add e2e tests are the limitations there with caching

PR #5126, opened 2023-10-13
chore: upgrade prettier and fix CI prettier check [INTEG-787]

whitelisab:
## Purpose

We had turned off our prettier check in CI several months ago, and we couldn't upgrade to `prettier@3` until we upgraded to `lerna@6` ([see this PR](https://github.com/contentful/apps/pull/4510)). 

This PR addresses several things:
- Upgrades Prettier
- Re-enables our Prettier CI check
- Removes additional Prettier configs so that the whole repo uses the same one
- Runs Prettier and makes updates to unformatted files

## Approach

I looked into using `nx affected` to run the prettier check, but that would require us to have prettier scripts for each app. I thought that having one prettier config for the whole repo would be cleaner. This is why I removed any other prettier configs besides the one at the root.

`prettier@3` updates the default value for `trailingComma` to `all` ([see here](https://prettier.io/blog/2023/07/05/3.0.0.html#breaking-changes)). Our CI script was running `prettier@3` even though we hadn't updated the dependency, which is why we were seeing all of those trailing comma issues in CI and not locally. Adding `"trailingComma": "es5"` to our `.prettierrc` retains the formatting we were used to, and that was previously added in [this PR](https://github.com/contentful/apps/pull/4496).

## Testing steps

I tested this by committing and pushing an unformatted file and saw that the CI check failed. We also still have the "kill switch" for the check if it becomes problematic.

## Breaking Changes

None, changes are just formatting updates.


lilbitner (approved):
Looks great! So awesome to have this fixed 🙌

PR #5127, opened 2023-10-13
fix: downgrade prettier

whitelisab:
## Purpose

We thought `prettier@3` would work with `lerna@6`, but we need to upgrade to `lerna@7` ([see here](https://github.com/lerna/lerna/releases/tag/v7.1.2))


matthew-gordon (approved):
👍

PR #5255, opened 2023-10-23
add command to run aicg ci tests

rusticpenguin:
## Purpose

Tests won't run for aicg

## Approach

make them run with the test:ci command


whitelisab (approved):
Nice! 🎊

PR #5260, opened 2023-10-24
feat: add initial Microsoft Teams app [INTEG-1468]

whitelisab:
## Purpose

This PR creates an initial app for Microsoft Teams.

## Approach

Utilized Create Contentful App using the Vite configuration. Also made some additional updates:

- Update test scripts
- Add linting
- Add support for absolute paths
- Add `.nvmrc`

## Testing steps

Ran app locally and was able to verify that the app rendered in our Microsoft Teams development environment. Also ran `test`, `lint`, and `build` scripts.

## Deployment

Deploy scripts will be added in a separate PR.

### DIFF
--- a/apps/microsoft-teams/frontend/.nvmrc
+++ b/apps/microsoft-teams/frontend/.nvmrc
@@ -1,0 +1,1 @@
+lts/hydrogen
--- a/apps/microsoft-teams/frontend/.nvmrc
+++ b/apps/microsoft-teams/frontend/.nvmrc
@@ -1,1 +1,1 @@
-lts/hydrogen
+v18.13.0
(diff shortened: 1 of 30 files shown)


apps/microsoft-teams/frontend/.nvmrc
```diff
@@ -0,0 +1 @@
+lts/hydrogen
```
jsdalton, line 1: should we specify the explicit node dependency or the LTS? i don't have a strong opinion about it, but i notice in most places at contentful we do the version number

jsdalton (approved):
Looks good to me 👍

apps/microsoft-teams/frontend/.nvmrc
whitelisab: Sounds good, updated to `v18.13.0`, which is the version used by several of our other apps.

PR #5280, opened 2023-10-26
Feature/add ms teams backend

jsdalton:
## Purpose

The bot needs a backend context to send notifications and handle authentication tasks from

## Approach

* Add a super basic backend to support hosted app actions
* Add a simple test action and script

TODO (future PRs)

* [ ] Add a top level build that packages up the frontend and backend actions in a single artifact to deploy

### DIFF
--- a/apps/microsoft-teams/backend/package.json
+++ b/apps/microsoft-teams/backend/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "@contentful/microsoft-teams-app",
+  "version": "1.0.0",
+  "private": true,
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "NODE_ENV=production tsc",
+    "test": "NODE_ENV=test TS_NODE_TRANSPILE_ONLY=1 mocha --exit",
+    "test:ci": "NODE_ENV=test TS_NODE_TRANSPILE_ONLY=1 CI=true mocha --exit",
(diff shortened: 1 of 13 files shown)


matthew-gordon (approved):
LGTM ✅

apps/microsoft-teams/backend/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@contentful/microsoft-teams-app",
```
whitelisab, line 2: small nit: maybe call this `@contentful/microsoft-teams-backend` to mirror what we have in the frontend. Then the  top level `package.json` could be named `@contentful/microsoft-teams-app` or `@contentful/microsoft-teams`

apps/microsoft-teams/backend/src/actions/send-test-notification.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+      parameters: {},
+    },
+  ];
+
+  beforeEach(() => {
+    cmaRequestStub = sinon.stub();
+    context = makeMockAppActionCallContext(cmaClientMockResponses, cmaRequestStub);
+  });
+
+  it('returns the images result', async () => {
```
whitelisab, line 32: This is the assertion nit that @matthew-gordon mentioned that should be updated eventually.

whitelisab (approved):
Just left some very small comments that can be addressed later.

PR #5306, opened 2023-10-27
fix all app builds

lilbitner:
## Purpose

Instead of creating a PR for each app build that fails because of the latest Node upgrade, I have tested the builds of all apps locally and updated the necessary scripts in this PR. 🤞 

See this [PR](https://github.com/contentful/apps/pull/5304) and this [PR](https://github.com/contentful/apps/pull/5305) for more context. We fixed Smartling and Typeform with individual PRs, this PR is for the rest of the apps!


jsdalton (approved):
💁

PR #5320, opened 2023-10-30
fix pipeline - netlify, brandfolder, dropbox builds

lilbitner:
## Purpose

In order to fix the current pipeline build failures, I have tested the builds (after running `npm i `😅 ) of all apps locally and updated the rest of the scripts in this PR - they should resemble the scripts that exist in most other apps now. 🤞 

See this [PR](https://github.com/contentful/apps/pull/5304) and this [PR](https://github.com/contentful/apps/pull/5305) for more context around this update - the origin of these changes are from a Node update last week.

### DIFF
--- a/apps/brandfolder/package.json
+++ b/apps/brandfolder/package.json
@@ -13,8 +13,8 @@
     "react-scripts": "4.0.3"
   },
   "scripts": {
-    "start": "cross-env BROWSER=none react-scripts start",
-    "build": "react-scripts build",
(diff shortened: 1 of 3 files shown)


ryunsong-contentful:
Could we also try the solution suggested in https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported

Where we instead run `npm audit fix --force` and get rid of the ssl legacy option?

PR #5321, opened 2023-10-30
Create a combined build for backend and frontend

jsdalton:
## Purpose

Since HAAs need to be deployed in the same artifact as the frontend build, our "true" build needs to be set up a top level which knows how to combine the two.

## Approach

* Like we did for AIIG, create a separate backend build and frontend build, and munge them together at the end

### Coming in follow up PR

* Test script combined with sandbox account to allow for local verification of HAA


matthew-gordon (approved):
✅ really beginning to enjoy this pattern personally

PR #5325, opened 2023-10-30
chore: add testable app action script

jsdalton:
## Purpose

We don't have an easy way to make sure our app actions work before deploying,.

## Approach

* Add a script to call app actions
* Create an app definition to act as a "sandbox" where we can easily test app action calls


whitelisab (approved):
👍

PR #5373, opened 2023-11-06
chore: rename backend to app-actions

jsdalton:
## Purpose

With the introduction of our separate standalone bot backend service it's become confusing to have a folder called "backend".

## Approach

* Rename top level folder to `app-actions` and adjust accordingly


whitelisab (approved):
The new naming makes sense to me!

PR #5416, opened 2023-11-08
Add endpoint to handle incoming activity from MS Teams

jsdalton:
## Purpose

We need an endpoint to handle incoming messages from the MS Teams framework.

## Approach

* Create a service to handle the various calls we'll be making to the teamsfx and botbuilder SDKs
* Use `ConversationBot.requestHandler` to handle the actual messages: https://learn.microsoft.com/en-us/javascript/api/@microsoft/teamsfx/conversationbot?view=msteams-client-js-latest#@microsoft-teamsfx-conversationbot-requesthandler This one is not deprecated and also used in a few examples.

### DIFF
--- a/apps/microsoft-teams/msteams-bot-service/package.json
+++ b/apps/microsoft-teams/msteams-bot-service/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "build": "NODE_ENV=production rimraf ./build && tsc",
     "lint": "eslint --max-warnings=0 .",
+    "start": "nodemon",
     "test": "NODE_ENV=test TS_NODE_TRANSPILE_ONLY=1 mocha",
     "test:ci": "NODE_ENV=test TS_NODE_TRANSPILE_ONLY=1 CI=true mocha",
     "test:debug": "NODE_ENV=test TS_NODE_TRANSPILE_ONLY=1 mocha -- --inspect --inspect-brk",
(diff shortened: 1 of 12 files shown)


apps/microsoft-teams/msteams-bot-service/package.json
```diff
@@ -19,17 +20,25 @@
...
     "@typescript-eslint/eslint-plugin": "^6.10.0",
     "@typescript-eslint/parser": "^6.10.0",
+    "botframework-connector": "^4.21.0",
     "chai": "^4.3.10",
     "chai-http": "^4.4.0",
     "eslint": "^8.53.0",
     "mocha": "^10.2.0",
+    "nodemon": "^3.0.1",
     "rimraf": "^5.0.5",
+    "sinon": "^17.0.1",
```
rusticpenguin, line 34: big fan of sinon!

apps/microsoft-teams/msteams-bot-service/src/services/ms-teams-conversation-service.ts
```diff
@@ -0,0 +1,36 @@
...
+    const conversationBot = new ConversationBot({
+      adapterConfig: {
+        MicrosoftAppId: botId,
+        MicrosoftAppPassword: botPassword,
+        MicrosoftAppType: 'MultiTenant',
+      },
+      notification: {
+        enabled: true,
+      },
+    });
```
rusticpenguin, line 19: Nit: What are your thoughts about moving this config into a separate file?

rusticpenguin (approved):
Looks great! Glad to see the stubbed service too! One nit, but doesn't block the merge

apps/microsoft-teams/msteams-bot-service/src/services/ms-teams-conversation-service.ts
jsdalton: Yeah, I played with exactly this idea and think it makes a lot of sense. We could eg initialize these directly in config and then pass them through vs. only initializing the env vars. I suggest we leave open for a future iteration, wdyt?

PR #5417, opened 2023-11-08
refactor: implement reducer to manage installation params [INTEG-1524]

whitelisab:
## Purpose

Updating the MS Teams app config page to use a reducer to manage the installation parameters and added a hook to initialize the parameters and dispatch them to the reducer. 

## Approach

Followed much of the pattern we established in the AICG app. I also implemented the ability to remove a notification, so now you can add and remove empty notifications on the config page.

## Testing steps

Run MS Teams (development) app locally. You should be able to edit the Tenant Id and add/remove notifications and save the app config page.

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
@@ -19,6 +19,11 @@ const NotificationsSection = (props: Props) => {
     dispatch({ type: actions.ADD_NOTIFICATION });
   };
 
+  const deleteNotification = (index: number) => {
+    notifications.splice(index, 1);
+    dispatch({ type: actions.UPDATE_NOTIFICATIONS, payload: notifications });
+  };
+
   return (
     <Box className={styles.box}>
       <Subheading>{notificationsSection.title}</Subheading>
@@ -27,7 +32,13 @@ const NotificationsSection = (props: Props) => {
         handleClick={createNewNotification}
       />
       {notifications.map((notification, index) => {
-        return <NotificationEditMode key={`notification-${index}`} index={index} />;
+        return (
+          <NotificationEditMode
+            key={`notification-${index}`}
+            index={index}
+            handleDelete={deleteNotification}
+          />
+        );
       })}
     </Box>
   );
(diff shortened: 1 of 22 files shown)


apps/microsoft-teams/frontend/src/components/config/parameterReducer.ts
```diff
@@ -0,0 +1,73 @@
...
+        notifications: action.payload,
+      };
+    case ADD_NOTIFICATION: {
+      return {
+        ...state,
+        notifications: [defaultNotification, ...state.notifications],
+      };
+    }
+    case APPLY_CONTENTFUL_PARAMETERS: {
+      const parameter = action.payload as AppInstallationParameters;
```
jsdalton, line 61: such a minor minor nit but singular `parameter` in the body of this action confused me. probably we mean `paramaters` plural?

jsdalton: Also i'm surrpised the type assertion is needed here, doesn't TS already know your payload is `AppInstallationParameters`?

jsdalton (approved):
Makes a lot of sense to me, especially as the complexity ramps up with managing notifications inside the parameters. I had a few small questions but overall looks brilliant! 🥇

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
```diff
@@ -1,17 +1,28 @@
...
 const NotificationsSection = (props: Props) => {
-  const { notifications, createNewNotification } = props;
+  const { notifications, dispatch } = props;
+
+  const createNewNotification = () => {
+    dispatch({ type: actions.ADD_NOTIFICATION });
+  };
+
+  const deleteNotification = (index: number) => {
+    notifications.splice(index, 1);
```
rusticpenguin, line 23: I feel we may want to either filter on the array or create a new one here. 
Eg.
```js 
const notificationsPayload = [...notifications];
notifcationsPayload.splice(index, 1);
```

Main reason being it's best practice in react to not mutate state directly. While this should hopefully be fine since we're dispatching right after, it could cause some unintended side-effects

rusticpenguin (changes requested):
Left one comment, otherwise looks great to me!

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
whitelisab: Great catch, updated!

apps/microsoft-teams/frontend/src/components/config/parameterReducer.ts
whitelisab: Yes, great catches! Made the changes.

matthew-gordon (approved):
💪 🌮 ✨ Looks great! Thank you for addressing the requested changes ✅

PR #5427, opened 2023-11-09
chore: error handling of incoming message handling

jsdalton:
## Purpose

When there's an error in the request handler we need to ensure that it's both logged and handled at the server level.

## Approach

* Add simple try/catch clause -- move to middleware later
* Capture message activity as mocks (may remove the unused ones later, but useful at this stage)

### DIFF
--- a/apps/microsoft-teams/msteams-bot-service/test/mocks.ts
+++ b/apps/microsoft-teams/msteams-bot-service/test/mocks.ts
@@ -24,3 +24,186 @@ export const makeMockConversationBot = (): sinon.SinonStubbedInstance<Conversati
   });
   return stubbedBot;
 };
+
+export const installationUpdateAdd = {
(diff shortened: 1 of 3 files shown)


apps/microsoft-teams/msteams-bot-service/test/mocks.ts
```diff
@@ -24,3 +24,186 @@ export const makeMockConversationBot = (): sinon.SinonStubbedInstance<Conversati
   });
   return stubbedBot;
 };
+
+export const installationUpdateAdd = {
```
whitelisab, line 28: Just more of a curiosity question: do these responses look any different for a team's general channel vs. a "named" channel? I had seem some differences when I was playing around with the sample bot.

whitelisab (approved):
Looks good! ⭐

PR #5428, opened 2023-11-09
feat: configure (local) dynamodb and docker

jsdalton:
## Purpose

For the bots service, we will implement the storage backend using DynamoDB. As a precursor, it would be great to configure docker for the service and ensure that the development server and tests all run on docker, including the local dynamodb.

## Approach

* Dockerize the service itself
* Add a docker compose file to manage the service, the dynamodb as well as a dynamodb admin and initializer (which creates the table for us if it doesn't already exit)
* Add a makefile to cover the main use cases (so we can do the same things we did before when working locally with the backend, but just via docker)
* Convert the ci tests to run inside docker. This allows us to use an actual dynamodb instance in tests vs mocking.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,7 @@ jobs:
       - node/install
       - vault/get-secrets:
           template-preset: 'aws-push-artifacts'
+      - setup_remote_docker
       - run:
           name: Setup circle and tools
           command: |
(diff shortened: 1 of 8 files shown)


matthew-gordon (approved):
LGTM ✨ , I remember pushback against the dynamodb admin as a callout but I found it useful 💪 ✅

PR #5438, opened 2023-11-10
feat: create channel selection modal for MS Teams [INTEG-1525]

whitelisab:
## Purpose

This PR creates the channel selection modal for the MS Teams app config page. 

## Approach

User can click "Add channel" to open the selection modal, where they see a list of channels/teams. Once they select an option, the selected channel is applied to the notification. The channel can be changed by clicking the edit icon. The user then clicks the "save" button in the notification to apply their changes to that notification. By saving the app config page, the channel is persisted in app installation parameters.

Note that currently the channels and teams are hardcoded sample data; we will update this later to fetch the actual list of channels/team where the app is installed for the tenant id.

Screenshot 2023-11-10 at 11 44 29 AM

## Testing steps

Run the app locally and navigate to an app config page.


apps/microsoft-teams/frontend/src/components/config/ChannelSelection/ChannelSelection.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const { unmount } = render(
+      <ChannelSelection notification={defaultNotification} handleNotificationEdit={vi.fn()} />
+    );
+
+    expect(screen.getByText(channelSelection.title)).toBeTruthy();
+    expect(screen.getByText(channelSelection.addButton)).toBeTruthy();
+    unmount();
+  });
+  it('mounts and renders an input when a channel is selected', () => {
+    render(
```
jsdalton, line 18: Just purely a curiousity / for my own learning question -- why do we need to unmount in the first test but not this one?

apps/microsoft-teams/frontend/src/components/config/ChannelSelection/ChannelSelection.tsx
```diff
@@ -0,0 +1,73 @@
...
+        handleNotificationEdit={handleNotificationEdit}
+        savedChannelId={notification.channelId}
+      />
+    ));
+  };
+
+  // TODO: update this when we start fetching channel installations
+  const getChannelName = (channelId: string) => {
+    const channel = mockChannels.find((channel) => channelId === channel.id);
+    const displayName = channel ? `${channel.name}, ${channel.teamName}` : '';
```
jsdalton, line 36: hm i wonder if silently swallowing the case where we don't find a missing channel is a good idea 🤔 

i don't need know that we need to error per se but what if we put `'unknown'` or `'channel not found'` here in the meantime?

jsdalton (approved):
Looks good like one or two minor non blocking comments / questions. Feel free to address on subsequent PRs

apps/microsoft-teams/frontend/src/components/config/ChannelSelection/ChannelSelection.spec.tsx
whitelisab: We learned with using vitest in AICG that if we have multiple tests in the same describe block, we need to unmount the component. So it was really just to make the second test in the suite run properly. I can address adding unmount in the second test in a follow up PR.

apps/microsoft-teams/frontend/src/components/config/ChannelSelection/ChannelSelection.tsx
whitelisab: Agreed, definitely something we will want to update once we are actually fetching the channel installations. This was feeling very temporary to me, but I can add in some copy to display in the meantime in a follow up!

PR #5468, opened 2023-11-15
feat: add notification view mode [INTEG-1552]

whitelisab:
## Purpose

This PR creates the "view mode" for configured notifications in the MS Teams app config page.

Screenshot 2023-11-15 at 8 34 46 AM

## Approach

A couple of additions with this PR:
- Only one notification can be in edit mode at a time. When a notification is in edit mode, other notifications have the `Edit` button disabled
- When a notification is in edit mode, save will only be enabled when all the fields have been filled out
- A cancel button has been added to edit mode

## Testing steps

Run the app locally in an MS Teams development app.


matthew-gordon (approved):
✨ 🌮 🥇 This all looks good to me ✅

PR #5469, opened 2023-11-15
chore: [ZEND-4323] - Update token setters for auth persistence

matthew-gordon:
## Purpose

This fixes a bug with perpetual auth call out for smartling sidebar app, users were required to re-auth everytime they navigated away from content or configuration. 

We were overwriting tokens in local storage with OR `||` logic for component state, this implicitly sets tokens to null and forces a refresh, or the per usual initial auth flow.

### DIFF
--- a/apps/smartling/frontend/src/Sidebar.tsx
+++ b/apps/smartling/frontend/src/Sidebar.tsx
@@ -23,8 +23,8 @@ interface Props {
 
 interface State {
   smartlingEntry: SmartlingContentfulEntry | null;
-  token: string;
-  refreshToken: string;
+  token: string | null;
+  refreshToken: string | null;
   showAllSubs: boolean;
   generalError: boolean;
 }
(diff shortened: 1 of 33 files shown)


apps/smartling/frontend/src/Sidebar.tsx
```diff
@@ -204,7 +207,12 @@ export default class Sidebar extends React.Component<Props, State> {
     const { sdk, projectId } = this.props;
     const { token } = this.state;
 
-    const res = await smartlingClient.getLinkedJobs(token, sdk.ids.space, projectId, sdk.ids.entry);
+    const res = await smartlingClient.getLinkedJobs(
```
rusticpenguin, line 210: Not sure which is the better solution there. But what do you think of:
```suggestion
    if (token === null) {
      this.runAuthFlow(true);
      return;
   }
 
    const res = await smartlingClient.getLinkedJobs(
```
This avoids sending a request if the token is null'd, and avoids the need for type casting token to string

ryunsong-contentful (approved):
If I understand this correctly, we need this to be null for the auth flow to not

mgoudy91 (approved):
👍 on the examples changes

PR #5476, opened 2023-11-15
Update null check and persist to localstorage properly

matthew-gordon:
## Purpose

This fixes a bug with perpetual auth call out for smartling sidebar app, users were required to re-auth everytime they navigated away from content or configuration.

We were overwriting tokens in local storage with OR || logic for component state, this implicitly sets tokens to null and forces a refresh, or the per usual initial auth flow.

### DIFF
--- a/apps/smartling/frontend/src/Sidebar.tsx
+++ b/apps/smartling/frontend/src/Sidebar.tsx
@@ -23,8 +23,8 @@ interface Props {
 
 interface State {
   smartlingEntry: SmartlingContentfulEntry | null;
-  token: string;
-  refreshToken: string;
+  token: string | null;
+  refreshToken: string | null;
   showAllSubs: boolean;
   generalError: boolean;
 }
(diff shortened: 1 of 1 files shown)


apps/smartling/frontend/src/Sidebar.tsx
```diff
@@ -204,6 +207,11 @@ export default class Sidebar extends React.Component<Props, State> {
     const { sdk, projectId } = this.props;
     const { token } = this.state;
 
+    if (!token) {
+      this.runAuthFlow(true);
+      return;
```
ryunsong-contentful, line 212: nit: do we want anything helpful to say/throw?

ryunsong-contentful (approved):

apps/smartling/frontend/src/Sidebar.tsx
matthew-gordon: No I think this is OK because it will kick back into the runAuthFlow, grab a token and come back through normally.. if it makes it passed here there is a logic gate on line 217 that fails through the same way

```typescript
if (res.code === 'AUTHENTICATION_ERROR') {
  this.runAuthFlow(true);
}
```

PR #5542, opened 2023-11-22
fix: bug [MONET-1607]

cemreyuksel:
We realized that EAP note was never able to get closed. This PR fixes the issue.  Once note is dismissed, it's not showed again. 

Related ticket: https://contentful.atlassian.net/browse/MONET-1607

### DIFF
--- a/packages/ecommerce-app-base/src/AppConfig/AppConfig.tsx
+++ b/packages/ecommerce-app-base/src/AppConfig/AppConfig.tsx
@@ -145,14 +145,15 @@ export default class AppConfig extends React.Component<Props, State> {
         parameters: toInputParameters(this.props.parameterDefinitions, parameters),
         fieldSkuTypes: (parameters as { skuTypes?: FieldsSkuTypes })?.skuTypes ?? {},
         appReady: true,
-        hideOrchestrationEapNote: false,
+        hideOrchestrationEapNote: parameters?.hideOrchestrationEapNote || false,
       },
       () => app.setReady()
     );
   };
 
   onAppConfigure = () => {
-    const { parameters, contentTypes, selectedFields, fieldSkuTypes } = this.state;
+    const { parameters, contentTypes, selectedFields, fieldSkuTypes, hideOrchestrationEapNote } =
+      this.state;
     const error = this.props.validateParameters(parameters);
 
     if (error) {
@@ -167,7 +168,7 @@ export default class AppConfig extends React.Component<Props, State> {
     }
 
     return {
-      parameters: updatedParameters,
+      parameters: { ...updatedParameters, hideOrchestrationEapNote },
       targetState: selectedFieldsToTargetState(contentTypes, selectedFields),
     };
   };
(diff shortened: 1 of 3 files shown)


packages/ecommerce-app-base/src/AppConfig/OrchestrationEapNote/OrchestrationEapNote.tsx
```diff
@@ -13,19 +13,18 @@ const styles = {
...
 
-export const OrchestrationEapNote = ({ hideOrchestrationEapNote, name }: Props) => {
-  const [, setHideOrchestrationEapNote] = useState(false);
+export const OrchestrationEapNote = ({ setHideOrchestrationEapNote, name }: Props) => {
   const title = `The ${name} app supports External references`;
+
+  const handleClose = () => {
+    setHideOrchestrationEapNote(true);
+  };
+
```
RichardM99, line 25: Why can't this component handle the state locally?

cemreyuksel: Because otherwise it does never get closed, the app state needs to be aware of it. (It was previously like that we realized that there is a bug. 🐞 )

packages/ecommerce-app-base/src/AppConfig/AppConfig.tsx
```diff
@@ -258,9 +259,9 @@ export default class AppConfig extends React.Component<Props, State> {
             </Form>
           </>
         )}
-        {isInOrchestrationEAP && (
+        {isInOrchestrationEAP && !hideOrchestrationEapNote && (
           <OrchestrationEapNote
-            hideOrchestrationEapNote={hideOrchestrationEapNote}
+            setHideOrchestrationEapNote={this.setHideOrchestrationEapNote.bind(this)}
```
RichardM99, line 264:
```suggestion
            onClose={this.setHideOrchestrationEapNote(true)}
```

And then just call it directly in the component or? i.e. 

` <Note className={styles.eapNote} withCloseButton={true} onClose={onClose} title={title}>`

cemreyuksel: Let me try but not sure this will work. 🤔

packages/ecommerce-app-base/src/AppConfig/OrchestrationEapNote/OrchestrationEapNote.tsx
```diff
@@ -13,19 +13,14 @@ const styles = {
...
 }
 
-export const OrchestrationEapNote = ({ hideOrchestrationEapNote, name }: Props) => {
-  const [, setHideOrchestrationEapNote] = useState(false);
+export const OrchestrationEapNote = ({ name, onClose }: Props) => {
   const title = `The ${name} app supports External references`;
+
   return (
-    <Note
-      style={{ display: hideOrchestrationEapNote ? 'none' : 'grid' }}
```
RichardM99, line 24: Why is this no longer needed?

cemreyuksel: Because we are checking this [here](https://github.com/contentful/apps/pull/5542/files#diff-debf4298077092d6fcc4bebd9daa39f85e1c9c77cc5de1b5acf8e5bcd37a468cR262)

packages/ecommerce-app-base/src/AppConfig/OrchestrationEapNote/OrchestrationEapNote.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import { MouseEventHandler } from 'react';
```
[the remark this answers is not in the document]
cemreyuksel: you're right, it's not used. 🤓

RichardM99 (approved):

PR #5559, opened 2023-11-27
feat: add delete confirmation modal [INTEG-1449]

whitelisab:
## Purpose

This PR adds the confirmation modal when a user tries to delete a notification.

## Approach

This is the confirmation modal that appears when a user clicks Delete for a notification.
Screenshot 2023-11-27 at 9 31 13 AM

## Testing steps

Delete a notification in the MS Teams (development) environment.


jsdalton (approved):
👍

PR #5578, opened 2023-11-29
feat: add empty state for channel selection and teams app help text [INTEG-1555]

whitelisab:
## Purpose

This PR adds an empty state for the channel selection modal and help text under the Tenant Id field to prompt users to install the Contentful MS Teams app.

## Approach

Added our `integration-component-library` as a dependency to utilize the `HyperLink` component. For now the link to Teams just brings you to Teams, but eventually we can update that to go directly to our app.

Empty State
Screenshot 2023-11-28 at 4 02 15 PM

Help Text
Screenshot 2023-11-29 at 8 53 17 AM

## Testing steps

Open the MS Teams development app. Change the mock channels so that it is an empty array in order to see the empty state.


apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.spec.tsx
```diff
@@ -1,13 +1,22 @@
...
 import { describe, expect, it, vi } from 'vitest';
 import { render, screen } from '@testing-library/react';
+import { accessSection } from '@constants/configCopy';
 
 const tenantIdValue = 'abc-123';
 
 describe('AccessSection component', () => {
   it('mounts with tenantId provided', () => {
-    render(<AccessSection dispatch={vi.fn()} tenantId={tenantIdValue} />);
+    const { unmount } = render(<AccessSection dispatch={vi.fn()} tenantId={tenantIdValue} />);
```
matthew-gordon, line 10: Is this for clean up (using the de-structred unmount) for memory leaks, or do we want to assert that these have been unmounted? If thats the case we can do a `.toBeNull()` check on the elements for example:

```typescript
const element= screen.getByText(accessSection.title);
unmount();
expect(element).toBeNull();
```

totally feel free to ignore this as well :D

matthew-gordon (approved):
LGTM, just one clarifier but not blocking

apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.spec.tsx
whitelisab: Good callout, this is for cleanup. We learned from AICG that with using `vitest` if you have multiple tests in the same `describe` block, you need to unmount the component between tests. If I remove the unmount, then the tests fail. So I don't think I necessarily need to assert that they have been unmounted, unless you think it would be helpful!

PR #5628, opened 2023-12-01
feat: add empty state for content type selection [INTEG-1579]

whitelisab:
## Purpose

This PR adds the empty state in the content type selection modal for MS Teams. This empty state will be visible when the app is installed in an environment where there are no content types.

Screenshot 2023-11-30 at 9 05 11 PM

## Approach

I made a few additional updates when adding this feature:
- I added a hook `useGetContentTypeConfigLink` to get the link to navigate to the content type config page within the web app
- I had to bump the `app-sdk` package in order to get `hostname`  from the sdk
- In order to reduce prop drilling the content types and the config link, I created context for the content type and the config link
- The Modal component cannot directly call the hooks to get the content types and config link, nor can it consume the context directly since it doesn't have the sdk context since it's not within the SDKProvider. So I did have to pass the content types and config link to the Modal component via props
- I created a helper to test the components that are consuming the content type context

## Testing steps

Set the content types to an empty array in our MS Teams environment to see the empty state.

### DIFF
--- a/apps/microsoft-teams/frontend/package-lock.json
+++ b/apps/microsoft-teams/frontend/package-lock.json
@@ -8,7 +8,7 @@
       "name": "@contentful/microsoft-teams-frontend",
       "version": "0.1.0",
       "dependencies": {
-        "@contentful/app-sdk": "4.22.0",
+        "@contentful/app-sdk": "^4.23.1",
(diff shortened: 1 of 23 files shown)


matthew-gordon (approved):
💪 Awesome LGTM

PR #5650, opened 2023-12-04
fix: data residency for netlify

rusticpenguin:
## Purpose

There was a missed api url in netlify for the data residency update.

## Approach

Use the host property in the appContextDetails to populate the url


jsdalton (approved):
👍

PR #5651, opened 2023-12-04
fix: lock down sharp to v0.32.6 due to installation errors in AWS lambda

jsdalton:
## Purpose

* Sharp introduced a new installation packaging set up where they distribute prebuilt binaries. Details here https://github.com/lovell/sharp/issues/3750. This was released in v0.33.0 which we upgraded to (via dependabot) a few days ago.
* The change broke the image generator with the following errors:

```
Error: Could not load the \"sharp\" module using the linux-x64 runtime
Possible solutions:
- Add platform-specific dependencies:
    npm install --os=linux --cpu=x64 sharp
  or
    npm install --force @img/sharp-linux-x64
- Consult the installation documentation: https://sharp.pixelplumbing.com/install
    at Object.<anonymous> (/var/task/7dFRjNacIIBFc6CVoVP822/node_modules/sharp/lib/sharp.js:85:9)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at Object.<anonymous> (/var/task/7dFRjNacIIBFc6CVoVP822/node_modules/sharp/lib/constructor.js:10:1)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)"
```

## Approach

* Downgrade and pin to 0.32.6 until we can get the installation working
* Local testing was not successful thus far

## Testing steps

* Tested old version locally in sandbox and it worked

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,6 +39,42 @@ updates:
         - ">=3"
 
 - package-ecosystem: npm
+  directory: "/apps/ai-image-generator"
+  schedule:
+    interval: daily
+    time: "03:00"
+    timezone: UTC
+  open-pull-requests-limit: 15
+  commit-message:
+    prefix: "fix"
+    prefix-development: "chore"
+    include: "scope"
+  ignore:
+    - dependency-name: "*"
+      update-types: ["version-update:semver-major"]
+    - dependency-name: "sharp"
+      versions:
+        - ">0.32.6"
+
+- package-ecosystem: npm
+  directory: "/apps/ai-image-generator/backend"
+  schedule:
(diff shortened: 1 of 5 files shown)


matthew-gordon (approved):
LGTM

PR #5674, opened 2023-12-05
chore: support sharp 0.33

jsdalton:
## Purpose

* We had to pin sharp to version 0.32.6 due to breaking changes with sharp's new installation process
* We want / need to unpin to continue to support new upstream sharp updates

## Approach

* Set package to 0.33 and remove dependabot exclusions
* Install the linux-x64 packages supplied by sharp during the build step
* Remove excess darwin* packages from node modules since they contain huge binaries that break our upload (our bundle size becomes too big)
* Make sure we copy over our new @img packages into the build artifact

## Testing steps

* Uploaded and tested this action on sandbox and confirmed it works ✨

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,38 +39,6 @@ updates:
         - ">=3"
 
 - package-ecosystem: npm
-  directory: "/apps/ai-image-generator"
-  schedule:
-    interval: daily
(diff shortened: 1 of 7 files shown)


whitelisab (approved):
Makes sense to me 👍

PR #5679, opened 2023-12-05
fix: cleaner sharp install to reduce folder size

jsdalton:
## Purpose

* Previous attempt to fix sharp (#5674) almost worked, but the build artifact was too big: https://app.circleci.com/pipelines/github/contentful/apps/26276/workflows/24c38da4-5537-40a8-bb96-78a70f62d160/jobs/81197?invite=true#step-117-5288_19

## Approach

* Instead of removing unneeded files, copy over only what's needed. This is a better approach and also has the advantage of minimzing the build size


whitelisab (approved):
👍

PR #5682, opened 2023-12-06
feat: support custom update and additional details display [INTEG-1611]

whitelisab:
## Purpose

This PR updates the `dam-app-base` with two new optional functions. These changes are being made to support feedback from customers to improve the Bynder app, which utilizes this package.

## Approach

I added two new optional functions to be passed into the `dam-app-base` setup integration. Since these functions are optional, adding these will not impact other apps that utilize this package.

1. `customUpdateStateValue`: changes the default save behavior. The current behavior is that newly selected assets are appended to the end of the selected assets. If you want to instead resave all the selected assets each time, this new function will enable you to do so.
2. `getAdditionalData`: passes additional data to display in the asset card. When this function is passed in, then you see a "more details" section on the asset card.

## Testing steps

Use `npm link` to run this package locally with Bynder to see the changes. I also tested with another app, Brandfolder, to ensure that these changes do not impact existing behavior.

## Breaking Changes

None, these additional functions are optional.

### DIFF
--- a/packages/dam-app-base/src/Editor/SortableComponent.tsx
+++ b/packages/dam-app-base/src/Editor/SortableComponent.tsx
@@ -1,11 +1,18 @@
-import { Card, IconButton } from '@contentful/f36-components';
+import { Card, Collapse, Flex, IconButton, TextLink } from '@contentful/f36-components';
 import { CloseIcon } from '@contentful/f36-icons';
 import tokens from '@contentful/f36-tokens';
 import { css } from 'emotion';
 import arrayMove from 'array-move';
 import * as React from 'react';
 import { SortableContainer, SortableElement, SortableHandle } from 'react-sortable-hoc';
-import { Asset, Config, DeleteFn, ThumbnailFn } from '../interfaces';
+import {
+  Asset,
+  Config,
+  DeleteFn,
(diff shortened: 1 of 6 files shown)


packages/dam-app-base/src/Editor/SortableComponent.tsx
```diff
@@ -128,7 +204,14 @@ const SortableList = SortableContainer<SortableContainerProps>((props: SortableC
   const { list } = props.resources.reduce(
     (acc, resource, index) => {
       const [url, alt] = props.makeThumbnail(resource, props.config);
-      const item = { url, alt, key: `url-unknown-${index}` };
+
+      let additionalData = null;
+
+      if (props.getAdditionalData) {
+        additionalData = props.getAdditionalData(resource);
+      }
```
rusticpenguin, line 212: I feel this can be simplified to:
```suggestion
      const additionalData = props.getAdditionalData?.(resource) || null;
```

rusticpenguin (changes requested):
Just one nit on some code simplification. Otherwise this looks great!

rusticpenguin (approved):
Looks great!!

PR #5684, opened 2023-12-06
fix: disable create notification button when in edit mode [INTEG-1610]

whitelisab:
## Purpose

The `Create notification` button should be disabled when you are viewing a notification in edit mode.

## Approach

See new disabled button state:
Screenshot 2023-12-06 at 1 51 07 PM

## Testing steps

Navigate to MS Teams (development) app.


matthew-gordon (approved):
Bueno! LGTM ✨ 🌮

PR #5685, opened 2023-12-06
feat: [INTEG-1479] - Trigger HAA channels list

matthew-gordon:
## Purpose

Adds hosted app action call to render list of installed ms teams channels, currently a static list is returned from the HAA 

ticket: https://contentful.atlassian.net/browse/INTEG-1479

## Approach

Screenshot 2023-12-06 at 2 01 40 PM

## Testing steps

HAA (hosted app actions) currently are only testable in a deployed state. Navigate to [ms-teams sandbox](https://app.contentful.com/spaces/a944qvnjghi2/apps/1qjevmbpYM840jFygQREbD):

Click on `Create Notification > Add Channel` to view list of retrieved/available channels


apps/microsoft-teams/frontend/src/customTypes/configPage.ts
```diff
@@ -18,6 +18,7 @@ export type SelectedEvents = {
 
 export interface TeamsChannel {
   id: string;
+  tenantId: string;
```
whitelisab, line 21: Do we need to have the `tenantId` as part of the channels response? Since the response will always be scoped to the `tenantId` that's passed into the app action call as a parameter.

apps/microsoft-teams/frontend/src/hooks/useGetTeamsChannels.ts
```diff
@@ -0,0 +1,44 @@
+import { useEffect, useCallback, useState } from 'react';
+import { useSDK } from '@contentful/react-apps-toolkit';
+import { ConfigAppSDK } from '@contentful/app-sdk';
+import { AppInstallationParameters, TeamsChannel } from '@customTypes/configPage';
+
+const useGetTeamsChannels = () => {
+  const [channels, setChannels] = useState<TeamsChannel[]>([]);
+  const sdk = useSDK<ConfigAppSDK>();
```
whitelisab, line 8: Ah, I was wondering why we were using the SDK in this hook and it's because we need to pass the `tenantId`. We actually might want to use the value from the parameters state in the `ConfigPage` component in case the user makes any updates to the tenant id.

apps/microsoft-teams/frontend/src/hooks/useGetTeamsChannels.ts
whitelisab, line 1: I really like having this logic in a hook!

apps/microsoft-teams/frontend/test/mocks/mockChannels.json
```diff
@@ -3,24 +3,28 @@
     "id": "19:e3a386bd1e0f4e00a286b4e86b0cfbe9@thread.tacv2",
     "name": "General",
     "teamId": "ed57f808-c14f-4a53-bf53-e36de0783385",
+    "tenantId": "666e56a6-1f2a-47c7-b88c-1ed9e1bb8668",
```
whitelisab, line 6: Same comment here about the `tenantId`, that wasn't in the original mock data and I'm curious why we added it. Doesn't really hurt anything to keep it 😄

apps/microsoft-teams/frontend/test/mocks/mockChannels.ts
whitelisab, line 1: Do we need both this file and the `json` file with the mock data?

whitelisab (commented):
Looks really great! I had some minor comments. I think this is a great start, and we can decide in a follow up task if we want to call the API just from the modal and save the channel ID names in installation params and/or update to use the Dialog location so that we can get over the issue of calling the SDK from the modal.

matthew-gordon:
> Looks really great! I had some minor comments. I think this is a great start, and we can decide in a follow up task if we want to call the API just from the modal and save the channel ID names in installation params and/or update to use the Dialog location so that we can get over the issue of calling the SDK from the modal.

Initially I was calling the api from the modal, but that was firing the call on every modal launch which would initially show the empty state component(s) and need to close and re-open the modal to get the list displaying as the HAA response shows `undefined` on the first go around which wasn't/isn't ideal. I am drilling the channels down from the notification "container" to account for this method which also needs the channels:

https://github.com/contentful/apps/blob/fc35c9c0a742492879600ee10fb687439f718277/apps/microsoft-teams/frontend/src/components/config/ChannelSelection/ChannelSelection.tsx#L45

`the channel ID names` could you expand more on this? We can connect later for clarity 🙏

apps/microsoft-teams/frontend/test/mocks/mockChannels.json
matthew-gordon: same answer from above, happy to remove it if this secondary index isn't being stored and we want to filter it on the call.

apps/microsoft-teams/frontend/src/customTypes/configPage.ts
matthew-gordon: We have decided to keep `tenantId` after talking with @jsdalton

apps/microsoft-teams/frontend/src/components/config/ChannelSelection/ChannelSelection.tsx
```diff
@@ -44,7 +42,7 @@ const ChannelSelection = (props: Props) => {
           <TextInput
             id="selected-channel"
             isDisabled={true}
-            value={getChannelName(notification.channelId, mockChannels, channelSelection.notFound)}
+            value={getChannelName(notification.channelId, channels, channelSelection.notFound)}
```
matthew-gordon, line 45: @whitelisab we can get rid of this function after adding `channelName` to `Notification` and pass `notification.channelName` as the value?

whitelisab (approved):
Looks good!! 🎊

apps/microsoft-teams/frontend/src/components/config/ChannelSelection/ChannelSelection.tsx
matthew-gordon: going to update this in a follow up PR 💪

PR #5761, opened 2023-12-13
feat: [INTEG-1587] - Add notification dedupe modal

matthew-gordon:
## Purpose

We want to avoid allowing the creation and configuration of duplicate notifications, this adds that logic and a modal for confirming or canceling editing of existing notification.

Ticket: https://contentful.atlassian.net/browse/INTEG-1587

## Testing steps

1.) Navigate to the development space, and start the app.

2.) Make sure there is a configured notification for content type and channel.

3.) Attempt to create a new notification with the same content type, and same channel.

4.) Observe this modal:

Screenshot 2023-12-13 at 3 26 59 PM

5.) Test CTA(s)
 - Click `Go back to editing` to cancel creation
 - Click `Update existing notification` to edit the existing notification
 
6.) Observe editing state for previously existing notification:
Screenshot 2023-12-13 at 3 27 11 PM

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/NotificationEditMode/NotificationEditMode.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/NotificationEditMode/NotificationEditMode.tsx
@@ -16,7 +16,11 @@ import CancelModal from '@components/config/CancelModal/CancelModal';
 interface Props {
   index: number;
   deleteNotification: (index: number) => void;
-  updateNotification: (index: number, editedNotification: Partial<Notification>) => void;
+  updateNotification: (
+    index: number,
+    editedNotification: Partial<Notification>,
+    isNew?: boolean
+  ) => void;
   notification: Notification;
   setNotificationIndexToEdit: Dispatch<SetStateAction<number | null>>;
   channels: TeamsChannel[];
@@ -43,8 +47,8 @@ const NotificationEditMode = (props: Props) => {
   };
 
   const handleSave = () => {
-    updateNotification(index, editedNotification);
-    setNotificationIndexToEdit(null);
+    const isNew = isNotificationNew(notification);
+    updateNotification(index, editedNotification, isNew);
   };
 
   const handleCancel = () => {
--- a/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
@@ -10,6 +10,7 @@ import { Notification } from '@customTypes/configPage';
 import { ParameterAction, actions } from '@components/config/parameterReducer';
 import useGetTeamsChannels from '@hooks/useGetTeamsChannels';
 import { ContentTypeContextProvider } from '@context/ContentTypeProvider';
+import DuplicateModal from '../DuplicateModal/DuplicateModal';
 
 interface Props {
   notifications: Notification[];
@@ -55,10 +56,45 @@ const NotificationsSection = (props: Props) => {
 
   const updateNotification = (index: number, editedNotification: Partial<Notification>) => {
     const notificationsPayload = [...notifications];
-    notificationsPayload[index] = { ...notificationsPayload[index], ...editedNotification };
+    const existingNotification = notificationsPayload[index];
+
+    // Update the notification at the specified index
+    const updatedNotification = { ...existingNotification, ...editedNotification };
+    notificationsPayload[index] = updatedNotification;
+
+    // Deduplicate based on content
+    const uniqueNotifications = notificationsPayload.filter(
+      (notification, idx) =>
+        notificationsPayload.findIndex(
+          (n) =>
+            n.channelId === notification.channelId && n.contentTypeId === notification.contentTypeId
+        ) === idx
+    );
+
+    // Check if the updated notification is unique
+    const isUnique = uniqueNotifications.length === notificationsPayload.length;
+
+    // If not unique, open the modal and set the index for further handling
+    if (!isUnique) {
+      ModalLauncher.open(({ isShown, onClose }) => {
+        return (
+          <DuplicateModal
+            isShown={isShown}
+            handleCancel={() => {
+              onClose(true);
+            }}
+            handleConfirm={() => {
+              setNotificationIndexToEdit(index);
+              onClose(true);
+            }}
+          />
+        );
+      });
+    }
+
     dispatch({
       type: actions.UPDATE_NOTIFICATIONS,
-      payload: notificationsPayload,
+      payload: uniqueNotifications,
     });
   };
 
--- a/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
@@ -63,7 +63,7 @@ const NotificationsSection = (props: Props) => {
     notificationsPayload[index] = updatedNotification;
 
     // Use a Set to keep track of unique keys
-    const uniqueKeys = new Set();
+    const uniqueKeys = new Set<string>();
 
     // Deduplicate based on content
     const uniqueNotifications = notificationsPayload.filter((notification) => {
--- a/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
@@ -6,11 +6,11 @@ import AddButton from '@components/config/AddButton/AddButton';
 import NotificationEditMode from '@components/config/NotificationEditMode/NotificationEditMode';
 import NotificationViewMode from '@components/config/NotificationViewMode/NotificationViewMode';
 import DeleteModal from '@components/config/DeleteModal/DeleteModal';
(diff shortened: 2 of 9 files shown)


apps/microsoft-teams/frontend/src/hooks/useGetTeamsChannels.ts
```diff
@@ -38,7 +39,7 @@ const useGetTeamsChannels = () => {
     getAllChannels();
   }, [getAllChannels]);
 
-  return channels;
+  return import.meta.env.DEV ? mockChannels : channels;
```
matthew-gordon, line 42: this is intentional at the moment to be able to work in development with the hosted app action call.. If we don't like this I can definitely remove it. What are your toughts @jsdalton, @lilbitner, @rusticpenguin ?

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
```diff
@@ -55,10 +56,45 @@ const NotificationsSection = (props: Props) => {
...
+    const uniqueNotifications = notificationsPayload.filter(
+      (notification, idx) =>
+        notificationsPayload.findIndex(
+          (n) =>
+            n.channelId === notification.channelId && n.contentTypeId === notification.contentTypeId
+        ) === idx
+    );
+
+    // Check if the updated notification is unique
+    const isUnique = uniqueNotifications.length === notificationsPayload.length;
```
lilbitner, line 75: I think this is a clever way to do this!! 👌  Do we perhaps not want to update the `notificationPayload` on line 63 before checking the uniqueness? If we just run a quick check before doing that, then maybe we don't have to iterate over the `notificationsPayload` "twice". Something like: 

`  notificationPayload.find((n, idx) => n.channelId === updatedNotification.channelId && n.contentTypeId === updatedNotification.contentTypeId) 
`
and then if that renders something ^ we render the modal? Of course there might be nuance here that I am missing with how we manage the notificationsPayload 🤔

lilbitner (commented):
Looks awesome. Added some code comments and will try and test locally as well :)

lilbitner (commented):
Working great 💯 ! I just noticed one thing while on the branch, that when the duplicate prevention modal opens, my current work is "lost", as in the notification I was editing is no longer there. Is that intentional or should I be able to return back to that notification with "Go back to editing"? really a workflow question :)

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
matthew-gordon: That's a great call out! With the way the state management is working on these notifications and the way they get persisted to app installation parameters, the `updatedNotification` NEEDS to be set for this to work as intended from what I gather.

I'm having a little trouble following it still but there is a default:

https://github.com/contentful/apps/blob/master/apps/microsoft-teams/frontend/src/constants/defaultParams.ts#L16-L21

default state in reducer
https://github.com/contentful/apps/blob/master/apps/microsoft-teams/frontend/src/components/config/parameterReducer.ts#L54-L67

dispatch
https://github.com/contentful/apps/blob/master/apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.tsx#L11

Essentially a blank one gets inserted and over written even on creation, so the UI jumps with an extra notification panel displaying the duplicate notification without filtering 😰

I'm trying to work through this snippet now 🤔 but it still needed filtering logic to avoid the ui skip:

```typescript
const updateNotification = (index: number, editedNotification: Partial<Notification>) => {
  const notificationsPayload = [...notifications];
  const existingNotification = notificationsPayload[index];

  // Update the notification at the specified index
  const updatedNotification = { ...existingNotification, ...editedNotification };

  // See if configured notification exists
  const foundNotificaiton = notificationsPayload.find(
    (n) =>
      n.channelId === updatedNotification.channelId &&
      n.contentTypeId === updatedNotification.contentTypeId
  );

  // If found, open the modal and set the index for further handling
  if (foundNotificaiton) {
    ModalLauncher.open(({ isShown, onClose }) => {
      return (
        <DuplicateModal
          isShown={isShown}
          handleCancel={() => {
            onClose(true);
          }}
          handleConfirm={() => {
            setNotificationIndexToEdit(index);
            onClose(true);
          }}
        />
      );
    });
  }

  notificationsPayload[index] = updatedNotification;

  dispatch({
    type: actions.UPDATE_NOTIFICATIONS,
    payload: notificationsPayload.filter(
      (notifcation) =>
        notifcation.channelId === foundNotificaiton?.channelId &&
        notifcation.contentTypeId === foundNotificaiton.contentTypeId
    ),
  });
};
```

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
```diff
@@ -62,14 +62,18 @@ const NotificationsSection = (props: Props) => {
     const updatedNotification = { ...existingNotification, ...editedNotification };
     notificationsPayload[index] = updatedNotification;
 
+    // Use a Set to keep track of unique keys
+    const uniqueKeys = new Set();
```
matthew-gordon, line 66: trying a compound key approach to avoid the heavy initial logic. @lilbitner this seems to also work as intended, will test it out further 🙇

matthew-gordon:
> Working great 💯 ! I just noticed one thing while on the branch, that when the duplicate prevention modal opens, my current work is "lost", as in the notification I was editing is no longer there. Is that intentional or should I be able to return back to that notification with "Go back to editing"? really a workflow question :)

Ah yeah I see what you mean, another great call out 🤔, might need to capture the "original" or its index and make sure to display that with the configured options state

apps/microsoft-teams/frontend/src/hooks/useGetTeamsChannels.ts
lilbitner: yeah I think this work - maybe worth adding a note above to adjust eventually?

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
lilbitner: ah okay interesting! I think the Set approach is honestly awesome - super easy to understand and very efficient!! 🥇

lilbitner:
> > Working great 💯 ! I just noticed one thing while on the branch, that when the duplicate prevention modal opens, my current work is "lost", as in the notification I was editing is no longer there. Is that intentional or should I be able to return back to that notification with "Go back to editing"? really a workflow question :)
> 
> Ah yeah I see what you mean, another great call out 🤔, might need to capture the "original" or its index and make sure to display that with the configured options state

Yeah I think this might be a blocking piece to this given that if I have an existing notification already all set up, I edit it to be a duplicate, then when the modal pops up that existing notification disappears. Happy to pair on this! But yeah seems like there needs to be some added logic around keeping track of that index

apps/microsoft-teams/frontend/src/components/config/NotificationEditMode/NotificationEditMode.tsx
```diff
@@ -43,8 +47,8 @@ const NotificationEditMode = (props: Props) => {
   };
 
   const handleSave = () => {
-    updateNotification(index, editedNotification);
-    setNotificationIndexToEdit(null);
```
whitelisab, line 47: Removing this line allows the user to go back to the notification that they're editing

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
```diff
@@ -53,13 +55,65 @@ const NotificationsSection = (props: Props) => {
...
+    if (!isUnique) {
+      ModalLauncher.open(({ isShown, onClose }) => {
+        return (
+          <DuplicateModal
+            isShown={isShown}
+            handleCancel={() => {
+              onClose(true);
+            }}
+            handleConfirm={() => {
+              if (isNew) {
```
whitelisab, line 84: If the duplicate notification is new and the user chooses to go to the existing duplicate notification, then this new one will go away. If it's an existing notification that is editing to be a duplicate of another, then that notification will reset back to its initial state when you navigate to the other notification.

lilbitner (commented):
Looking great!!!  Love that I can click "Go back to editing" and the notification I was working on is preserved 👌 . I am wondering one more thing and this might be splitting hairs at this point so feel free to ignore or address at a later time! When I instead click on "Update existing notification", I guess I sort of expected that the original notification will have the changes I had on the one I was just editing i.e if I added another event for example. However the existing notification that I am directed to still has the original configuration. Does that make sense? and if so, wdyt?

whitelisab:
> Looking great!!! Love that I can click "Go back to editing" and the notification I was working on is preserved 👌 . I am wondering one more thing and this might be splitting hairs at this point so feel free to ignore or address at a later time! When I instead click on "Update existing notification", I guess I sort of expected that the original notification will have the changes I had on the one I was just editing i.e if I added another event for example. However the existing notification that I am directed to still has the original configuration. Does that make sense? and if so, wdyt?

Great point @lilbitner! I think this would be a good thing to bring up with Emily and we can decide if we want to carry over the selected events to the original notification. Then we can address it in a follow up task. For now @matthew-gordon I will approve this PR so you can get it merged in!

whitelisab (approved):

PR #5787, opened 2023-12-19
fix: hide notifications config if no tenant id [INTEG-1609]

whitelisab:
## Purpose

When a user is installing the MS Teams app, they should not see the Notifications Configuration section before the app is saved/installed. Also, this adds validation so that the app config page cannot be saved if there is no tenant id.

## Approach

See screenshot showing error message when user tries to save the app config without a tenant id. Also note that the Notifications section is not visible yet:

Screenshot 2023-12-19 at 12 26 12 PM

## Testing steps

In our MS Teams development space, uninstall the app and reinstall. Try to save the app without entering a tenant id.


apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.spec.tsx
```diff
@@ -1,18 +1,23 @@
...
 import { render, screen } from '@testing-library/react';
-import { accessSection, notificationsSection } from '@constants/configCopy';
+import { accessSection } from '@constants/configCopy';
 import { mockSdk } from '@test/mocks';
 
 vi.mock('@contentful/react-apps-toolkit', () => ({
   useSDK: () => mockSdk,
 }));
 
+vi.mock('@components/config/NotificationsSection/NotificationsSection', () => ({
```
whitelisab, line 11: I added a mock for this child component because I was seeing errors from the NotificationsSection component when it calls our hooks for getting the content types and channels. These errors were only happening when this test was running in CI.

jsdalton (approved):
Looks good!

PR #5788, opened 2023-12-19
chore: update copy and links [INTEG-1648]

whitelisab:
## Purpose

This PR updates copy and links in the MS Teams app config page based on feedback in our Mob QA session.

## Approach

Updated copy and links. Also ensured that when the empty state is visible for the content type and channel selection modals, that the "select" button is not visible, see example below:

Screenshot 2023-12-19 at 1 28 16 PM

## Testing steps

Navigate to the MS Teams development app.


jsdalton (approved):
Copy reads much better now ☺️

PR #5826, opened 2023-12-28
refactor: add channel context and update installation params [INTEG-1669]

whitelisab:
## Purpose

This PR makes a few changes to channels in the config page:
- Previously we were only storing channel id and we instead want to store all the channel info returned from the app action
- We are prop drilling channels from the `NotificationsSection` component and this PR refactors that to use context to store the channel info
- Removes a helper function that was previously used to get the channel name

## Approach

In a follow up task we can add additional error handling for channels that are deleted, can't be found, etc.

## Testing steps

This may require you to uninstall and reinstall the app in our development environment to test the changes locally (or at least delete all of the configured notifications).

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/NotificationViewMode/NotificationViewMode.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/NotificationViewMode/NotificationViewMode.tsx
@@ -11,13 +11,9 @@ import {
 } from '@contentful/f36-components';
 import { MoreHorizontalIcon } from '@contentful/f36-icons';
 import { styles } from './NotificationViewMode.styles';
-import { getContentTypeName, getChannelName } from '@helpers/configHelpers';
-import { Notification, TeamsChannel } from '@customTypes/configPage';
-import {
-  channelSelection,
-  contentTypeSelection,
-  notificationsSection,
-} from '@constants/configCopy';
+import { getContentTypeName } from '@helpers/configHelpers';
+import { Notification } from '@customTypes/configPage';
+import { contentTypeSelection, notificationsSection } from '@constants/configCopy';
 // TODO: update this when we start fetching channel installations
 
 interface Props {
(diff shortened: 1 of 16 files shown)


apps/microsoft-teams/frontend/src/components/config/NotificationViewMode/NotificationViewMode.tsx
```diff
@@ -11,13 +11,9 @@ import {
...
-import { Notification, TeamsChannel } from '@customTypes/configPage';
-import {
-  channelSelection,
-  contentTypeSelection,
-  notificationsSection,
-} from '@constants/configCopy';
+import { getContentTypeName } from '@helpers/configHelpers';
+import { Notification } from '@customTypes/configPage';
+import { contentTypeSelection, notificationsSection } from '@constants/configCopy';
 // TODO: update this when we start fetching channel installations
```
matthew-gordon, line 17: I might have missed this, is the TODO here still necessary or can we 🧹 ?

apps/microsoft-teams/frontend/src/components/config/NotificationViewMode/NotificationViewMode.tsx
```diff
@@ -58,7 +46,7 @@ const NotificationViewMode = (props: Props) => {
             )}
           </Subheading>
           <Paragraph marginBottom="none">
-            {getChannelName(notification.channelId, channels, channelSelection.notFound)}
+            {`${notification.channel.name}, ${notification.channel.teamName}`}
```
matthew-gordon, line 49: 🎉

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
```diff
@@ -127,36 +126,36 @@ const NotificationsSection = (props: Props) => {
         />
       </Box>
       <ContentTypeContextProvider>
-        {notifications.map((notification, index) => {
-          const inEditMode = notificationIndexToEdit === index;
+        <ChannelContextProvider>
```
matthew-gordon, line 129: It might be worth discussing lifting these providers into a `Providers` component and wrapping higher in component tree 🤔 just a call out since we have multiple providers now, good code huddle topic maybe

matthew-gordon (approved):
Looks great, and works as expected. 💪 🚀

apps/microsoft-teams/frontend/src/components/config/NotificationViewMode/NotificationViewMode.tsx
whitelisab: Good catch! Removed it

apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx
whitelisab: Agreed! I'll keep it as is for now, and we can discuss an approach later. We can clean up a bit when we address some additional error handling later

PR #5082, opened 2023-10-07
fix: Netlify app: correct format pattern

benomatis:
Replace date-fns format pattern `a..aa` with `a`

Fixes https://github.com/contentful/apps/issues/5081

Issue #5081, opened 2023-10-07
Netlify app: format pattern "a..aa" returns "AM..AM"


benomatis:
In date-fns v2.28.0, despite what's indicated in the [format patterns list of the documentation](https://date-fns.org/v2.30.0/docs/format), the format pattern `a..aa` returns `AM..AM` instead of just `AM`. It looks like using simply `a` is what returns `AM`.

I reported this on `date-fns` but you might want to correct it here to use `a` not having to wait for `date-fns` to fix it - not sure if that's the right approach: https://github.com/date-fns/date-fns/issues/3539

https://github.com/contentful/apps/blob/73d8caa4279a81fdbf8089facb10c15b5e218f99/apps/netlify/frontend/src/sidebar/message-processor.js#L80

<img width="338" alt="image" src="https://github.com/contentful/apps/assets/5822748/92c7c0de-fb87-4655-9bfb-9d8fb882e0a5">

benomatis:
This is a misunderstanding of the doc: "a..aa" is a notation meant to say "a" till "aa" (ie. either "a", or "aa").

Now [`date-fns` have a PR suggesting a correction to their doc](https://github.com/date-fns/date-fns/pull/3547/files), so I'm closing this issue here.

PR #5829, opened 2023-12-28
fix: correct Netlify date pattern [INTEG-1670]

whitelisab:
## Purpose

See this [original PR](https://github.com/contentful/apps/pull/5082) and [issue](https://github.com/contentful/apps/issues/5081). Our CI checks were not passing on the original PR, so I cherry-picked the commit to open this new PR.

## Approach

Replace date-fns format pattern `a..aa` with `a`. Tests were passing even with the incorrect date pattern, so I also updated the tests to be more specific, so that they will fail with the old pattern.


lilbitner (approved):
Nice 🙌

PR #5981, opened 2024-01-11
feat: trigger hosted app action to send test notification [INTEG-1655]

whitelisab:
## Purpose

This PR creates a hosted app action that calls the MS Teams Bot Service to send a test notification to a configured channel. It also updates the frontend to call the HAA when you click on the "Test" button for a notification.

## Approach

Followed some of the response/error handling patterns from other apps, and passing in environment variables in the same way that the Merge App does. There can be additional follow up work when we add other HAAs to improve some error handling.

## Testing steps

1. Follow the instructions in the README to set development environment variables
2. Run `npm run build` at the top level
3. Follow the instructions in the README to deploy the build to the sandbox app

Note: at this point the notification will not actually be sent, you should see an error "Team not found". This is because we are still using sample data for the channels instead of pulling in actual installations. You can actually send a message by manually changing the channel and team id to something where your development bot is installed.

## Breaking Changes

None


matthew-gordon (approved):
🚀 LGTM

apps/microsoft-teams/app-actions/src/actions/send-test.spec.ts
```diff
@@ -29,8 +39,11 @@ describe('sendTestNotification.handler', () => {
     context = makeMockAppActionCallContext(cmaClientMockResponses, cmaRequestStub);
   });
 
-  it('returns the ok result', async () => {
-    const result = await handler({ channel: 'foo' }, context);
-    expect(result).to.have.property('ok', true);
+  it('calls the cma to get the tenant id from app installation params', async () => {
```
jsdalton, line 42: In addition to the test you have here (which verifies the side effect of calling the handler) I always find it nice to include a test that verifies the output (in this case the JSON blog with the ok: true). Would be nice to have this test here.

apps/microsoft-teams/app-actions/src/actions/send-test.ts
```diff
@@ -1,16 +1,88 @@
...
 
 interface AppActionCallParameters {
-  channel: string;
+  channelId: string;
+  teamId: string;
+  contentTypeId: string;
+  spaceName: string;
+}
+
+export interface BotServiceResponse {
```
jsdalton, line 13: We have some types that are going to work a lot better for what we want to do here 😄 .

See https://github.com/contentful/apps/blob/master/apps/ai-image-generator/backend/src/types.ts#L32-L48

You can see an example of how we type the return of our handler here:

https://github.com/contentful/apps/blob/master/apps/ai-image-generator/backend/src/actions/aiig-generate-image.ts#L18

This type is going to work better for disambiguating between success and error results.

We really need a backend toolkit (or "hosted app actions toolkit"? 🤔 ) where we can stick these common types and utilties instead of copying and pasting. 😢

apps/microsoft-teams/app-actions/src/actions/send-test.ts
```diff
@@ -1,16 +1,88 @@
...
   _context: AppActionCallContext
-): Promise<AppActionCallResponse<null>> => {
+): Promise<AppActionCallResponse<string>> => {
+  const { channelId, teamId, contentTypeId, spaceName } = _payload;
+  const {
+    cma,
+    appActionCallContext: { appInstallationId },
+  } = _context;
+
+  const config = {
```
jsdalton, line 29: Let's add a todo to break this into its own file (ie something like config.ts) since we'll want to share it across actions later.

apps/microsoft-teams/app-actions/src/helpers/send-test-notification.ts
```diff
@@ -0,0 +1,28 @@
...
+  const res = await fetch(`${botServiceUrl}/api/notifications/test_notification`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'x-api-key': apiKey,
+    },
+    body: JSON.stringify(payload),
+  });
+
+  const response = (await res.json()) as BotServiceResponse;
```
jsdalton, line 25: Let's add a TODO to parse the response instead of the assertion here.

apps/microsoft-teams/build-actions.js
```diff
@@ -5,6 +5,8 @@ const { hideBin } = require('yargs/helpers');
 
 const manifest = require('./contentful-app-manifest.json');
 
+require('dotenv').config({ path: './app-actions/.env' });
```
jsdalton, line 8: Maybe an easier way (per my other comments above) would be to have a default .env template (instead of excluding it in git) that sets the env vars to be lame text strings and then we override with "real" values in the build steps? 🤔 

Since we're already relying on .env here I think we might as well.

apps/microsoft-teams/app-actions/src/actions/send-test.ts
whitelisab: Awesome, thank you! Agreed on adding this to backend toolkit that has common types and helpers for HAA and lambdas!

apps/microsoft-teams/app-actions/src/actions/send-test.spec.ts
whitelisab: Yes 💯! I will address that in a follow up

apps/microsoft-teams/build-actions.js
whitelisab: I updated this so that we have the .env with default text. For local development, you can set these to actual values. Then for CI, I created new contexts in Circle CI for apps staging and production and moved the environment variables there. This approach seems cleaner to me compared to the first iteration, but I'd love your feedback to see if this is a better direction!

jsdalton: As discussed let's follow up later to require dotenv in tests, and also create a .env.development example file

jsdalton (approved):

Issue #5599, opened 2023-11-30
sap-commerce-cloud


FilipeAPV:
**Describe the bug**

1. The total number of pages is not being updated in `Dialog.tsx` state.
The problem seems to come from the method `updateTotalPages` that is not using the function parameter "totalPages":
```
updateTotalPages = (totalPages: number) => {
    // this.setState({ totalPages: this.state.totalPages }); -> bug
    this.setState({ totalPages: totalPages }); 
  };
```

2. Two bugs in the following code also from `Dialog.tsx`:

```
  nextPageButtonEvent = () => {
    this.setState({ page: this.state.page - 1 });
    this.load();
  };

  prevPageButtonEvent = () => {
    this.setState({ page: this.state.page - 1 });
    this.load();
  };
```
- `nextPageButtonEvent ` should be + 1, to retrieve the next page
- setState is async, so this.load() is going to be called with the non-updated value.

Maybe something like this would be the solution:
```
nextPageButtonEvent = () => {
  this.setState({ page: this.state.page + 1 }, () => {
    this.load();
  });
};
```

**To Reproduce**
1. Install and configure connector
2. Open connector dialog
3. In the first bug, the pagination btn's will not be shown. In the second one, you'll be working with the previous value of "page".

rusticpenguin:
Thank you so much for triaging this Filipe! I'll be working to get these fixes in tomorrow!

lilbitner:
Hi @FilipeAPV, thank you for raising this bug! We have pushed a fix and it should be resolved now. Please feel free to open another Github issue if you find issues on your end with the resolution.

PR #5986, opened 2024-01-12
fix bugs in sap-commerce cloud dialog

rusticpenguin:
## Purpose

Bugs were pointed out in #5599. This ticket is to address them

This had to do with pagination and total page size


lilbitner (approved):
Looks good!

PR #6009, opened 2024-01-16
feat: add validation for commercetools [MONET-1737]

cemreyuksel:
## Purpose

[Ticket](https://contentful.atlassian.net/browse/MONET-1737)

## Approach

When the API URL contains a trailing / the 3rd-party API request in graph-api fails. This PR removes the '/' if there is one, inside `validateParameters` function. 

## Testing steps

I tested locally by running this app. Only when you refresh the page, the value on the form changes but I think this is not a problem for the bug fix. There were no tests for the application.

### DIFF
--- a/apps/commercetools-without-search/src/config.ts
+++ b/apps/commercetools-without-search/src/config.ts
@@ -79,5 +79,9 @@ export function validateParameters(parameters: ConfigurationParameters): string 
     return 'Provide the commercetools data locale.';
   }
 
+  if (parameters.authApiEndpoint?.endsWith('/')) {
+    parameters.apiEndpoint = parameters.apiEndpoint?.slice(0, -1);
+  }
+
   return null;
 }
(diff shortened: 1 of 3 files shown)


kdamball (approved):

kdamball (commented):
Can we make sure it doesn't break the frontend app

apps/commercetools-without-search/src/config.ts
```diff
@@ -79,5 +79,9 @@ export function validateParameters(parameters: ConfigurationParameters): string
     return 'Provide the commercetools data locale.';
   }
 
+  if (parameters.authApiEndpoint?.endsWith('/')) {
+    parameters.apiEndpoint = parameters.apiEndpoint?.slice(0, -1);
```
mrlaessig, line 83: Is this correct to check `authApiEndpoint` and then overwrite `apiEndpoint`?

cemreyuksel: No, this is copy paste error. Pushed the right commit now, sorry.

mrlaessig (approved):

matthew-gordon (approved):
LGTM, feel free to merge 🚀

PR #6033, opened 2024-01-18
feat: introduce provider with custom api for self installs

jsdalton:
## Purpose

We want to support the ability for the MS Teams app to trigger a save of configurations directly.

## Approach

* Expose the undisclosed API (implemented in experience packages) that allows for self installs to be triggered via the SDK (via an undocumented ability to provide a custom API to the SDK set up)
* Provide a custom SDK provider which leverages the custom API
* Provide a hook to expose the API itself
* Named it `saveConfiguration` in the API we use to make it more clear what's actually happening


whitelisab (approved):
Makes sense to me! ⭐

PR #6034, opened 2024-01-18
chore: remove msteams service (implemented in separate repo)

jsdalton:
## Purpose

SInce the msteams-bot-service is implemented in a separate repo, we can get rid of some old code.


whitelisab (approved):
Thank you for the cleanup! 🧹

PR #6044, opened 2024-01-19
feat: add HAA to get channels list [INTEG-1633]

whitelisab:
## Purpose

This PR adds a hosted app action to call our MS Teams Bot Service to get the list of channels for the teams where the Contentful app is installed for a given tenant id. This eventually displays in the channel selection modal.

## Approach

We already had this stubbed out using mock data, so I just implemented actually calling the bot service to get the channels. I also decided to have the HAA get the tenant id from the installation parameters, rather than passing it from the frontend. In order to utilize some types from MS Teams, I installed the `botbuilder` package.

## Testing steps

Deployed to our sandbox space and you are able to see the list of actual channels.

Screenshot 2024-01-18 at 11 01 14 AM

### DIFF
--- a/apps/microsoft-teams/app-actions/package.json
+++ b/apps/microsoft-teams/app-actions/package.json
@@ -13,7 +13,6 @@
   },
   "dependencies": {
     "@contentful/node-apps-toolkit": "^2.6.0",
-    "botbuilder": "^4.21.4",
     "contentful-management": "^11.3.0"
   },
   "devDependencies": {
@@ -22,6 +21,7 @@
     "@types/mocha": "^10.0.1",
     "@types/sinon": "^10.0.16",
     "@types/sinon-chai": "^3.2.9",
+    "botbuilder": "^4.21.4",
     "chai": "^4.3.7",
     "mocha": "^10.2.0",
     "sinon": "^17.0.0",
--- a/apps/microsoft-teams/app-actions/package.json
+++ b/apps/microsoft-teams/app-actions/package.json
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@contentful/node-apps-toolkit": "^2.6.0",
+    "botbuilder": "^4.21.4",
     "contentful-management": "^11.3.0"
   },
   "devDependencies": {
--- a/apps/microsoft-teams/app-actions/src/actions/list-channels.ts
+++ b/apps/microsoft-teams/app-actions/src/actions/list-channels.ts
@@ -1,16 +1,59 @@
 import { AppActionCallContext } from '@contentful/node-apps-toolkit';
 import { AppActionCallResponse, Channel } from '../types';
-import { mockChannels } from '../../test/fixtures/mockChannels';
-
-interface AppActionCallParameters {
-  tenantId: string;
-}
+import { fetchTenantId } from '../utils';
+import helpers from '../helpers';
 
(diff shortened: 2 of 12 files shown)


apps/microsoft-teams/app-actions/src/actions/list-channels.ts
```diff
@@ -1,16 +1,54 @@
...
+    if (config.botServiceUrl === undefined) {
+      throw new Error('MS Teams Bot Service URL not provided.');
+    }
+
+    if (config.apiKey === undefined) {
+      throw new Error('MS Teams Bot Service API Key not provided.');
+    }
+
+    const tenantId = await fetchTenantId(cma, appInstallationId);
+    channels = await helpers.getChannelsList(config.botServiceUrl, config.apiKey, tenantId);
```
jsdalton, line 28: I think this is okay for now but to promote reuse I actually recommend we adopt the "client pattern" (not an official term, i just made that up) here.

basically instead of buiding ad hoc fetch requests every time we want to pull data from our MS Teams service we should build a dedicated client class around that API and implement each different kind of request as a method.

this also makes it much easier to test and later to evolve.

apps/microsoft-teams/app-actions/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@contentful/node-apps-toolkit": "^2.6.0",
+    "botbuilder": "^4.21.4",
```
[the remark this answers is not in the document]
matthew-gordon: If this is just for types then can probably just move this to dev dependencies since they are ripped out at compile time

apps/microsoft-teams/app-actions/src/helpers/get-channels-list.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  if (!response.ok) {
+    throw new Error(response.errors?.[0]?.message ?? 'Failed to get channels');
+  }
+
+  const channelsList = transformInstallationsToChannelsList(response.data, tenantId);
+  return channelsList;
+};
+
+export const transformInstallationsToChannelsList = (
```
jsdalton, line 27: I mentioned the idea of converting the `getChannelsList` function to a method of an API client. But I think if we do that it's better to keep that method "dumb" (just return the data without transforming it to our own data model in the app action).

but that means we _should_ keep this transformation function here as a helper and then take the list we get back and transform it to the format we need.

apps/microsoft-teams/app-actions/src/actions/list-channels.ts
whitelisab: I created a follow up ticket for us to address this refactoring later!

apps/microsoft-teams/app-actions/src/actions/list-channels.ts
```diff
@@ -1,16 +1,54 @@
...
 ): Promise<AppActionCallResponse<Channel[]>> => {
-  const channels = mockChannels;
+  const {
+    cma,
+    appActionCallContext: { appInstallationId },
+  } = context;
+
+  let channels: Channel[];
+
+  try {
```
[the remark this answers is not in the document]
whitelisab: Done! Updated the config file based on our discussion

jsdalton (approved):
🎉

PR #6045, opened 2024-01-19
fix: separate custom api and sdk contexts

jsdalton:
## Purpose

The `useSDK` hook was no longer working after we introduced the `SdkWithCustomApiProvider`. The problem was that the `SDKContext` needed to be populated with a provider, which was missing.

## Approach

* Just split the two contexts and include the SDK provider in the custom provider as a child element

Screenshot of config actually working:

image


whitelisab (approved):
Nice!

PR #6084, opened 2024-01-23
feat: [INTEG-1686] - Add vercel base app

matthew-gordon:
## Purpose

- Adds base Vercel app and configuration screen


apps/vercel/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@contentful/react-apps-toolkit": "1.2.16",
+    "contentful-management": "10.38.3",
+    "emotion": "10.0.27",
+    "react": "18.2.0",
+    "react-dom": "18.2.0"
+  },
+  "scripts": {
+    "start": "vite",
+    "build": "tsc && vite build",
+    "test": "vitest",
```
jsdalton, line 18: let's makes sure to add a `test:ci` script so that our tests run in CI

apps/vercel/src/components/common/VercelIcon.tsx
```diff
@@ -0,0 +1,12 @@
+export const VercelBrand = {
+  primaryColor: '#000',
+  url: '',
+};
+
+export default function AmplifyIcon(props: any) {
```
jsdalton, line 6: maybe add a TODO (If this is in fact the amplify icon) or rename (if it's just misnamed)

jsdalton (approved):
Small things from me. Obviously there'w a lot of work in progress stuff (lorem ipsum content, blank README) but i think they are obivous enough that we can fix later.

apps/vercel/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,118 @@
...
+                spellCheck={false}
+                name="accessToken"
+                type="password"
+                placeholder={'ex. atE2sdftcIp01O1isdfXc3QTdT4...'}
+                value={accessToken}
+                onChange={handleAccessTokenChange}
+              />
+              <HelpText>
+                Follow{' '}
+                <TextLink
```
whitelisab, line 95: Eventually we can use some of the components in the `integration-frontend-toolkit`, like the `HyperLink` component for this 😄

apps/vercel/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,118 @@
...
+import VercelIcon from '../components/common/VercelIcon';
+
+export interface AppInstallationParameters {
+  accessToken?: string;
+}
+
+const ConfigScreen = () => {
+  const sdk = useSDK<ConfigAppSDK>();
+  const [parameters, setParameters] = useState<AppInstallationParameters>({});
+  const [accessToken, setAccessToken] = useState<string>('');
```
rusticpenguin, line 26: I feel like we could cut down on duplicate data & some moving parts by removing accessToken and just using parameters instead

apps/vercel/src/components/config-screen/ConfigScreen.styles.tsx
```diff
@@ -0,0 +1,34 @@
+import tokens from '@contentful/f36-tokens';
+import { css } from 'emotion';
+import { VercelBrand } from '../common/VercelIcon';
+
+export const styles = {
```
whitelisab, line 5: Is this the right design for the background of the app config page? Maybe I'm looking at the wrong design file, but I think it was supposed to look more like this (no black background, no image at the bottom, etc.): 
Screenshot 2024-01-23 at 4 38 27 PM

whitelisab (approved):
Looks good! I had one question about the overall design on the app config page, but that can be addressed in a follow up if needed.

apps/vercel/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,118 @@
...
+  TextInput,
+  TextLink,
+} from '@contentful/f36-components';
+import { useSDK } from '@contentful/react-apps-toolkit';
+import { ExternalLinkIcon } from '@contentful/f36-icons';
+import { styles } from '../components/config-screen/ConfigScreen.styles';
+import VercelIcon from '../components/common/VercelIcon';
+
+export interface AppInstallationParameters {
+  accessToken?: string;
```
rusticpenguin, line 20: Instead of optional fields, I feel we could make sure everything is always defined. This will lead to a nicer dev experience as we won't have to null check when trying to access properties on `parameters`.

When initializing the AppInstallation parameters we can instantiate it with a default object. 
```ts
const [parameters, setParameters] = useState<AppInstallationParameters>({
  accessToken: ''
});
```

apps/vercel/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+const ConfigScreen = () => {
+  const sdk = useSDK<ConfigAppSDK>();
+  const [parameters, setParameters] = useState<AppInstallationParameters>({});
```
rusticpenguin, line 25: Not necessary for this current PR, but we may want to switch this over to a reducer rather than state. 

That way we can easily utilize field validation and have a centralized place for managing state.

apps/vercel/src/components/config-screen/ConfigScreen.styles.tsx
matthew-gordon: We aren't currently doing an OATUH flow for this, so for consistency sake its following the previously established config screen designs (amplify) for base configuration since its not ready for prime time and designs aren't finalized

apps/vercel/src/locations/ConfigScreen.tsx
matthew-gordon: yeah intention was not to over complicate this since we are moving towards na OAUTH flow, right now the only requirements are to be able to store the access token in parameters and using a controlled input to manage the value

apps/vercel/src/locations/ConfigScreen.tsx
matthew-gordon: yeah sounds good 👍 will address this tomorrow

apps/vercel/src/locations/ConfigScreen.tsx
matthew-gordon: yep makes sense, that just for the initial controlled input honestly as we are going to implement an OAUTH flow

apps/vercel/src/locations/ConfigScreen.tsx
matthew-gordon: yes indeed, definitely makes sense. This is totally a placeholder app at the moment.

apps/vercel/src/locations/ConfigScreen.tsx
rusticpenguin: With the OAUTH flow, will we be removing this field? If so then I agree, completely a moot point!

PR #6094, opened 2024-01-24
Rename app event topics

jsdalton:
## Purpose

The current topic names we chose for event subscription in the app config are somewhat ambiguous and could cause us problems if we decide to support notifications for other resources in the future. It's also a bit more cumbersome to compare incoming events with the object where those notification subscriptions are specified.

## Approach

* Rename the events to the app event topics
* A bit of reshuffling of files

### DIFF
--- a/apps/microsoft-teams/frontend/src/customTypes/configPage.ts
+++ b/apps/microsoft-teams/frontend/src/customTypes/configPage.ts
@@ -1,14 +1,3 @@
-export const TOPIC_ACTION_MAP = {
-  'ContentManagement.Entry.create': 'created',
-  'ContentManagement.Entry.save': 'saved',
-  'ContentManagement.Entry.auto_save': 'auto saved',
-  'ContentManagement.Entry.archive': 'archived',
-  'ContentManagement.Entry.unarchive': 'unarchived',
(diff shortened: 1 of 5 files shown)


apps/microsoft-teams/frontend/src/customTypes/configPage.ts
```diff
@@ -1,4 +1,22 @@
-import { AppEventKey } from '@constants/configCopy';
+export const TOPIC_ACTION_MAP = {
+  'ContentManagement.Entry.create': 'created',
+  'ContentManagement.Entry.save': 'saved',
```
whitelisab, line 3: I assume we'll just keep `save` and `auto_save` in here for now even though we aren't supporting those events in the UI currently?

whitelisab (approved):
LGTM! 💯

apps/microsoft-teams/frontend/src/customTypes/configPage.ts
jsdalton: Ha ha no actually, this is a good catch.

I was originally thinking about aligning our types (and using the topic action map) but it was turning into too much trouble with too little benefit so I unwinded. But then I forgot to delete this constant. :)

PR #6097, opened 2024-01-24
refactor: change notification to message and utilize new endpoint [INTEG-1713]

whitelisab:
## Purpose

We are renaming notification to message to describe what is sent to the MS Teams channel. This PR updates the names of the HAA and helper functions. This also utilizes the refactored endpoint created in the bot service as well as updating to reflect the new payload and response.

## Approach

- Refactors to follow the pattern of the entry activity message endpoint (e.g. change notifications/test_notification to tenant/:tenantId/test_messages)
- Removes the space name from the test message adaptive card
- Changes the response shape so that `data` is an object, not a string
- Changes request shape so that `teamId` and `channelId` are part of the `channel` object (same pattern as entry activity message endpoint)

## Testing steps

Tested locally in sandbox and with bot service changes.

## Breaking Changes

This PR needs to be merged in conjunction with the bot service changes


matthew-gordon (approved):
LGTM, definitely more clear around the naming 🚀

PR #6098, opened 2024-01-24
feat: add loading state for test message [INTEG-1745]

whitelisab:
## Purpose

This PR adds a loading state for the test message button, since the user didn't have any feedback that the test message had been sent while waiting for a response.

## Approach

- Updated button copy to be more descriptive
- Button now has loading state and is disabled until it receives a response back from the HAA

Screenshot 2024-01-24 at 3 39 12 PM

## Testing steps

Pushed to sandbox to test a successful and failure test message.

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/NotificationEditModeFooter/NotificationEditModeFooter.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/NotificationEditModeFooter/NotificationEditModeFooter.tsx
@@ -7,16 +7,21 @@ interface Props {
   handleCancel: () => void;
   handleSave: () => void;
   isSaveDisabled: boolean;
+  isTestLoading: boolean;
 }
 
 const NotificationEditModeFooter = (props: Props) => {
-  const { handleTest, handleCancel, handleSave, isSaveDisabled } = props;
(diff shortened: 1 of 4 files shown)


matthew-gordon (approved):
✨ 🌮 🎊

apps/microsoft-teams/frontend/src/components/config/NotificationEditModeFooter/NotificationEditModeFooter.tsx
```diff
@@ -7,16 +7,21 @@ interface Props {
...
   return (
     <Box className={styles.footer}>
       <Flex justifyContent="flex-end" margin="spacingS">
         <ButtonGroup variant="spaced" spacing="spacingS">
-          <Button variant="transparent" onClick={handleTest}>
+          <Button
+            variant="transparent"
+            onClick={handleTest}
+            isLoading={isTestLoading}
+            isDisabled={isTestLoading}>
```
matthew-gordon, line 24: Would it make more sense  if `isTestLoading` to use `Sending` verbage instead of `Send test message`? No biggie open to either.

whitelisab: I like changing this to `isTestSending`, thanks for the suggestion! 🙌

PR #6118, opened 2024-01-25
feat: add support for actually sending messages

jsdalton:
## Purpose

Our handle app event knows what message data it needs to send and which channels it needs to send it to. The last step is to _actually_ call the API and send the messages.

## Approach

* made an intentional change to error object in the result type

### DIFF
--- a/apps/microsoft-teams/app-actions/src/services/msteams-bot-service.ts
+++ b/apps/microsoft-teams/app-actions/src/services/msteams-bot-service.ts
@@ -1,0 +1,32 @@
+import { EntryActivityMessage, SendMessageResult } from '../types';
+
+export class MsTeamsBotService {
+  constructor(public readonly botServiceUrl: string, public readonly apiKey: string) {}
+
+  async sendEntryActivityMessage(
+    entryActivitytMessage: EntryActivityMessage,
+    tenantId: string
+  ): Promise<SendMessageResult> {
+    const res = await fetch(
+      `${this.botServiceUrl}/api/tenant/${tenantId}/entry_activity_messages`,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'x-api-key': this.apiKey,
+        },
+        body: JSON.stringify(entryActivitytMessage),
+      }
+    );
+    const responseBody = await res.json();
+    this.assertSendMessageResult(responseBody);
+    return responseBody;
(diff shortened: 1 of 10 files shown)


apps/microsoft-teams/app-actions/src/services/msteams-bot-service.ts
```diff
@@ -0,0 +1,32 @@
+import { EntryActivityMessage, SendMessageResult } from '../types';
+
+export class MsTeamsBotService {
+  constructor(public readonly botServiceUrl: string, public readonly apiKey: string) {}
+
+  async sendEntryActivityMessage(
+    entryActivitytMessage: EntryActivityMessage,
```
whitelisab, line 7: just a small typo in this parameter -- `entryActivitytMessage ` should be `entryActivityMessage `

apps/microsoft-teams/app-actions/src/services/msteams-bot-service.ts
```diff
@@ -0,0 +1,32 @@
...
+    const responseBody = await res.json();
+    this.assertSendMessageResult(responseBody);
+    return responseBody;
+  }
+
+  private assertSendMessageResult(value: unknown): asserts value is SendMessageResult {
+    if (typeof value !== 'object' || !value)
+      throw new TypeError('invalid type returned from MsTeamsBotService');
+    if (!('ok' in value))
+      throw new TypeError('malformed SendMessageResult returrnd from MsTeamsBotService API');
```
whitelisab, line 30: change `returrnd` to `returned`

apps/microsoft-teams/app-actions/src/types.ts
```diff
@@ -15,7 +15,7 @@ export interface AppActionCallResponseSuccess<TResult> {
 
 export interface AppActionCallResponseError {
   ok: false;
-  errors: ActionError[];
+  error: ActionError;
```
whitelisab, line 18: Yes, I like this change!

apps/microsoft-teams/app-actions/src/services/msteams-bot-service.ts
whitelisab, line 1: Nice, I like this MsTeamsBotService. We can eventually move the test message here, too.

whitelisab (approved):
Awesome! 🏅

PR #6119, opened 2024-01-25
feat: standardize error handling

jsdalton:
## Purpose

Our existing error handling is a bit repetitive and non-robust (errors could still be thrown by certain actiions and not "handled".

## Approach

* Provide a wrapper for our handlers


whitelisab (approved):
Yes! I really like centralizing all of this ⭐

PR #6121, opened 2024-01-25
chore: refactor app actions to use ms teams bot service client

jsdalton:
## Purpose

We decided that code would be cleaner if we standardized around the `MsTeamsBotService` client in our calls to that API


apps/microsoft-teams/app-actions/test/mocks.ts
```diff
@@ -285,3 +285,26 @@ export const mockContentType: ContentTypeProps = {
     },
   ],
 };
+
+export const mockTeamInstallation: TeamInstallation = {
```
whitelisab, line 289: There's also a `mockInstallation` [here](https://github.com/contentful/apps/blob/master/apps/microsoft-teams/app-actions/test/fixtures/mockInstallations.ts). We should probably decide if we want to keep everything in one mock file or have these as separate fixtures.

apps/microsoft-teams/app-actions/src/actions/send-test.ts
```diff
@@ -35,20 +31,17 @@ export const handler = withAsyncAppActionErrorHandling(
...
-      tenantId,
-      testMessagePayload
+    const msTeamsBotServiceResponse = await config.msTeamsBotService.sendTestMessage(
+      testMessagePayload,
+      tenantId
     );
 
-    if (!response.ok) {
-      throw new Error(response.error ?? 'Failed to send test message');
+    // if we align the error types from the bot service and our app action error we could just return
```
whitelisab, line 39: Yes! When we update the bot service error to be an object, not a string, then we can do this.

whitelisab (approved):
Looks great, much cleaner! 💪

apps/microsoft-teams/app-actions/test/mocks.ts
jsdalton: Fixed. I removed the fixture to standardize around the "mocks" mega file.

I guess we're pretty loosy goosey with what's a fixture v.s what's a mock. We should align on that topic :)

PR #6139, opened 2024-01-26
chore: clean up user stuff

jsdalton:
## Purpose

Since we can't get users from an app, we don't need the code that tries!


whitelisab (approved):
🧹 💯

PR #6151, opened 2024-01-29
feat: sort channels returned from HAA [INTEG-1744]

whitelisab:
## Purpose

The channel selection modal should display the channels in a sorted order.

## Approach

Updated the HAA to returned sorted channels. Channels are now displayed sorted by team name, then by channel name, but always with the General channel first. Also, I moved all the mocks to the same file. See example of sorted channels in the modal:

Screenshot 2024-01-29 at 9 00 39 AM

## Testing steps

Tested in staging.


matthew-gordon (approved):
Nice, love the filter logic ✨

PR #6154, opened 2024-01-29
chore: upgrade swr to 2.2.3

jsdalton:
## Purpose

Side Notes App 🎶 is stuck on an earlier version of swr do to some type changes.

Problem identified here https://github.com/vercel/swr/issues/2826

## Approach

* Add a third empty argument to match a different overload. See https://github.com/vercel/swr/issues/2826#issuecomment-1804077705


matthew-gordon (approved):
🚀

PR #6169, opened 2024-01-30
chore: upgrade forma36 components

jsdalton:
## Purpose

* Unblock dependabot upgrade https://github.com/contentful/apps/pull/6038

## Approach

* Need to update to react-scripts v5


matthew-gordon (approved):
🏅

PR #6170, opened 2024-01-30
chore: remove ecommerce app

jsdalton:
## Purpose

* The ecommerce app POC was never released nor is there any intention to release it. So we should remove it from our repository

## Approach

* I created a branch and pushed it up https://github.com/contentful/apps/tree/ecommerce-app-poc so we could look at the code for reference if we need to (or run the app locally)

TODO

* [ ] Delete app definition
* [ ] Delete spaces


rusticpenguin (approved):
RIP 💻 💰. You will be missed

matthew-gordon (approved):
The has been that never was 🖖

PR #6172, opened 2024-01-30
chore: upgrade react scripts for brandfolder

jsdalton:
## Purpose

* Unblock https://github.com/contentful/apps/pull/6171

## Approach

* Need to upgrade to react-scripts 5 for a compatible build script version


matthew-gordon (approved):
🔥

PR #6173, opened 2024-01-30
chore: upgrade react-script to v5 throughout

jsdalton:
## Purpose

* An old version of react-scripts is behind this dependabot https://github.com/contentful/apps/pull/6171

## Approach

* Rather than playing whackamole just update them all at once


matthew-gordon (approved):
✅

jjolton-contentful:
is this build failure potentially related to https://contentful.slack.com/archives/C01KSUNUJHX/p1706609830032219 ?

jsdalton:
Good question @jjolton-contentful . In this case I don't believe so. It has to do with a few layers of dependencies being out of date.

jsdalton:
This change is revealing some other weirdness that I'm not ready to tackle at the moment. I'm going to have to circle back and try these one by one. Closing for the time being.

PR #6193, opened 2024-01-31
feat: add headers to be passed to bots service

lilbitner:
## Purpose

The purpose of this PR is to pass headers from the app actions to the ms-teams-bots-service API. 

## Approach

Generally we would expect headers that describe the user and browser Contentful details, but since we are using app actions, we have to pass these headers a bit more manually. App actions already have a request context that we can utilize for this to get some information. 

I decided to add `requestContext` as an argument for each method calling the MSTeamsBotService, instead of instantiating the service with the requestContext, to ensure the service remains flexible, and so that we don't need to instantiate it multiple times to save an extra argument . Though I would be happy to adjust, it seems to me that there are pros and cons to both approaches.

### DIFF
--- a/apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
+++ b/apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
@@ -24,7 +24,7 @@ export const handler = withAsyncAppActionErrorHandling(
   ): Promise<AppActionCallResponse<SendEntryActivityMessageResult[]>> => {
     const {
       cma,
-      appActionCallContext: { appInstallationId },
+      appActionCallContext: { appInstallationId, environmentId, spaceId, userId },
     } = context;
 
     const { payload, topic: topicString, eventDatetime } = parameters;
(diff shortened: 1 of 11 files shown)


jsdalton (approved):

apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
```diff
@@ -24,7 +24,7 @@ export const handler = withAsyncAppActionErrorHandling(
   ): Promise<AppActionCallResponse<SendEntryActivityMessageResult[]>> => {
     const {
       cma,
-      appActionCallContext: { appInstallationId },
+      appActionCallContext: { appInstallationId, environmentId, spaceId, userId },
```
jsdalton, line 27: it'll be interesting to see what value we get for `userId` here, since this action is triggered by an app backend, not a user 🤔

lilbitner: Yeah good question, so we are actually passing these headers in via the frontend, so it should ideally be client-side information captured about the user if I am understanding correctly!

PR #6204, opened 2024-02-01
chore: pin prettier explicitly to installed version

jsdalton:
## Purpose

There's a problem that can arise for developers in this repository where:

* You haven't installed the root app (i.e. `npm i` at the root level)
* You make some code changes
* Your globally installed prettier has a different version than the version for this repo
* Your prettier post commit hook defers to the globally installed version (since no local is present)
* Your code gets autoformatted in a way that's incompatible with the prettier checks for this repo

## Approach

* Use `npx` for all prettier invocations (directly or indirectly) and fix these to match the declared version
* Prevent dependabot from upgrading prettier and instead force a manual upgrade


rusticpenguin (approved):
👍

PR #6234, opened 2024-02-01
chore: upgrade f36 for mux

jsdalton:
## Purpose

* This mega PR is failing so to fix we will upgrade each app one by one https://github.com/contentful/apps/pull/6200

## Approach

* Needed to upgrade react-scripts to 5.0.1 because webpack +5 is required for f36 components latest
* webpack5 doesn't work with node modules, these need to be compiled independently
* react-app-rewired is the recommended solution for customing webpack for create-react-app apps
* config-overrides.js is where we implement the required browser version of node modules. mostly these are dependencies of `jsonwebtokens` (which is a node module itself, not a browser module)
* the tests appear to never have been run or configured for this app. so added a `test:ci` script to actually run the tests
* to fix a broken test had to add a setupTests.ts file and polyfill a working ResizeObserver
* Also needed to change a mock which never functioned in tests

## Testing steps

Manually confirmed after all these upgrades that the app itself still works:

image


apps/mux/package.json
```diff
@@ -5,37 +5,43 @@
   "devDependencies": {
     "@contentful/app-scripts": "^1.2.0",
     "@testing-library/jest-dom": "5.17.0",
-    "@testing-library/react": "12.1.5",
+    "@testing-library/react": "12.1.4",
```
lilbitner, line 8: is this downgrade intentional?

lilbitner (approved):
Confirmed it works locally myself, and code looks good 👍 just one pending question! on the dependency changes!

PR #6245, opened 2024-02-02
feat: add auth flow for MS Teams [INTEG-1761]

whitelisab:
## Purpose

This PR adds an authorization flow for the MS Teams app so that we can get the tenant id from an authorized MS user rather than having the user paste it in.

## Approach

I utilized the [Microsoft Authentication Library for React](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/README.md#samples). I implemented an `AuthProvider` which allows us to use hooks and methods from this library to authenticate a user. 

When a user authenticates on initial installation, I am using our `customApi.saveConfiguration` to trigger an installation and save the tenant id in installation parameters. 

The `Disconnect` button prompts a modal asking the user to confirm that they want to sign out, then goes into the MS logout flow. If a user is not authenticated, then the notifications section won't show.

For now, I am displaying the name and username of the authenticated user. We have a follow up ticket to get the organization name and logo.

## Testing steps

You will need to add `VITE_MS_APP_CLIENT_ID` environment variable (see 1password).


apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.tsx
```diff
@@ -46,13 +52,22 @@ const ConfigPage = () => {
...
+        <Heading>{headerSection.title}</Heading>
+        <Paragraph>{headerSection.description}</Paragraph>
+        <hr className={styles.splitter} />
+        <AccessSection
+          dispatch={dispatchParameters}
+          parameters={parameters}
+          isAppInstalled={isAppInstalled}
+        />
+      </Box>
+      {isAppInstalled && accounts.length ? (
```
lilbitner, line 65: nit: if you have `isAppInstalled && accounts.length && (<NotificationsSection .... />)` here does this cause the weird issue of showing the "0"? I just tend to prefer avoiding a return of `null`

apps/microsoft-teams/frontend/src/components/config/DisconnectModal/DisconnectModal.spec.tsx
```diff
@@ -0,0 +1,12 @@
+import DisconnectModal from './DisconnectModal';
+import { describe, expect, it, vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { accessSection } from '@constants/configCopy';
+
+describe('DisconnectModal component', () => {
+  it('mounts and renders the correct content', () => {
+    render(<DisconnectModal isShown={true} handleCancel={vi.fn()} handleDisconnect={vi.fn()} />);
+
+    expect(screen.getByText(accessSection.disconnectModal.confirmDisonnect)).toBeTruthy();
```
lilbitner, line 10: nit: I am pretty new to vitest, but sometimes I find that asserting `toBeTruthy` is just asserting that the node exists, and not that it is visible (like `toBeVisible`). is that the intent here?

apps/microsoft-teams/frontend/src/constants/configCopy.ts
```diff
@@ -11,10 +11,20 @@ const appDeepLink = 'https://teams.microsoft.com/';
...
+  login: 'Connect to Teams',
+  logout: 'Disconnect',
   teamsAppInfo:
     'Install the Contentful app in Microsoft Teams channels where you would like to receive notifications.  View Teams docs',
   teamsAppLink: 'View Teams docs',
+  disconnectModal: {
+    confirmDisonnect: 'Disconnect',
+    goBack: 'Go back',
+    description:
+      'By disconnecting, you will no longer see configured notifications for this tenant. Are you sure you want to disconnect?',
```
lilbitner, line 26: this is a great call to warn the users 👍

apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
```diff
@@ -1,57 +1,128 @@
...
+        await customApi.saveConfiguration({ ...parameters, tenantId: authResult.tenantId });
+        dispatch({
+          type: actions.UPDATE_TENANT_ID,
+          payload: authResult.tenantId,
+        });
+      }
+    } catch (e) {
+      const message = e instanceof Error ? e.message : 'Failed to authenticate with Microsoft';
+      sdk.notifier.error(message);
+      console.log(e);
```
lilbitner, line 49: nit: do we maybe want this to be a `console.error`?

apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
```diff
@@ -1,57 +1,128 @@
...
+  const accessComponent = () => {
+    // Display if not authorized and if logout is not in progress
+    if (!accounts.length && inProgress !== 'logout') {
+      return (
+        <>
+          <Paragraph>{description}</Paragraph>
+          <Box>
+            <Button
+              onClick={() => handleLogin()}
+              isDisabled={inProgress === 'login'}
```
lilbitner, line 81: nit: could we maybe pull out the 'inProgress === 'login' into a variable (like `loginInProgress` or something) that is descriptive and can be referenced where it is used throughout here? sometimes I find that helps with readability and staying DRY! Same for the logout as well?

apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
```diff
@@ -1,57 +1,128 @@
...
+          </Box>
+        </>
+      );
+    } else {
+      return (
+        <>
+          <Card padding="large">
+            <Flex justifyContent="space-between">
+              <Flex flexDirection="column">
+                <Subheading marginBottom="none">{accounts[0]?.name}</Subheading>
```
lilbitner, line 94: just my curiosity, is the accounts type just always an array, which is why we have to index it here?

lilbitner (approved):
Its looking and feeling really great so far, way to crush this Lisa 🚀 !!! Just have some superrrr nit comments, otherwise looking great to me!

apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.tsx
whitelisab: Yes, it did cause the weird "0" issue, which is why I changed it from `isAppInstalled && accounts.length && (<NotificationsSection .... />)` to what I have here 😄

apps/microsoft-teams/frontend/src/components/config/DisconnectModal/DisconnectModal.spec.tsx
whitelisab: Good question! I am also still getting comfortable with vitest, but using `toBeTruthy` is something that we used in other apps with vitest to assert that the node exists. If I try to use `toBeVisible` here, it says `Property 'toBeVisible' does not exist on type 'Assertion<HTMLElement>'`. I wonder if there is a a library we can import or a way to set up vitest to have more assertions that we can use. Something to look into in the future!

apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
whitelisab: Yup, good call to keep it DRY! 😄

apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
whitelisab: Yes, accounts comes from the `useMsal` hook and the type is `AccountInfo[]`. So it's an array of authenticated accounts with info like name, username, tenantid, etc. https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/src/MsalContext.ts#L18

PR #6247, opened 2024-02-02
fix: silence excess output

jsdalton:
## Purpose

In the recent prettier fix we introduced a problem.

When we switched from calling prettier directly to calling it indirectly via `npm run` we failed to realize that npm run sends data to STDOUT which messed with the logic of our script here: https://github.com/contentful/apps/blob/master/.circleci/prettier-check.sh#L18

Basically our script expects NO OUTPUT when the prettier check is okay.

## Approach

* add the `--silent` flag to remove echoing of the command


whitelisab (approved):
🌟

PR #6248, opened 2024-02-02
chore: update sentry app name to support consistency

lilbitner:
## Purpose

The purpose of this PR is to update the app name that will be passed on all Sentry exceptions, to be consistent with the backend implementation. 

Perhaps eventually this can be a constant, for all apps, in the frontend-toolkit AND a constant in the future node-toolkit. For now, I have noted how these should be consistent in [this](https://contentful.atlassian.net/wiki/spaces/ECO/pages/4479189315/WIP+Sentry+for+Integrations) WIP documentation.


rusticpenguin (approved):
Looks good!

PR #6250, opened 2024-02-02
chore: remove ms teams sample bot [INTEG-1763]

whitelisab:
## Purpose

The sample bot was a helpful guide during earlier stages of development for the MS Teams app, we don't need it anymore!

## Approach

🧹


jsdalton (approved):
Bye bye! 👋

PR #6263, opened 2024-02-05
feat: improve logout flow for MS Teams [INTEG-1785]

whitelisab:
## Purpose

Ideally we don't want to rely on a logout flow from MS where you have to select which account you are logging out of if you want to switch accounts on the MS Teams app config page. After digging around the MSAL documentation, I same across this [promptless logout](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/e7a55511680aec602310f63db0bb5b7d2b07fab3/lib/msal-browser/docs/logout.md#promptless-logout)

## Approach

Implemented promptless logout. This required adding the `login_hint` optional claim for our app in the MS Entra admin center. Then, I passed in the logged in account to the `logoutPopup` method in order to create a logout experience where you don't need to select an account. You still do see a popup, but I could not find a away to work around this.

In addition, I ensured that the parameters were being updated on logout and when you change accounts. These are not being persisted in the installation parameters; saving the updated tenant id still requires the user to click "Save" at the moment, until we figure out next steps with triggering save from the app config page.

Promptless logout experience:
Screenshot 2024-02-05 at 12 25 51 PM

## Testing steps

Tested logging in and out of two accounts locally.

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/e7a55511680aec602310f63db0bb5b7d2b07fab3/lib/msal-browser/docs/logout.md#promptless-logout

### DIFF
--- a/apps/microsoft-teams/frontend/package.json
+++ b/apps/microsoft-teams/frontend/package.json
@@ -19,6 +19,7 @@
   },
   "scripts": {
     "start": "vite",
+    "dev": "vite --force",
     "build": "tsc && vite build",
     "lint": "eslint . --max-warnings 0",
     "test": "vitest --logHeapUsage --coverage",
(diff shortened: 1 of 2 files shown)


matthew-gordon (approved):
🚀

PR #6265, opened 2024-02-05
feat: [INTEG-1692] - Add status under token field

matthew-gordon:
## Purpose

We want a naive way to verify the status of a vercel access token, this adds that implmentation. This is intentionally very simple, as we flush out technical implementations the network call can be adjusted and better seperated.

Ticket: https://contentful.atlassian.net/browse/INTEG-1692 

## Testing steps

1.) Uninstall the app in development if installed, and re-install the custom app
Screenshot 2024-02-05 at 1 37 10 PM

2.) Confirm error is thrown when trying to install without a token
Screenshot 2024-02-05 at 1 38 34 PM

3.) Follow the detailed instructions for creating a vercel token, and attempt to install after token has been created

NOTE: Token cannot be accessed again after creation
Screenshot 2024-02-05 at 1 39 13 PM

4.) Malform the saved vercel token, and verify error status is displayed

Screenshot 2024-02-05 at 1 44 00 PM

### DIFF
--- a/apps/vercel/src/locations/ConfigScreen.tsx
+++ b/apps/vercel/src/locations/ConfigScreen.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState, useEffect, ChangeEvent } from 'react';
+import React, { useCallback, useState, useEffect, useReducer, ChangeEvent } from 'react';
 import { ConfigAppSDK } from '@contentful/app-sdk';
 import {
   Box,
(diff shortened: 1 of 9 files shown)


apps/vercel/src/locations/ConfigScreen.tsx
```diff
@@ -103,6 +140,22 @@ const ConfigScreen = () => {
                 to create an access token in the Vercel dashboard.
               </HelpText>
             </FormControl>
+            <Box style={{ width: '100%' }}>
```
lilbitner, line 143: nit: I know a lot of things are in progress, but perhaps these inline styles can be in separate style files in the future?

apps/vercel/src/locations/ConfigScreen.tsx
```diff
@@ -103,6 +140,22 @@ const ConfigScreen = () => {
...
+                  Status
+                </Text>
+                <Box>
+                  {appInstalled && parameters.vercelAccessToken && tokenValid ? (
+                    <Badge variant="positive">Valid access token</Badge>
+                  ) : tokenError ? (
+                    <Badge variant="negative">Invalid access token</Badge>
+                  ) : (
+                    <Badge variant="warning">Token not configured</Badge>
+                  )}
```
lilbitner, line 155: nit: Could this maybe be abstracted into an if/else or switch statement? Just a person preference but I find many conditionals more readable that way!

lilbitner (commented):
Looking great! 

There is one workflow I tried out, and its placing an invalid token and clicking install. For me it presented the "Successfully installed" message at the bottom, but then displays the "Invalid access token" message. Is that expected as of right now? Also could be a user error 😅

matthew-gordon:
> Looking great!
> 
> There is one workflow I tried out, and its placing an invalid token and clicking install. For me it presented the "Successfully installed" message at the bottom, but then displays the "Invalid access token" message. Is that expected as of right now? Also could be a user error 😅

Great call out, technically it beats the installation logic with any value pasted into the input, but presents the correct status badge after it fires the deployments call. I didn't add the network call on installation because I'm not sure that we are keeping this authentication implementation for the token, and the interface is still somewhat a WIP. Something I can bring up with the group at large for sure though 🤔

lilbitner (approved):
💯

PR #6266, opened 2024-02-06
fix: handle linebreaks in stream parser correctly

jsdalton:
## Purpose

There's a bug in AICG where a customer is reporting a parse error in the console.

This is reproducible. The error is

```
useAI.tsx?t=1707179108609:72 SyntaxError: Unexpected token 'd', "data: {"id"... is not valid JSON
    at JSON.parse (<anonymous>)
    at streamToParsedText (aiHelpers.ts?t=1707178867600:21:27)
    at Array.reduce (<anonymous>)
    at AI.parseStream (index.ts?t=1707178867600:57:28)
    at async generateMessage (useAI.tsx?t=1707179108609:64:30)
    at async handleGenerate (CommonGenerator.tsx?t=1707179108609:83:7)
```

The problem seems to trace back to the `streamToParsedText` function, which takes the raw string from the stream, parses it as JSON, and pulls out the content from the parsed line.

The stream lines are actually coming back in two forms:

Without an initial "line break":

```
data: {"id":"chatcmpl-8p3eILXh3j1n0SKA7XEXosKP2ndx…ent":","},"logprobs":null,"finish_reason":null}]}'}
```

With an initial "line break":

```
\ndata: {"id":"chatcmpl-8p3eILXh3j1n0SKA7XEXosKP2nd…ent":","},"logprobs":null,"finish_reason":null}]}
```

The problem is that sometimes the "data:" portion is not being stripped out, and thus invalid JSON is being passed to the parser.

## Approach

Upon closer examination, the bug is a small one in the regex: `/(\n)?^data:\s*/`. The "line start" is appearing in front of data, which appears like it should match correctly when there's a line break at the front. But this is not an _actual_ line break, instead it's just the control character. So when the `\n` is present, the regex is not matching, and the string is not getting removed, breaking the JSON.

The solution is just to move the `^` character to the front of the regex which now matches the data correctly.

### DIFF
--- a/apps/ai-content-generator/src/utils/aiApi/aiHelpers.ts
+++ b/apps/ai-content-generator/src/utils/aiApi/aiHelpers.ts
@@ -50,9 +50,6 @@ const isEmptyOrDone = (line: string) => line === '' || line === '[DONE]';
  * @returns
  */
 const streamToParsedText = (acc: string, line: string) => {
-  console.log('hi');
-  console.log({ acc });
-  console.log({ line });
   const dataPrepend = /(\n)?data:\s*/;
   const textWithoutDataPrepend = line.replace(dataPrepend, '').trim();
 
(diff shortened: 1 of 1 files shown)


rusticpenguin (approved):
Thanks for catching this!

PR #6330, opened 2024-02-08
feat: add notifiers when config page should be saved [INTEG-1802]

whitelisab:
## Purpose

In order to ensure that space admins are saving their changes after creating/editing/deleting notifications, we want to give a user a warning that they need to save the config page in order for their changes to persist. Note that this is an interim solution while we work on triggering a save from the app config page.

## Approach

Per our designs, after a user creates, edits, or deletes a notification, they will see a confirmation that notifications are updated, followed by a warning that they need to save the app config page.

Screenshot 2024-02-08 at 8 24 17 AM

I also implemented this for the access section/tenant id section:

Screenshot 2024-02-08 at 8 38 44 AM

## Testing steps

In our development space, make changes to the notifications section and authenticate to different tenants to see the notifications.


matthew-gordon (approved):
🌮 ✨ LGTM

PR #6355, opened 2024-02-12
ContentTypeSelectionModal & ChannelSelectionModalProps - Clicking entire row should select the option.

ethan-ozelius-contentful:
## Purpose
When selecting a Channel or Content Type from the `<ContentTypeSelectionModal>` or `<ChannelSelectionModalProps>`, clicking anywhere inside the table's row should select that option.

channels

### Context
This is from a Mob QA feedback session demo'ing MS-teams.

## Approach
Instead of adding the `onClick` callback to the `<Radio>`, add it to the `<Table.Row>`.  React will complain that you did not provide a `onChange` handler to the `<Radio>`, so just add a noop to placate react.  If there is a better solution, I'm all ears :)

## Testing steps
Tested in "Apps - Development and Testing Org", "Microsoft Teams (development)", master env.

## Breaking Changes
None

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.styles.ts
+++ b/apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.styles.ts
@@ -3,8 +3,12 @@ import tokens from '@contentful/f36-tokens';
 
 export const styles = {
   table: css({
-    borderRadius: 'none !important',
+    borderRadius: '0 !important',
     boxShadow: 'none !important',
     borderBottom: `1px solid ${tokens.gray200}`,
+  }),
+
+  tableRow: css({
+    cursor: 'pointer',
   }),
 };
--- a/apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.spec.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.spec.tsx
@@ -1,8 +1,9 @@
 import ChannelSelectionModal from './ChannelSelectionModal';
 import { describe, expect, it, vi } from 'vitest';
-import { render, screen } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import { channelSelection } from '@constants/configCopy';
 import { defaultNotification } from '@constants/defaultParams';
+import { mockChannels } from '@test/mocks/mockChannels';
 
 describe('ChannelSelectionModal component', () => {
   it('mounts and renders the empty state when no channels are present', () => {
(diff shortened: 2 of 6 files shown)


apps/microsoft-teams/frontend/test/mocks/mockChannels.ts
```diff
@@ -24,7 +24,7 @@ const mockChannels: TeamsChannel[] = [
   },
   {
     id: '19:3bccfda604454e63bd839399e6752ba3@thread.tacv2',
-    name: 'General',
+    name: 'Sales General',
```
ethan-ozelius-contentful, line 27: Add Sales to make each item of this mock unique.  This helps when using react testing lib `getBy...`

apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.styles.ts
```diff
@@ -3,8 +3,12 @@ import tokens from '@contentful/f36-tokens';
 
 export const styles = {
   table: css({
-    borderRadius: 'none !important',
+    borderRadius: '0 !important',
     boxShadow: 'none !important',
     borderBottom: `1px solid ${tokens.gray200}`,
   }),
+
+  tableRow: css({
```
whitelisab, line 11: Good call for adding this! We should also ensure that the cursor is a pointer for the `ContentTypeSelectionModal`, since I don't see the styles file for that component updated.

apps/microsoft-teams/frontend/test/mocks/mockChannels.ts
whitelisab: Makes sense! However there is a reason we have this named `General`. In Teams, each Team must have at least 1 channel, and that channel's name is always `General`. So in this case the `Sales Team` Team needs to have a channel named `General`. I know this is only a mock, but just wanted to explain why we have it this way!

whitelisab (approved):
Awesome! 🔥 Approved, just two minor comments to consider

apps/microsoft-teams/frontend/test/mocks/mockChannels.ts
ethan-ozelius-contentful: Gotcha, that makes sense.  With that in mind, I reverted the changes to this file and updated the unit test that needed it to use `teamName` instead of name.

Thanks, good callout.

apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.styles.ts
ethan-ozelius-contentful: Good call, I have updated `ContentTypeSelectionModal.styles.ts` and added the `classNames={styles.tableRow}` to `ContentTypeSelectionModal.tsx`

PR #6381, opened 2024-02-14
feat: display MS org name logo [INTEG-1782]

whitelisab:
## Purpose

This PR updates the access section of the MS Teams app config page to display the organization name and logo after a user authenticates with Microsoft. 

Screenshot 2024-02-14 at 9 38 06 AM

## Approach

While the tenant Id and the username can be accessed from the MSAL hook, we need to call the MS Graph API in order to get the organization name and logo. I created a new `msGraphApi` util class to work with the API. This class handles getting the access token for the API request, calling the correct endpoints, and error handling. 

If there is an error getting the organization name and logo, these items won't display and a notification will show saying that there is error getting organization details

Screenshot 2024-02-14 at 8 45 38 AM

In order to create tests for `msGraphApi`, I utilized MSW to mock the HTTP requests. 

## Testing steps

Navigate to the MS Teams development space and authenticate to our sandbox account to see the org name and logo.

See my notes in confluence for more information about working with the MS Graph API.

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.styles.ts
+++ b/apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.styles.ts
@@ -7,4 +7,8 @@ export const styles = {
     width: 16px;
   }
 `),
+  orgLogo: css({
+    /** Setting size to default size for favicon in MS organizational branding */
+    height: 32,
+  }),
 };
--- a/apps/microsoft-teams/frontend/src/utils/msGraphApi.ts
+++ b/apps/microsoft-teams/frontend/src/utils/msGraphApi.ts
@@ -15,7 +15,7 @@ export class MsGraphApiError extends Error {
 
   constructor(res: MsGraphApiErrorType) {
     super(res.message ?? '');
-    this.status = res.status ?? 0;
+    this.status = res.status;
   }
 }
 
@@ -86,7 +86,7 @@ class MsGraph {
 
     const options = {
       method: 'GET',
-      headers: headers,
+      headers,
     };
 
     return options;
--- a/apps/microsoft-teams/frontend/src/utils/msGraphApi.ts
+++ b/apps/microsoft-teams/frontend/src/utils/msGraphApi.ts
@@ -1,0 +1,180 @@
+import {
+  AccountInfo,
+  IPublicClientApplication,
+  InteractionRequiredAuthError,
+} from '@azure/msal-browser';
+import { graphConfig } from '@configs/authConfig';
+
+export interface MsGraphApiErrorType {
+  message: string;
+  status: number;
+}
+
+export class MsGraphApiError extends Error {
+  status: number;
+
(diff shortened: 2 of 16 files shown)


apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.styles.ts
```diff
@@ -7,4 +7,8 @@ export const styles = {
     width: 16px;
   }
 `),
+  orgLogo: css({
+    /** Setting size to default size for favicon in MS organizational branding */
+    height: 32,
```
matthew-gordon, line 12: this is more of a general question, but curious if it makes sense for us to start using tokens for sizing values? Can't remember where I saw that suggested recently, but what are your thoughts?

matthew-gordon (approved):
This looks good! Thank you for walking through it yesterday!

apps/microsoft-teams/frontend/src/utils/msGraphApi.ts
```diff
@@ -0,0 +1,164 @@
...
+  readonly tenantId: string;
+
+  constructor(instance: IPublicClientApplication, account: AccountInfo) {
+    this.baseUrl = graphConfig.graphOrgEndpoint;
+    this.instance = instance;
+    this.account = account;
+    this.tenantId = account.tenantId;
+  }
+
+  /**
```
ethan-ozelius-contentful, line 48: love to see the jsdocs!

apps/microsoft-teams/frontend/src/utils/msGraphApi.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private validateResponseStatus = (response: Response, responseJson: any) => {
+    if (response.status >= 400) {
+      const apiErrorResponse = {
+        status: response.status,
+        message: responseJson?.error?.message,
+      };
+
+      throw new MsGraphApiError(apiErrorResponse);
```
ethan-ozelius-contentful, line 103: Does this get caught anywhere?  If not, is there a risk of causing a runtime error that might make MS teams app unusable?

ethan-ozelius-contentful (approved):
Nice work!

apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.styles.ts
whitelisab: Yes, I've definitely used the [Forma36 tokens](https://f36.contentful.com/tokens/color-system), and we should try to use those whenever possible. They only have spacing values or sizing on specific Forma components, so I'm not sure it's the best use case here unless I'm missing something!

apps/microsoft-teams/frontend/src/utils/msGraphApi.ts
whitelisab: Yes, it was a great suggestion from you for us to use this more!

PR #6382, opened 2024-02-14
Clarify update/cancel duplicate notification instructions for MS Teams

ethan-ozelius-contentful:
## Purpose
Update the copy of `<DuplicateModal>` to better explain to a user their options when they are attempting to make changes to an existing notification.

1. Confirm option => "If yes, edits to entry actions will apply to your existing notification."
2. Cancel option => "If you prefer to cancel, you can then edit the content type or Teams channel."

### New
Screenshot 2024-02-14 at 11 40 16 AM

### Old
Screenshot 2024-02-14 at 11 41 42 AM

This is part of the Mob QA feedback for MS Teams.


whitelisab (approved):
🚢

PR #6399, opened 2024-02-15
fix: example package versions and add dependabot [MONET-1825]

cemreyuksel:
Functions weren't being being shown on `user_interface`. This PR 

- updates the packages 
- adds dependabot so that we prevent this bug happening again in the future.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -129,6 +129,96 @@ updates:
       versions:
         - ">=11"
 
+- package-ecosystem: npm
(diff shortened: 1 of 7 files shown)


ghepting:
Can you provide more context on what the issue with emotion latest is/was?

cemreyuksel:
@ghepting we realized none of the examples with functions were working because the packages didn't pick the updates from the `@contentful/app-scripts`. Every time there is a dependency updates, the examples are not automatically updated and this might cause similar issues in the future.

PR #6406, opened 2024-02-16
Refactor(MS-Teams) - Lift "notification index to edit" state from <NotificationSection> to <ConfigPage>

ethan-ozelius-contentful:
### No Functional change, just refactor

Lifting the state up <NotificationSection> up to <ConfigPage> will enable the <ConfigPage> to be run validations on a currently open (in edit mode) notification.

## Purpose
There is an [existing concept in MS Teams of whether or not a notification is being edited](https://github.com/contentful/apps/blob/master/apps/microsoft-teams/frontend/src/components/config/NotificationsSection/NotificationsSection.tsx#L29).  As part of the effort to improve validations for MS-Teams, we want to check to see if there are any "pending changes" when a user attempts to save by clicking the save button in the top right.  In order to detect any pending changes from the `<ConfigPage>`, the `notificationIndexToEdit` state needs to be lifted up one level.

## Screen recording of a notification being put into "edit" mode
editting

## Testing steps

Tested manually and with unit tests.

## Breaking Changes

None


rusticpenguin (approved):
Looks good! Makes sense to have this living at the root for your follow on work.

PR #6407, opened 2024-02-16
MS Teams - validate "Pending Changes" when User clicks top-right (widgetRenderer?) save.

ethan-ozelius-contentful:
## Purpose
Add a validation to MS Teams `onConfigure()` method that checks to see if there are any "pending changes".  If there are, inform the user via modal that they need to cancel/save their changes and return false (do no update configuration).  

### Pending Changes detected, configuration save blocked
The User has pending changes because they are in the process of editing a notification, but have not saved their work.
validate

### Pending Change detected, pending changes canceled, configuration save successful.
new_validate

## Approach
Use the existing concept of "pending changes", see `notificationIndexToEdit`.  If so launch a modal and return false.

## Testing steps
1. Launch MS Teams FE locally and navigate to [MS Teams development space](https://app.contentful.com/spaces/zgiu00cobrsj/apps/4BaHsPHryAsKkDuNFdKRVM).  Invoke pending changes by editing an existing notification, or creating a new one.  Even if the notification is valid, if it is not saved we should inform the user and require them to save/cancel.

## Breaking Changes
n/a

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.spec.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.spec.tsx
@@ -1,11 +1,15 @@
 import ConfigPage from './ConfigPage';
 import { describe, expect, it, vi } from 'vitest';
-import { render, screen } from '@testing-library/react';
+import { render, screen, waitFor, fireEvent } from '@testing-library/react';
 import { mockSdk } from '@test/mocks';
-import { headerSection } from '@constants/configCopy';
+import { headerSection, notificationsSection } from '@constants/configCopy';
 
 vi.mock('@contentful/react-apps-toolkit', () => ({
   useSDK: () => mockSdk,
+}));
+
+vi.mock('@azure/msal-react', () => ({
+  useMsal: vi.fn().mockReturnValue({ accounts: [{ asdf: 'asdf' }] }),
 }));
 
 vi.mock('@components/config/AccessSection/AccessSection', () => ({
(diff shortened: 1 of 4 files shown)


apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.spec.tsx
```diff
@@ -1,29 +1,82 @@
...
+
+    unmount();
+  });
+
+  describe('Saving via WidgetRenderer', () => {
+    it('prompts the User to save or cancel their pending changes when they have an open Notification modal', async () => {
+      const { unmount } = render(<ConfigPage />);
+
+      // Open an existing notification to edit it, i.e. pending change
+      let editButton;
```
jsdalton, line 56: i am not a vitest expert but per https://vitest.dev/api/vi#vi-waitfor-0-34-5 couldn't we restructure this a bit to just return the edit button from the function?

```
const editButton = await waitFor(() => {
        // waitFor async callbacks to resolve inside useEffects
        return screen.getByTestId('edit-notification-btn');
      });
      
  if (editButton) // ...
```

apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.spec.tsx
```diff
@@ -1,29 +1,82 @@
...
+      await waitFor(() => {
+        // waitFor async callbacks to resolve inside useEffects
+        editButton = screen.getByTestId('edit-notification-btn');
+        return !!editButton;
```
jsdalton, line 60: My suggestion above stemmed from not understanding why we need to return a boolean here. Just wasn't sure what that was needed. 😄

jsdalton (approved):
Looks good to me! 🏅 One small comment around some code that looked it could be cleaned up in the tests. And then also yeah whatever changes make sense on the confirm copy.

apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.spec.tsx
rusticpenguin: +1 to this solution. Keeping everything `const` is always nice

apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.spec.tsx
ethan-ozelius-contentful: To be entirely honest, I'm not sure why I was returning a boolean here, Friday was a long time ago.

I think I was attempting to placate typescript, because TS was concerned that the button was possibly undefined.

**Previous implementation**
Screenshot 2024-02-20 at 11 44 48 AM

I have implemented the above suggested code pattern, which is much cleaner and makes TS happy :) (which is what life is all about).

apps/microsoft-teams/frontend/src/components/config/ConfigPage/ConfigPage.spec.tsx
ethan-ozelius-contentful: Nice, this is much better!  

- [ ] implement this.

PR #6444, opened 2024-02-20
feat: save org details as params from auth flow [INTEG-1818]

whitelisab:
## Purpose

In order to prevent a user from having to regularly authenticate to Microsoft on the app config page, this PR updates the app installation parameters so that the org details (name, logo, and username) are persisted. 

## Approach

Persist additional information from Microsoft in the installation parameters. Also updated error handling for the Auth flow to align with designs. 

Here's the error a user sees if they are unable to authenticate to Microsoft
Screenshot 2024-02-20 at 9 42 51 AM

Here's the error a user sees if they cannot get the organization details from Microsoft
Screenshot 2024-02-19 at 12 43 42 PM

Here's an updated version of what successful authentication looks like (removed username)
Screenshot 2024-02-20 at 9 40 16 AM

## Testing steps

Run the MS Teams frontend app locally in our testing environment

## Breaking Changes

This PR introduces new installation parameters, so an app might need to be reinstalled to see the full changes.

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
@@ -1,13 +1,5 @@
-import { Dispatch, useEffect, useState } from 'react';
-import {
-  Box,
-  Button,
-  Card,
-  Flex,
-  ModalLauncher,
-  Paragraph,
-  Subheading,
-} from '@contentful/f36-components';
+import { Dispatch } from 'react';
+import { Box, Button, Flex, ModalLauncher, Paragraph } from '@contentful/f36-components';
 import { useMsal } from '@azure/msal-react';
 import { loginRequest } from '@configs/authConfig';
 import { accessSection } from '@constants/configCopy';
@@ -22,6 +14,8 @@ import { ConfigAppSDK } from '@contentful/app-sdk';
 import DisconnectModal from '@components/config/DisconnectModal/DisconnectModal';
 import { displayConfirmationNotifications } from '@helpers/configHelpers';
 import MsGraph from '@utils/msGraphApi';
+import { AccountInfo } from '@azure/msal-browser';
+import AccessSectionCard from '@components/config/AccessSectionCard/AccessSectionCard';
 
 interface Props {
   dispatch: Dispatch<ParameterAction>;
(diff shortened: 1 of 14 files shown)


matthew-gordon (approved):
This LGTM! I will ping you about auth specific questions to consider for future us!

apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
```diff
@@ -87,93 +88,84 @@ const AccessSection = (props: Props) => {
...
+              payload: msAccountInfo,
             });
           }}
         />
       );
     });
   };
 
-  const getOrgDetails = async () => {
+  const getOrgDetails = async (account: AccountInfo): Promise<typeof defaultOrgDetails> => {
```
ethan-ozelius-contentful, line 110: I noticed in the previous implementation, inside a `useEffect`, we were checking the validity of `accounts` and `tenantId`.  I'm assuming that check is no longer necessary?

```js
if (accounts.length && parameters.tenantId) {
```

Probably because this would throw an error, if tenantId and account were invalid?  Is that a true statement?

```js
const authResult = await instance.loginPopup(loginRequest);
const { tenantId, account } = authResult;
```

apps/microsoft-teams/frontend/src/components/config/AccessSectionCard/AccessSectionCard.spec.tsx
```diff
@@ -0,0 +1,64 @@
...
+        handleLogout={vi.fn()}
+      />
+    );
+
+    expect(screen.getByText(mockParameters.orgName)).toBeTruthy();
+    expect(screen.getByAltText('logo')).toBeTruthy();
+    unmount();
+  });
+
+  it('displays correct copy when unauthorized', () => {
```
ethan-ozelius-contentful, line 24: Should this be ... when ~~un~~ authorized

ethan-ozelius-contentful (approved):
Left a non-blocking question, and a non-blocking possible copy/paste typo.  Otherwise looks good, nice work!

apps/microsoft-teams/frontend/src/components/config/AccessSectionCard/AccessSectionCard.spec.tsx
whitelisab: Yes thanks for the catch! Should be `displays correct copy when authorized with no org details`

PR #6500, opened 2024-02-21
feat: disable test message and update test errors [INTEG-1812]

whitelisab:
## Purpose

This PR ensures that the `Test Message` button is disabled when a notification is incomplete. Also, it updates the error handling for sending the message to ensure that an error message is displayed.

## Approach

`Test Message` button should be disabled for incomplete notifications. The error handling for the test message was built on a previous implementation where the body had an errors array, this is now updated to be a single error object. Note that there is additional work that needs to be done in another PR in the bot service to display the proper error message, currently the error message says `malformed MsTeamsBotServiceResponse returned from MsTeamsBotService API`.

## Testing steps

Deployed to sandbox and sent a test to a deleted channel to see an error message.


rusticpenguin (approved):
Looks good!!

PR #6513, opened 2024-02-22
feat: update app save behavior for auth [INTEG-1804]

whitelisab:
## Purpose

This PR addresses a few items:

- Some small copy changes
- Updating the app config save when auth information changes 
- Prevent a message from being sent if the tenant id doesn't match

## Approach

Here is the updated approach for the auth flow:

- On initial install if auth is successful, the the app config page will be saved with the tenant id and installed
- If org details are not found and the `Retry Auth` button is pressed and org details are successful, save the app
- If the `Disconnect` button is clicked and the user confirms, then all parameters are reset and config is saved

## Testing steps

Test the auth flow and ensure that the app config page and parameters are saved.

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
@@ -1,4 +1,4 @@
-import { Dispatch } from 'react';
+import { Dispatch, useEffect, useState } from 'react';
 import { Box, Button, Flex, ModalLauncher, Paragraph } from '@contentful/f36-components';
 import { useMsal } from '@azure/msal-react';
 import { loginRequest } from '@configs/authConfig';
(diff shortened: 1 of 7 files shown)


apps/microsoft-teams/frontend/src/components/config/AccessSection/AccessSection.tsx
```diff
@@ -39,10 +44,56 @@ const AccessSection = (props: Props) => {
...
+  // saves the app config page after dispatch completes and parameters are updated
+  const saveMsAccountInfo = async () => {
+    await customApi.saveConfiguration();
+    setHasUpdatedMsAuth(false);
+    setHasUpdatedOrgDetails(false);
+    setHasDisconnected(false);
+    setUpdatedMsAccountInfo(parameters);
+  };
+
+  useEffect(() => {
```
whitelisab, line 56: My intention here was to ensure that we were calling `saveConfiguration` _after_ the new parameter data had been dispatched. This is because `saveConfiguration` uses what's in parameters to save, rather than us passing in the new parameters to the `saveConfiguration` call. Definitely open to feedback on this if there is a better way!

rusticpenguin (approved):
Good copy changes! And looks clean 👍

PR #6531, opened 2024-02-23
feat: Improve ColorPicker Field

Toxicom:
## Summary
   - Replace deprecated `FieldExtensionSDK`;
   - Replace deprecated `AppExtensionSDK`;
   - Display only validated colors in a validated field;
   - Show hex value in dropdown.

## Purpose

The purpose is to improve UX for the content editor by simplifying the color picker field.

## Approach

I have also considered exposing a boolean in the configuration screen to validate against the label (`white` instead of `#ffffff`) but that would introduce breaking changes. I am thinking about supporting both and converting label to value behind the scenes, but I will leave this for a future PR.

## Testing steps

1. Create two text fields with the Color Picker appearance
2. In one of the fields, add validation for some of your theme values
3. Confirm that in the validated field's dropdown you only see the validated items available for selection

## Breaking Changes

App should still function as previously; if there is no validation present, `validatedColors` return the `theme.colors` array unchanged.

N/A

## Deployment

N/A

### Unvalidated Color Picker field
image

### Validated Color Picker field
image


whitelisab:
Hi @Toxicom! Thanks for opening this PR with this enhancement for the color picker app. We are going to review internally with our designer and will hopefully get back to you by the end of next week with some feedback.

Toxicom:
Hi @whitelisab thanks for responding!

> if you already have theme colors, but you want only a subset of those theme colors to be available for a content type, is that correct?

Yes that's correct. Say I have my whole color design system themed out in the Color Picker app as multiple colors with different shades ([a tonal palette](https://m3.material.io/styles/color/system/how-the-system-works#094adbe5-d41e-49b4-8dff-906d6094668d)), so `blue-100` to `blue-900`, `orange-100` to `orange-900` etc., but I only want for that specific content type to be able to select only a few of them, like `blue-400` and `orange-500` only.

> Additionally I did some testing to ensure that the app still worked as expected in these different situations:
> 
>     * There are no theme colors configured and no validation
> 
>     * There are theme colors and no validation
> 
>     * There is validation and no theme colors
> 
>     * There are additional validations on the field besides just "Accept only specified values"

Thank you for testing, as I understand the app still works as expected? :)

> While the app can be used for both JSON object and short text field types, this feature will only work for short text field types

That is correct, sadly there is no validation available for a JSON field type.

> If you set "Allow custom value" to Yes when editing the field appearance, it kind of defeats the purpose of this because then it shows the specified values in the validation and the custom value picker, so it's important that it is set to No

Correct, ideally a developer would set up the color picker field. I also realized you can't correctly publish a custom value for a validated field, so it's counter-intuitive for it to show up there. I have made changes so on validated fields the custom color option does not show up.

> We will work to update our documentation after, so that's why I just wanted to be clear on all of this.

Sounds good!

> Finally, I am requesting that you make a small change to update the spacing between the color name and hex value for the menu items (see PR comment). Thank you and let me know if you have any questions!

I've updated spacing to match what's in the `SelectColorButton` - `spacing2Xs`

whitelisab (approved):
The changes look good and I've validated that the app works as expected in the different scenarios I outlined, so I'm approving this PR.

Now we need to work on the build failures. I tried rerunning the checks and they're still failing. @Toxicom can you please pull the latest mater and rebase? Then we can see how the builds do after that. Thank you!

Toxicom:
Oh lord my rebasing went wrong, let me try and fix that..

Toxicom:
So sorry @whitelisab I'm very new to working with forks so somehow I've messed up the rebasing. I have opened a new PR that's up to date with upstream `master` branch and has the changes I had here: https://github.com/contentful/apps/pull/6724

PR #6554, opened 2024-02-26
fix: support non-image file types in Dropbox

whitelisab:
## Purpose

This solves a customer issue where they are not able to add non-image file types to their Contentful entry using the Dropbox app.

## Approach

Remove the file type restrictions on the Dropbox picker and ensure that if there isn't a thumbnail image, the app won't break. Confirmed that a non-image file can be added to an entry and here's how it renders.

Screenshot 2024-02-26 at 3 58 03 PM

Eventually we can update this to have the same fallback images that are used in the Bynder app.

## Testing steps
Tested locally.

### DIFF
--- a/apps/dropbox/src/index.js
+++ b/apps/dropbox/src/index.js
@@ -15,7 +15,8 @@ const FIELDS_TO_PERSIST = [
 ];
 
 function makeThumbnail(file) {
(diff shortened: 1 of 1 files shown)


rusticpenguin (approved):
Looks good!

PR #6584, opened 2024-02-27
Microsoft Teams - add fuzzy search for content types

ethan-ozelius-contentful:
## Todo
- [x] Add Magnifying glass to search input to match figma designs.

## Purpose
Add the ability for Users to search for a specific content type by searching for the content type's `id, displayField, name, or description` attributes.

ms-search

## Approach
This PR introduces a client-side lightweight fuzzy search package called [fuse.js](https://www.fusejs.io/).

## Limitations
Since this is a client side search, the complete searchable set is the exact same set of content types that were fetched via `useGetContentTypes()` and exposed via `ContentTypeContext`.  

There are currently 2 known limitations: 
- No real time updates.
1. User opens the modal
2. Adds an new content type in a separate type
3. The new content type will **not** be rendered.

- 100 results limit
Contentful limits the number of results returned from the API to 100.  If the content type that the user is searching for, was not returned as part of the initial 100 results, they will not be able to find it :( .  This will be address in another PR [ticket tbd]

## Testing steps
Tested locally and with unit tests.

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
@@ -1,5 +1,13 @@
-import { useState } from 'react';
-import { Button, FormControl, Modal, Radio, Table } from '@contentful/f36-components';
+import { useState, useMemo, useCallback } from 'react';
+import {
+  Box,
+  Button,
+  FormControl,
+  IconButton,
+  Modal,
+  Radio,
+  Table,
+} from '@contentful/f36-components';
 import { contentTypeSelection } from '@constants/configCopy';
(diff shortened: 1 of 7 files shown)


apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
```diff
@@ -44,14 +72,15 @@ const ContentTypeSelectionModal = (props: Props) => {
       );
     }
 
-    if (contentTypes.length) {
```
ethan-ozelius-contentful, line 47: renamed `contentTypes` to `filteredContentTypes` since we no longer using just the exact set that is provided via `useGetContentTypes() & ContentType context`, now we want to possibly render a subset of that data.  Open to suggestions of better name than `filtered...`

PR #6609, opened 2024-02-28
MS-Teams - batch requests to fetch content types.

ethan-ozelius-contentful:
## Purpose
In order to support customers who have 100+ content types, we need to batch multiple requests (of 100 at a time, configurable) to retrieve all content types.  This also will be needed for pagination.

batches

## Approach
I chose to use a `while` loop, despite the fact that I usually despise `while` loops.  But in this case I think it makes the most sense, since we know we will have to enter a loop at least one time to get the first 100 content types, and then _if_ we have to batch request more, then we can restart the loop.  I'm definitely open to other ways of solving this.

## Testing steps
Unit testing and locally

## Follow up PR
This hook is a good candidate to be moved to the [integration-frontend-toolkit](https://github.com/contentful/integration-frontend-toolkit/tree/master).  In the name of speed/scope this PR will implement the hook directly, but we should abstract this hook to that library and also ensure that it will support `async/await` as well as `.then()` patterns.  We were lucky to already have a robust `{ loading, error }` state, but not all apps that could use this hook might want to pause program execution with `await`.  I think, this theory might need some investigation.

### DIFF
--- a/apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.spec.ts
+++ b/apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.spec.ts
@@ -3,6 +3,10 @@ import { describe, expect, it, vi } from 'vitest';
 import useGetContentTypes from './useGetContentTypes';
 import { mockCma, mockSdk, mockGetManyContentType } from '@test/mocks';
 
+vi.mock('@constants/contentTypeLimit', () => ({
+  CONTENT_TYPE_LIMIT: 1,
+}));
+
 vi.mock('@contentful/react-apps-toolkit', () => ({
   useSDK: () => mockSdk,
   useCMA: () => mockCma,
@@ -16,6 +20,8 @@ describe('useGetContentTypes', () => {
       expect(result.current.contentTypes).toEqual(mockGetManyContentType.items);
       expect(result.current.loading).toEqual(false);
     });
+
+    vi.restoreAllMocks();
   });
 
   it('should return generalized error if error is thrown', async () => {
@@ -26,5 +32,21 @@ describe('useGetContentTypes', () => {
       expect(result.current.error?.message).toEqual('Unable to get content types');
       expect(result.current.loading).toEqual(false);
     });
+
+    vi.restoreAllMocks();
+  });
+
+  it('batches requests if the number of available content types exceeds the request limit', async () => {
+    const getManyResponse = { ...mockGetManyContentType, total: 2 };
+    getManyResponse.total = 2;
+    const getManySpy = (mockSdk.cma.contentType.getMany = vi.fn().mockReturnValue(getManyResponse));
+
+    renderHook(() => useGetContentTypes());
+
+    await waitFor(() => {
+      expect(getManySpy).toHaveBeenCalledTimes(2);
+    });
+
+    vi.restoreAllMocks();
   });
 });
--- a/apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.ts
+++ b/apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.ts
@@ -2,6 +2,7 @@ import { useEffect, useCallback, useState } from 'react';
 import { useSDK } from '@contentful/react-apps-toolkit';
 import { ConfigAppSDK } from '@contentful/app-sdk';
 import { ContentTypeProps } from 'contentful-management';
+import { CONTENT_TYPE_LIMIT } from '@constants/contentTypeLimit';
 
 /**
  * This hook is used to get all the content types for the space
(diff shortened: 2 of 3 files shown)


apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.ts
```diff
@@ -18,17 +19,41 @@ const useGetContentTypes = () => {
...
+
+        const totalNumOfContentTypes = contentTypesResponse.total;
+        contentTypesSuperset.push(...contentTypesResponse.items);
+
+        if (contentTypesSuperset.length >= totalNumOfContentTypes) {
+          // we have fetched all possible content types, exit while
+          needToFetchNextBatch = false;
+        } else {
+          // we need to fetch the next batch of content types, re-execute while
+          skip += CONTENT_TYPE_LIMIT;
```
rusticpenguin, line 44: Just a quick callout, I would check to see if we're skipping the 100th index every call. 

Depending on how the skip is setup we could be getting 0 - 99, then on the next call getting 101-200, etc. Or this could be 100% fine and a non issue

ethan-ozelius-contentful: Good call out, I tested this out by setting the `CONTENT_TYPE_LIMIT` to 2 and adding/subtracting 1 on this line.
 
`skip += CONTENT_TYPE_LIMIT - 1` will fetch records `[1, 2] , [2, 3] , [3, 4]`
`skip += CONTENT_TYPE_LIMIT + 1` will fetch records `[1, 2] , [4, 5], [7]`

So we should be good with `skip += CONTENT_TYPE_LIMIT`, which I varifyed fetches records `[1,2] [3,4] [5, 6] [7]` in batches of 2, it works similarly for batches of 3..4...n

apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.spec.ts
```diff
@@ -26,5 +33,21 @@ describe('useGetContentTypes', () => {
       expect(result.current.error?.message).toEqual('Unable to get content types');
       expect(result.current.loading).toEqual(false);
     });
+
+    vi.restoreAllMocks();
+  });
+
+  it('batches requests if the number of available content types exceeds the request limit', async () => {
+    const getManyResponse = cloneDeep(mockGetManyContentType);
```
jsdalton, line 41: Small thing but these days I think cloning via es6 destructuring is just as easy as lodash and fewer unnecessary external dependencies.

```js
const getManyResponse = { ...mockGetManyContentType }
```

apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.spec.ts
```diff
@@ -26,5 +33,21 @@ describe('useGetContentTypes', () => {
...
+    getManyResponse.total = 2;
```
jsdalton, line 42: Yeah and then you can even combine this line with the above

```js
const getManyResponse = { ...mockGetManyContentType, total: 2 }
```

apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.ts
```diff
@@ -18,17 +19,41 @@ const useGetContentTypes = () => {
...
       setLoading(true);
-      // TODO: Implement pagination for content types
-      const contentTypesResponse = await sdk.cma.contentType.getMany({});
-      setAllContentTypes(contentTypesResponse.items || []);
+
+      const contentTypesSuperset: ContentTypeProps[] = [];
+      let skip = 0;
+      let needToFetchNextBatch = true;
+
+      while (needToFetchNextBatch) {
```
jsdalton, line 27: I think a while loop is quite appropriate here.

For me (subjective opinion) it cleans things up / gets rid of a tiny bit of noise to embrace explicit control flow and ditch managing the extra mutable variable. Something like:

```js
while (true) {
  // ... fetch all the things and push
  if (contentTypesSuperset.length >= totalNumOfContentTypes) break;
  skip += CONTENT_TYPE_LIMIT;
}
```

Just an optional suggestion for your consideration.

jsdalton (approved):
Looks good! 💯 Just a few (optional) suggestions

apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.ts
```diff
@@ -18,17 +19,40 @@ const useGetContentTypes = () => {
...
+
+      setAllContentTypes(contentTypesSuperset);
     } catch (error) {
       const err = new Error('Unable to get content types');
       setError(err);
       console.error(error);
     }
 
     setLoading(false);
-  }, [sdk.cma.contentType]);
```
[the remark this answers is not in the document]
ethan-ozelius-contentful: Discussed with Lisa, we're not entirely sure why the linter did/didn't complain about `sdk.cma.contentType` or `sdk.cma.contentType.getMany`... but basically this was just to placate the linter.  There's 0 risk here, since the SDK is static, so I'll just leave this as is.

- [ ] leave it be.

apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.spec.ts
ethan-ozelius-contentful: Yeah, I like destructuring for accessing arrays/objects, but with objects (and maybe array?) it's not a _true_ clone utility, in the sense that nested object properties for the cloned object, also get updated for the object that was cloned, and vice versa.

```js
const a = { abc: 1, nested: { def: "def" } }
const b = { ...a }
b.abc = 2
console.log(a)
// { abc: 1, nested: { def: "def" } }. 👍 no issues!

b.nested.def = "uh oh!!"
console.log(a)
// { abc: 1, nested: { def: "uh oh!!" } } a.nested.def was updated.
```

So when I actually want to deep (any nested props) clone, I like lodash cloneDeep or even `JSON.parse(JSON.stringify(foo))`.

apps/microsoft-teams/frontend/src/hooks/useGetContentTypes.ts
ethan-ozelius-contentful: I could definitely go either way, I slightly prefer keeping the `needToFetchNextBatch`, just because it explains what the entry condition of the while loop is in english.  But, not a deal breaker.

I'm gonna keep it for now, but I appreciate the suggestion 👍

PR #6626, opened 2024-02-29
Microsoft Teams - add fuzzy search for content types

ethan-ozelius-contentful:
## Purpose
Add the ability for Users to search for a specific content type by searching for the content type's `id, displayField, name, or description` attributes.

ms-search

### UI States
1. No content types were returned from `useGetContentTypes()`.  In this case we should show the "No content types" empty state, with the search input disabled.

Screenshot 2024-03-04 at 2 49 49 PM

2. Ka-boom, an error occurred while attempting to retrieve content types.  In this case, we should show the Error state, without the search input.

Screenshot 2024-03-04 at 2 49 49 PM

3. N content types were returned from `useGetContentTypes()` and a search query has filtered the set down to a subset (could possibly be entire set).  In this case, we should render the search input.
 
Screenshot 2024-03-04 at 2 49 49 PM

4. N content types were returned from `useGetContentTypes()` and a search query has filtered the set down to 0 results, i.e. failed search.  In this case, render 0 content types, and the search modal.

Screenshot 2024-03-04 at 2 49 49 PM

## Seeking review on
I walked through `<SearchableList>` in code huddle in particular how it is abstract and a good candidate to be added to the react-frontend toolkit.  But any suggestions on how to make it more extensible or clear would be appreciated.  I'm not completely in love with the `renderItem()` prop that defines how to render s searchable item, I think I would prefer to use `props.children`, but the code required to do that got pretty ugly pretty fast.  So I'm going with "render props" for now, but there might be some improvements to be made, idk.

## Approach
This PR introduces a client-side lightweight fuzzy search package called [fuse.js](https://www.fusejs.io/).

## Testing steps
Tested locally and with unit tests.

## Limitations
- real time updates - if a new content type is created after the config is loaded, the config page will need to be refreshed to retrieve the new content types.  This shouldn't be a deal breaker/blocker since there is a workaround, and creating new content types for a mature space is not overwhelmingly common.

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
@@ -7,7 +7,9 @@ import ModalHeader from '@components/config/ModalHeader/ModalHeader';
 import { ContentTypeProps } from 'contentful-management';
 import EmptyState from '@components/config/EmptyState/EmptyState';
 import WebApp from '@components/config/EmptyState/WebApp';
+import DebouncedSearchInput from '@components/config/DebouncedSearchInput/DebouncedSearchInput';
 import ErrorMessage from '../ErrorMessage/ErrorMessage';
+import SearchableList from '../SearchableList/SearchableList';
 
 interface Props {
   isShown: boolean;
@@ -31,9 +33,40 @@ const ContentTypeSelectionModal = (props: Props) => {
   } = props;
 
   const [selectedContentTypeId, setSelectedContentTypeId] = useState(savedContentTypeId ?? '');
+  const [searchQuery, setSearchQuery] = useState<string>('');
 
-  const { title, button, link, emptyContent, emptyHeading, errorMessage } =
+  const { title, button, link, emptyContent, emptyHeading, errorMessage, searchPlaceholder } =
     contentTypeSelection.modal;
+
+  /**
+   * The search input and the list of content types must be kept separate so that the rerendering
+   * of the list does unnecessarily re-render input (and lose focus/state). So we are maintaining
+   * search query in this component that is received from <DebouncedSearchInput> and passing it
+   * to <SearchableList>
+   */
+  const handleSearchQueryUpdate = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(e.target.value);
+  };
+
+  const renderRow = (contentType: ContentTypeProps) => {
+    return (
+      <Table.Row
+        key={contentType.sys.id}
+        onClick={() => setSelectedContentTypeId(contentType.sys.id)}
+        className={styles.tableRow}>
+        <Table.Cell>
+          <Radio
+            id={contentType.sys.id}
+            isChecked={selectedContentTypeId === contentType.sys.id}
+            onChange={() => {
+              /* clicking entire row checks this radio, i.e. do nothing here */
+            }}>
+            {contentType.name}
+          </Radio>
+        </Table.Cell>
+      </Table.Row>
+    );
+  };
 
   const renderModalContent = () => {
     if (error) {
@@ -49,25 +82,18 @@ const ContentTypeSelectionModal = (props: Props) => {
         <>
           <Modal.Content>
             <FormControl as="fieldset" marginBottom="none">
+              <DebouncedSearchInput
+                placeholder={searchPlaceholder}
+                onChange={handleSearchQueryUpdate}
+              />
               <Table className={styles.table}>
                 <Table.Body>
-                  {contentTypes.map((contentType) => (
-                    <Table.Row
-                      key={contentType.sys.id}
-                      onClick={() => setSelectedContentTypeId(contentType.sys.id)}
-                      className={styles.tableRow}>
-                      <Table.Cell>
-                        <Radio
-                          id={contentType.sys.id}
-                          isChecked={selectedContentTypeId === contentType.sys.id}
-                          onChange={() => {
-                            /* clicking entire row checks this radio, i.e. do nothing here */
-                          }}>
-                          {contentType.name}
-                        </Radio>
-                      </Table.Cell>
-                    </Table.Row>
-                  ))}
+                  <SearchableList
+                    list={contentTypes}
+                    searchQuery={searchQuery}
+                    renderListItem={renderRow}
+                    searchKeys={['sys.id', 'displayField', 'name', 'description']}
+                  />
                 </Table.Body>
               </Table>
             </FormControl>
@@ -90,6 +116,11 @@ const ContentTypeSelectionModal = (props: Props) => {
 
     return (
       <Modal.Content>
+        <DebouncedSearchInput
+          placeholder={searchPlaceholder}
+          onChange={handleSearchQueryUpdate}
+          disabled={true}
+        />
         <EmptyState
           image={<WebApp />}
           heading={emptyHeading}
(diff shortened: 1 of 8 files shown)


apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
```diff
@@ -31,10 +33,32 @@ const ContentTypeSelectionModal = (props: Props) => {
   } = props;
 
   const [selectedContentTypeId, setSelectedContentTypeId] = useState(savedContentTypeId ?? '');
-
+  const [filteredContentTypes, setFilteredContentTypes] =
+    useState<ContentTypeProps[]>(contentTypes);
   const { title, button, link, emptyContent, emptyHeading, errorMessage } =
     contentTypeSelection.modal;
 
+  const handleSearch = useCallback(async (e: React.ChangeEvent<HTMLInputElement>) => {
```
ethan-ozelius-contentful, line 41: - [ ] not async!

apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
```diff
@@ -31,10 +33,32 @@ const ContentTypeSelectionModal = (props: Props) => {
...
+    const fuseOptions = {
+      isCaseSensitive: false,
+      keys: ['sys.id', 'displayField', 'name', 'description'],
+    };
+
+    const fuse = new Fuse(filteredContentTypes, fuseOptions);
+    const searchPattern = e.target.value;
+    const fuseSearchResultObj = fuse.search(searchPattern);
+    // extract actual contentType objects from fuse search result object;
+    const extractedSearchResults = fuseSearchResultObj.map((result) => result.item);
```
ethan-ozelius-contentful, line 57: Open to suggestions for a better name than `extractedSearchResults`.  Basically the object that gets returned is a fuse object with `{....todo....}`

apps/microsoft-teams/frontend/src/components/config/ContentTypeSearch/ContentTypeSearch.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Box, IconButton, TextInput } from '@contentful/f36-components';
+import { SearchIcon } from '@contentful/f36-icons';
+import { contentTypeSelection } from '@constants/configCopy';
+import { debounce } from 'lodash';
+
+type Props = {
+  handleChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+};
+
+const ContentTypeSearch = ({ handleChange }: Props) => {
```
ethan-ozelius-contentful, line 11: Abstract this component to be used for both content type searching and teams channel searching.

- [ ]

apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
```diff
@@ -30,11 +31,15 @@ const ContentTypeSelectionModal = (props: Props) => {
...
   } = props;
 
+  const [searchQuery, setSearchQuery] = useState<string>('');
   const [selectedContentTypeId, setSelectedContentTypeId] = useState(savedContentTypeId ?? '');
-
-  const { title, button, link, emptyContent, emptyHeading, errorMessage } =
+  const { title, button, link, emptyContent, emptyHeading, errorMessage, searchPlaceholder } =
     contentTypeSelection.modal;
 
+  const handleSearchQueryUpdate = (e: React.ChangeEvent<HTMLInputElement>) => {
```
ethan-ozelius-contentful, line 39: probably worth adding a comment explaining how this component is brokering the search query between `<DeboundedSearchInput>` and `<ResultsList>` and why it is necessary: updating state of either of those 2 components should not force the other to rerender, so we needed to push some state down to child components.

apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
```diff
@@ -48,26 +81,19 @@ const ContentTypeSelectionModal = (props: Props) => {
       return (
         <>
           <Modal.Content>
+            <DebouncedSearchInput
+              placeholder={searchPlaceholder}
+              onChange={handleSearchQueryUpdate}
+            />
             <FormControl as="fieldset" marginBottom="none">
               <Table className={styles.table}>
                 <Table.Body>
```
ethan-ozelius-contentful, line 90: This is a little confusing how we define a `<Table>` wrapper and then the table rows are delegated to `<SearchableList>`... It's not the end of the world since it's defined in this component, I just wonder if passing `{ children }` would read easier 🤔

apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
```diff
@@ -48,26 +81,19 @@ const ContentTypeSelectionModal = (props: Props) => {
       return (
         <>
           <Modal.Content>
+            <DebouncedSearchInput
```
ethan-ozelius-contentful, line 84: - [ ] move input inside of `<formControl>`

apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
jsdalton: I don't have a strong opinion here. But I think it works nicely the way you've got it now and you haven't named things that couples it to a specific HTML element choice 👍

apps/microsoft-teams/frontend/src/components/config/SearchableList/SearchableList.tsx
```diff
@@ -0,0 +1,56 @@
...
+      setFilteredList(list);
+      return;
+    }
+
+    const fuseOptions = {
+      isCaseSensitive: false,
+      keys: searchKeys,
+    };
+
+    const fuse = new Fuse(list, fuseOptions);
```
jsdalton, line 32: Would it make sense to instantiate this list at the outer scope of your component function instead of each time a search is executed?

apps/microsoft-teams/frontend/src/components/config/SearchableList/SearchableList.tsx
```diff
@@ -0,0 +1,56 @@
...
+    // const searchPattern = e.target.value;
```
jsdalton, line 33: You probably meant to delete this commented out code?

jsdalton (commented):
Looks great! I had one question but I really like the solution here. Presumably it's even worth extracting to our frontend toolkit?

apps/microsoft-teams/frontend/src/components/config/SearchableList/SearchableList.tsx
ethan-ozelius-contentful: Yes, it definitely would :)

- [ ] move fuse instantiation outside of component

apps/microsoft-teams/frontend/src/components/config/SearchableList/SearchableList.tsx
ethan-ozelius-contentful: Yep, this was a carry-over from when this was more tightly coupled to a search input, as opposed to just passing `searchQuery`.  Good call.

- [ ] remove comment

apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
ethan-ozelius-contentful: Sounds good, we can always tweak the implementation later if we want but I'm happy with this solution as is.

jsdalton (approved):
Great! ⭐

PR #6628, opened 2024-03-01
fix: use jose instead of jsonwebtoken

jsdalton:
## Purpose

jsonwebtoken is not designed for nor does it work reliably in the browser. It's intended as a node library.

Newer versions of react-script require a version of webpack for which it is no longer reliable.

## Approach

* Replace with jose which is designed to work in the browser. See https://jwt.io/libraries?language=JavaScript


jsdalton:
This isn't going to work. We will need to resolve this by moving token signing to a backend process (ie HAA)

PR #6648, opened 2024-03-01
HAAs for mux

jsdalton:
## Purpose

In order to move JWT token signing to a backend context, we want to set up the Mux app to use hosted app actions.

As a preliminary step, this PR focuses on reorganizing the app to include a front end along side the app actions, with a parent project that's responsible for the building and deployment of the whole thing.

## Approach

* Front end is moved to `frontend` folder
* HAA is created in `app-actions` folder
* Root contains a manifest pointing to a single "static" HAA (`muxGetSignedUrlTokens`)
* Action is implemented to return static values
* Copied and pasted a handful of boilerplate files etc. to operationalize the app action

Note: Signed URLs do not yet work in this version. Those will be fixed in a follow up.

## Testing steps

This PR was built and deployed to a sandbox space to verify that everything is still "correct":

image

A manual test of the app action in that space shows the action itself is deployed and works as expected.

```
ACCESS_TOKEN={redacted} npm run call-app-action -- -a muxGetSignedUrlTokens -s 0h02439dt6mi

> @contentful/mux-app@1.0.0 call-app-action
> npx ts-node scripts/call-app-action.ts -a muxGetSignedUrlTokens -s 0h02439dt6mi

APP ACTION RESULT
{
  sys: {
    id: '5bsgVGKFfgpPe7jnGj0DnA',
    type: 'AppActionCallDetails',
    space: { sys: { type: 'Link', linkType: 'Space', id: '0h02439dt6mi' } },
    sys: { type: 'Link', linkType: 'Environment', id: 'master' },
    createdBy: {
      sys: {
        type: 'Link',
        linkType: 'AppAction',
        id: 'muxGetSignedUrlTokens'
      }
    },
    createdAt: '2024-03-13T14:25:34.142Z'
  },
  request: { function: 'actions/get-jwt-tokens.js', parameters: {} },
  response: {
    body: '{"ok":true,"data":{"playbackToken":"token","posterToken":"token","storyboardToken":"token"}}',
    statusCode: 200
  },
  statusCode: 200,
  responseAt: '2024-03-13T14:25:34.756Z',
  errors: [],
  eventType: 'call',
  url: 'hosted',
  requestAt: '2024-03-13T14:25:34.142Z',
  environmentId: 'master'
}
Result: success!

```


apps/mux/build-actions.js
```diff
@@ -0,0 +1,82 @@
...
+
+    result[fileName] = resolve(__dirname, action.entryFile);
+
+    return result;
+  }, {});
+};
+
+const main = async (watch = false) => {
+  try {
+    console.log('Building app actions');
```
lilbitner, line 55: seeing a few` console.logs` throughout the file, but assume these will be removed when the dust settles!

apps/mux/build-actions.js
```diff
@@ -0,0 +1,82 @@
...
+
+const env = process.env.NODE_ENV;
+const path = `./app-actions/.env${env === 'development' ? `.${env}` : ''}`;
+require('dotenv').config({ path });
+
+const argv = yargs(hideBin(process.argv)).argv;
+
+const validateActions = () => {
+  const requiredProperties = ['id', 'path', 'entryFile'];
+  const uniqueValues = new Set();
```
lilbitner, line 16: overall awesome JS here with the validation logic and use of `Set` 🔥

apps/mux/frontend/public/index.html
```diff
@@ -5,6 +5,5 @@
   </head>
   <body>
     <div id="root"></div>
-    <script type="text/javascript" src="./index.tsx"></script>
```
lilbitner, line 8: just out of curiosity here, why is this being removed?

lilbitner (approved):
Looks great!  A few nit/non-blocking comments!

apps/mux/scripts/call-app-action.ts
```diff
@@ -0,0 +1,145 @@
...
+    callParams
+  );
+
+  return appActionRunner.run();
+};
+main().then((result) => {
+  if (result) {
+    console.log('APP ACTION RESULT');
+    console.log(util.inspect(result, { showHidden: false, depth: null, colors: true }));
+    console.log('Result: success!');
```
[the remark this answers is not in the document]
jsdalton: yeah exactly. this actually _IS_ just a CLI tool to test out app actions locally ... we don't use it in anywhere in the app runtime code itself. You can see the result of the output in the PR description above ☝️

apps/mux/package.json
```diff
@@ -1,61 +1,26 @@
...
-    "eject": "react-app-rewired eject",
-    "lint": "eslint --max-warnings=0 .",
-    "create-app-definition": "contentful-app-scripts create-app-definition"
+    "create-app-definition": "contentful-app-scripts create-app-definition",
+    "build": "rimraf ./build && npm run build-frontend && npm run build-app-actions",
+    "build-app-actions": "cd app-actions && npm ci && cd .. && node build-actions.js",
+    "build-frontend": "cd frontend && npm ci --include=dev && BUILD_PATH=../build npm run build",
+    "build:dev": "rimraf ./build && npm run build-frontend && npm run build-app-actions:dev",
+    "build-app-actions:dev": "cd app-actions && npm ci && cd .. && NODE_ENV=development node build-actions.js",
+    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEV_TESTING_ORG_ID} --definition-id 2Z5q0FnYZjRN79MjTE4ofn --token ${TEST_CMA_TOKEN}",
```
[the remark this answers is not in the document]
jsdalton: yes. we actually deleted the `deploy` script awhile back to prevent deploys. this is one of the frustrating points about JSON -- you can't comment! If that were possible we would have just commented out the old deploy script with a note which would have been much clearer.

apps/mux/build-actions.js
jsdalton: this is another one where the logs are intentional since this is really a build (CLI) tool. so the purpose of these is to get output on the screen when it's run. 😄

apps/mux/frontend/public/index.html
jsdalton: Good catch that looks like a mistaken deletion while experimenting (though interestingly the app still runs just fine without that line 🤔 )

jsdalton: E.g. I'm comparing to https://github.com/contentful/apps/blob/master/apps/ai-image-generator/frontend/public/index.html which does not have this line but  https://github.com/contentful/apps/blob/master/apps/microsoft-teams/frontend/index.html does

jsdalton: After chatting internally we're inclined to keep this deletion since it's not needed (nor very clear what if any purpose it would serve).

PR #6670, opened 2024-03-05
feat(ms-teams): make teams channels searchable

ethan-ozelius-contentful:
## Purpose
Make Microsoft teams channels searchable, by channel name or team name

channels

## Approach
This PR removes `ChannelSelectionSupplementalModalContent` abstraction.  I tried every way that I could think of to avoid this, but react re-rendering had different plans.

The issue that I kept running into, was that the Search input was unnecessarily re-rendering, causing the input to lose focus and any previous value.  The reason for that is that we were passing `<DebouncedSearchInput>` as a `child` prop to `<ChannelSelectionSupplementalModalContent>`, and the entire `child` prop (ReactNode) was being rerendered.  I attempted to move `<DebouncedSearchInput>` up to be a sibling of `<ChannelSelectionSupplementalModalContent>`, but that violated the figma, since the search input was above the description paragraph.

## Testing steps
Tested locally, unit tests and some Mob QA feedback.

### March 6th Mob QA feedback items:
- [x] bug: input loses focus on re-render
- [x] bug: input loses previous value on re-render
- [ ] Make the magnifying glass an icon, instead of a button
- [ ] **investigating** if 0 search results are returned from fuzzy search, show a message indicating that.
- [x] fine tune debounce time to 300ms

**Follow up ticket(s)**: if a channel is already selected, it should be “pinned/hardcoded“ to the list of filtered channels
- [ ] Ticket: TBD
- [ ] **investigating** if a channel is selected, and the modal is “re-opened“ the selected channel should be pinned to the top.

## Limitations
1. real time updates - if the modal is opened, and then a new channel created or contentful app is installed in a channel, the user will need to refresh the page to see the newly available channel.


apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
```diff
@@ -22,23 +31,58 @@ interface ChannelSelectionModalProps {
...
 
-  const SupplementalModalContent = ({ children }: { children: React.ReactNode }) => (
-    <ChannelSelectionSupplementalModalContent
-      onClose={onClose}
-      handleNotificationEdit={handleNotificationEdit}
-      selectedChannel={selectedChannel}>
-      {children}
-    </ChannelSelectionSupplementalModalContent>
+  /**
+   * The search input and the list of content types must be kept separate so that the rerendering
```
jsdalton, line 55: looks like some copypasta from the previous PR 😁

apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
```diff
@@ -22,23 +31,58 @@ interface ChannelSelectionModalProps {
...
+        <Table.Row
+          key={channel.id}
+          onClick={() => setSelectedChannel(channel)}
+          className={styles.tableRow}>
+          <Table.Cell>
+            <Radio
+              id={channel.id}
+              isChecked={selectedChannel.id === channel.id}
+              onChange={() => {
+                /* clicking entire row checks this radio, i.e. do nothing here */
```
jsdalton, line 76: i guess I missed this on the last PR, but curious why we need to even include the empty function. Is the `onChange` attribute reuqired or something?

jsdalton (approved):
Looks good to me. One or two small comments / unimportant questions.

apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
ethan-ozelius-contentful: Yes, this is copy/pasta from contentType selection modal, but I still think it's relevant since this component is also proxying/brokering the search query from `<DebouncedSearchInput>` to `<SearchableList>` and I thought it was helpful to add a comment explaining that.  Re-reading it now, I'm not sure it would useful it is... 🤔 

And yeah, the first line mentions `content types` which is inaccurate.  So at the very least I will fix that :)

apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
ethan-ozelius-contentful: Yep, onChange is a required prop, so placating Forma-36.

PR #6723, opened 2024-03-07
feat(ms-teams): seachableList accepts an optional "pinnedItems" list …

ethan-ozelius-contentful:
…to render at top of filtered list, regardless of fuzzy-search matching

## Purpose
`<SearchableList>` now accepts an optional list of "Pinned" items that will always be rendered regardless of whether or not the fuzzy search matched those "pinned" items.  This is useful for MS teams channel & content type selection modals, when a user has already made a selection, then execute a search query that does not match that selected channel or content type. (see Screenshots)

### Pinned Channel
pinned

### pinned Content Type
pinned_contentTypes

## Approach
`<SearchableList>` now accepts an optional list of "Pinned" items, when determining the "filtered" list (list that will actually be rendered), prepend any pinned items to filtered list.  Note: the filtered list must also be unique, because it is possible that the pinned item can also occur in the original list, to solve this we are using `Set`, this could have also been solved with lodash `uniq()` but the less deps the better imo.

## Testing steps
tested locally & unit tested.

### DIFF
--- a/apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
@@ -66,6 +66,7 @@ const ChannelSelectionModal = (props: ChannelSelectionModalProps) => {
             <Radio
               id={channel.id}
               isChecked={selectedChannel.id === channel.id}
+              onChange={() => null /* noop, since clicking entire row selects this radio */}
               helpText={channel.teamName}>
               {channel.name}
             </Radio>
@@ -84,6 +85,8 @@ const ChannelSelectionModal = (props: ChannelSelectionModalProps) => {
         </Modal.Content>
       );
     }
+
+    const pinnedChannels: TeamsChannel[] = channels.filter((c) => c.id === selectedChannel.id);
 
     if (channels.length) {
       return (
@@ -104,7 +107,8 @@ const ChannelSelectionModal = (props: ChannelSelectionModalProps) => {
               <Table className={styles.table}>
                 <Table.Body>
                   <SearchableList
-                    list={channels}
+                    items={channels}
+                    pinnedItems={pinnedChannels}
                     renderListItem={renderRow}
                     searchKeys={['name', 'teamName']}
                     searchQuery={searchQuery}
--- a/apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
+++ b/apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
@@ -50,7 +50,10 @@ const ContentTypeSelectionModal = (props: Props) => {
           onClick={() => setSelectedContentTypeId(contentType.sys.id)}
           className={styles.tableRow}>
           <Table.Cell>
-            <Radio id={contentType.sys.id} isChecked={selectedContentTypeId === contentType.sys.id}>
+            <Radio
+              id={contentType.sys.id}
+              isChecked={selectedContentTypeId === contentType.sys.id}
+              onChange={() => null /* noop, since clicking entire row selects this radio */}>
               {contentType.name}
             </Radio>
           </Table.Cell>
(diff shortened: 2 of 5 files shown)


apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
```diff
@@ -85,6 +86,13 @@ const ChannelSelectionModal = (props: ChannelSelectionModalProps) => {
       );
     }
 
+    const pinnedChannels: TeamsChannel[] = [];
+    const selectedChannelToPin = channels.find((c) => c.id === selectedChannel.id);
```
jsdalton, line 90: btw I _think_ all of the code in this section can be replaced with `filter`:

```ts
const pinnedChannels: TeamsChannel[] = channels.filter((c) => c.id === selectedChannel.id);
```

So this would return an empty array (if nothing selected) or an array of one item (with one selected) or an array with multiple items (in the "bug in our code somewhere" case of 2 or more selected items)

apps/microsoft-teams/frontend/src/components/config/ContentTypeSelectionModal/ContentTypeSelectionModal.tsx
```diff
@@ -69,6 +72,16 @@ const ContentTypeSelectionModal = (props: Props) => {
       );
     }
 
+    let selectedContentType: undefined | ContentTypeProps = undefined;
+    const pinnedContentTypes: ContentTypeProps[] = [];
```
jsdalton, line 76: Same things above. Can all be simplified down to:

```ts
const pinnedContentTypes: ContentTypeProps[] = contentTypes.filter((c) => c.sys.id === selectedContentTypeId);
```

apps/microsoft-teams/frontend/src/components/config/SearchableList/SearchableList.tsx
```diff
@@ -13,7 +14,13 @@ interface Props<T> {
  * implementation of the search options, as well as the rendering of filtered items is left up to
  * the consumer of this component.
  */
-const SearchableList = <T,>({ list, searchQuery, searchKeys, renderListItem }: Props<T>) => {
+const SearchableList = <T,>({
+  list,
```
jsdalton, line 18: I wonder if at this stage it's a worth a rename of this prop to `items`? I feel like it'd be a lot more clear and also make explicit the connection between items and pinned items. And also "render list item" 😁 

Purely a subjective naming suggestion, totally optional if you think it makes sense. (But it would open a can of worms to rename elsewhere in the implementation)

jsdalton (approved):
I just had one or two suggestions. Overall I like the improvement! 💯

apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
ethan-ozelius-contentful: Hard to argue with this visual :) Great suggestion.  Yeah, while I was writing this I specifically thought to myself, there's definitely a much cleaner way to consolidate this code.  Thanks for the pointer!

Screenshot 2024-03-08 at 11 35 11 AM

Screenshot 2024-03-08 at 11 35 25 AM

apps/microsoft-teams/frontend/src/components/config/SearchableList/SearchableList.tsx
ethan-ozelius-contentful: Hadn't thought about that, but definitely makes sense.  I'll update the name, it feels correct to have all 3 props named in the same ballpark.

- [ ] rename `list` to `items`

apps/microsoft-teams/frontend/src/components/config/ChannelSelectionModal/ChannelSelectionModal.tsx
```diff
@@ -66,6 +66,7 @@ const ChannelSelectionModal = (props: ChannelSelectionModalProps) => {
             <Radio
               id={channel.id}
               isChecked={selectedChannel.id === channel.id}
+              onChange={() => null /* noop, since clicking entire row selects this radio */}
```
ethan-ozelius-contentful, line 69: Adding this back 🤦  I removed this in an unrelated PR, because I thought it was an unnecessary prop that forma-36 was enforcing, but it is actually a browser/react warning that is pretty obnoxious in the console.  [conversation](https://github.com/contentful/apps/pull/6670#discussion_r1515232011)

PR #6724, opened 2024-03-07
feat: Improve ColorPicker Field

Toxicom:
## Summary
   - Replace deprecated `AppExtensionSDK`;
   - Replace deprecated `FieldExtensionSDK`;
   - Display only validated colors in a validated field;
   - Show hex value in dropdown;
   - Disable custom values in a validated field.

## Purpose

The purpose is to improve UX for the content editor by simplifying the color picker field.

## Approach

I have also considered exposing a boolean in the configuration screen to validate against the label (`white` instead of `#ffffff`) but that would introduce breaking changes. I am thinking about supporting both and converting label to value behind the scenes, but I will leave this for a future PR.

## Testing steps

1. Create two text fields with the Color Picker appearance
2. In one of the fields, add validation for some of your theme values
3. Confirm that in the validated field's dropdown you only see the validated items available for selection

## Breaking Changes

App should still function as previously; if there is no validation present, `validatedColors` return the `theme.colors` array unchanged.

N/A

## Deployment

N/A

### Unvalidated Color Picker field
image

### Validated Color Picker field
image

### DIFF
--- a/apps/color-picker/src/locations/ConfigScreen.tsx
+++ b/apps/color-picker/src/locations/ConfigScreen.tsx
@@ -1,4 +1,4 @@
-import { AppExtensionSDK } from '@contentful/app-sdk';
+import { ConfigAppSDK } from '@contentful/app-sdk';
 import {
   Button,
   Card,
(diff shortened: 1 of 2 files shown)


Toxicom:
Not sure about tests failing but I can confirm build completed without issues:
image

whitelisab (approved):
Thanks for recreating this PR that's up to date with master @Toxicom! I reran the build on this PR and it's passing so we're all set and I'll get this merged in ✅

PR #6755, opened 2024-03-12
feat: adds js + ts examples for appevent.filter function [EXT-5023]

mgoudy91:


ghepting (approved):
LGTM even though we'll be making more updates soon

PR #6767, opened 2024-03-13
chore: [] - fix vercel client

matthew-gordon:
## Purpose

- Fixes broken Vercel client


apps/vercel/src/locations/ConfigScreen.tsx
```diff
@@ -98,6 +101,21 @@ const ConfigScreen = () => {
     getContentTypes();
   }, []);
 
+  useEffect(() => {
```
matthew-gordon, line 104: I lifted this (and `contentTypes` and we can talk about following up with injecting these lists if that makes sense @jsdalton

jsdalton (approved):

PR #6872, opened 2024-03-15
fix: handle DeletedEntry entry type

jsdalton:
## Purpose

When we unpublish or delete an entry, the payload we get from the AppEvent is _not_ a regular Entry. Instead it's of type `DeletedEntry`. It looks like this:

```json
{
  "sys": {
    "type": "DeletedEntry",
    "id": "19bRpH5BASo0WIu9i2vH6Z",
    "space": {
      "sys": {
        "type": "Link",
        "linkType": "Space",
        "id": "zl9er5nqimd0"
      }
    },
    "environment": {
      "sys": {
        "id": "master",
        "type": "Link",
        "linkType": "Environment"
      }
    },
    "contentType": {
      "sys": {
        "type": "Link",
        "linkType": "ContentType",
        "id": "blogPost"
      }
    },
    "revision": 7,
    "createdAt": "2024-03-15T18:37:58.806Z",
    "updatedAt": "2024-03-15T18:37:58.806Z",
    "deletedAt": "2024-03-15T18:37:58.806Z"
  }
}
```

The main difference is it's missing fields and thus missing a title.

Our logic was choking on the missing fields.

## Approach

* Provide a backup title if fields are missing (e.g. what we do when entry is created)

### DIFF
--- a/apps/microsoft-teams/app-actions/src/helpers/entry-activity.spec.ts
+++ b/apps/microsoft-teams/app-actions/src/helpers/entry-activity.spec.ts
@@ -4,6 +4,7 @@ import { buildEntryActivity } from './entry-activity';
 import {
   CollectionProp,
   ContentTypeProps,
+  EntryProps,
   LocaleProps,
   PlainClientAPI,
   UserProps,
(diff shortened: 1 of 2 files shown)


whitelisab (approved):
💯

PR #6887, opened 2024-03-18
feat: add mux token signing as HAA

jsdalton:
## Purpose

As part of the effort to get the Mux app functional again, we need to move token signing to the backend (since jsonwebtoken is no longer supported in browsers).

This PR introduces token signing in a backend app action. Later, the front end will be updated to call these values instead of attempting to sign on the frontend.

## Approach

* Use jsonwebtoken and move signing to app action


lilbitner (approved):
Looks good! 💯

PR #6888, opened 2024-03-19
Extend "handle-app-event" to support workflows

ethan-ozelius-contentful:
**DO NOT MERGE**

- [x] Do not merge this PR before the supporting [ms-teams-bot-service PR](https://github.com/contentful/msteams-bot-service/pull/352)

## Purpose
Extend the existing `"handle-app-event"` app action to also handle Workflows.  

## Approach
In the interest of time, as well as prototyping/learning we made the explicit choice
to re-use/extend this "app-events" app-action to handle "workflow" events as well.

This block of code to handle workflow topics is intentionally repetative and intrusive.
The end goal is to create a new app action to specifically handle "workflow" events.
Ideally, at that time, it will be a simple copy/paste operation, leaving this
app-action unchanged.

So, obviously an alternative would be to create a distinct app-action solely to handle workflows.  I'm not sure what the pros/cons of that decision would be beyond cleaner code. 

## Testing steps
Unit tested.

### Open questions
1. Exact structure/format of payload coming to this app from `workflow-consumer-api`.  
~~- [ ] todo: get example payload~~ 🚢  it

2. Ensure that the payload that this app sends to `config.msTeamsBotService.sendWorkflowUpdateMessage()` contains everything that we will need to render an adaptive card. **With the exception** of "colors" see [colors jira ticket](https://contentful.atlassian.net/browse/INTEG-1899)

## Deployment
Although it would not a catastrophic ka-boom to merge this PR before the supporting [ms-teams-bot-service PR](https://github.com/contentful/msteams-bot-service/pull/352) has been merged, it's a good habit (imo) to merge code only once the dependencies (in our case, `config.msTeamsBotService.sendWorkflowUpdateMessage()`) have been implemented/satisfied.

### DIFF
--- a/apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
+++ b/apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
@@ -3,14 +3,18 @@ import {
   AppActionCallResponse,
   EntryActivityMessage,
   SendEntryActivityMessageResult,
+  SendWorkflowUpdateMessageResult,
   Topic,
+  WorkflowUpdateMessage,
 } from '../types';
 import { EntryProps } from 'contentful-management/types';
 import helpers from '../helpers';
 import { parametersFromAppInstallation } from '../helpers/app-installation';
 import { config } from '../config';
 import { withAsyncAppActionErrorHandling } from '../helpers/error-handling';
+import { TOPIC_ACTION_MAP } from '../constants';
 
+// Todo: expand this to describe Workflows payload
 interface AppActionCallParameters {
   payload: string;
   topic: string;
@@ -21,13 +25,62 @@ export const handler = withAsyncAppActionErrorHandling(
   async (
     parameters: AppActionCallParameters,
     context: AppActionCallContext
-  ): Promise<AppActionCallResponse<SendEntryActivityMessageResult[]>> => {
+  ): Promise<
+    AppActionCallResponse<SendEntryActivityMessageResult[] | SendWorkflowUpdateMessageResult>
+  > => {
     const {
       cma,
       appActionCallContext: { appInstallationId, environmentId, spaceId, userId, cmaHost },
     } = context;
 
     const { payload, topic: topicString, eventDatetime } = parameters;
+
+    /**
+     * In the interest of time, as well as prototyping/learning we made the explicit choice
+     * to re-use/extend this "app-events" app-action to handle "workflow" events as well.
+     *
+     * This block of code to handle workflow topics is intentionally repetative and intrusive.
+     * The end goal is to create a new app action to specifically handle "workflow" events.
+     * Ideally, at that time, it will be a simple copy/paste operation, leaving this
+     * app-action unchanged.
+     */
+    if (topicString === TOPIC_ACTION_MAP['Workflow.Step.notifyMicrosoftTeams']) {
+      const appInstallation = await cma.appInstallation.get({ appDefinitionId: appInstallationId });
+      const { tenantId } = parametersFromAppInstallation(appInstallation);
+
+      // request body
+      const entry = JSON.parse(payload) as EntryProps;
+
+      const entryActivity = await helpers.buildEntryActivity(
+        { entry, topic: topicString, eventDatetime },
+        cma,
+        cmaHost
+      );
+
+      const title = entryActivity.entryTitle;
+      const contentTypeId = entry.sys.contentType.sys.id;
+
+      const workflowUpdateMessage: WorkflowUpdateMessage = {
+        title: title,
+        contentType: contentTypeId,
+        currentStep: 'currentStep', // hard-coded - to be replaced with correct step once payload from workflow-consumer-api is determined.
+        previousStep: 'previousStep', // hard-coded - to be replaced with correct step once payload from workflow-consumer-api is determined.
+        callToActionUrl: entryActivity.entryUrl,
+        updateDateTime: entryActivity.eventDatetime,
+      };
+
+      const sendWorkflowUpdateResult = await config.msTeamsBotService.sendWorkflowUpdateMessage(
+        workflowUpdateMessage,
+        tenantId,
+        { appInstallationId, environmentId, userId, spaceId }
+      );
+
+      return {
+        ok: true,
+        data: { sendWorkflowUpdateResult, workflowUpdateMessage },
+      };
+    }
+    /** END workflows block */
 
     // TODO parse entry and topic
     const entry = JSON.parse(payload) as EntryProps;
@@ -51,7 +104,11 @@ export const handler = withAsyncAppActionErrorHandling(
       if (notification.contentTypeId !== contentTypeId) return false;
 
       // don't send if the topic is not "checked" in the notification subscription
-      if (!notification.selectedEvents[topic]) return false;
+      if (
+        topic !== TOPIC_ACTION_MAP['Workflow.Step.notifyMicrosoftTeams'] &&
+        !notification.selectedEvents[topic]
+      )
+        return false;
 
       return true;
     });
--- a/apps/microsoft-teams/app-actions/src/constants.ts
+++ b/apps/microsoft-teams/app-actions/src/constants.ts
@@ -5,4 +5,5 @@ export const TOPIC_ACTION_MAP = {
   'ContentManagement.Entry.publish': 'published',
   'ContentManagement.Entry.unpublish': 'unpublished',
   'ContentManagement.Entry.delete': 'deleted',
+  'Workflow.Step.notifyMicrosoftTeams': 'Workflow.Step.notifyMicrosoftTeams',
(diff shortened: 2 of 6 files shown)


apps/microsoft-teams/app-actions/src/constants.ts
```diff
@@ -5,4 +5,5 @@ export const TOPIC_ACTION_MAP = {
   'ContentManagement.Entry.publish': 'published',
   'ContentManagement.Entry.unpublish': 'unpublished',
   'ContentManagement.Entry.delete': 'deleted',
```
ethan-ozelius-contentful, line 7: This update is a little confusing to me... I don't think we necessarily have a concept of a workflow that has a value... if that makes sense.  There aren't named changes in state in workflows, it's just a transition from one step to the next. 

For that reason, I just copied the Topic key also to its value.  I'd definitely be interested to figure out any ways to improve this code.

cc: @jsdalton

apps/microsoft-teams/app-actions/src/services/msteams-bot-service.ts
```diff
@@ -5,11 +5,34 @@ import {
   MsTeamsBotServiceResponse,
   TeamInstallation,
   TestMessage,
+  WorkflowUpdateMessage,
 } from '../types';
 
 export class MsTeamsBotService {
   constructor(public readonly botServiceUrl: string, public readonly apiKey: string) {}
```
ethan-ozelius-contentful, line 13: fyi: This method is pretty much an exact copy/pasta from `sendEntryActivityMessage()` the only real difference is the URL.

apps/microsoft-teams/app-actions/src/types.ts
```diff
@@ -28,6 +28,9 @@ export enum AppEventKey {
   ENTRY_DELETED = 'ContentManagement.Entry.delete',
   ENTRY_ARCHIVE = 'ContentManagement.Entry.archive',
   ENTRY_UNARCHIVE = 'ContentManagement.Entry.unarchive',
```
ethan-ozelius-contentful, line 30: In the same ballpark as [this comment](https://github.com/contentful/apps/pull/6888/files#r1529519858), This feels like a code smell, seeing as a workflow update (workflow event?  I kinda prefer update.  Sounds more like a state change, which is the model I think they're going for.  IMO.) is not the same as an entry event. (publish, create etc.)

apps/microsoft-teams/app-actions/src/types.ts
```diff
@@ -82,6 +85,15 @@ export interface EntryActivity {
   entryUrl: string;
 }
```
ethan-ozelius-contentful, line 87: payload structure subject to change, pending an example payload coming from `workflow-consumer-api` service.

apps/microsoft-teams/frontend/src/hooks/useGetTeamsChannels.ts
```diff
@@ -8,7 +8,6 @@ const useGetTeamsChannels = () => {
   const [error, setError] = useState<Error>();
   const [loading, setLoading] = useState<boolean>(true);
   const sdk = useSDK<ConfigAppSDK>();
-  const { tenantId } = sdk.parameters.installation as AppInstallationParameters;
```
ethan-ozelius-contentful, line 11: unrelated to this PR, but this is an unnecessary dependency.  `tenantId` is only used in the `useCallback()` change deps.  See line 46

apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
```diff
@@ -21,14 +25,63 @@ export const handler = withAsyncAppActionErrorHandling(
...
+    /**
+     * In the interest of time, as well as prototyping/learning we made the explicit choice
+     * to re-use/extend this "app-events" app-action to handle "workflow" events as well.
+     *
+     * This block of code to handle workflow topics is intentionally repetative and intrusive.
+     * The end goal is to create a new app action to specifically handle "workflow" events.
+     * Ideally, at that time, it will be a simple copy/paste operation, leaving this
+     * app-action unchanged.
+     */
+    if (topicString === TOPIC_ACTION_MAP['Workflow.Step.notifyMicrosoftTeams']) {
```
jsdalton, line 47: Hmm. Since this is really a "hack" I suggest we we just compare to the literal value here `Workflow.Step.notifyMicrosoftTeams` and not try to add this to the TOPIC_ACTION_MAP. It doesn't belong there.

apps/microsoft-teams/app-actions/src/constants.ts
```diff
@@ -5,4 +5,5 @@ export const TOPIC_ACTION_MAP = {
...
+  'Workflow.Step.notifyMicrosoftTeams': 'Workflow.Step.notifyMicrosoftTeams',
```
jsdalton, line 8: Right I see your copy. Let's just remove this line and check the topic string against the literal value as I commented above.

apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
```diff
@@ -21,14 +25,63 @@ export const handler = withAsyncAppActionErrorHandling(
...
+        cmaHost
+      );
+
+      const title = entryActivity.entryTitle;
+      const contentTypeId = entry.sys.contentType.sys.id;
+
+      const workflowUpdateMessage: WorkflowUpdateMessage = {
+        title: title,
+        contentType: contentTypeId,
+        currentStep: 'currentStep',
```
jsdalton, line 66: Mabye drop a comment or TODO here like "TODO use actual values from workflow action"

apps/microsoft-teams/app-actions/src/types.ts
```diff
@@ -28,6 +28,9 @@ export enum AppEventKey {
   ENTRY_DELETED = 'ContentManagement.Entry.delete',
   ENTRY_ARCHIVE = 'ContentManagement.Entry.archive',
   ENTRY_UNARCHIVE = 'ContentManagement.Entry.unarchive',
+
+  // This is technically not an app event, obviously.  It is a "workflow" app topic.
+  WORKFLOW_UPDATE = 'Workflow.Step.notifyMicrosoftTeams',
```
jsdalton, line 33: Yeah I think we don't need it.

apps/microsoft-teams/app-actions/src/constants.ts
jsdalton: We could just make it an official cosntant though (separate from TOPIC_ACTION_MAP)

```ts
export const WORKFLOW_STEP_EVENT = 'Workflow.Step.notifyMicrosoftTeams'
```

apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
ethan-ozelius-contentful: Sounds good

- [ ] "todo: ..."

apps/microsoft-teams/app-actions/src/constants.ts
ethan-ozelius-contentful: Yeah, that seems like a reasonable step in the right direction if we need an official const/enum/type elsewhere in the codebase.  Let me take a look.

- [ ] investigate applicabilty of `export const WORKFLOW_STEP_EVENT = 'Workflow.Step.notifyMicrosoftTeams'`
- [ ] implement if needed.

apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
ethan-ozelius-contentful: Actually, I'll investigate how this would actually be composed, i.e. what's the exact payload coming from `workflow-consumer-api`.  If it's already working, it might make sense to grab the actual data. 🤞

apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
ethan-ozelius-contentful: Hmmm... In my weakly held opinion, I'd still prefer to add `Workflow.Step.notifyMicrosoftTeams` to the list of topics that this app (not app action) can expect.  The downside of that is that a workflow topic, doesn't really _map_ to anything in the way that app events do.

All that being said, I'm getting a bunch of type errors, because `selectedEvents` and `helpers.buildEntryActivity` are expecting their topic param to be an `app-event` topic.  In the interest of time, I'm going to make the tradeoff to shoe-horn `Workflow.Step.notifyMicrosoftTeams` into the list of _until now_ app-event topics.  Hopefully this can be cleaned up when we address the above comment/todo.

jsdalton (approved):
One thing I suggest fixing. But overall I think your approach is fine. In general the problems I've highlighted are more closely related to the fact that we're hacking the app event handler. If/when we improve that we won't need to worry about those issues anymore.

PR #6903, opened 2024-03-19
Feature/vercel preview envs

jsdalton:
## Purpose

In order to introduce dynamically generated preview URLs for side by side previews with Vercel, we need an app action that returns all of the data needed to populate a preview environment.

## Approach

* For now just return a static value. Later we will do the actual work of dynamically generating the URL etc (from the app installation params as well as data from Vercel's API)
* Rather than a single value return an array. TBD whether it makes more sense to return all the preview envs for all content types in the space or simply for a given content type. For now assume a content type id is required but this could change later.


matthew-gordon (approved):
👍

PR #6987, opened 2024-03-22
fix: pipeline errors from misaligned packages

lilbitner:
## Purpose

The purpose of this PR is to fix the misaligned packages and their package.locks. It looks like this issue came from a a few dependabot updates. 

## Approach

I went into each app and ran the `npm ci` command, which is the command that is throwing [this](https://app.circleci.com/pipelines/github/contentful/apps/29609/workflows/cd39ccc4-213a-4d2d-b331-eaf54cd544bc/jobs/96794) pipeline error. If the package and package.lock were not aligned, I ensured they were. 

I am curious how this happens 🤔 I am assuming the dependabot updates happen so quickly that they merge one after the other and do not "rebase" or pull from master to ensure the packages are in sync, but am curious if others have a hypothesis.

### DIFF
--- a/apps/saleor/package-lock.json
+++ b/apps/saleor/package-lock.json
@@ -19397,6 +19397,18 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
+    "node_modules/tempy/node_modules/type-fest": {
+      "version": "0.16.0",
+      "resolved": "https://registry.npmjs.org/type-fest/-/type-fest-0.16.0.tgz",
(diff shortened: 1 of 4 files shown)


ethan-ozelius-contentful (approved):

apps/saleor/package-lock.json
```diff
@@ -19397,6 +19397,18 @@
         "url": "https://github.com/sponsors/sindresorhus"
```
ethan-ozelius-contentful, line 19397: definitely not blocking, but this change rolls forward 4 major versions, which makes me think we might need to make more changes in this app?

lilbitner: Ok! Will keep an eye on it!

PR #7068, opened 2024-03-26
feat: Add ContentType and preview path selection on config page

lilbitner:
## Purpose

This PR adds the beginnings of the ContentType/PreviewPath selection section on the config page. 

Designs: https://www.figma.com/file/3zhaUVyNKPZOFjVPL7mG00/Vercel?type=design&node-id=758-42795&mode=design&t=ipC1WnpgyJcr4PdA-4

## Approach

I created separate components for the `contentType` and `previewPath` row of selection and the associated list, which is rendered from the stored reducer state. 

https://github.com/contentful/apps/assets/58186851/609cc2e3-86af-45b5-9169-a993ea65c7e1

## Testing steps

Render the Vercel (development) config page using this branch locally. Go ahead and add a contentType and previewPath configuration, play around with removing rows and adding them, and save the configuration. The rows should then persist if you refresh the page. 

## Follow-up work

The following are TO DO items to improve on as we build on top of the steel thread, after consulting with Emily & team: 
1. Flesh out the workflow around when to enable the Add Content Type button. Currently, it is disabled if the current row is not filled out, and the and is disabled if there exists an added row already. This prevents the user from adding a bunch of unnecessary rows if they have not filled out the ones already added, but am open to making tweaks to this flow! 
2. Add logic to either to potentially prevent configuring multiple `previewPaths` per `contentType`.


apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSection.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { styles } from '../ConfigScreen.styles';
+import { AppInstallationParameters } from '../../../types';
+
+interface Props {
+  parameters: AppInstallationParameters;
+  dispatchParameters: Dispatch<SetStateAction<any>>;
+}
+
+export const ContentTypePreviewPathSection = ({ parameters, dispatchParameters }: Props) => {
+  // TO DO: Adjust logic to limit content type duplication of preview path and token
```
lilbitner, line 14: notes in the PR description, follow-up work to determine if we want to support multiple `previewPaths` per `contentTypes` selected. At this stage, we do not have logic to prevent that.

apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionList/ContentTypePreviewPathSelectionList.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const handleRemoveRow = (parameters: ContentTypePreviewPathSelection) => {
+    if (addRow.length) setAddRow([]);
+
+    dispatchParameters({
+      type: actions.REMOVE_CONTENT_TYPE_PREVIEW_PATH_SELECTION,
+      payload: parameters,
+    });
+  };
+
+  // TO DO: Adjust AddRow logic to boolean if we want to enforce only one row appearing at once
```
lilbitner, line 42: notes in the PR description, will adjust this logic once we experiment with the workflow of when the Add ContentType button appears i.e if users can endlessly add this, if it is disabled until the user has filled out the current row etc. For now, a user can only add one row at a time, and it is disabled if the current row is empty.

apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionList/ContentTypePreviewPathSelectionList.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const handleAddRow = () => {
+    setAddRow([...addRow, addRow.length]);
+  };
+
+  const isAddButtonDisabled =
+    Boolean(addRow.length) || !Boolean(contentTypePreviewPathSelections.length);
+
+  const renderSelectionRow = () => {
+    // TO DO: Handle case where contentTypes are not present - do not render add button etc.
```
lilbitner, line 51: **design follow-up:**  for how to handle no content types. We render something within the dropdown right now when there are not content types. Does that suffice? I feel the 'Add Content Type' button might be confusing in these cases.

matthew-gordon (approved):
This looks good to me and is working as described/expected! We can follow up on design/comments 💪

PR #7081, opened 2024-03-27
feat: add entry id to EntryActivity [INTEG-1922]

whitelisab:
## Purpose

For `create`, `unpublish`, and `delete` events, we do not have access to the entry title. This needs to be addressed to unblock us on the testing/validation from Microsoft since this was labeled as a mandatory fix from them.

## Approach

So we are no going to pass `entryId` as part of the payload for the `entry_activity_message` endpoint. There is a companion PR in the bot service to implement changes in the adaptive card.

We construct a title with the entry ID in the app action and displayed that in the adaptive card. I kept this login in there as a fallback.

### DIFF
--- a/apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
+++ b/apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
@@ -131,13 +131,13 @@ export const handler = withAsyncAppActionErrorHandling(
       return { sendMessageResult, entryActivityMessage };
     });
 
-    const sendEntryActiviyMessageResult: SendEntryActivityMessageResult[] = await Promise.all(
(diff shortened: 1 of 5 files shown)


apps/microsoft-teams/app-actions/src/actions/handle-app-event.ts
```diff
@@ -131,13 +131,13 @@ export const handler = withAsyncAppActionErrorHandling(
       return { sendMessageResult, entryActivityMessage };
     });
 
-    const sendEntryActiviyMessageResult: SendEntryActivityMessageResult[] = await Promise.all(
+    const sendEntryActivityMessageResult: SendEntryActivityMessageResult[] = await Promise.all(
```
whitelisab, line 134: This is just fixing a typo 😄

PR #7101, opened 2024-03-28
feat: Add informational table modal to config page

lilbitner:
## Purpose

This PR adds the InformationalModal that will help guide the user when configuring their content types and preview paths on the Vercel config page. 

Designs: https://www.figma.com/file/3zhaUVyNKPZOFjVPL7mG00/Vercel?type=design&node-id=758-43738&mode=design&t=VTfQ2mppRDRTqXiM-4
## Approach

I created several new components to create this modal, some reusable and some very specific. 

## Testing steps
 
Render the modal from the config page link "View more examples".

End result screenshot (can't quite include the footer 😅 )

Screenshot 2024-03-28 at 2 23 46 PM

### TO DO

1. Determine if there should be user feedback onClick of the copy buttons that are throughout the modal. Right now the copy functionality works, but there is nothing telling the user that something has been copied to their clipboard.

### DIFF
--- a/apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/InformationalModal/InformationalTables/InfoTable/InfoTable.styles.ts
+++ b/apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/InformationalModal/InformationalTables/InfoTable/InfoTable.styles.ts
@@ -1,6 +1,6 @@
 import tokens from '@contentful/f36-tokens';
 import { css } from 'emotion';
-import { INFORMATIONAL_MODAL_COLUMN_WIDTH } from '../../../../../../constants/styles';
+import { INFORMATIONAL_MODAL_COLUMN_WIDTH } from '@constants/styles';
 
 export const styles = {
   headersWrapper: css({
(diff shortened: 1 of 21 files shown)


apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/InformationalModal/InformationalTables/InfoTable/InfoTable.styles.ts
```diff
@@ -0,0 +1,38 @@
+import tokens from '@contentful/f36-tokens';
+import { css } from 'emotion';
+import { INFORMATIONAL_MODAL_COLUMN_WIDTH } from '../../../../../../constants/styles';
```
jsdalton, line 3: Don't we / shouldn't we use the thing where we have named top level packages to avoid these long directory traversals?

jsdalton (approved):

apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/InformationalModal/InformationalTables/InfoTable/InfoTable.styles.ts
lilbitner: yeah great callout, I have added the aliases to our `tsconfig` here, and will continue to refine other files with this as I continue frontend vercel work 👍

PR #7117, opened 2024-03-29
feat: display error message when content type or channel is not available [INTEG-1597]

whitelisab:
## Purpose

For configured notifications, if the selected content type or channel is no longer available (e.g. it was deleted), then the user should see an error message on that notification.

## Approach

I first updated the app installation parameters so that the content type name would be saved. This allows the name to still display on the notification after the content type is deleted, rather than it showing "Content type not found" as the name.

I updated the existing `ErrorMessage` component to add additional props to make it reusable for this use case as well as the existing one in the selection modals.

I did not any indication of a failing notification for the notification "view" mode, as I wanted to check in with design on that. For now, the error only displays when you open a configured notification. See below for the errors displayed when the content type or channel saved in a notification is not found:

Screenshot 2024-03-29 at 9 22 57 AM

## Testing steps

Tested deleting a configured content type and channel to see the new error message.

## Breaking Changes

Because this updates the app installation parameters, the app will likely need to be reinstalled in order for it to work correctly and not throw an error.


apps/microsoft-teams/frontend/src/components/config/ContentTypeSelection/ContentTypeSelection.spec.tsx
```diff
@@ -22,7 +22,11 @@ describe('ContentTypeSelection component', () => {
   it('mounts and renders an input when a content type is selected', () => {
     const { unmount } = ContentTypeCustomRender(
       <ContentTypeSelection
-        notification={{ ...defaultNotification, contentTypeId: 'blogPost' }}
```
ethan-ozelius-contentful, line 25: not blocking, but a quick negative unit test to assert that an error message is rendered when `isItemValid()` returns false might be a small improvement.

ethan-ozelius-contentful (approved):
Left one very optional minor suggestion, but otherwise looks really good. 

🙌 nice work!

apps/microsoft-teams/frontend/src/components/config/ContentTypeSelection/ContentTypeSelection.spec.tsx
whitelisab: Great suggestion! I'll add an additional test for the `ContentTypeSelection` and `ChannelSelection` components 👍

PR #7118, opened 2024-03-29
feat: Add new components, fix linting, remove unecessary install params

lilbitner:
## Purpose

The PR accomplishes the following: 
1. Adds linting config to Vercel (and fixes subsequent lint errors - hence size of this PR)
2. Aligns the designs with the current state of the Config page, specifically changes within the Project Selection section. Really this was just a few copy changes, some code re-org, and some testing was added.
3. Adjusts some props and variable names i.e `dispatchParameters => dispatch` to remain consistent. 
4. Adjusts import aliases for the `types` file. Interesting learning, you can't have `@types` because it thinks its a declaration file, so needed to name the directory `customTypes`, like we do in MS Teams. 
5. Removes the Sidebar as a location for the app (and deletes associated code 😢 can always resurrect if needed!)
6. Removes `contentTypes` and `projects` as part of the `appInstallation` parameters. These values can just be held in state and passed down as props (I don't think its necessary to extract to a context or anything more complex at this stage).


apps/vercel/frontend/src/components/common/GrayInfoBox/GrayInfoBox.styles.ts
```diff
@@ -1,7 +1,8 @@
 import tokens from '@contentful/f36-tokens';
 import { css } from 'emotion';
 
-export const useStyles = (rootOptions?: {}) => ({
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const useStyles = (rootOptions: any) => ({
```
lilbitner, line 5: this was a really hard type to figure out and I time-boxed it since I don't think its imperative, though would be nice to have.

jsdalton (approved):
Great! 💯

PR #7119, opened 2024-03-29
fix: update workflow message app action to match workflow consumer [INTEG-1943]

whitelisab:
## Purpose

Update the `msteamsHandleAppEvent` app action to reflect the changes in the `workflow-consumer-service` that calls this action.

## Approach

Update types to match payload in consumer.


ethan-ozelius-contentful (approved):
looks good to me!

PR #7133, opened 2024-04-02
feat: enforce contentType requirement when configuring Vercel

lilbitner:
## Purpose

The purpose of this PR is to improve the current approach to saving the `contentType` and `previewPath` parameters on the Vercel config page, from a user and technical perspective. 

## Approach

The original approach was using an awkward debounce (it would lose focus on the input when the debounce ended, blah!), and required that the user input a `contentType` AND `previewPath` to save the parameters (even though this wasn't obvious). We are now requiring that the user input a `contentType`, and sequentially enabling the `previewPath` input field once this `contentType` is filled out. It is not mandatory to have a `previewPath` per `contentType` in order to save the config. Having this mandatory `contentType` helps immensely from a technical standpoint in order to update data and app installation parameters appropriately, and it more importantly encourages the user to configure the app in a logical step-by-step process. 

https://github.com/contentful/apps/assets/58186851/b976c228-f96b-424b-acaf-33b332c8638a

## Testing steps

Play around on the Vercel config page with adding `contentType` and `previewPath` selections, as shown in the video above. 

## Follow-up work
1. Show informative messages (warning or error) when the user saves the selections with empty `previewPaths`. Even though this is allowed, its important to inform the user that this will not 'work' when actually using the app. 
2. Add more specific path validations of the `previewPath` itself i.e should be a valid path, not a random string.


jsdalton (approved):
😎

PR #7185, opened 2024-04-05
Add .env file to gitignore of app examples []

BasKiers:
## Purpose

I learned recently that using `npm run upload` will automatically create an `.env` file with the OrganizationId, AppDefinitionId and **Access Token**.

We don't want customers to accidentally upload their access token to their git repository as that would create a serious security risk.

Therefore I updated all `.gitignore` files in the app examples to ignore `.env` files.

### DIFF
--- a/examples/asset-upload/.gitignore
+++ b/examples/asset-upload/.gitignore
@@ -7,6 +7,11 @@
 # production
 /dist
 
+# dotenv environment variables file
+.env
+.env.*
+!.env*.example
+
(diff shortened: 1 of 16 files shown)


matthew-gordon:
`npm run upload will automatically create an .env file with the OrganizationId, AppDefinitionId and Access Token.`

Yikes, thank you for catching this! LGTM

matthew-gordon (approved):

PR #7187, opened 2024-04-05
fix: don't include configuration if preview path not specified

jsdalton:
## Purpose

We allow users to save a confiiguration but not specify a preview path for a given content type (ie they can leave it blank). In this case, we want to make sure the configuration doesn't get included in the provided preview environment.

## Approach

* Don't include if the (trimmed) preview path is not found


lilbitner (approved):
Looks good 👍

PR #7188, opened 2024-04-05
feat: add color fields to workflows payload

rusticpenguin:
## Purpose

We would like to show cards in the adaptive card. This is followup to our Workflows  changes


lilbitner (approved):
👍

PR #7204, opened 2024-04-08
chore: Add functionality to fetch apiPaths

lilbitner:
## Purpose

The purpose of this PR is to render the list of api-paths dynamically on the config page. 

## Approach

We first fetch the latest `deploymentId`, scoped by the selected projectId, and use this `deploymentId` to then fetch a `deploymentSummary`. We filter this data based on `serverlessFunctions` that are of type `Page` and include 'api' within the path. 

https://github.com/contentful/apps/assets/58186851/8d235ef1-802e-4fd0-bec1-67588499c845

## Testing steps

Paste a valid token, select a project, and select an api path on the vercel config page. You should be able to change projects and render the new list of paths as well. 

## Follow-up

As you will notice in the comments I have added there are a few follow-up items: 
1. Ensure we are scoping the `listDeploymentSummary` endpoint appropriately by team. It seems that if we scope by project, this should suffice. 
2. Ensure we grabbing the latest `deploymentId` when fetching it using the `getLatestDeploymentId`. 
3. Cleanup the behavior of the dropdown. i.e ensure that an appropriate default is selected or just render the placeholder, especially when switching between projects.

### DIFF
--- a/apps/vercel/frontend/src/clients/Vercel.ts
+++ b/apps/vercel/frontend/src/clients/Vercel.ts
@@ -1,4 +1,11 @@
-import { CreateDeploymentInput, Deployment, ListProjectsResponse } from '@customTypes/configPage';
+import {
+  ApiPath,
+  CreateDeploymentInput,
+  Deployment,
+  ListDeploymentSummaryResponse,
+  ListProjectsResponse,
+  ServerlessFunction,
+} from '@customTypes/configPage';
 
 interface VercelAPIClient {
   checkToken: () => Promise<boolean>;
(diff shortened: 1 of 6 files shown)


apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -37,6 +44,44 @@ export default class VercelClient implements VercelAPIClient {
     return data;
   }
 
+  async listDeploymentSummaries(
+    projectId: string,
+    deploymentId?: string
+  ): Promise<ListDeploymentSummaryResponse> {
+    const latestDeploymentId = deploymentId || (await this.getLatestDeploymentId(projectId));
+    const res = await fetch(
+      `${this.baseEndpoint}/v6/deployments/${latestDeploymentId}/files/outputs?file=..%2Fdeploy_metadata.json`,
```
lilbitner, line 53: Have added a ticket to scope this fetch even more-so by `teamId` (if necessary)

apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -37,6 +44,44 @@ export default class VercelClient implements VercelAPIClient {
...
+
+  async getLatestDeploymentId(projectId: string): Promise<string> {
+    const res = await fetch(`${this.baseEndpoint}/v6/deployments?projectId=${projectId}`, {
+      headers: this.buildHeaders(),
+      method: 'GET',
+    });
+
+    const data = await res.json();
+
+    return data.deployments[0].uid;
```
lilbitner, line 82: Have added a ticket to further investigate if this is the "way" to grab the latest deploymentId, or if we can more accurately target this.

apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -63,4 +108,20 @@ export default class VercelClient implements VercelAPIClient {
...
+      (file: ServerlessFunction) => file.type === 'Page' && file.path.includes('api')
+    );
+  }
+
+  private formatApiPaths(data: ServerlessFunction[]): ApiPath[] {
+    return data.map((file: ServerlessFunction) => {
+      const filePath = file.path;
+      return {
+        name: filePath,
+        id: filePath,
```
lilbitner, line 123: there isn't an id associated with the `serverlessFunction` object returned, so these are the same thing for now. Thoughts?

matthew-gordon: I think this is okay for now, maybe we can create a compound key for `id` with

```typescript
id: `${filePath}-${uuid()}`
``` 

Or something similar, we can brainstorm for sure.

matthew-gordon (approved):
AWESOME this makes sense to me 😄 Look good overall, I am wondering if we should go ahead and make the paths prefixed with `/` i.e.: `/api/enable-draft` instead of `api/enable-draft` here 🤔 Not blocking at all but something we need to consider here or user interface I believe

lilbitner:
> AWESOME this makes sense to me 😄 Looks really good overall, amazing work! I'm wondering if we should go ahead and make the paths prefixed with `/` i.e.: `/api/enable-draft` instead of `api/enable-draft` here 🤔 Not blocking at all but something we need to consider or update user interface/app action I believe

Yeah super good point! Maybe worth discussing a bit further in terms of determining if we reserve that logic for user-interface to manipulate as needed. Or perhaps we add it here for visual sake. Will capture in follow-up work!

apps/vercel/frontend/src/clients/Vercel.ts
lilbitner: Ah yes, good idea. Will be sure to circle back to this 👍 I have added a note within the code here

PR #7217, opened 2024-04-09
chore: Vercel API path selection clean-up

lilbitner:
## Purpose

The purpose of this PR is to do some minor cleanup with the configuration of Api paths on the Vercel configuration page. 

## Approach

1. I ensured that when the user changes the project, but there is an already selected API path, then it clears this state so as not to render API paths that may not existed with the new project. 
2. I added a broader test for the Config page
3. I tested pretty thoroughly that the list of deployments returned from Vercel does in fact return the deployments in order of most recent to least recent. I added a note within the code about this.

### DIFF
--- a/apps/vercel/frontend/src/clients/Vercel.ts
+++ b/apps/vercel/frontend/src/clients/Vercel.ts
@@ -79,6 +79,7 @@ export default class VercelClient implements VercelAPIClient {
 
     const data = await res.json();
 
+    // Vercel returns deployments sorted by date in descending order
     return data.deployments[0].uid;
   }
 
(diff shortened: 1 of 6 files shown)


apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -79,6 +79,7 @@ export default class VercelClient implements VercelAPIClient {
 
     const data = await res.json();
 
+    // Vercel returns deployments sorted by date in descending order
```
ethan-ozelius-contentful, line 82: 🙌

ethan-ozelius-contentful (approved):
Nice!

PR #7219, opened 2024-04-09
chore: add design updates for Vercel config page

lilbitner:
## Purpose

The purpose of this PR is to accomplish the following: 
1. Always render the Add Content type button, but disable it when there are no more content types to select (Mob QA feedback). 
2. Consolidate the sections of the Config page into fewer distinct sections based on design updates. 
3. Add functioning href's into the Config page links. 
4. Adjust some of the copies and layouts on the config page per design updates (small amount of spacing here and there). Most copies are not yet finalized so there is a follow-up ticket to finalize copies. Ensured there is a `marginBottom` to the entire page itself. 

## Approach

I abstracted the authentication logic into its own section/component, as I foresee much more logic to be added and isolated here as we flesh this section out. I also adjusted some file structure and adjusted some testing. 

New consolidated look can be seen below:
Screenshot 2024-04-09 at 3 10 20 PM

## Testing steps

The only functionality that changes is the first item listed above. Ensure that the Add Content type button is always present on the config page in the bottom section, just disabled when there is an empty row present OR all content types have been selected. 

## Follow-up work

There are several more design updates to make to this page, this is the first batch!

### DIFF
--- a/apps/vercel/frontend/src/components/config-screen/ApiPathSelectionSection/ApiPathSelectionSection.styles.ts
+++ b/apps/vercel/frontend/src/components/config-screen/ApiPathSelectionSection/ApiPathSelectionSection.styles.ts
@@ -1,0 +1,11 @@
+import tokens from '@contentful/f36-tokens';
+import { css } from 'emotion';
+
+export const styles = {
+  box: css({
+    width: '100%',
+  }),
+  helpText: css({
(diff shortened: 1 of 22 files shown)


matthew-gordon:
This makes sense to me and looks/works as expected 💪 👍

matthew-gordon (approved):

PR #7238, opened 2024-04-10
feat: add notification for user when copying to clipboard

lilbitner:
## Purpose

The purpose of this PR is to add a small notification for the user when they copy something to the clipboard. 

## Approach
I created a hook, `useNotifier`, to encapsulated this copy notification logic, as we can use the same hook for other notification as we see fit (and can expand the return value of the hook). With introducing this hook, it utilizes the App `SDK`, so I had to subsequently mock the `SDK` in tests and add the `SDKProvider` to the modal itself. 

Screenshot 2024-04-10 at 8 06 26 AM

## Testing steps

Copy one of the "copy-able" parts of the config page and notice the notification appear!


apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/InformationalModal/InformationalModal.tsx
```diff
@@ -16,58 +17,62 @@ export const InformationalModal = ({ onClose, isShown }: Props) => {
...
-        <>
-          <Modal.Header title={title} onClose={onClose} />
-          <Modal.Content>
-            <Flex flexDirection="column" alignItems="baseline" gap={tokens.spacingS}>
-              <Flex flexDirection="column" gap={tokens.spacingL}>
-                <Box>
-                  <Paragraph className={styles.firstParagraph}>{exampleOne.description}</Paragraph>
-                  <GrayInfoBox>{exampleOne.example}</GrayInfoBox>
-                </Box>
+    <SDKProvider>
```
lilbitner, line 20: just needed to add the SDKProvider to the modal (since this is a sibling node to the app itself when rendered).

matthew-gordon (approved):

PR #7239, opened 2024-04-10
feat: conditionally render config page sections

lilbitner:
## Purpose

The purpose of this PR is to progressively render the sections on the config page as the user fills out certain steps. Additionally, I adjusted some of the copies and spacing on the informative modal on the config page to match the designs. 
## Approach

The user will not be able to select a project until there is a valid access token, they will not be able to select an API path until there is a project selected, and they will not be able to configure content types and preview paths until all the above is complete. 

The spacing changes within the modal can be seen below:

#### Before
Screenshot 2024-04-10 at 1 09 37 PM

#### After
Screenshot 2024-04-10 at 1 09 05 PM

## Testing steps

Follow the workflow as described above ^

### DIFF
--- a/apps/vercel/frontend/src/locations/ConfigScreen.tsx
+++ b/apps/vercel/frontend/src/locations/ConfigScreen.tsx
@@ -14,25 +14,15 @@ import { ApiPathSelectionSection } from '@components/config-screen/ApiPathSelect
 import { AuthenticationSection } from '@components/config-screen/AuthenticationSection/AuthenticationSection';
 
 const ConfigScreen = () => {
-  const [parameters, dispatchParameters] = useReducer(parameterReducer, initialParameters);
-  const [isAppInstalled, setIsAppInstalled] = useState(false);
+  const [isTokenValid, setIsTokenValid] = useState(false);
   const [contentTypes, setContentTypes] = useState<ContentType[]>([]);
   const [projects, setProjects] = useState<Project[]>([]);
   const [apiPaths, setApiPaths] = useState<ApiPath[]>([]);
+
+  const [parameters, dispatchParameters] = useReducer(parameterReducer, initialParameters);
   const sdk = useSDK<ConfigAppSDK>();
 
   useInitializeParameters(dispatchParameters);
-
-  const getIsAppInstalled = useCallback(async () => {
-    const isInstalled = await sdk.app.isInstalled();
-
-    setIsAppInstalled(isInstalled);
-  }, [sdk]);
-
-  useEffect(() => {
-    getIsAppInstalled();
-    sdk.app.onConfigurationCompleted(() => setIsAppInstalled(true));
-  }, [sdk]);
 
   const onConfigure = useCallback(async () => {
     const currentState = await sdk.app.getCurrentState();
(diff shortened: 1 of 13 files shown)


apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -14,26 +14,16 @@ import { ApiPathSelectionSection } from '@components/config-screen/ApiPathSelect
 import { AuthenticationSection } from '@components/config-screen/AuthenticationSection/AuthenticationSection';
 
 const ConfigScreen = () => {
-  const [parameters, dispatchParameters] = useReducer(parameterReducer, initialParameters);
-  const [isAppInstalled, setIsAppInstalled] = useState(false);
```
lilbitner, line 18: @matthew-gordon I would love your perspective/context on this logic that I removed. It seemed to me that we want to support the same functionality on the config page no matter if the user is configuring it before installed or if it is installed and they are editing it, or is there another workflow here to consider?

apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -57,21 +47,10 @@ const ConfigScreen = () => {
...
-        dispatchParameters({
-          type: actions.UPDATE_VERCEL_ACCESS_TOKEN_STATUS,
-          payload: true,
-        });
-      } else {
-        dispatchParameters({
-          type: actions.UPDATE_VERCEL_ACCESS_TOKEN_STATUS,
-          payload: false,
-        });
-      }
```
lilbitner, line 71: I also removed this logic to update the status of the token within the installation parameters @matthew-gordon. Would appreciate any insight here that I might be overlooking. My approach was that the knowledge around if the token is valid or not does not need to be persisted in params, but rather can be handled in state on render.

apps/vercel/frontend/src/locations/ConfigScreen.tsx
matthew-gordon: Yeah this was a relic of the "loose requirements" as this started organically with only the input field initially existing for the access token, so just a bit of naive blocking logic to make ourselves go create access tokens while flushing out customer interviews/feedback.

apps/vercel/frontend/src/locations/ConfigScreen.tsx
matthew-gordon: Mostly same answer above, this was just the mans to update based on an arbitrary status check to a `/tokens` endpoint in vercel and keep the ui up to date from early iterations, you are absolutely correct in your assumptions

apps/vercel/frontend/src/locations/ConfigScreen.tsx
lilbitner: Awesome! makes total sense 💯

apps/vercel/frontend/src/locations/ConfigScreen.tsx
lilbitner: Perfect!

matthew-gordon (approved):
Makes sense to me 😄 awesome work! Thank you for the clarifying clean up! 💪 💪

PR #7249, opened 2024-04-11
feat: add authentication token error to field

lilbitner:
## Purpose

With respect to the Vercel config page, the purpose of this PR is the following: 
1. Adjust the error state for an invalid authentication pasted token
2. Adjust the empty state for when projects do not exist
3. Update all of the page copies to align with designs
4. Update layout of Content type section

## Approach

The approach is fairly straightforward - removed the original "Status" badge that was used for the authentication section and instead placed the error message as per the [designs](https://www.figma.com/file/3zhaUVyNKPZOFjVPL7mG00/Vercel?type=design&node-id=857-15326&mode=design&t=8txI8tLGkelAsTuO-4). See below:

### New error state

Screenshot 2024-04-12 at 8 21 23 AM

Additionally, ensured that the Content type field and Preview path field each have their own labels (previously it was one label per row- see below). 

### Before
Screenshot 2024-04-12 at 8 22 37 AM

### After
Screenshot 2024-04-12 at 8 20 42 AM

## Follow-up work
1. More error states and empty states to be added in follow-up work!

### DIFF
--- a/apps/vercel/frontend/src/components/common/Select/Select.styles.ts
+++ b/apps/vercel/frontend/src/components/common/Select/Select.styles.ts
@@ -1,8 +1,0 @@
-import tokens from '@contentful/f36-tokens';
-import { css } from 'emotion';
-
-export const styles = {
-  formControl: css({
-    marginBottom: tokens.spacingS,
-  }),
-};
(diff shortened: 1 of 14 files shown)


matthew-gordon (approved):
Looking great! 🚀 💪

PR #7284, opened 2024-04-12
feat: Add error states when selected project or API path no longer exists on vercel config page

lilbitner:
## Purpose

The purpose of this PR is to add logic that renders an error message when a selected project or API path no longer exists within Vercel itself. 

## Approach

Because the project section and API path section is converging so much, I refactored a lot of logic to create a fairly flexible SelectSection component. This is where dispatch and error logic is handled. Designs originally only called for an error state when a project doesn't exist, but I figured this could be a useful state in case API paths change within Vercel as well. 

Notice the new error states below: 
Screenshot 2024-04-12 at 1 07 52 PM

Screenshot 2024-04-12 at 1 08 02 PM


apps/vercel/frontend/src/components/common/SelectSection/SelectSection.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Select } from '@components/common/Select/Select';
+import { Path, Project } from '@customTypes/configPage';
+import { copies } from '@constants/copies';
+import { FormControl } from '@contentful/f36-components';
+import { actions } from '@constants/enums';
+
+type CopySection = Extract<
+  keyof typeof copies.configPage,
+  'projectSelectionSection' | 'pathSelectionSection'
+>;
```
lilbitner, line 13: checkout some lessons from TotalTypescript!! @matthew-gordon

lilbitner:
> I had one question about the error state (posted on Slack) but the implementation itself looks good! 💯

Awesome! I will follow-up with any work with respect to the comment you added in Slack!

PR #7299, opened 2024-04-15
Account for empty content type and empty preview path

lilbitner:
## Purpose

This PR accomplishes the following: 
1. Updates the Select error state to also include the red outline of the field (follow-up from last PR)
2. Adds a tooltip to the Add Content Type button when all content types have been configured.
3. Adds an error state to the preview path input that appears when the field is empty and the user has saved/installed the page. This error message disappears when the user interacts with that section once again. 
4. Creates a context that can be used to manage props being passed from the Config page to child components. 
5. Fixes a few imports that were not compiling correctly. 

### Updated invalid state on Project and API path select component:

Screenshot 2024-04-15 at 4 22 00 PM

### Error state for preview path: 

Screenshot 2024-04-15 at 4 21 16 PM

### Tooltip for Add Content Type button: 

Screenshot 2024-04-15 at 1 11 53 PM

### DIFF
--- a/apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
+++ b/apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
@@ -81,13 +81,20 @@ export const ContentTypePreviewPathSelectionRow = ({
     return isAppConfigurationSaved && !configuredPreviewPath && !!configuredContentType;
   }, [isAppConfigurationSaved, configuredPreviewPath, configuredContentType]);
 
+  const itemAlignment = useMemo(() => {
+    if (isPreviewPathInvalid) {
+      return !renderLabel ? 'baseline' : 'normal';
+    }
+    return 'flex-end';
+  }, [isPreviewPathInvalid, renderLabel]);
+
   return (
     <Box className={styles.wrapper}>
       <FormControl marginBottom="spacingM" id="contentTypePreviewPathSelection">
         <Flex
           flexDirection="row"
           justifyContent="space-evenly"
-          alignItems={isPreviewPathInvalid ? 'normal' : 'flex-end'}
+          alignItems={itemAlignment}
           gap={tokens.spacingXs}>
           <Box className={styles.inputWrapper}>
             <Select
--- a/apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
+++ b/apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
@@ -41,7 +41,7 @@ export const ContentTypePreviewPathSelectionRow = ({
 
   const { inputs } = copies.configPage.contentTypePreviewPathSection;
 
-  const { isAppConfigurationSaved, parameters } = useContext(ConfigPageContext);
+  const { isAppConfigurationSaved } = useContext(ConfigPageContext);
 
   const handlePreviewPathInput = (event: React.ChangeEvent<HTMLInputElement>) => {
     onParameterUpdate({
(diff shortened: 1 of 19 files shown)


apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -112,8 +116,17 @@ const ConfigScreen = () => {
     });
   };
 
+  const clearIsAppConfigureCalled = () => {
+    setIsAppConfigureCalled(false);
+  };
```
lilbitner, line 121: not super sold on this name. thoughts? This essentially clears the ephemeral state of knowing that the Install/Save button has been clicked.

apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
```diff
@@ -67,13 +77,17 @@ export const ContentTypePreviewPathSelectionRow = ({
     [contentTypes]
   );
 
+  const isPreviewPathInvalid = useMemo(() => {
+    return isAppConfigureCalled && !configuredPreviewPath && !!configuredContentType;
```
jsdalton, line 81: Yeah this is all fairly readable as code but is doing my head in as far as why we consider it invalid under these conditions.

jsdalton (commented):
Okay I did indeed find the whole "is app configured" state a bit confusing and in a comment I posted some thoughts on how to address it. Hit me up directly if you want to discuss further!

apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -14,12 +14,14 @@ import { ApiPathSelectionSection } from '@components/config-screen/ApiPathSelect
...
 import { copies } from '@constants/copies';
 import { actions } from '@constants/enums';
+import { ConfigPageProvider } from '@contexts/ConfigPageProvider';
 
 const ConfigScreen = () => {
   const [isTokenValid, setIsTokenValid] = useState(false);
   const [contentTypes, setContentTypes] = useState<ContentType[]>([]);
   const [projects, setProjects] = useState<Project[]>([]);
   const [apiPaths, setApiPaths] = useState<ApiPath[]>([]);
+  const [isAppConfigureCalled, setIsAppConfigureCalled] = useState(false);
```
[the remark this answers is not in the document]
lilbitner: yeah this may be a naming mishap on my end! but this state essentially is allowing the Preview path component to know that the user has "saved/installed" the current configuration in order to render that warning message. We only know this at the top level when `onConfigure` is called and the preview path warning message only appears on save/install. So this initial state is false, and then is an ephemeral true, until the user then interacts with the Content type/preview path section, which then calls the `clearIsAppConfigureCalled` function to reset this state.

apps/vercel/frontend/src/locations/ConfigScreen.tsx
lilbitner: Yeah totally valid assumptions based on the naming. [This](https://github.com/contentful/apps/pull/7299#discussion_r1567855792) comment explains as well but essentially I needed to create logic that allows child components to know when the user has clicked "save/install" and the isAppConfigureCalled is the state telling the components this. The aformentioned child components are specifically the preview path inputs, that when the user saves/installs, and the input is empty, the warning message appears. This allows the user to save/install without inputing those values but still informs the user that its not a functioning value. Where I took some liberty is that this isAppConfigureCalled is an ephemeral "true" once the user saves/installs, and when the user then interacts with the Content type section, this then resets using this `clearIsAppConfigureCalled` state, to the default false. I can certainly think a bit harder about naming these to ensure that this confusion is limited, which I had a feeling would happen 😅 I also think perhaps engaging anywhere on the config page could reset this state, as opposed to the user just clicking in the Content type section.

apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
lilbitner: Have removed one condition here to simplify it slightly!

apps/vercel/frontend/src/locations/ConfigScreen.tsx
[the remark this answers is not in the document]
lilbitner: This has been adjusted, to render as `true` in the beginning 👍

jsdalton (approved):
🥇

PR #7320, opened 2024-04-16
feat: fetch mux tokens from HAA instead of from the client

jsdalton:
## Purpose

This is the nearly final step toward unblocking mux deployments. Now that we have the ability sign tokens on the backend, we simply need to use those tokens from the frontend and remove the code that was building them in the client.

## Approach

* Fetch tokens from the HAA
* Await the tokens when uploading a new video so the video doesn't try to load without them being ready
* Remove all the no longer needed client side token utils
* Remove all the no longer needed dependencies and react-app-rewired stuff

Note: A bug was found during testing unrelated to this PR (bug exists on master) that causes videos uploaded through the app itself to not load correctly at first. This will need to be fixed independent of this PR

## Testing steps

mux-with-signed

Viewable in the sandbox: https://app.contentful.com/spaces/0h02439dt6mi/entries/Ic5wZuMDO8TrObj49kWig?focusedField=muxVideo


rusticpenguin (approved):
Looks good!

PR #7321, opened 2024-04-16
fix: improve loading experience of vercel config page

lilbitner:
## Purpose

The purpose of this PR is to resolve the flashing error messages that may occur when the Vercel config page originally renders. This is really an optimization of state and data fetching management on the page itself, mostly with respect to validating the authentication token. 

## Approach

1. I created a loading state that reflects API loading when checking the token, fetching projects, and fetching API paths. This ensures we can not prematurely render error messages or empty messages in any of these sections.
2. I also separated the logic of validating the token when it is updated and also validating it on render of the page. Also this introduces some duplicate code, this ensures that we are not just relying on useEffect updates to validate this logic, which tends to introduce complexity with various variables updating at various times, which ultimately causes the flashing behavior. I created the `hasTokenBeenUpdated` state to support the relationship between these two functions, to ensure we only validate the token once. 
3. I also updates some copies per Mob QA feedback. 

### Before: 

https://github.com/contentful/apps/assets/58186851/ca96cf87-d1b3-43ae-9f5e-79917b354368

### After: 

https://github.com/contentful/apps/assets/58186851/6973c920-a88a-487e-9769-841f719a59a2

## Testing steps

Render the loading page, with either information already persisted, or from the start. Ensure that updating the access token does not cause unwanted side effects of error messages or empty states rendering for a flash of a second.

### DIFF
--- a/apps/vercel/frontend/src/locations/ConfigScreen.tsx
+++ b/apps/vercel/frontend/src/locations/ConfigScreen.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState, useEffect, useReducer, ChangeEvent } from 'react';
+import { useCallback, useState, useEffect, useReducer, ChangeEvent, isValidElement } from 'react';
 import { ConfigAppSDK, ContentType } from '@contentful/app-sdk';
 import { Box, Heading, Paragraph, Stack } from '@contentful/f36-components';
 import { useSDK } from '@contentful/react-apps-toolkit';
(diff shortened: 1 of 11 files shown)


apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -46,19 +48,24 @@ const ConfigScreen = () => {
     };
   }, [parameters, sdk]);
 
+  const vercelClient = new VercelClient(parameters.vercelAccessToken);
```
rusticpenguin, line 51: Thoughts on if we want to store this in a state and update it in a useEffect?  That way it's not creating a new vercel client every rerender

matthew-gordon (approved):
✨ 🚀

apps/vercel/frontend/src/locations/ConfigScreen.tsx
lilbitner: Good idea George! I will follow-up up with some adjustments in subsequent PRs :)!!

PR #7335, opened 2024-04-17
chore: reenable mux production deploys

jsdalton:
## Purpose

Re-enable the Mux app deploy script, to start deploying Mux app to the production app definition. Currently, Mux is "pinned" to a build from Feb 5 while we rearchitected how tokens were being generated for signed URLs.

## Approach

* Add back the production deploy script to start deploying again

## Testing steps

For piece of mind I recorded a walkthrough of an end to end walkthrough of the Mux app, both with and without signed URLs. This was recorded from our sandbox space, which I built and deployed the current `master` too. So theoretically it's tested against the same artifact we will start deploying in this PR.

https://www.loom.com/share/d095b751fe1246829f6b6f4c1060f89a


matthew-gordon (approved):
🖖

PR #7336, opened 2024-04-17
feat: add preview path input validation on save

lilbitner:
## Purpose

The purpose of this PR is to provide validation within the preview path input when a user saves the configuration. 

## Approach

The following rules apply for the validation: 
1. The input must start with a '/' and must contain at least one set of closed curly braces i.e '{}' somewhere within the string. 
The validation of the preview path input will only run when the user has saved the configuration.

I also adjusted the messages that appear for preview path error messages (either empty or invalid). Would love input on those!

Screenshot 2024-04-18 at 7 31 34 AM

## Testing steps

Input an invalid preview path, save the config, and notice the error message that appears. The error message should then clear once you interact with the section.

### DIFF
--- a/apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
+++ b/apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
@@ -88,7 +88,8 @@ export const ContentTypePreviewPathSelectionRow = ({
   }, [isAppConfigurationSaved, configuredPreviewPath, configuredContentType]);
 
   useEffect(() => {
-    if (configuredPreviewPath && isAppConfigurationSaved) {
+    if (!isAppConfigurationSaved) setIsPreviewPathInvalid(false);
+    else if (configuredPreviewPath) {
       const isPreviewPathValid = validatePreviewPath(configuredPreviewPath);
       setIsPreviewPathInvalid(!isPreviewPathValid);
     }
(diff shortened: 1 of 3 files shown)


apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
```diff
@@ -63,6 +64,9 @@ export const ContentTypePreviewPathSelectionRow = ({
     onRemoveRow(configuredContentTypePreviewPathSelection);
   };
 
+  const validatePreviewPath = (previewPath: string) =>
+    [/^\/.*{([^}]+)}.*$/].some((regex) => regex.test(previewPath));
```
lilbitner, line 68: require the string to start with a '/' and include at least one set of '{}' within the string. Thoughts on this logic for the validation? Do we want to include or exclude a pattern?

jsdalton: Hm, the rules we are validating and the logic make sense but I think going with a single regex means we lack meaningful information about why the particular string fails validation.

I recommend we isolate our rules, test them separately, and collect error messages indepedently. So right now we have 3 rules:

* Path cannot be blank
* Path must start with a `/` (i'm not 100% sure about this rule btw, since we can always join the path to the main domain in our redirect and add the '/' ourselves)
* Path must contain at least one "token" (still think we need a better word for this)

apps/vercel/frontend/src/constants/copies.ts
```diff
@@ -44,7 +44,8 @@ export const copies = {
         previewPath: {
           label: 'Preview path',
           placeholder: 'Set preview path and token',
-          errorMessage: 'This field is empty and not saved',
+          emptyErrorMessage: 'Field is empty and omitted from configuration',
+          invalidFormattingMessage: 'Formatting is invalid and omitted from configuration',
```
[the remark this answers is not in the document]
jsdalton: Yeah I do think we need to tweak this message and presentation a bit. My thoughts:

* I think we should remove "omitted from configuration" entirely on both. Just dispaly the problem whatever it might be ("Field is empty" or "Formatting is invalid")
* Maybe display a warning instead of an error?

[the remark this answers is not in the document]
lilbitner: The copies have been adjusted to be more succinct!

apps/vercel/frontend/src/components/config-screen/ContentTypePreviewPathSection/ContentTypePreviewPathSelectionRow/ContentTypePreviewPathSelectionRow.tsx
lilbitner: I have adjusted the wording of the main validation! The word "token" is a TBD adjustment

jsdalton (approved):
💯

PR #7337, opened 2024-04-17
feat: make SkuPicker search placeholder and help text configurable [INTEG-2017]

whitelisab:
## Purpose

The SkuPicker component had the placeholder text hardcoded to "Search for a Product...", which is not helpful if your ecommerce app supports multiple sku types (e.g. categories). Also, this PR adds the ability to create help text under the search box. This is the first step in adding in category search functionality in the commercetools app, and also addresses bugs in apps with multiple sku types, since the placeholder text was incorrect.

## Approach

I added two new functions: `MakeSearchHelpText` and `MakeSearchPlaceholderText`. I ensured that these were backwards compatible if the app doesn't pass in these functions. Otherwise, the app consuming this library will need to pass in these functions in order to have configured placeholder and help text. Here's a branch where I am testing these changes with the commercetools app: https://github.com/contentful/apps/compare/master...feat/commercetools-category-search

I also made some small UI improvements, like making the search box wider and making the help text and total count the same font.

Here are the various states now supported:

Default placeholder with no help text or total results:
Screenshot 2024-04-17 at 2 42 45 PM

Default placeholder with total results:
Screenshot 2024-04-17 at 2 43 32 PM

Custom placeholder with help text:
Screenshot 2024-04-17 at 2 43 06 PM

Custom placeholder with help text and total results:
Screenshot 2024-04-17 at 2 36 06 PM

I also ensured that this works with the "Search only by SKU option" in the commercetools app:
Screenshot 2024-04-17 at 3 45 48 PM

## Testing steps

I tested this in Storybook and also locally by linking the package in the commercetools app.

## Breaking Changes

There should not be any breaking changes introduced here

### DIFF
--- a/packages/ecommerce-app-base/src/SkuPicker/SkuPicker.tsx
+++ b/packages/ecommerce-app-base/src/SkuPicker/SkuPicker.tsx
@@ -8,6 +8,7 @@ import { Paginator } from './Paginator';
 import {
   MakeSaveBtnTextFn,
   MakeSearchPlaceholderText,
+  MakeSearchHelpText,
   Pagination,
   Product,
   ProductPreviewsFn,
@@ -17,7 +18,7 @@ import { ProductSelectionList } from './ProductSelectionList';
 import { styles } from './styles';
 import { mapSort } from '../utils';
 
-import { Button, Checkbox, Text, TextInput } from '@contentful/f36-components';
+import { Button, Checkbox, Flex, Text, TextInput } from '@contentful/f36-components';
 
 import { SearchIcon } from '@contentful/f36-icons';
 
@@ -29,6 +30,7 @@ export interface Props {
   skuType?: string;
   makeSaveBtnText?: MakeSaveBtnTextFn;
   makeSearchPlaceholderText?: MakeSearchPlaceholderText;
+  makeSearchHelpText?: MakeSearchHelpText;
   hideSearch?: boolean;
   showSearchBySkuOption?: boolean;
 }
@@ -45,10 +47,6 @@ interface State {
 
 const DEFAULT_SEARCH_DELAY = 250;
 
-function defaultMakeSearchPlaceholderText(_skuType?: string): string {
-  return 'Search for a product...';
-}
-
 function defaultGetSaveBtnText(selectedSKUs: string[]): string {
   switch (selectedSKUs.length) {
     case 0:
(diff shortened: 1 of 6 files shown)


packages/ecommerce-app-base/src/SkuPicker/SkuPicker.tsx
```diff
@@ -48,6 +58,10 @@ function defaultGetSaveBtnText(selectedSKUs: string[]): string {
   }
 }
 
+function defaultMakeSearchPlaceholderText(_skuType?: string): string {
```
ethan-ozelius-contentful, line 61: Definitely not blocking, but since we have the nice cleanly defined type: `MakeSearchPlaceholderText`, would it make sense to type this function?  Like I said, not blocking, no impact to customers but since we have the type, maybe it would save a future developer 30 seconds?  Not sure.

I could go either way :)

packages/ecommerce-app-base/src/SkuPicker/SkuPicker.tsx
```diff
@@ -154,11 +168,37 @@ export class SkuPicker extends Component<Props, State> {
...
   };
 
+  constructSearchAdditionalInfo = (
+    makeSearchHelpText: MakeSearchHelpText | undefined,
+    skuType: string | undefined,
+    paginationTotal: number
+  ) => {
+    const helpText = makeSearchHelpText ? makeSearchHelpText(skuType) : '';
+    const totalResults = !!paginationTotal
+      ? `${paginationTotal.toLocaleString()} total results`
```
ethan-ozelius-contentful, line 178: today I learned, `Number.toLocaleString()` super cool.

ethan-ozelius-contentful: Also what if `paginationTotal` is `0`?  would we still want to render `0 results`?  Or is it ok to render `''`.  Which correct me if I'm misreading this, I think that's what this code would do.

ethan-ozelius-contentful (approved):

packages/ecommerce-app-base/src/SkuPicker/SkuPicker.tsx
whitelisab: Good question! I just checked and the existing behavior is to render `''` instead of `0 results`. I'm just going to keep it as is for now.

PR #7363, opened 2024-04-22
fix: ensure authentication token logic is handled consistently 

lilbitner:
## Purpose

The purpose of this PR is to do the following: 
1. Ensure that the VercelClient is never instantiated without the applied access token. 
2. Ensure that the loading and "validity" states that are dictating the rendering and error states on the Config page are being adjusted based on valid calls and responses from API communication. 
3. Ensure you cannot save an invalid token. 

## Approach

You will notice comments throughout the code, but I have made a few conditional adjustments within the ConfigPage code. Transparently, I am not excited by the ever-increasing complexity of our loading states and conditional statements, so I have a follow-up task to really abstract some of this, but feel its higher priority to work on other things at this time. 

## Testing steps

Work through the following testing flows: 
1. Clean install the Vercel app, and iterate on adding an invalid as well as a valid token. Play around with reloading the frame for the various states and observe there shouldn't be flashing of any error state, and notice that you cannot install when the token is invalid. You may notice a small flash of a component that renders and then unmounts when the token being applied is invalid, but I feel this is a small price to pay when everything else is working great right now. Not letting good get in the way of perfect at the moment 😅 . 
2. Also play around in the same way as above, but with an already installed app.

### DIFF
--- a/apps/vercel/frontend/src/clients/Vercel.ts
+++ b/apps/vercel/frontend/src/clients/Vercel.ts
@@ -7,8 +7,12 @@ import {
   ServerlessFunction,
 } from '@customTypes/configPage';
 
+interface CheckTokenResponse {
+  ok: boolean;
+}
+
 interface VercelAPIClient {
-  checkToken: () => Promise<boolean>;
+  checkToken: () => Promise<CheckTokenResponse>;
   listProjects: () => Promise<ListProjectsResponse>;
   createDeployment: (input: CreateDeploymentInput) => Promise<Deployment>;
   getDeploymentById: (deploymentId: string) => Promise<Deployment>;
@@ -24,13 +28,13 @@ export default class VercelClient implements VercelAPIClient {
     });
   }
 
-  async checkToken(): Promise<boolean> {
+  async checkToken(): Promise<CheckTokenResponse> {
     const res = await fetch(`${this.baseEndpoint}/v5/user/tokens`, {
       headers: this.buildHeaders(),
       method: 'GET',
     });
 
-    return res.ok;
+    return { ok: res.ok };
   }
 
   async listProjects(): Promise<ListProjectsResponse> {
--- a/apps/vercel/frontend/src/locations/ConfigScreen.tsx
+++ b/apps/vercel/frontend/src/locations/ConfigScreen.tsx
@@ -71,8 +71,8 @@ const ConfigScreen = () => {
     }
 
     if (!parameters.vercelAccessToken) {
-      const isTokenEmpty = true;
-      updateTokenValidityState(isTokenEmpty);
+      // if there is no value set for the access token we will consider it valid
+      updateTokenValidityState(true);
     } else if (!hasTokenBeenValidated) {
       checkToken();
     }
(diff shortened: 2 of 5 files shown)


apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -24,13 +28,13 @@ export default class VercelClient implements VercelAPIClient {
...
 
-  async checkToken(): Promise<boolean> {
+  async checkToken(): Promise<CheckTokenResponse> {
     const res = await fetch(`${this.baseEndpoint}/v5/user/tokens`, {
       headers: this.buildHeaders(),
       method: 'GET',
     });
 
-    return res.ok;
+    return { ok: res.ok };
```
lilbitner, line 37: I adjusted the return of this to ensure that we can determine when the call is finished. This was difficult to do when the return of the call was boolean itself.

apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -36,7 +37,7 @@ const ConfigScreen = () => {
   const onConfigure = useCallback(async () => {
     const currentState = await sdk.app.getCurrentState();
 
-    if (!parameters.vercelAccessToken) {
+    if (!parameters.vercelAccessToken || !isTokenValid) {
```
lilbitner, line 40: this ensures a user cannot save the configuration if the token is not valid.

apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -47,9 +48,13 @@ const ConfigScreen = () => {
...
     };
-  }, [parameters, sdk]);
+  }, [parameters, sdk, isTokenValid]);
 
-  const vercelClient = new VercelClient(parameters.vercelAccessToken);
+  useEffect(() => {
+    if (parameters.vercelAccessToken) {
+      setVercelClient(new VercelClient(parameters.vercelAccessToken));
+    }
+  }, [parameters.vercelAccessToken]);
```
lilbitner, line 57: this was a great suggestion from @rusticpenguin in a pervious PR, to ensure we only instantiate this Client once. This also ensures that we only instantiate when an access token exists, to avoid the error that occurs when communicating with the Vercel API without a token of any sort.

jsdalton (approved):
I had one small comment just to make things a tad more clear, but otherwise looking good 👌

apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -59,14 +64,16 @@ const ConfigScreen = () => {
...
+      if (vercelClient) {
+        const response = await vercelClient.checkToken();
+        if (response) updateTokenValidityState(response.ok);
+      }
     }
 
-    if (!hasTokenBeenValidated) {
+    if (!parameters.vercelAccessToken) {
+      const isTokenEmpty = true;
+      updateTokenValidityState(isTokenEmpty);
```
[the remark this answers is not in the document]
lilbitner: ah good point yeah I totally see how this was confusing, have made this adjustment! 🚀

PR #7385, opened 2024-04-23
feat: add fallback component for an API path

lilbitner:
## Purpose

The purpose of this PR is to add the fallback component that will render if the fetched API paths are 1. not feasibly fetched or 2. if the results are empty.

## Approach

If there is an error with the API path endpoint or if the results are empty, this fallback input is rendered. 

Rendering this fallback component when the results are empty is something I am wavering on, as it seems to me that if the results are empty, their application is not configured correctly in the first place, so should we evens support this? But I also could perhaps since this API path endpoint is potentially unreliable, the endpoint just may at some point not return anything or return differing data, even if the user has set-up their application appropriately. 

The error handling I have added also denotes the data as "invalid" if there is any presence of an empty array with the `apiPaths` state on the `ConfigScreen`. 

I scope this PR to just work on handling edge cases with this API path and no other endpoints or functionality. 

Screenshot 2024-04-23 at 12 07 55 PM

## Testing steps

Select a project on the Vercel config page that does not have any of these routes configured (there is at least one). Notice how the API path Select field is actually an input instead. You should be able to save and update as before.

### DIFF
--- a/apps/vercel/frontend/src/locations/ConfigScreen.tsx
+++ b/apps/vercel/frontend/src/locations/ConfigScreen.tsx
@@ -117,17 +117,19 @@ const ConfigScreen = () => {
     }
 
     if (parameters.selectedProject) {
-      // reset the selected api path only when the project changes
-      if (apiPaths.length) {
-        dispatchParameters({
-          type: actions.APPLY_API_PATH,
-          payload: '',
-        });
-      }
-
       getApiPaths();
     }
   }, [parameters.selectedProject, vercelClient]);
+
+  useEffect(() => {
+    if (parameters.selectedProject && !isLoading && !isAppConfigurationSaved) {
+      // reset the selected api path only when the project changes
+      dispatchParameters({
+        type: actions.APPLY_API_PATH,
+        payload: '',
+      });
+    }
+  }, [parameters.selectedProject, isLoading, isAppConfigurationSaved]);
 
   const updateTokenValidityState = (tokenValidity: boolean) => {
     setIsLoading(false);
--- a/apps/vercel/frontend/src/locations/ConfigScreen.tsx
+++ b/apps/vercel/frontend/src/locations/ConfigScreen.tsx
@@ -110,8 +110,13 @@ const ConfigScreen = () => {
     async function getApiPaths() {
       setIsLoading(true);
       if (vercelClient) {
-        const data = await vercelClient.listApiPaths(parameters.selectedProject);
-        setApiPaths(validateApiPathData(data) ? data : []);
+        try {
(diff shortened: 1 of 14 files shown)


apps/vercel/frontend/src/utils/validateApiPathData/validateApiPathData.ts
```diff
@@ -0,0 +1,4 @@
+import { ApiPath } from '@customTypes/configPage';
+
+export const validateApiPathData = (apiPathData: ApiPath[]) =>
+  Array.isArray(apiPathData) && apiPathData.length > 0;
```
lilbitner, line 4: this validation is what i would like the most input on. Should we show the input field instead of the select field if the array is empty? See my thought process in the PR description, but I think there are pros and cons here.

apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -109,26 +110,33 @@ const ConfigScreen = () => {
...
       setIsLoading(true);
       if (vercelClient) {
-        const data = await vercelClient.listApiPaths(parameters.selectedProject);
-        setApiPaths(data || []);
+        try {
+          const data = await vercelClient.listApiPaths(parameters.selectedProject);
+          setApiPaths(validateApiPathData(data) ? data : []);
+        } catch (e) {
+          console.error(e);
+          setApiPaths([]);
```
lilbitner, line 118: the `apiPaths` data will be an empty array if there is an error or if the data is not "valid"

apps/vercel/frontend/src/utils/validateApiPathData/validateApiPathData.ts
jsdalton: My thoughts:

* Agreed that showing the input field if paths is empty or we hit an error (either)
* But I think we should have a warning message for each and it should be different:
  * if error: "⚠️ We had trouble fetching routes for this Vercel project."
  * if empty: " ⚠️ It looks like you haven't configured any routes yet."
* both case: display instruction text

apps/vercel/frontend/src/locations/ConfigScreen.tsx
jsdalton: See below. I _think_ we might want to throw an error here and then catch and it handle upstream

jsdalton (approved):

apps/vercel/frontend/src/utils/validateApiPathData/validateApiPathData.ts
lilbitner: Okay I like the idea of giving the user instruction 👍 will add a follow-up task here

PR #7430, opened 2024-04-26
feat: add link to help text for api path

jsdalton:
## Purpose

The naming and help text for choosing your draft mode route handler are confusing at the moment.

## Approach

* Update the copy and provide a link to the guide
* Refactor components a bit to support providing a react node and not just a string for help text
* UPDATE LINK is intentionally there and ugly, to remind us that we need to update it with the real link once we publish the guide

## Testing steps

Before:

image

After:

image


matthew-gordon (approved):
LGTM ✅

PR #7450, opened 2024-04-29
feat: scope token to teamId and add frontend error handling functionality for Vercel config page

lilbitner:
## Purpose

The purpose of this PR is to accomplish the following: 
1. Ensure that we are scoping our API communication with Vercel to the provided `teamId` that is associated with the access token. 
2. Communicate all errors/warnings effectively to the user.

## Approach

1. In order to implement the `teamId` scope, I created another method within the `VercelClient` to fetch the "current" token metadata. With this data, if the token metadata contains a `teamId`, this is stored as part of the installation state, and subsequently used as a param when fetching the projects and API paths. If the token does not contain a `teamId`, an error is thrown and appropriately shown to the user. 
2. In order to effectively manage errors on the frontend, I implemented an `errorsReducer`, similar to our current `parametersReducer`. Originally errors were handled on a one-off basis, but this was not scalable, especially with edge cases continually being found. Instead, we can now easily manage the strictly typed states of errors appropriately by using the store as the single source of truth. 

With the above work I created a handful of new hooks that abstract logic from jumbled component code: 
1. `useError` hook returns the appropriate error message based on the current error supplied, and indicates if an error exists at all. 
2. `useFetchAndValidateData` hook returns three methods to effectively handle communication with the token, project, and API path API. 

## Testing steps

Access token section: 
1. Paste an invalid token (i.e '1234') and notice the error message that appears. 
2. Paste a token that is scoped to the entire account (not a team) and notice the error message that appears. 
3. Paste a valid token scoped to a team, and notice no error message. 

Project and API path sections are a bit tricky to test, but you can create a project within Vercel, configure that project on the configuration page, and then delete the project to notice an error message. 

Content type/preview path section:
1. Save an empty preview path, and notice the error message that appears. 
2. Save another preview path that is formatted incorrectly, and notice the error message that appears. 

## Follow-up work

- I have not yet added much nuanced handling of when there are no API paths retrieved (and we instead render a text input). There is a follow-up ticket to capture this work.


apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -28,57 +29,56 @@ export default class VercelClient implements VercelAPIClient {
...
+  private buildTeamIdQueryParam(teamId?: string): string {
+    return teamId ? `&teamId=${teamId}` : '';
+  }
 
-    return { ok: res.ok };
+  async checkToken(): Promise<GetToken> {
+    const res = await this.getToken();
+    const { token } = await res.json();
+
+    if (!res.ok) throw new Error(errorTypes.INVALID_TOKEN);
```
jsdalton, line 40: Should we move this one line above? In case eg there wasn't a json response? Not sure it matters.

apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -87,43 +87,51 @@ export default class VercelClient implements VercelAPIClient {
...
 
     return data;
   }
 
-  async getDeploymentById(deploymentId: string): Promise<Deployment> {
-    const res = await fetch(`${this.baseEndpoint}/v13/deployments/${deploymentId}`, {
-      headers: this.buildHeaders(),
-      method: 'GET',
-    });
+  async getLatestDeploymentId(projectId: string, teamId?: string): Promise<string> {
```
jsdalton, line 120: maybe we want this to be called `getLatestDeploymentById` (include the word "By')

apps/vercel/frontend/src/hooks/useError/useError.tsx
```diff
@@ -0,0 +1,51 @@
...
+  | keyof Errors['apiPathSelection']
+  | keyof Errors['projectSelection']
+  | keyof Omit<PreviewPathError, 'contentType'>;
+
+export const getErrorMessage = (error: Error, contentType?: string) => {
+  const currentError = pickBy(error);
+  let errorKey: ErrorKeys;
+
+  if (contentType) {
+    errorKey = Object.keys(currentError)[1] as ErrorKeys;
```
jsdalton, line 23: pretty minor but could this whole block be a bit cleaner with a ternary operator instead of the mutable variable and if clause?

jsdalton (approved):
Okay, I had just a few small suggestions but nothing blocking. Looking good!

apps/vercel/frontend/src/clients/Vercel.ts
lilbitner: oh good idea - this here is grabbing the latest deployment ID itself, so not the latest deployment by ID. Though maybe I will make it more generically return the latest deployment, and the consuming method can grab the ID as it sees fit 👍 will adjust!

apps/vercel/frontend/src/hooks/useError/useError.tsx
lilbitner: sure! Have adjusted this 👍

apps/vercel/frontend/src/clients/Vercel.ts
lilbitner: Yeah good catch 🦅 , have adjusted this!

matthew-gordon (approved):
Re-read the resolved issues and checked against current files, things make sense and look great to me 👍 ✅

PR #7453, opened 2024-04-29
fix: fix flashing bug of components when clicking on Content type selection

lilbitner:
## Purpose

The purpose of this PR is to quickly resolve a bug that is occurring on the Vercel config page, when a user interacts with the content type/preview path selection section, the component itself umounts and Draft mode path selection is cleared. This unmounting should not be happening. 

## Approach

I simply removed the problem `useEffect` on the Config page and moved it to the logic where the project is being changed. This logic is to essentially clear the API path parameter when the project is changed, so it has just been moved closer to that logic.


matthew-gordon (approved):
Ahh great catch! LGTM! ✅

PR #7566, opened 2024-05-03
feat: add error handling for selected projects that do not have protection enabled

rusticpenguin:
## Purpose
When a user selects a project, that project may not be setup correctly to be displayed in our Preview. We require the project to have protection bypass enabled.

## Approach

- Add logic around checking if the protection bypass is enabled
- Add case to the error reducer

## Testing steps

1. Go to Vercel and disable protection bypass for your project
1. Go to your app's config page
1. Select the project with the disabled protection bypass
1. Verify the error message appears below the project selector


apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -76,7 +76,7 @@ const ConfigScreen = () => {
 
     async function checkToken() {
       if (vercelClient) {
-        validateToken(updateTokenValidityState);
+        await validateToken(updateTokenValidityState);
```
lilbitner, line 79: thank you for adding these awaits 💯 !!

lilbitner (changes requested):
Code looks great George! Thanks for adding those awaits and for always adding clean and readable code. 
In terms of functionality, when I test it out locally, I am seeing some clobbering logic within the `ProjectSelectionSection` file. The two `useEffects` are essentially overriding each other so the error is not persisting in the UI (as the bypass error is cleared by the `validateProjectEnv` method. I think we should still follow an approach that there should only exist one error at a time for the project selection step, so perhaps we can combine the `useEffects` and make the steps sequential in some way? i.e we check that the project has a valid spaceId and then if so, check the bypass token? or vise versa? Also feel free to adjust the logic in the `validateProjectEnv` method, as it clears all errors if the `spaceId` is valid, which may not be necessary?

apps/vercel/frontend/src/locations/ConfigScreen.tsx
```diff
@@ -158,6 +153,7 @@ const ConfigScreen = () => {
     renderPostAuthComponents &&
     !errors.projectSelection.projectNotFound &&
     !errors.projectSelection.cannotFetchProjects &&
+    !errors.projectSelection.protectionBypassIsDisabled &&
```
lilbitner, line 156: I think it might be fine to follow the pattern here that we still render the next section even if this error exists, because it is simply a setting they have to change. Just like we allow them to continue to select the next section if the `projectSelection.invalidSpaceId` is true (which is a setting they need to change). Or perhaps we choose one line of logic or the other! wdyt? See video below for more context (on the weird standard we have for one but not the other). We kind of already have a pattern where things will NOT work with certain configurations but we still allow them to save most of the changes (i.e in the content type/preview path section we allow them to save empty fields even though that clearly won't work). 

https://github.com/contentful/apps/assets/58186851/76661919-44ac-4a3c-9228-2534e296f62f

apps/vercel/frontend/src/hooks/useFetchData/useFetchData.tsx
```diff
@@ -103,20 +103,30 @@ export const useFetchData = ({
...
         }
+
+        const protectionBypassIsDisabled = selectedProject && !selectedProject.protectionBypass;
+        if (protectionBypassIsDisabled) {
+          throw new Error(errorTypes.PROTECTION_BYPASS_IS_DISABLED);
+        }
+
+        dispatchErrors({
+          type: errorsActions.RESET_PROJECT_SELECTION_ERRORS,
+        });
```
lilbitner, line 129: awesome consolidation!!!

lilbitner (commented):
Just leaving one comment, curious your thoughts!! (more a UX thing). Awesome work!

apps/vercel/frontend/src/locations/ConfigScreen.tsx
rusticpenguin: Sounds good with me!

lilbitner: perhaps we remove this line from the conditional then?

apps/vercel/frontend/src/constants/errorMessages.ts
```diff
@@ -4,6 +4,8 @@ export const errorMessages = {
   expiredToken: 'Token has expired.',
   projectNotFound:
     'The project you have configured is no longer available. Please select another one.',
+  protectionBypassIsDisabled:
+    'The project you have configured does not have protection bypass enabled. Please enable it.',
```
jsdalton, line 8: Can we update the error message here? 

> The project you have configured does not have protection bypass enabled. Please enable it. Note that your Vercel project will need to be redeployed in order for the change to take effect.

jsdalton (changes requested):

jsdalton (approved):

whitelisab (approved):

PR #7587, opened 2024-05-06
feat: add error message when project not aligned to spaceId

lilbitner:
## Purpose

The purpose of this PR is to add an error message when the user selects a project that is not configured to be in the space where the current app is installed/being installed. 

## Approach

I added the below error message on render of the page AND when the user changes project, and the selected project contains an incompatible `CONTENTFUL_SPACE_ID` variable. The error lingers until the user changes to a valid project, but they are still able to install and save changes. 

Screenshot 2024-05-07 at 7 07 56 AM

## Testing steps

Within [this](https://app.contentful.com/spaces/5rotc6q2lcs7/apps/656HN4kwcFsgd1lmZ5BhvP) space, I have created one Vercel project that is compatible with the space:` nextjs-lb-4`. The rest of the projects are not compatible, so you should see warnings for all other project selections.

### DIFF
--- a/apps/vercel/frontend/src/clients/Vercel.ts
+++ b/apps/vercel/frontend/src/clients/Vercel.ts
@@ -162,7 +162,7 @@ export default class VercelClient implements VercelAPIClient {
     teamId: string
   ) {
     try {
-      const data = await this.listProject(projectId, teamId);
+      const data = await this.getProject(projectId, teamId);
       const envs = data.env;
       const contentfulSpaceIdEnv = envs.find((env) => env.key === CONTENTFUL_SPACE_ID);
       if (contentfulSpaceIdEnv) {
@@ -187,7 +187,7 @@ export default class VercelClient implements VercelAPIClient {
     return false;
   }
 
-  private async listProject(projectId: string, teamId: string): Promise<Project> {
+  private async getProject(projectId: string, teamId: string): Promise<Project> {
     let projectData: Response;
     try {
       projectData = await fetch(
--- a/apps/vercel/frontend/src/clients/Vercel.ts
+++ b/apps/vercel/frontend/src/clients/Vercel.ts
@@ -167,7 +167,9 @@ export default class VercelClient implements VercelAPIClient {
       const contentfulSpaceIdEnv = envs.find((env) => env.key === CONTENTFUL_SPACE_ID);
       if (contentfulSpaceIdEnv) {
         const res = await fetch(
-          `${this.baseEndpoint}/v1/projects/${projectId}/env/${contentfulSpaceIdEnv.id}`,
+          `${this.baseEndpoint}/v1/projects/${projectId}/env/${
+            contentfulSpaceIdEnv.id
+          }?${this.buildTeamIdQueryParam(teamId)}`,
           {
             headers: this.buildHeaders(),
             method: 'GET',
--- a/apps/vercel/frontend/src/clients/Vercel.ts
+++ b/apps/vercel/frontend/src/clients/Vercel.ts
@@ -180,7 +180,7 @@ export default class VercelClient implements VercelAPIClient {
         return data.value === currentSpaceId;
       }
     } catch (e) {
-      // let user continue if there is an validating env value
+      // let user continue if there is an error validating env value
       console.error(e);
     }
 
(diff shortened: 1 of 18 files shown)


apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -146,6 +156,56 @@ export default class VercelClient implements VercelAPIClient {
...
+            headers: this.buildHeaders(),
+            method: 'GET',
+          }
+        );
+
+        const data = await res.json();
+        return data.value === currentSpaceId;
+      }
+    } catch (e) {
+      // let user continue if there is an error validating env value
```
lilbitner, line 183: if there is an error using these endpoints to actually validate the value at some point, I opted to let the user continue. Reason being that this is a nice bonus in the background, and if it is not working correctly, there doesn't seem to be a need for the user to know about it.

apps/vercel/frontend/src/clients/Vercel.ts
```diff
@@ -146,9 +156,61 @@ export default class VercelClient implements VercelAPIClient {
...
+      console.error(e);
+    }
+
+    return false;
+  }
+
+  private async listProject(projectId: string, teamId: string): Promise<Project> {
```
jsdalton, line 190: Super small naming thing but I tend to think of "list" as returning a collection and would suggest a better verb like `getProject` or `fetchProject`.

apps/vercel/frontend/src/hooks/useFetchData/useFetchData.tsx
```diff
@@ -102,5 +102,30 @@ export const useFetchData = ({
...
+        );
+        if (!isProjectSelectionValid) throw new Error(errorTypes.INVALID_SPACE_ID);
+        else {
+          dispatchErrors({
+            type: errorsActions.RESET_PROJECT_SELECTION_ERRORS,
+          });
+        }
+      } catch (e) {
+        const err = e as Error;
+        if (err.message === errorTypes.INVALID_SPACE_ID) {
```
jsdalton, line 121: Not a super big deal but I tend to like the pattern of creating a custom error type and then handling that error type:

```ts
// custom error class
class InvalidProjectSpaceId extends Error {}

// execution
if (!isProjectSelectionValid) throw new InvalidProjectSpaceId(`The current space ID '${currentSpaceId}' doesn't match the space id defined for the project`)

// handling
if (err instanceof InvalidProjectSpaceId) {
  // handle it
}
```

Even if you don't handle the message now it's just an easier way to do something with it in the future

jsdalton (commented):
Code looks good, just one or two small suggestions.

A few questions about the error message:

* Should this be a warning? I feel like we don't have enough confidence that it's wrong to show a true error.
* Probably would be nice to show the value of that env var in the message? I guess it might help allay confusion.

apps/vercel/frontend/src/hooks/useFetchData/useFetchData.tsx
lilbitner: Yeah awesome suggestion! I will definitely note this for the future 👍

apps/vercel/frontend/src/clients/Vercel.ts
lilbitner: great callout, have adjusted this!

lilbitner:
> Code looks good, just one or two small suggestions.
> 
> A few questions about the error message:
> 
> * Should this be a warning? I feel like we don't have enough confidence that it's wrong to show a true error.
> * Probably would be nice to show the value of that env var in the message? I guess it might help allay confusion.

Yeah good questions. 
1. In my opinion, we do have a solid amount of confidence that there is a mismatch of `spaceIds` here and that the app will not work without this being fixed (per my understanding, I could be missing something). Along with this, there really isn't precedent in our design system for a "warning" type for the Select component. It is fairly binary (valid or invalid). I could certainly create custom component to add here instead, but I do think it is breaking enough to warrant an error.
2. Yeah I also considered this. This adds a bit more complexity to the error handling, but I could do this as part of some follow-up work, to make it easier for the user to adjust the spaceId if they need to 👍

matthew-gordon (approved):
Looks good and makes sense to me 🤌 ✨

PR #7631, opened 2024-05-07
feat: use fallback url [MONET-2032]

cemreyuksel:
## Purpose

[TICKET](https://contentful.atlassian.net/browse/MONET-2032)

This PR adds a fallback url so that users do not get confused with configuring it.


matthew-gordon (approved):
🙏 makes sense to me! ✅ ✨

pciazynski (approved):
LGTM!

PR #7633, opened 2024-05-07
fix: set installationUuid parameter on initial install [INTEG-1768]

whitelisab:
## Purpose

Currently, the slack workflow on the configuration page requires a user to save the installation after authenticating with the Slack workspace in order to successfully render the listed channels. If the user does not save the configuration page before trying to configure a notification, the channels modal shows a warning that channels cannot be listed. 

## Approach

When fetching channels, the app first needs to fetch the auth token from the auth token repository. We were receiving an error "Auth token not found" when trying to get channels as we were hitting [this line](https://github.com/contentful/apps/blob/master/apps/slack/lambda/lib/routes/auth-token/repository.ts#L100). In this get method, we are getting the installation parameters (including the installationUuid) from the CMA. I was able to verify that this was coming back as undefined after initial installation, so we were trying to look for the auth token based on the spaceEnvUuid and were not finding the auth token.  

In order to ensure that the installation uuid is available in the installation parameters after initial authentication/installation, we need to set this parameter. The Slack app currently installs the app after successfully authenticating using a custom api method `api.install`. In this method you can pass in parameters to save, and previously the only being passed in was the workspace id. This PR updates that to also set the installation uuid upon successful authentication. 

Screenshot 2024-05-07 at 8 57 28 AM

## Testing steps

I tested this locally when initially installing the app and was able to verify that I could see the channels in the modal without having to "save" the app. I also tested this when the Slack app is already installed and when you have to re-authenticate to ensure that the app still behaves as expected.

## Breaking Changes

This should not introduce any breaking changes.

### DIFF
--- a/apps/slack/frontend/src/useConnect.ts
+++ b/apps/slack/frontend/src/useConnect.ts
@@ -28,8 +28,8 @@ export const useConnect = () => {
 
       if (token) {
         try {
-          await api.install({ workspaces: [workspace.id] });
           const installationUuid = uuidv4();
+          await api.install({ workspaces: [workspace.id], installationUuid });
           setInstallationUuid(installationUuid);
 
           await apiClient.createAuthToken(sdk, cma, token, installationUuid);


ethan-ozelius-contentful (approved):
good find!

lilbitner (approved):
Awesome Lisa! This makes a lot of sense 👍 totally makes sense that it needs that reference to the UUID. 

As an unrelated aside, I started testing locally but ran into some issues with NGROK (which I think you had already mentioned at some point given we lost our private account). I think you already have a ticket to update the NGROK instructions for the Slack app? 

Great job!!! 👏

whitelisab:
> Awesome Lisa! This makes a lot of sense 👍 totally makes sense that it needs that reference to the UUID.
> 
> As an unrelated aside, I started testing locally but ran into some issues with NGROK (which I think you had already mentioned at some point given we lost our private account). I think you already have a ticket to update the NGROK instructions for the Slack app?
> 
> Great job!!! 👏

Yes, I'll make sure we have a ticket for cleaning up the instructions for running the app locally! 🧹

PR #7636, opened 2024-05-07
chore(dropbox): update to dam-app-base v3.0.0 and update to react v18

ethan-ozelius-contentful:
## Purpose
Update the Dropbox app's version of react to 18, as well as update dam-app-base to v3.0.0

## Breaking Changes
v3 of the dam-app-base is a breaking change for consumers that are on a version of react less than 18.0.0.  Fortunately for the dropbox app, we don't render any actual react code, instead depending on dam-app-base to render react code.  So no changes to the dropbox app necessary.

See [dam-app-base changelog](https://github.com/contentful/apps/blob/master/packages/dam-app-base/CHANGELOG.md#300-2024-05-07)

### DIFF
--- a/apps/dropbox/package-lock.json
+++ b/apps/dropbox/package-lock.json
@@ -8,9 +8,9 @@
       "name": "@contentful/dropbox-assets",
       "version": "1.6.74",
       "dependencies": {
-        "@contentful/dam-app-base": "2.0.55",
-        "react": "16.12.0",
-        "react-dom": "16.12.0"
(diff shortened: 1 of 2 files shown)


ethan-ozelius-contentful:
@lilbitner 

> This makes sense to me 👍 were you able to test that all is working normally with Dropbox locally?

Yep, I ran an `npm pack` in dam-app-base, and then installed that tar file in dropbox.  That's how I found this little CSS bug that will be fixed in the [PR](https://github.com/contentful/apps/pull/7637).

PR #7637, opened 2024-05-07
fix(dam-app-base): sortable images should have min height

ethan-ozelius-contentful:
## Purpose
**Bug:** A minor CSS bug was discovered with the new version of dndkit that replaced react-sortable-hoc for sortable components in dam-app-base.  When dragging/sorting components, if the heights of the components vary, it creates a jarring UX.  See video:

**Broken**
borked-sort

Resolution is a quick style fix to sortable cards.  Something must have changed when migrating from react-sortable-hoc to dndkit.  just a few lines of CSS.

**Fixed**
fixed-sort

## Testing steps
Tested locally, see videos.


lilbitner (approved):
Great fix!! 👏

PR #7672, opened 2024-05-10
APPX-8397 Correction for prettier-check fail error for app-test in contentful repo

Usha-Kundapur:
## Purpose
- Correction for prettier-check fail error for app-test in contentful repo


Usha-Kundapur:
Updated in original PR here https://github.com/contentful/apps/pull/7669

PR #7632, opened 2024-05-07
feat: add a search input for Potter characters []

Vinz93:
## Purpose

You can search your favourite character now! 🪄 🎉 
Screenshot 2024-05-07 at 15 14 59

PR #7713, opened 2024-05-14
feat: add characters search bar to potter gql []

Vinz93:
## Purpose

Adds search bar same as our [Rest example ](https://github.com/contentful/apps/pull/7632)


cemreyuksel (approved):
🎉

PR #7903, opened 2024-05-30
feat: add contentful preview secret input

malewis5:
## Purpose

We want live preview to work for Hobby, Pro and Enterprise accounts. Currently, we use Vercel Protection Bypass Secret to validate the request from Contentful to Vercel. However, Hobby accounts do not have access to Vercel Protection Bypass. Therefore, we need to set an environment variable that's responsible for validating the request came from Contentful.

## Approach

Adding this input for Contentful Preview Secret allows a user to install the integration on their Contentful Space and update their Vercel project's environment variables at the same time. This input will check to make sure the environment variable doesn't already exist, explain to the user that they need to delete it in their Vercel project if it does exist, and then will save the secret to Contentful and Vercel to make sure it's in sync and the same value.

## Testing steps

Initial Setup:
1. Install the app
2. Create access token and open that project in Vercel
3. Go to environment variables for that project
4. Make sure there is not a CONTENTFUL_PREVIEW_SECRET variable.

Fresh Install Test:
1. Install the app
2. Create access token and open that project in Vercel
3. Select the same project in Contentful
4. Enter a random value for the Contentful Preview Secret
5. Save
6. You should see a disabled field in Contentful saying the variable exists
7. You should see the variable in your Vercel environment variables for the project
8. These two values should be equal

Possible Collision Test:
1. Install the app
2. Create access token and open that project in Vercel
3. Create an environment variable on Vercel in that project called CONTENTFUL_PREVIEW_SECRET with any value
4. Select that project in Contentful
5. You should see a disabled field AND an error telling you to delete the variable in Vercel.
6. Delete the variable in Vercel
7. Click "Retry"
8. The field should be enabled, blank and allow you to save
9. Environment variable should be present in Vercel after you save.

## Breaking Changes

Yes, this change is also dependent on the SDK being updated. First, we need to send this CONTENTFUL_PREVIEW_SECRET value in a header (best case) or query param (worst case). Then, the SDK needs to validate the request using this secret instead of the Vercel Protection Bypass cookie.

Talk to @jsdalton or @malewis5

## Deployment

Do not deploy this change until the SDK has been tested alongside it.
We should also write better tests, this isn't my strong suite.

### DIFF
--- a/apps/vercel/frontend/src/clients/Vercel.ts
+++ b/apps/vercel/frontend/src/clients/Vercel.ts
@@ -30,7 +30,11 @@ interface ListEnvironmentVariables {
 interface VercelAPIClient {
   checkToken: () => Promise<GetToken>;
   getToken: () => Promise<Response>;
-  updateEnvironmentVariable: (variable: string, variableName: string, projectId: string) => Promise<UpdateEnvironmentVariable>;
+  updateEnvironmentVariable: (
+    variable: string,
+    variableName: string,
+    projectId: string
+  ) => Promise<UpdateEnvironmentVariable>;
   listProjects: (teamId?: string) => Promise<ListProjectsResponse>;
   listApiPaths: (projectId: string, teamId?: string) => Promise<ApiPath[]>;
   validateProjectContentfulSpaceId: (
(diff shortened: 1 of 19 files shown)


whitelisab:
Closing this PR since Protection Bypass for Automation is available on all Vercel plans.

PR #7431, opened 2024-04-26
feat(dam-app-base): replace react-searchable-hoc with dndkit

ethan-ozelius-contentful:

PR #7932, opened 2024-06-03
fix: use index for deleting asset [ZEND-5025]

whitelisab:
## Purpose

[This PR](https://github.com/contentful/apps/pull/7431/files) to get the `dam-app-base` ready to support React 18 introduced some changes to the `deleteItem` login in the `SortableComponent`. These changes caused issues with deleting, where when clicking on an item in the list to delete, the first item instead got deleted.

## Approach

The `key` was being used to splice the item from the list, but this turned out to be the URL of the asset, so this was getting passed as `NaN` and then removing the first item in the list. This PR updates the `deleteItem` logic to instead use index instead of key.

## Testing steps

I tested this locally using `npm pack` to test the changes to the `dam-app-base` within the Bynder app.

## Deployment

After updating the `dam-app-base` we will open a PR for the Bynder app to bump the package version.

### DIFF
--- a/packages/dam-app-base/src/Editor/SortableComponent.tsx
+++ b/packages/dam-app-base/src/Editor/SortableComponent.tsx
@@ -242,12 +242,12 @@ export const SortableComponent = ({
   const [items, setItems] = useState<AssetWithId[]>([]);
 
   const deleteItem = useCallback(
-    (index: UniqueIdentifier) => {
+    (index: number) => {
       const myResources = [...resources];
-      myResources.splice(Number(index), 1);
+      myResources.splice(index, 1);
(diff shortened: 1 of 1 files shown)


ethan-ozelius-contentful (approved):

ethan-ozelius-contentful:
Looks solid, really good find.  Sorry about introducing this bug when upgrading from react 16 || 17 - 18.  Obviously non-blocking since this bug in currently in production and should be fixed asap, but this functionality feels like something that can be unit-tested pretty easily.  Might be worth a 1-pointer task to follow up someday.

mgoudy91 (approved):

PR #7933, opened 2024-06-03
fix: handle teamId in Vercel API call

jsdalton:
## Purpose

* Team ID is required in Vercel API calls. We've added it to all our frontend API calls but missed updating the app action
* We should handle error responses from the Vercel API (at least acknowledge what went wrong and capture the error)

## Approach

* Add the teamId as a query param. Since it's "required" by the frontend now we will assume it's present in the app installation parameter
* Add a rudimentary error handler to the app action Vercel api service

### DIFF
--- a/apps/vercel/app-actions/src/services/vercel-service.ts
+++ b/apps/vercel/app-actions/src/services/vercel-service.ts
@@ -8,7 +8,7 @@ export class VercelService {
       method: 'GET',
       headers: this.buildRequestHeaders(),
     });
-    await this.maybeThrowApiError(response);
+    await this.handleApiError(response);
     const vercelProject = await response.json();
     this.assertVercelProject(vercelProject);
     return vercelProject;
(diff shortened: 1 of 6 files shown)


apps/vercel/app-actions/src/services/vercel-service.ts
```diff
@@ -1,18 +1,29 @@
...
     });
+    await this.maybeThrowApiError(response);
     const vercelProject = await response.json();
     this.assertVercelProject(vercelProject);
     return vercelProject;
   }
 
+  // cheap and dirty error handling -- could later provide a more structured error response
+  // when API errors are encountered
+  private async maybeThrowApiError(response: Response): Promise<void> {
```
lilbitner, line 19: this error handling definitely works for now 💯 nit: wondering if the name of this method could best reflect what it is doing, i.e `handleApiError` or `checkResponseStatus`, instead of what the results _potentially_ are? totally fine with the name as is now, just as I was reading the method on line 11, I definitely had to look at the method itself to fully understand what it was doing.

lilbitner (approved):
Great 🥇 !

apps/vercel/app-actions/src/services/vercel-service.ts
jsdalton: It's a good point. I think `handleApiError` is a lot more obvious and close enough to what it does.

t-col (approved):

PR #7666, opened 2024-05-09
build(react18)!: drop support for react v16 and v17

ethan-ozelius-contentful:
## Purpose
1. Update Ecommerce-app-base to react v18
2. implement new style root.render(), as opposed to render(, node). [see react article](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis)
3. Swap out `react v16 >=` `react-sortable-hoc` library with new dndkit that supports react v18 <=

**BREAKING CHANGE**: Dropping support for `react-dom <= 17` and `react <= 17`.  Consumers of ecommerce-app-base are required to implement react v18.

#### Shopify
Tested locally by `npm pack`'ing a `ecommerce-app-base` tar file, and installing it locally inside shopify.
shopify

### Also tested with Storybook
sortable-storybook

PR #7944, opened 2024-06-04
Fix ecommerce-app-base bug in drag-n-drop refetch logic

ethan-ozelius-contentful:
## Purpose
Fix a bug that was introduced in this [PR](https://github.com/contentful/apps/pull/7666) that was causing an infinite loop rerender to occur.

**Root cause**
1. On initial render, `usePreviousSkus()` hook is invoked to track `skus` from previous mount, 
2. This [useEffect](https://github.com/contentful/apps/blob/master/packages/ecommerce-app-base/src/Editor/SortableComponent.tsx#L75C3-L82) invokes `fetchProductPreviews()` on mount, as expected.
3. `fetchProductPreviews()` retrieves product previews via network request
4. `fetchProductPreviews()` updates the `productPreviews` state, causing a rerender, as expected.
5. on rerender, `usePreviousSkus()` runs and retrieves the list of skus from previous mount, as well as updates the current ref to the most current list of skus (list of skus received via props).
6. since the useEffect (see number 1) lists `previousSkus` as a dependency, the useEffect runs again.  Setting off an **infinite loop**.

### Screen capture of working solution
shopify-fixed

### DIFF
--- a/packages/ecommerce-app-base/src/AppConfig/AppConfig.spec.tsx
+++ b/packages/ecommerce-app-base/src/AppConfig/AppConfig.spec.tsx
@@ -5,53 +5,7 @@ import { ConfigAppSDK } from '@contentful/app-sdk';
 
 import AppConfig from './AppConfig';
 import { definitions } from './parameters.spec';
-
-const contentTypes = [
-  {
-    sys: { id: 'ct1' },
-    name: 'CT1',
-    fields: [
-      { id: 'product_x', name: 'Product X', type: 'Symbol' },
-      { id: 'y', name: 'Y', type: 'Object' },
-    ],
-  },
-  {
-    sys: { id: 'ct2' },
-    name: 'CT2',
-    fields: [
-      { id: 'foo', name: 'FOO', type: 'Text' },
-      { id: 'z', name: 'Z', type: 'Array', items: { type: 'Symbol' } },
-    ],
-  },
-  {
-    sys: { id: 'ct3' },
-    name: 'CT3',
-    fields: [
-      { id: 'bar', name: 'BAR', type: 'Object' },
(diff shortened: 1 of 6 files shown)


packages/ecommerce-app-base/src/AppConfig/AppConfig.spec.tsx
```diff
@@ -5,53 +5,7 @@ import { ConfigAppSDK } from '@contentful/app-sdk';
 
 import AppConfig from './AppConfig';
 import { definitions } from './parameters.spec';
-
```
ethan-ozelius-contentful, line 8: Copied/pasted this code out of `AppConfig` into `__mocks__` dir so that other test files can re-use this functionality.

packages/ecommerce-app-base/src/Editor/SortableComponent.tsx
```diff
@@ -57,28 +57,31 @@ export const SortableComponent: FC<Props> = ({
   const [productPreviews, setProductPreviews] = useState<Product[]>([]);
   const previousSkus = usePreviousSkus(skus);
 
-  const getProductPreviews = useCallback(
```
ethan-ozelius-contentful, line 60: removed `shouldRefetch` arg from this callback and reversed the dependency, so that consumers of this callback are responsible for determining `shouldRefetch`.  See [line 82](https://github.com/contentful/apps/pull/7944/files#diff-83200e90393b598c4e2f6fe5c314900c465f4d735d01a22a63fefa4f0c5b64edR82)

t-col (approved):

whitelisab (approved):
Looks good! Thank you for the helpful comments and tests in this PR 🚀

PR #8117, opened 2024-06-18
fix upgrading ws

david-shibley-contentful:
## Purpose

To resolve security vulnerabilities


sarahlessner:
I am going to close this one since it's quite old, and we have done a lot of other dependency updates across our apps causing lots of conflict here. Additionally, most of the `ws` vulnerabilities I've updated have involved updating underlying packages to resolve issues.

PR #8326, opened 2024-07-15
fix: revert forma36 version update in examples

whitelisab:
## Purpose

The `4.67.2` version of `@contentful/f36-components` has an issue that is causing builds to fail. This is a temporary fix to revert bumping this package in our examples. Our examples are being used by `create-contentful-app` and so using this broken version of the forma package is creating issues for apps created using CCA.


whitelisab:
Closing since we are going to implement this solution instead: https://github.com/contentful/apps/pull/8327

PR #8345, opened 2024-07-17
fix: update package-lock in examples to remove gh packages references []

whitelisab:
## Purpose

Several of our examples had `package-lock.json` files that had dependencies installed from github packages instead of from npm. I changed my `.npmrc` to point to the npm registry and re-generated these `package-lock.json` files. This also upgraded several dependencies in these examples.

Updated to also include a change to the Dropbox `package-lock.json` since it also was installing packages from github instead of npm.

Also updated root .npmrc with npm registry.

## Testing steps

Tested Dropbox app locally to ensure it still functions as expected

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 engine-strict = true
+@contentful:registry=https://registry.npmjs.org
--- a/apps/dropbox/package-lock.json
+++ b/apps/dropbox/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "@contentful/dropbox-assets",
   "version": "1.6.74",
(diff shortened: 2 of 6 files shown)


t-col (commented):
maybe we also update the repo level `.npmrc` so this doesn't happen on accident again?

whitelisab:
> maybe we also update the repo level `.npmrc` so this doesn't happen on accident again?

Good call, I updated the root `.npmrc` and I also caught that our Dropbox app was installing from github packages, so I regenerated that `package-lock.json` as well

t-col (approved):

PR #8380, opened 2024-07-22
fix: remove .rsc files from api paths.

malewis5:
## Purpose
We added more information to the Vercel deployment metadata. This is an non-versioned API route so that's why it was an undocumented breaking change. The previous team was aware this was an non-versioned API route.

## Approach
Added a filter to remove .rsc files.

## Testing steps
Configure app and under Draft Mode route handler you shouldn't see any .rsc options.

## Breaking Changes
No

N/A
## Deployment
No


whitelisab:
Hey @malewis5! Thanks for this contribution. I just re-ran our CI checks on this PR and it looks like there are prettier violations in this `apps/vercel/frontend/src/clients/Vercel.ts`. Could you address those and then I'll get this approved and merged?

malewis5:
@whitelisab I believe it's all pretty now 😄.

whitelisab (approved):

PR #8394, opened 2024-07-23
Feat: dynamic models [EXT-5427]

david-shibley-contentful:
## Purpose

Using dynamic models ensures the latest available models are options in the dropdown

image


apps/ai-content-generator/src/components/config/model/Model.tsx
```diff
@@ -1,21 +1,34 @@
...
+      const models = await getGptModels(apiKey);
+      setGptModels(models.sort((a, b) => a.id.localeCompare(b.id)));
+    }
+    fetchGptModels();
+  }, [gptModels]);
 
   const modelList = gptModels.map((model) => (
     <Select.Option key={model.id} value={model.id}>
-      {model.name}
+      {model.id}
```
whitelisab, line 31: There was some intentionality for how these are displaying on the app config page, according the "models" section of this brand guidelines: https://openai.com/brand/

apps/ai-content-generator/src/configs/ai/gptModels.ts
```diff
@@ -1,10 +1,19 @@
...
-  { id: 'gpt-3.5-turbo', name: 'GPT-3.5', textLimit: 12000 },
-  { id: 'gpt-3.5-turbo-16k', name: 'GPT-3.5 16k', textLimit: 12000 },
-  { id: 'gpt-4', name: 'GPT-4' },
-  { id: 'gpt-4-32k', name: 'GPT-4 32k' },
-];
+import OpenAI from 'openai';
 
-const defaultModelId = gptModels[0].id;
+async function getGptModels(apiKey: string) {
+  const openai = new OpenAI({ apiKey, dangerouslyAllowBrowser: true });
```
whitelisab, line 4: I'm just curious, this looks different from how we are calling the api in other parts of the app, adding `dangerouslyAllowBrowser: true` and then calling openai.models.list vs using the modelsBaseUrl (https://github.com/contentful/apps/blob/master/apps/ai-content-generator/src/configs/ai/baseUrl.ts). Is this due to the version bump of the openai package?

whitelisab (commented):
Added some comments for you to consider!

apps/ai-content-generator/src/components/config/model/Model.tsx
```diff
@@ -1,21 +1,34 @@
...
+
+  useEffect(() => {
+    async function fetchGptModels() {
+      if (!apiKey) return;
+      if (gptModels.length) return;
+      const models = await getGptModels(apiKey);
+      setGptModels(models.sort((a, b) => a.id.localeCompare(b.id)));
+    }
+    fetchGptModels();
+  }, [gptModels]);
```
mgoudy91, line 27: When I run this locally it complains about a non-exhaustive dependency list (missing apiKey), which I think is correct as it completely fails to load the list of models for me without this

mgoudy91 (changes requested):
I think this is close! I share a lot of the questions Lisa has asked, as well as an addition suggestion that I believe is needed for this to work right. Let me know what you think around those.

apps/ai-content-generator/src/components/config/model/Model.tsx
mgoudy91, line 27:
```suggestion
  }, [apiKey, gptModels]);
```

apps/ai-content-generator/src/components/app/dialog/common-generator/output/output-text-panels/original-text-panel/OriginalTextPanel.tsx
```diff
@@ -41,9 +38,7 @@ const OriginalTextPanel = (props: Props) => {
     generate();
   };
 
-  const sdk = useSDK<DialogAppSDK<AppInstallationParameters>>();
-  const model = gptModels.find((model) => model.id === sdk.parameters.installation.model);
-  const textLimit = model?.textLimit;
+  const textLimit = DEFAULT_TEXT_LIMIT;
```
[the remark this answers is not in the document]
david-shibley-contentful: Probably not a safe assumption. What would you suggest instead?

apps/ai-content-generator/package.json
```diff
@@ -12,7 +12,7 @@
     "@emotion/react": "^11.11.1",
     "@segment/analytics-next": "^1.55.0",
     "contentful-management": "11.14.0",
-    "openai": "^3.3.0",
+    "openai": "^4.53.0",
```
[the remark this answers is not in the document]
david-shibley-contentful: I will undo the version bump

apps/ai-content-generator/src/components/config/model/Model.tsx
david-shibley-contentful: Name doesn't exist on the model, just id

apps/ai-content-generator/src/configs/ai/gptModels.ts
david-shibley-contentful: Will get rid of

apps/ai-content-generator/src/components/config/model/Model.tsx
mgoudy91: I think the new ids are the same as the old names, it's just a different data access model

apps/ai-content-generator/src/components/app/dialog/common-generator/output/output-text-panels/original-text-panel/OriginalTextPanel.tsx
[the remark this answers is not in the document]
mgoudy91: yeah this is weird and I cant find a great answer online, but I think it's a pretty high limit. I'm okay to go with it for me and revisit as needed.

mgoudy91 (approved):
looks good

PR #8555, opened 2024-08-13
refactor: double deployment script [ext-5484]

mgoudy91:
## Purpose

This change is a foundation for a rewrite of the SAP Commerce Cloud (+ AIR header) app. The AIR header variant of the app has not ever worked meaningfully, and contains:
- A frontend app that is almost entirely a copy of the regular SAP app
- A completely non-functional lambda app that should be replaced by an app-action approach.

As such, we're going to completely delete the existing app **source code** and publish the regular SAP app as both. Later we'll add the app action, as well as some branching logic that will detect an app id, and attach the AIR header to appropriate requests.

This change sets the groundwork for this effort with the double-publish job and removing the old SAP-AIR app.

## Alternate approach

We also considered creating a new 'SAP-base' package, that would maintain a majority of the functionality for both apps. However, the gains from this approach felt too minimal given how small our intended differentiation between the two apps will be, and felt that we can always return to this approach later if it feels like the differences will be more substantial.

image

image


apps/sap-commerce-cloud/package.json
mgoudy91, line 1: This is the only non-delete change in the PR

sarahlessner (approved):
image
💅

t-col (approved):

PR #8556, opened 2024-08-13
feat: standardize css in SAP app

sarahlessner:
## Purpose

Consistent CSS for better app maintainability

## Approach

Followed conventions of more modern apps where we have CSS files using emotion that export a style object for display components to import. 

## Testing steps

Everything should look as it did before, testing that no CSS looks broken.

### DIFF
--- a/apps/sap-commerce-cloud/src/components/Dialog/Dialog.styles.ts
+++ b/apps/sap-commerce-cloud/src/components/Dialog/Dialog.styles.ts
@@ -1,0 +1,12 @@
+import { css } from 'emotion';
+import tokens from '@contentful/forma-36-tokens';
+
+export const styles = {
+  textInput: css({ width: '250px' }),
+  grid: css({ marginTop: tokens.spacingM, paddingLeft: tokens.spacingL }),
+  table: css({ padding: tokens.spacingL }),
+  tableCell: css({ width: '10%' }),
+  pagination: css({ margin: tokens.spacingL }),
+  nextButton: (page: number) => css({ marginLeft: page > 0 ? tokens.spacingL : '' }),
+  selectProductsButton: css({ margin: tokens.spacingL }),
+};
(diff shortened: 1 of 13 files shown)


apps/sap-commerce-cloud/src/index.tsx
```diff
@@ -7,8 +7,8 @@ import '@contentful/forma-36-tokens/dist/css/index.css';
 import './index.css';
 import AppConfig from './AppConfig/AppConfig';
 
-import Field from './components/Field';
-import Dialog from './components/Dialog';
+import Field from './components/Field/Field';
```
sarahlessner, line 10: I moved both of these to their own folders to follow the pattern of the component + css file living together in a folder

apps/sap-commerce-cloud/src/components/Dialog/Dialog.styles.ts
```diff
@@ -0,0 +1,12 @@
+import { css } from 'emotion';
+import tokens from '@contentful/forma-36-tokens';
+
+export const styles = {
+  textInput: css({ width: '250px' }),
+  grid: css({ marginTop: tokens.spacingM, paddingLeft: tokens.spacingL }),
+  table: css({ padding: tokens.spacingL }),
+  tableCell: css({ width: '10%' }),
+  pagination: css({ margin: tokens.spacingL }),
+  nextButton: (page: number) => css({ marginLeft: page > 0 ? tokens.spacingL : '' }),
```
sarahlessner, line 10: I noticed with this next page pagination button, the left margin is always present currently. We only want that spacing when there's also a previous page button so I added in this new logic to handle.

apps/sap-commerce-cloud/src/components/ProductPreviews/SortableListItem.styles.ts
```diff
@@ -0,0 +1,93 @@
+import { css } from 'emotion';
+import tokens from '@contentful/forma-36-tokens';
+import { Product } from '../../interfaces';
+
+const IMAGE_SIZE = 48;
```
david-shibley-contentful, line 5: we could export this and import it in `SortableListItem.tsx`

mgoudy91 (approved):
looks clean, no concerns

PR #8558, opened 2024-08-13
feat: remove react-sortable-hoc and add dnd-kit [EXT-5593]

whitelisab:
## Purpose

This PR removes the `react-sortable-hoc` dependency and replaces it with `dnd-kit` for the drag and drop interface on the Field location for this app. We are doing this so that we can upgrade this app to React 18 and also to fix a bug with this app where drag and drop was not working properly

## Approach

Updated the files in `ProductPreviews` directory to utilize the new dnd library. I also removed the Category files that hd the old library since the app was not using this feature anyway.

I also added a `productUrl` property to the products since this is what was being used as the unique SKU id, but it was hard to access from some places.

## Testing steps

I tested the app locally and was able to drag and drop the selected products and delete them.

Screenshot 2024-08-13 at 11 45 25 AM

Resources for migrating `react-sortable-hoc` to `dnd-kit` that might be useful for context:

- See this PR as an example of when this work was done for the `ecommerce-app-base`: https://github.com/contentful/apps/pull/7666. 
- https://edwardsmoses.com/migrating-to-dnd-kit-from-react-sortable-hoc


apps/sap-commerce-cloud/package.json
```diff
@@ -6,18 +6,18 @@
   "dependencies": {
     "@contentful/app-scripts": "^1.7.13",
     "@contentful/app-sdk": "^4.23.0",
-    "@contentful/ecommerce-app-base": "1.0.7",
```
whitelisab, line 9: This was just an extra import, we are not using this dependency

apps/sap-commerce-cloud/src/api/dataTransformers.ts
```diff
@@ -1,39 +1,33 @@
...
-    return {
-      id,
-      name: get(item, ['name', locale === undefined ? '' : locale], ''),
-      slug: get(item, ['slug', locale === undefined ? '' : locale], ''),
-      isMissing: false,
-      sku: '',
-      image: '',
-      externalLink,
+      sku,
+      productUrl: productUrl,
```
whitelisab, line 30: Added `productUrl` so that we would have access to this throughout the app. This is what the app is using as a unique id saved to the field. See more context here as to why we aren't just saving the SKU as many other apps do: https://github.com/contentful/apps/pull/3062

apps/sap-commerce-cloud/src/components/CategoryPreviews/CategoryPreviews.tsx
whitelisab, line 1: Removed all files in the `CategoryPreviews` directory since this feature of selecting categories to save never worked.

mgoudy91 (approved):
looks great to me, checked and seems to be working great locally

PR #8559, opened 2024-08-13
feat: replace react scripts with vite [EXT-5596]

sarahlessner:
## Purpose

Replacing react-scripts with vite


apps/sap-commerce-cloud/vite.config.ts
```diff
@@ -0,0 +1,16 @@
+import react from '@vitejs/plugin-react';
+import { defineConfig } from 'vite';
+
+export default defineConfig(() => ({
```
sarahlessner, line 4: I was checking the MS teams vite config to base most of this off of. We didn't have to do anything special with the build since this is purely a frontend app. I'm particularly unsure if the `test` config setup is a correct/generic setup or if that is something special to the MS teams app and we want to do something different here.

apps/sap-commerce-cloud/package.json
```diff
@@ -20,13 +20,12 @@
     "react-dom": "^17.0.1"
   },
   "scripts": {
-    "start": "cross-env BROWSER=none react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test",
-    "test:ci": "CI=true react-scripts test",
+    "start": "vite",
+    "build": "vite build",
+    "test": "vitest --logHeapUsage --coverage",
```
sarahlessner, line 25: another thing I copied from MS teams that I'm not sure we need. Most of the other things I copied from online migration tutorials.

apps/sap-commerce-cloud/package.json
```diff
@@ -49,10 +48,11 @@
     "@testing-library/user-event": "^12.7.2",
     "@types/jest": "^26.0.20",
     "@types/lodash.get": "^4.4.6",
-    "@types/node": "^14.14.31",
+    "@types/node": "^20.0.0",
```
sarahlessner, line 51: I had to update node in order to support the vite package installation. I picked this one because it looked like the latest stable version but if you're aware of a reason we should be using a different lower version of Node instead I wanted to call that out.

apps/sap-commerce-cloud/package-lock.json
mgoudy91, line 1: `3,819 additions, 18,470 deletions` 👀 exciting

apps/sap-commerce-cloud/package.json
mgoudy91: being on 14 before is wild...

I think 20 is reasonable (assuming the type versioning is in lock step with node itself, which it probably is)

apps/sap-commerce-cloud/src/AppConfig/AppConfig.spec.tsx
```diff
@@ -2,15 +2,16 @@ import React from 'react';
 import { render } from '@testing-library/react';
 import { ParameterDefinition } from '../interfaces';
 import AppConfig from './AppConfig';
+import { vi } from 'vitest';
```
sarahlessner, line 5: shoutout to @david-shibley-contentful for figuring out that we needed to replace jest with vi 😄

PR #8561, opened 2024-08-13
SAP refactor

mgoudy91:
### Complete rework of the SAP app
- Removes the old "SAP air" app code, and replaces with with a double-deployment where the SAP app code is built and deployed to both the SAP and SAP air Contentful apps.
- SAP works almost the same as before
- SAP AIR no longer uses AWS lambdas for its backend, and instead proxies (to avoid SAP CORS issue) an app action with the contentful-specific AIR header attached to all requests.

### MANY fixes, some notables:
- drag and drop works as expected
- uses more modern useSDK pattern for app locations
- modernization of other contentful app building patterns like forma components, loading states, and reworked class structure

image

image


BobHemphill76:
apps/sap-commerce-cloud-with-air/lambda/package-lock.json

looks like something might have changed for the lambda sub dir in the old sap-commerce-cloud-with-air?

sarahlessner (approved):
🚢

PR #8571, opened 2024-08-14
feat: update forma36 packages [EXT-5594]

sarahlessner:
## Purpose

Updating the forma 36 packages to the latest versions

## Testing steps

Comparing the testing version for development against the current version of the sap testing site to see if there was any CSS issues.


apps/sap-commerce-cloud/src/components/Dialog/Dialog.styles.ts
```diff
@@ -1,9 +1,9 @@
 import { css } from 'emotion';
-import tokens from '@contentful/forma-36-tokens';
+import tokens from '@contentful/f36-tokens';
 
 export const styles = {
   textInput: css({ width: '250px' }),
-  grid: css({ marginTop: tokens.spacingM, paddingLeft: tokens.spacingL }),
+  grid: css({ margin: `${tokens.spacingM} 0`, paddingLeft: tokens.spacingL }),
```
mgoudy91, line 6: This looks like a style change, was it intentional?

mgoudy91 (commented):
Looks good so far! Happy to re-review and approve when the last part gets in!

apps/sap-commerce-cloud/src/components/Dialog/Dialog.styles.ts
sarahlessner: Yep. When reviewing the portions of the app we updated, we noticed that there was some spacing built into the card before that isn't there with the new component library. We needed to have this margin apply to the top and bottom now that the element below this input section doesn't have top spacing.

apps/sap-commerce-cloud/package.json
```diff
@@ -6,11 +6,12 @@
   "dependencies": {
     "@contentful/app-scripts": "^1.7.13",
     "@contentful/app-sdk": "^4.23.0",
+    "@contentful/f36-button": "^4.68.1",
```
whitelisab, line 9: Curious if it makes sense to remove this dependency and just use the button from the `@contentful/f36-components` package?

whitelisab (approved):
One small comment/question, otherwise looks awesome. Great work! 🌟

apps/sap-commerce-cloud/package.json
sarahlessner: I saw an example in the migration guide for the Icon Button that imported it from there so I just didn't realize I could get it from the main component library as well. I agree that makes sense then and I'll make that change.

PR #6769, opened 2024-03-13
Feature/vercel haas

jsdalton:
## Purpose

We'll need HAAs to dynamically construct URLs for preview environments. This PR introduces some of the boilerplate and project structure needed to support HAAs along side the frontend.

## Approach

* Set up app actions as a folder
* Move frontend to subfolder
* Set up top level project to handle deploy and build
* Add a basic status HAA as a smoke test

## Testing steps

Front end still works:

image

Hosted app action call output:

```
ACCESS_TOKEN={redacted} npm run call-app-action -- -a vercelGetPreviewEnvironment -s q5wv5ajna1af

> @contentful/vercel-app@1.0.0 call-app-action
> npx ts-node scripts/call-app-action.ts -a vercelGetPreviewEnvironment -s q5wv5ajna1af

APP ACTION RESULT
{
  sys: {
    id: '7a9cGi1GVeAHDUbYnEV0ZJ',
    type: 'AppActionCallDetails',
    space: { sys: { type: 'Link', linkType: 'Space', id: 'q5wv5ajna1af' } },
    sys: { type: 'Link', linkType: 'Environment', id: 'master' },
    createdBy: {
      sys: {
        type: 'Link',
        linkType: 'AppAction',
        id: 'vercelGetPreviewEnvironment'
      }
    },
    createdAt: '2024-03-13T17:25:34.110Z'
  },
  request: { function: 'actions/get-preview-env.js', parameters: {} },
  response: { body: '{"ok":true,"data":{"url":"url"}}', statusCode: 200 },
  statusCode: 200,
  responseAt: '2024-03-13T17:25:34.603Z',
  errors: [],
  eventType: 'call',
  url: 'hosted',
  requestAt: '2024-03-13T17:25:34.110Z',
  environmentId: 'master'
}
Result: success!

```

PR #8603, opened 2024-08-16
feat: add hosted app actions for sap [EXT-5612]

whitelisab:
## Purpose

We will use hosted app actions to make api calls with the AIR header for the SAP AIR app. This PR introduces a lot of the boilerplate around hosted app actions.

## Approach

Updated the SAP app to follow the project structure and boilerplate used in the MS Teams and Vercel apps. See this PR as an example for Vercel: https://github.com/contentful/apps/pull/6769

Specifically, here is a summary of the changes:

- Move all frontend app code into a new `frontend` directory
    - Update the `vite.config` to support parent project build outDir location
- Introduce a new `app-actions` directory
    - Copied the 3 hosted app actions used in this old branch: https://github.com/contentful/apps/tree/feat/sap-air-original
    - Added testing and other boilerplate used in other apps for HAAs
- Set up top level project to handle the double deploy and build for both apps (public marketplace and private AIR header)
    - Add manifest file for the new actions
    - Add build actions and call actions scripts

## Testing steps

I confirmed that I could successfully build the frontend and actions and deployed to both staging apps. Confirmed that the app config page and field locations are rendering for both staging apps. I am not able to see products in the Dialog, but that's also a problem I am seeing on our feature branch.

Screenshot 2024-08-16 at 12 38 10 PM


apps/sap-commerce-cloud/app-actions/src/actions/testHandler.ts
whitelisab, line 1: I added this test handler and .spec file for it to just have a test that is passing in this folder. We can remove this down the road.

apps/sap-commerce-cloud/app-actions/src/actions/fetchBaseSites.ts
```diff
@@ -0,0 +1,41 @@
...
+        'Content-Type': 'application/json',
+        'application-interface-key': apiKey,
+      },
+    });
+
+    const res = await req.json();
+
+    return {
+      status: 'Success',
+      // @ts-ignore
```
whitelisab, line 25: You'll see this in a few places in these app actions. I just opted to ignore ts errors for now since these will need to be refactored to work anyway.

apps/sap-commerce-cloud/frontend/package.json
whitelisab, line 1: Changes in this file: I renamed the frontend app and removed all of the deploy scripts; those are now in the top level `package.json`

apps/sap-commerce-cloud/scripts/call-app-action.ts
whitelisab, line 1: This is something we've used in other apps and it's useful for testing app actions

apps/sap-commerce-cloud/package.json
```diff
@@ -1,66 +1,28 @@
...
-    "start": "vite",
-    "build": "vite build",
-    "test": "vitest --logHeapUsage --coverage",
-    "test:ci": "vitest",
+    "create-app-definition": "contentful-app-scripts create-app-definition",
+    "build": "rimraf ./build && npm run build-frontend && npm run build-app-actions",
+    "build-app-actions": "cd app-actions && npm ci && cd .. && node build-actions.js",
+    "build-frontend": "cd frontend && npm ci --include=dev && BUILD_PATH=../build npm run build",
+    "build:dev": "rimraf ./build && npm run build-frontend && npm run build-app-actions:dev",
+    "build-app-actions:dev": "cd app-actions && npm ci && cd .. && NODE_ENV=development node build-actions.js",
```
mgoudy91, line 10: What was the thinking here? is vite not an option for apps with app actions?

whitelisab: The diff makes this look weird, but all I did here was move what had been at in the `package.json` at the root to this file: `apps/sap-commerce-cloud/frontend/package.json`, and that file now contains all of the vite stuff. Then I created a new `package.json` at the root that has scripts responsible for building the app actions and the frontend and then deploying them. Does that help explain it, or am I missing something in your question?

jjolton-contentful (approved):

apps/sap-commerce-cloud/package.json
mgoudy91: no that's basically it, thanks for that :)

mgoudy91 (approved):

PR #8605, opened 2024-08-16
Feat: convert dialog component from class to function [EXT-5599]

david-shibley-contentful:

### DIFF
--- a/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
+++ b/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, ChangeEvent } from 'react';
+import React, { useEffect, useState, ChangeEvent, useCallback } from 'react';
 import {
   Button,
   Grid,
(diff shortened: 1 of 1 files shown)


apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -29,269 +29,220 @@ interface DialogProps {
...
+  const [selectedProducts, setSelectedProducts] = useState<string[]>([]);
+  const [errors, setErrors] = useState<ErrorType[]>([]);
+
+  useEffect(() => {
+    const initialize = async () => {
+      await loadBaseSites();
+      load();
+    };
+    initialize();
+  }, []);
```
BobHemphill76, line 48: Shouldn't we have loadBaseSites and load as deps here?

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -29,269 +29,220 @@ interface DialogProps {
...
+
+  const load = async () => {
```
BobHemphill76, line 50: this might be a good candidate for useCallback if we make it a dep of useEffect

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -29,269 +29,220 @@ interface DialogProps {
...
-      selectedProducts: this.state.selectedProducts,
-      errors: errors,
-    });
+    setProducts(products);
+    setErrors(errors);
   };
 
-  loadBaseSites = async () => {
-    const baseSites = await fetchBaseSites(this.props.sdk.parameters as SAPParameters);
+  const loadBaseSites = async () => {
```
BobHemphill76, line 62: same as above

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -29,269 +29,220 @@ interface DialogProps {
...
-      this.state.baseSite,
-      this.state.query,
-      this.state.page,
-      this.props.sdk.parameters as SAPParameters,
-      this.updateTotalPages
+      baseSite,
+      query,
+      page,
+      sdk.parameters as SAPParameters,
+      setTotalPages
```
whitelisab, line 56: Is `setTotalPages` the same as `updateTotalPages`? Just making sure that there's not difference in functionality

BobHemphill76:
Overall, just had a comment about hook deps and maybe migrating some of our async functions to useCallback.  They're not required, especially if it diverges too much from our other apps.

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
david-shibley-contentful: good idea!

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
david-shibley-contentful: yes, but it will now be `load, sdk.parameters`

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
david-shibley-contentful: yeah :-)

PR #8606, opened 2024-08-17
feat:adding tests [EXT-5600]

david-shibley-contentful:
## Purpose

Tests are important

## Approach

Before:
Screenshot 2024-08-16 at 12 16 12 PM

After:
image


apps/sap-commerce-cloud/frontend/src/components/Field/Field.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { render } from '@testing-library/react';
+import { AppParameters } from '../../interfaces';
+import { vi } from 'vitest';
+
+describe('Field', () => {
+  it('should render', () => {
+    const { container } = render(
+      <Field
+        {...{
+          sdk: {
```
whitelisab, line 12: In a couple of these components it looks like you are creating a mock sdk to pass as a prop. We already have a `mockSdk` that can be used here: https://github.com/contentful/apps/blob/feat/sap-app-refactor-ext5483/apps/sap-commerce-cloud/frontend/src/__mocks__/mockSdk.ts. Maybe to reduce some duplicative code you could update this mock and use this instead?

apps/sap-commerce-cloud/frontend/src/components/ProductPreviews/SortableComponent.spec.tsx
whitelisab, line 1: Just curious to learn more about the changes in this file, why you changed where `act` is imported from and removed that assertion?

apps/sap-commerce-cloud/frontend/src/components/fields.spec.tsx
whitelisab, line 1: Is this the same as the `AppConfig/fields.spec.tsx` file? I see we have that `fields.ts` file in both places and I'm not sure that we should, maybe worth cleaning up at some point? Maybe an accidental copy/paste of the file at some point?

apps/sap-commerce-cloud/frontend/src/api/dataTransformers.spec.tsx
```diff
@@ -0,0 +1,33 @@
+import { productTransformer, baseSiteTransformer } from './dataTransformers';
+
+describe('dataTransformers', () => {
+  describe('productTransformer', () => {
+    it('transforms product data', () => {
+      const apiEndpoint = 'http://localhost:9002';
+      const skuIdsToSkusMap = { '123': 'product1' };
+      const baseSite = 'electronics-spa';
+      const item = {
+        id: '123',
```
whitelisab, line 10: Just a small suggestion to use the existing product mock here: https://github.com/contentful/apps/blob/feat/sap-app-refactor-ext5483/apps/sap-commerce-cloud/frontend/src/__mocks__/mockProductPreview.ts and to see if there are any other mocks used in a lot of these tests that you can extract into the mocks directory?

whitelisab (commented):
Awesome work on this, so great to see that comprehensive test coverage!  🤩 Added a few small comments for you to consider

apps/sap-commerce-cloud/frontend/src/api/dataTransformers.spec.tsx
david-shibley-contentful: can do! thanks :-D

apps/sap-commerce-cloud/frontend/src/components/Field/Field.spec.tsx
david-shibley-contentful: will do 👍

apps/sap-commerce-cloud/frontend/src/components/ProductPreviews/SortableComponent.spec.tsx
david-shibley-contentful: there was a warning about how act is deprecated and should be imported from react instead

david-shibley-contentful: I dont remember why I removed the assertion, its probably a mistake. I will put it back

apps/sap-commerce-cloud/frontend/src/api/dataTransformers.spec.tsx
david-shibley-contentful: honestly I think A LOT of stuff could be extracted

apps/sap-commerce-cloud/frontend/src/components/fields.spec.tsx
david-shibley-contentful: The files are sliiiiightly different but yeah the are basically the same

whitelisab (approved):
The mocks look great, thanks for doing that! I think ideally we might want to be on the most recent versions of those testing dependencies and it looks like you backed out of those changes to get things working. Perhaps something we can address in the ticket to upgrade dependencies?

PR #8618, opened 2024-08-19
Feat/adding eslint 2

david-shibley-contentful:
## Purpose

[EXT-5621](https://contentful.atlassian.net/browse/EXT-5621)

## Deployment

[EXT-5621]: https://contentful.atlassian.net/browse/EXT-5621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/sap-commerce-cloud/frontend/package.json
+++ b/apps/sap-commerce-cloud/frontend/package.json
@@ -23,7 +23,9 @@
     "start": "vite",
     "build": "vite build",
     "test": "vitest --logHeapUsage --coverage",
-    "test:ci": "vitest"
+    "test:ci": "vitest",
+    "lint": "eslint 'src/**/*.{js,jsx,ts,tsx}'",
+    "lint:fix": "eslint 'src/**/*.{js,jsx,ts,tsx}' --fix"
   },
   "eslintConfig": {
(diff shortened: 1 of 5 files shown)


apps/sap-commerce-cloud/frontend/package.json
```diff
@@ -23,7 +23,9 @@
     "start": "vite",
     "build": "vite build",
     "test": "vitest --logHeapUsage --coverage",
-    "test:ci": "vitest"
+    "test:ci": "vitest",
+    "lint": "eslint 'src/**/*.{js,jsx,ts,tsx}'",
```
whitelisab, line 27: In several other apps, we have the `--max-warnings 0` option so that it will fail on CI if there are any warnings. Thoughts on adding that?

whitelisab (approved):
One small comment, otherwise looks good! 💪

apps/sap-commerce-cloud/frontend/package.json
david-shibley-contentful: great idea, thanks!

PR #8631, opened 2024-08-20
fix: sap small fixes []

whitelisab:
## Purpose

Removing console.logs, fixing typos, and making the linter happy.

Also fixed some issues in the `Field` and `Dialog` components that were causing tests to fail

### DIFF
--- a/apps/sap-commerce-cloud/frontend/package.json
+++ b/apps/sap-commerce-cloud/frontend/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@contenful/sap-commerce-cloud-frontend",
+  "name": "@contentful/sap-commerce-cloud-frontend",
   "version": "1.0.0",
   "private": true,
   "homepage": "./",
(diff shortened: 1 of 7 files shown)


jjolton-contentful (approved):
🧼

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -39,19 +39,16 @@ const Dialog: React.FC<DialogProps> = ({ sdk }) => {
...
-        query,
-        page,
-        sdk.parameters as SAPParameters,
-        setTotalPages
-      );
-      setProducts(products);
-      setErrors(errors);
-    };
-    return load;
+  const load = useCallback(async () => {
```
whitelisab, line 42: The Dialog tests were failing since this component wasn't actually loading since we weren't calling the load function

apps/sap-commerce-cloud/frontend/src/components/Field/Field.spec.tsx
```diff
@@ -16,16 +16,4 @@ describe('Field', () => {
     );
     expect(container).toMatchSnapshot();
   });
-
-  it('should update state value', () => {
```
whitelisab, line 20: These tests were failing with `TypeError: default is not a constructor`. I opted to just remove them since our tests should mimic behavior to trigger the function rather than call the internal methods directly

apps/sap-commerce-cloud/frontend/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@contenful/sap-commerce-cloud-frontend",
```
sarahlessner, line 2: 😬

sarahlessner (approved):
Looks good nice!!

PR #8653, opened 2024-08-21
fix: remove dependencies that cause infinite rerenders in AppConfig and add key to form control 

sarahlessner:
## Purpose

On the app config page we have an infinite re-render that required a fix. We also have a unique key prop warning to address.


apps/sap-commerce-cloud/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -39,18 +39,39 @@ interface Props {
   description: string;
 }
 
-export default function AppConfig(props: Props) {
+export default function AppConfig({
+  sdk,
```
sarahlessner, line 43: None of this is related to a fix I just like having the props destructured this way and noticed it

apps/sap-commerce-cloud/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -39,18 +39,39 @@ interface Props {
...
-    toInputParameters(props.parameterDefinitions, null)
+    toInputParameters(parameterDefinitions, null)
   );
 
+  const { space, app, ids } = sdk;
+
   useEffect(() => {
-    const init = async () => {
-      const { space, app, ids } = props.sdk;
+    const onAppConfigure = () => {
```
sarahlessner, line 61: moved this into the `useEffect` hook trying to resolve dependency array related issues, but just left it here since it's only called on mount via the `init` function anyway.

apps/sap-commerce-cloud/frontend/src/AppConfig/AppConfig.tsx
```diff
@@ -71,26 +92,14 @@ export default function AppConfig(props: Props) {
...
-
-      return {
-        parameters: toAppParameters(props.parameterDefinitions, parameters),
-        targetState: selectedFieldsToTargetState(contentTypes, selectedFields),
-      };
-    };
 
     init();
-  }, [props, contentTypes, parameters, selectedFields]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
sarahlessner, line 101: All of these dependencies that the linter suggests we add are causing the infinite rerender issue, since these deps change each time `init` and subsequently `onAppConfigure` get called.  I messed around with it for quite a while, and sadly couldn't find a better solution than just disabling this linter rule. None of the other `AppConfig` files I could find for other apps were functions components to look to as a possible example.  

That said, I do feel confident that we're doing the right thing here, or at least that this is the same as what we were doing before the conversion to a function component. The way this file works currently (as a class component) we are simply calling `init` from `componentDidMount`. There were no other lifecycle methods being used like `componentDidUpdate` that we need to try to account for with other callback functions or handling changes via the dependency array. So I do feel confident that disabling this and saying "hey thanks for the tip, but I know I just want this all to run one time, on mount" is accurate. 

That said if you have a better idea.... 👂 👂 👂

whitelisab: Maybe one suggestion would be to have two `useEffect` -- one for `app.onConfigure` and one for all the initial init stuff (getting the parameters, content types, editor interface, etc.). You can see an example here in AWS Amplify: https://github.com/contentful/apps/blob/master/apps/aws-amplify/src/locations/ConfigScreen.tsx. Not sure if that will help you at all!

whitelisab (approved):
Approving, just added one small suggestion to look into!

apps/sap-commerce-cloud/frontend/src/AppConfig/AppConfig.tsx
sarahlessner: Ah yeah I had tried two `useEffect` hooks + a callback, unsuccessfully. But I don't think I was following the flow of the app initialization correctly. That example you shared made it easy to follow and got it working! I do see how updating our SDK will clean some stuff up though.

PR #8655, opened 2024-08-21
feat: debounce search and dialog component fixes [EXT-5636]

sarahlessner:
## Purpose

The dialog component is currently working both on keystroke and also has a search button. This PR removes that search button in favor of a search-as-you-type approach. A debounce was added to avoid fetching too often. 

Also in the Dialog component, a fix was implemented for being unable to change the selected basesite. 

Screenshot here showing the removal of the search button + the fact that the electronics-spa base site is actually shown as selected with the correct product list rendering. 

Screenshot 2024-08-21 at 4 05 44 PM

### DIFF
--- a/apps/sap-commerce-cloud/frontend/package.json
+++ b/apps/sap-commerce-cloud/frontend/package.json
@@ -17,7 +17,8 @@
     "cross-env": "^7.0.3",
     "lodash.get": "^4.4.2",
     "react": "^18.3.1",
-    "react-dom": "^18.3.1"
+    "react-dom": "^18.3.1",
+    "use-debounce": "^10.0.3"
   },
   "scripts": {
     "start": "vite",
--- a/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.styles.ts
+++ b/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.styles.ts
@@ -9,4 +9,5 @@ export const styles = {
   pagination: css({ margin: tokens.spacingL }),
   nextButton: (page: number) => css({ marginLeft: page > 0 ? tokens.spacingL : '' }),
   selectProductsButton: css({ margin: tokens.spacingL }),
+  selectButton: css({ justifySelf: 'flex-end' }),
 };
--- a/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
+++ b/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
@@ -127,7 +127,7 @@ const Dialog: React.FC<DialogProps> = ({ sdk }) => {
         <GridItem>
           <TextInput
             type="text"
-            placeholder={'Type to search...'}
+            placeholder={'Type to search products'}
             className={cx(styles.textInput, 'f36-margin-bottom--m')}
             value={query}
             onChange={updateSearchTerm}
--- a/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
+++ b/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
@@ -23,7 +23,7 @@ import union from 'lodash/union';
 import { formatProductUrl } from '../../utils';
 import { styles } from './Dialog.styles';
 import { cx } from '@emotion/css';
-import { DoneIcon, SearchIcon } from '@contentful/f36-icons';
+import { DoneIcon } from '@contentful/f36-icons';
 import { useDebounce } from 'use-debounce';
 
 interface DialogProps {
--- a/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
+++ b/apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
@@ -24,6 +24,7 @@ import { formatProductUrl } from '../../utils';
 import { styles } from './Dialog.styles';
 import { cx } from '@emotion/css';
 import { DoneIcon, SearchIcon } from '@contentful/f36-icons';
+import { useDebounce } from 'use-debounce';
 
 interface DialogProps {
   sdk: DialogAppSDK<AppParameters>;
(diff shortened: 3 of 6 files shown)


apps/sap-commerce-cloud/frontend/package.json
```diff
@@ -17,7 +17,8 @@
     "cross-env": "^7.0.3",
     "lodash.get": "^4.4.2",
     "react": "^18.3.1",
-    "react-dom": "^18.3.1"
+    "react-dom": "^18.3.1",
+    "use-debounce": "^10.0.3"
```
sarahlessner, line 21: Popular library been around a while, saw a similar thing we were using for other apps - a bit more straightforward to implement than the lodash one with a callback.

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.styles.ts
```diff
@@ -9,4 +9,5 @@ export const styles = {
   pagination: css({ margin: tokens.spacingL }),
   nextButton: (page: number) => css({ marginLeft: page > 0 ? tokens.spacingL : '' }),
   selectProductsButton: css({ margin: tokens.spacingL }),
+  selectButton: css({ justifySelf: 'flex-end' }),
```
sarahlessner, line 12: Some CSS adjustment for the Select Products button now that there's no search button

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -42,14 +44,18 @@ const Dialog: React.FC<DialogProps> = ({ sdk }) => {
...
       page,
       sdk.parameters as SAPParameters,
       setTotalPages
     );
     setProducts(products);
     setErrors(errors);
-  }, [baseSite, query, page, sdk.parameters]);
+  }, [baseSite, debouncedQuery, page, sdk.parameters]);
+
+  useEffect(() => {
```
sarahlessner, line 56: I broke this out so we now have two `useEffect` hooks, which is part of resolving the issue with being able to change the basesite. 

What was happening is that our original `useEffect` hook fetched all of the basesites and set our state value for `baseSite` to whatever was at the first index of the basesites list. Then the hook would call `load` which fetches the products. Because of that, `load` was a dependency for the `useEffect` hook and anytime we called `load` we wound up triggering the `useEffect` hook. That caused the `baseSite` state value to get overwritten. 

We needed to separate that logic so that we only fetch/set the basesites on mount. The `load` function is a callback and it has the `baseSite` as a dependency so we don't need to explicitly call `load` again here after we initially fetch those baseSites.

That's also why I was able to remove the `load` calls elsewhere in the file. Overall I believe we were fetching the product list a lot more often than necessary.

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -137,15 +127,10 @@ const Dialog: React.FC<DialogProps> = ({ sdk }) => {
         <GridItem>
           <TextInput
             type="text"
-            placeholder={'Search Term...'}
+            placeholder={'Type to search products'}
```
sarahlessner, line 130: I thought to change this placeholder just as a tiny thing to help make it a little more clear that the way to search is to type in a term. Let me know if there's a better convention here.

apps/sap-commerce-cloud/frontend/package.json
david-shibley-contentful: we could also just implement our own debounce function:

```
function debounce(func, delay) {
  let timeoutId;

  return function(...args) {
    clearTimeout(timeoutId);

    timeoutId = setTimeout(() => {
      func.apply(this, args);
    }, delay);
  };
}
```

david-shibley-contentful (approved):
LGTM

apps/sap-commerce-cloud/frontend/src/components/Dialog/Dialog.tsx
mgoudy91: very cool, appreciate the write up

mgoudy91 (approved):
Looks good to me! I think there's _some_ possibility the type-to-search UX might be an annoyance for some users, but I'm okay giving this model a try and seeing if we have any complaints.

I was also wondering, is there any general suggestion for search UX when it comes to accessibility? I saw some conflicting answers online so it seems like it might not be a settled question.

sarahlessner:
> I was also wondering, is there any general suggestion for search UX when it comes to accessibility? I saw some conflicting answers online so it seems like it might not be a settled question.

@mgoudy91 that's a great question. I'll do a little research and implement some accessibility tweaks for anything that seems obvious. I don't seen an aria label on the search bar, as an example, so at the very least we should improve some small things like that. I also wonder if the input bar should be a focused element when the dialog opens.

mgoudy91:
> > I was also wondering, is there any general suggestion for search UX when it comes to accessibility? I saw some conflicting answers online so it seems like it might not be a settled question.
> 
> @mgoudy91 that's a great question. I'll do a little research and implement some accessibility tweaks for anything that seems obvious. I don't seen an aria label on the search bar, as an example, so at the very least we should improve some small things like that. I also wonder if the input bar should be a focused element when the dialog opens.

Sounds great, and only if that's interesting to you, certainly not a blocking concern at all. 

I will say though (as this is something I randomly learned _literally_ yesterday) that apparently autofocusing input fields is _not_ desirable from an accessibility standpoint and often a big problem for screen readers

sarahlessner:
> I will say though (as this is something I randomly learned literally yesterday) that apparently autofocusing input fields is not desirable from an accessibility standpoint and often a big problem for screen readers

Good to know, I was just trying to read up on that because we used to do it at LD all the time but I wasn't sure if it was actually good or not... so that's a bad example haha. But I do think it's important to spend some time on accessibility it's too often an afterthought so I'll spend some time there to just make it a little bit better.

sarahlessner:
Going to merge and added https://contentful.atlassian.net/browse/EXT-5645 for making accessibility improvements outside of just this component.

PR #8693, opened 2024-08-23
chore: group dependency upgrades

t-col:
## Purpose

group dependency upgrades for dev and production dependencies


jjolton-contentful (approved):
🙇

PR #8730, opened 2024-08-26
Fix/ext 5655 loading state

mgoudy91:
## Purpose

- adds loading state spinner for product requests to SAP
- Attempts to fix an infinite request bug with the way we memoize our api

image


sarahlessner (approved):
LGTM 🏅

PR #8731, opened 2024-08-26
feat: SAP app cleanup - move all component files into components folder [EXT-5652]

sarahlessner:
## Purpose

The AppConfig folder with several components lived outside of the `components` folder. Doing this step first since it's a more messy refactor.


apps/sap-commerce-cloud/frontend/tsconfig.json
```diff
@@ -15,7 +15,15 @@
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "react-jsx",
-    "types": ["vite/client"]
+    "types": ["vite/client"], 
+    "paths": {
```
sarahlessner, line 19: I noticed in the guidelines for app developments not to use too many relative paths like `../../` and while I moved the AppConfig folder, I noticed I needed to update several imports relative paths. I tried spending a little time to resolve that and add these paths like I saw for other apps, but I couldn't get this to work - none of these paths resolve when I try to use them such as `@hooks/useApi` won't resolve. Anyone have more experience with this to know what step might be missing?

mgoudy91 (approved):

PR #8890, opened 2024-09-12
feat: Enable 3rd party libs using globals [MONET-2325]

kdamball:
## Purpose

Polyfill global & node modules by default to avoid bugs with globals on 3rd party libs

## Approach

Some 3rd party libraries use node globals/builtin modules which our runtimes do not always provide by default. We could polyfill one by one but it becomes impractical when using libraries. As a result, we should provide a default API regardless of runtime.

## Testing steps

Deploying and running functions should just work just like before. There might be some bundle size problems however


mrlaessig (approved):

jjolton-contentful (approved):

examples/function-appaction/build-functions.js
```diff
@@ -61,13 +63,15 @@ const main = async (watch = false) => {
     const config = {
       entryPoints: getEntryPoints(),
       bundle: true,
-      minify: true,
-      platform: 'node',
       outdir: 'build',
-      logLevel: 'info',
       format: 'esm',
-      target: 'es6',
-      external: ['node:*'],
```
BasKiers, line 70: I believe we need to retain `'node:*'` imports as external such that it builds without those dependencies and expect them to be provided by the runtime.

BasKiers (approved):
Awesome fix! Great work 😄 I'm only wondering whether we'd want to keep the `external: ['node:*']` configuration

examples/function-appaction/build-functions.js
kdamball: I dont think it's needed anymore now that we include prebuilt node modules. I just confirmed by including it in the configuration and the file checksum is the same

PR #8918, opened 2024-09-17
refactor: improve appaction function example types [EXT-5711]

jjolton-contentful:
## Purpose

We recently introduced better typing support for AppActionRequest -> users can now specify the Category of their App Action, as well as the types of their Custom App Action Parameters. https://github.com/contentful/node-apps-toolkit/pull/702

This PR provides an example of using this functionality in the function-appaction example app.


whitelisab (approved):
🔥

PR #8949, opened 2024-09-24
fix(function-appaction-example): specify minimal node version for upload, add allowNetworks [EXT-00]

jjolton-contentful:
## Purpose

The function-appaction example requires node v18 for the upload command to work properly, and was missing "api.contentful.com" from the `allowNetworks` list


BobHemphill76 (approved):
Might be worth looking into adding an engines section to the package.json: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

PR #8954, opened 2024-09-24
chore: Remove Typeform vulnerable dependencies [EXT-5806]

sarahlessner:
## Purpose

Replaces react-scripts with vite, remove emotion old library, remove old forma libraries


apps/typeform/frontend/src/Auth/TypeformOAuth.tsx
```diff
@@ -1,13 +1,13 @@
...
-import { Button } from '@contentful/forma-36-react-components';
+import { Button } from '@contentful/f36-components';
 import { BASE_URL, CLIENT_ID } from '../constants';
 
 interface Props {
   sdk?: AppExtensionSDK;
   expireSoon?: boolean;
   isFullWidth: boolean;
-  buttonType?: 'primary' | 'positive' | 'negative' | 'muted' | 'naked' | undefined;
+  buttonType?: 'primary' | 'positive' | 'negative' | 'secondary' | 'transparent' | undefined;
```
mgoudy91, line 10: 👍

apps/typeform/frontend/src/constants.ts
```diff
@@ -2,6 +2,6 @@ const SDK_WINDOW_HEIGHT = 450;
 const BASE_URL = 'https://api.typeform.com';
 
 // TODO Move to env var
-const CLIENT_ID = 'HC3UDnoiaP1UCMqJ7kCAyTFdHrDt8nLtXx4BKRJxom2M';
+const CLIENT_ID = '2vyzzT2AjqrtfWKmaigvZjF8oYwUXrJABmcS5WK4MPJg';
```
mgoudy91, line 5: is this intentional?

mgoudy91 (approved):
lgtm, qq on the client id

apps/typeform/frontend/src/constants.ts
sarahlessner: Thanks actually I should move that back. There's some weird stuff in the instructions where this client ID is pointed to the prodcution app. but why is it just hardcoded like this in the first place? haha there's a TODO around moving stuff to env variables which seems like we should do that eventually but didn't yet. I'd also like to test this one better on staging before merging so let's chat about that later too.

sarahlessner: changed that back and ready for rereview @mgoudy91

mgoudy91 (approved):

PR #8967, opened 2024-09-25
chore: removing vulnerable dependencies from Image Focal Point app

david-shibley-contentful:


apps/image-focal-point/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
```
t-col, line 6:
```suggestion
    <title>Image Focal Point</title>
```
?

apps/image-focal-point/index.html
```diff
@@ -0,0 +1,13 @@
...
+</head>
+<body>
+    <div id="root"></div>
+
+    <script type="module" src="./src/index.jsx"></script>
```
t-col, line 11:
```suggestion
    <script type="module" src="./src/index.js"></script>
```
doesn't look like there's a jsx file in this app

PR #8978, opened 2024-09-26
chore: remove vulnerable dependencies from gatsby app

david-shibley-contentful:
## Purpose

[EXT-5787](https://contentful.atlassian.net/browse/EXT-5787)


apps/gatsby/src/index.spec.jsx
```diff
@@ -79,12 +78,14 @@ describe('Gatsby Preview entry point', () => {
...
     render.mockClear();
-    jest.unmock('@contentful/app-sdk');
+    vi.unmock('@contentful/app-sdk');
   });
 
-  it('should initialize the app', () => {
-    loadEntryPoint();
+  it('should initialize the app', async () => {
+    await import('./index');
+    // loadEntryPoint();
```
sarahlessner, line 86: Why is this commented out now? Do we not need to test it anymore?

apps/gatsby/vite.config.js
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'vite';
+
+export default defineConfig({
+  server: {
+    port: 3000,
+  },
+  base: './',
```
sarahlessner, line 7: Was there a reason the base is written like this instead of an empty string?

david-shibley-contentful: I like it better this way cause it is more explicit, but really its 6 in one half dozen in the other. I could easily be talked into an empty string if you prefer

apps/gatsby/src/index.spec.jsx
```diff
@@ -79,12 +78,14 @@ describe('Gatsby Preview entry point', () => {
...
 
+    console.log(render.mock.calls);
```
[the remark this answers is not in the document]
david-shibley-contentful: nope. good catch

apps/gatsby/src/index.spec.jsx
david-shibley-contentful: that needs to be uncommented, will fix

sarahlessner (approved):

PR #9035, opened 2024-10-02
Chore/vuln deps image focal point

david-shibley-contentful:
## Purpose

[EXT-5794](https://contentful.atlassian.net/browse/EXT-5794)

## Deployment

[EXT-5794]: https://contentful.atlassian.net/browse/EXT-5794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sarahlessner (approved):

apps/image-focal-point/package.json
```diff
@@ -5,28 +5,30 @@
...
+    "@contentful/f36-tokens": "^4.0.5",
+    "@emotion/css": "^11.13.0",
+    "@vitejs/plugin-react": "^4.3.2",
     "hex-rgb": "4.3.0",
     "lodash.camelcase": "4.3.0",
     "prop-types": "15.8.1",
-    "react": "16.8.6",
-    "react-dom": "16.8.6",
-    "react-scripts": "5.0.1"
+    "react": "^18.3.1",
```
sarahlessner, line 28: Nice! Curious if anything necessitated the update to react to get stuff working, or you just saw the opportunity to do it after updating forma?

apps/image-focal-point/package.json
```diff
@@ -36,5 +38,8 @@
     "> 1%",
     "not ie <= 11"
   ],
-  "homepage": "."
+  "homepage": ".",
+  "optionalDependencies": {
+    "@rollup/rollup-linux-x64-gnu": "4.9.5"
```
sarahlessner, line 43: Curious about this too since I've been having some weird rollup issues with optimizely and wondering if you were running into rollup stuff that this fixed?

apps/image-focal-point/src/components/AppView/AppView.jsx
```diff
@@ -143,7 +143,7 @@ export class AppView extends Component {
       <>
         <div className={styles.background} />
         <div className={styles.body}>
-          <Typography>
+          <React.Fragment>
```
sarahlessner, line 146: I was thinking that `React.Fragment` is an old method that was replaced with just the empty brackets `<></>` so we probably don't need this? I think `Typography` was basically deprecated in Forma for the same reason, that we just use empty fragment tags now.

apps/image-focal-point/src/index.jsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
-import { Note } from '@contentful/forma-36-react-components';
+import { Note } from '@contentful/f36-components';
```
sarahlessner, line 4: I think one of the updates with `note` was the switch to use a generic `variant` prop instead of the component specific `noteType` prop. I didn't see that change in here and I'm figuring that's because this app isn't in typescript and so it didn't yell about it? But I wonder if the note's variant is actually being respected since we're passing in a deprecated prop.

apps/image-focal-point/package.json
david-shibley-contentful: just saw the opportunity to do it :-)

apps/image-focal-point/package.json
david-shibley-contentful: yeah I was, this fixed it :-)

apps/image-focal-point/src/components/AppView/AppView.jsx
david-shibley-contentful: Can change 👍

apps/image-focal-point/src/index.jsx
david-shibley-contentful: good to know! will fix

apps/image-focal-point/package.json
david-shibley-contentful: the error I was getting that this resolved was 
`Cannot find module @rollup/rollup-linux-x64-gnu`

sarahlessner (approved):
LGTM! 🥇

PR #9056, opened 2024-10-08
feat: allow autotagger to be called by an app action [EXT-00]

jjolton-contentful:
## Purpose

Contentful Functions can now be linked to App Actions (see https://www.contentful.com/developers/docs/extensibility/app-framework/functions/#use-case-app-actions)

This changeset extends the Autotagger example app, so that users can now trigger autotagging on demand using an App Action Function, as opposed to or in addition to using the existing App Event Function.

### DIFF
--- a/examples/autotagger/src/locations/ConfigScreen.tsx
+++ b/examples/autotagger/src/locations/ConfigScreen.tsx
@@ -35,10 +35,20 @@ const ConfigScreen: React.FC = () => {
 
   const onConfigure = useCallback(async () => {
     const currentState = await sdk.app.getCurrentState();
-
+    const targetState = {
+      EditorInterface: {
+        ...currentState?.EditorInterface,
+      },
+    };
+    parameters.contentTypes?.split(',').forEach((id) => {
+      // assign the app to the sidebar on all selected content types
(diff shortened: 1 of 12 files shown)


examples/autotagger/src/locations/ConfigScreen.tsx
```diff
@@ -35,10 +35,20 @@ const ConfigScreen: React.FC = () => {
 
   const onConfigure = useCallback(async () => {
     const currentState = await sdk.app.getCurrentState();
-
+    const targetState = {
```
whitelisab, line 38: 🔥

examples/autotagger/src/locations/Home.tsx
```diff
@@ -5,11 +5,6 @@ import { /* useCMA, */ useSDK } from '@contentful/react-apps-toolkit';
 
 const Home = () => {
   const sdk = useSDK<HomeAppSDK>();
-  /*
```
whitelisab, line 8: nit: maybe also get rid of the commented out `useCMA` on line 4 like you did in the other files?

whitelisab (approved):
Nice!

t-col (approved):

PR #9077, opened 2024-10-10
Chore: Update Optimizely app - react update, react vite, emotion, forma libraries

sarahlessner:


apps/optimizely/src/EditorPage/subcomponents/references-section.jsx
```diff
@@ -100,7 +100,3 @@ ReferencesSection.propTypes = {
   references: PropTypes.array,
   sdk: PropTypes.object.isRequired,
 };
-
-ReferencesSection.defaultProps = {
```
sarahlessner, line 104: Getting warnings in the console about this being deprecated and to just define the default when passing in props instead, so for some components I removed this usage, and then destructured the props if they weren't already to be able to define the default that way.

apps/optimizely/src/ConnectButton.spec.jsx
```diff
@@ -2,20 +2,15 @@ import React from 'react';
...
 
 import ConnectButton from '../src/ConnectButton';
+import { vi } from 'vitest';
 
 configure({ testIdAttribute: 'data-test-id' });
 
 describe('ConnectButton', () => {
   afterEach(cleanup);
 
-  it('should match snapshot', () => {
```
sarahlessner, line 11: I just removed this snapshot since it didn't seem very useful, and I had the problem with the snapshot constantly changing due to the forma button classNames. It wasn't testing anything specific. The other test checks for the test id that should be found if this connect button renders successfully and I think that's sufficient?

apps/optimizely/package.json
```diff
@@ -4,35 +4,40 @@
...
+    "start": "vite",
+    "build": "rm -rf build && vite build",
     "build:local": "npm run env:local && npm run build",
     "env:local": "echo REACT_APP_OPTIMIZELY_CLIENT_APP_ID=${REACT_APP_OPTIMIZELY_CLIENT_APP_ID} > .env.local",
     "deploy": "aws s3 sync ./build ${STATIC_S3_BASE}/optimizely --acl public-read",
     "create-app-definition": "contentful-app-scripts create-app-definition",
-    "deploy:test": "npm run deploy",
-    "test": "TZ=UTC react-scripts test",
-    "test:ci": "TZ=UTC react-scripts test"
+    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id ${DEV_TESTING_ORG_ID} --definition-id 245M1niKUUrRjipRq6Jn1k --token ${TEST_CMA_TOKEN}",
```
mgoudy91, line 38: this line doesnt really do anything so we can probably yank it
```
```

mgoudy91:
```suggestion
```

mgoudy91 (approved):

PR #9101, opened 2024-10-15
feat: prd / test / dev serverless config

BobHemphill76:

### DIFF
--- a/apps/jira/functions/config/serverless-env.dev.yml
+++ b/apps/jira/functions/config/serverless-env.dev.yml
@@ -1,0 +1,4 @@
+frontendUrl: https://app.jira-test.ctfapps.net
+oauthRedirectUri: https://api.jira-test.ctfapps.net/auth
+baseUrl: https://api.jira-test.ctfapps.net
+domainName: api.jira-test.ctfapps.net
--- a/apps/jira/functions/config/serverless-env.prd.yml
+++ b/apps/jira/functions/config/serverless-env.prd.yml
@@ -1,0 +1,4 @@
+frontendUrl: https://app.jira.ctfapps.net
+oauthRedirectUri: https://api.jira.ctfapps.net/auth
+baseUrl: https://api.jira.ctfapps.net
+domainName: api.jira.ctfapps.net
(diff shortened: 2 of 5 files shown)


sarahlessner:
From Bob on Slack:

We currently have a cert in our cf-proddev-apps-production aws account wildcarded with *.[jira.ctfapps.net](http://jira.ctfapps.net/). We use this cert for the CloudFront distribution associated with the API GW custom domain name '[api.jira.ctfapps.net](http://api.jira.ctfapps.net/)'.  However, the naming convention(applied to 15ish apps) we landed on for our staging apps custom domain names is something like api.{appName}-[test.ctfapps.net](http://test.ctfapps.net/), which I believe means we cannot use that same cert for the CF distro associated with that staging custom domain.
 Do you have any concerns around using the *.[cftapps.net](http://cftapps.net/) cert for our [api.jira-test.ctfapps.net](http://api.jira-test.ctfapps.net/) custom domain?
This app comes with a companion jira app that needs to be installed.  I can see that we have the app installed in our jira account, but I can't see it's configuration.  At the very least there are a few OAuth clients configured for dev / prd.
Does this app or Jira own those OAuth clients?  And if so, how would I go about creating another client for staging?
Is there any other configuration (allowedOrigins) that we would need to configure to allow our new lambda to connect to this jira app
For OAuth redirect flow, do we need any configuration updates to the jira app / the new client created above to allow another redirectUri?
We currently host the front end application via  s3 -> CloudFront Distro ([app.jira.ctfapps.net](http://app.jira.ctfapps.net/)) (using the cert mentioned above).  But for our other staging apps we have just been creating private apps within the org mentioned above. 
Provided we know where the FE is hosted (for redirectUri purposes), do you see any other potential concerns with this?

sarahlessner:
After the OAuth talk we had today
Provided we know where the FE is hosted (for redirectUri purposes), do you see any other potential concerns with this?
yes there are concerns here because we expect hosted apps to have signed urls that will clash with the jira oauth client static redirect uri.  So, we will most likely need to host this FE at something like [app.jira-test.ctfapps.net](http://app.jira-test.ctfapps.net/), which I think implies we just add another s3 -> CF distro ([app.jira-test.ctfapps.net](http://app.jira-test.ctfapps.net/)) and our deploy:test process is uploading static resources to s3 rather than targeting a private app in the org.  But this does imply that this distro is most likely public :shrug: or at least the same level of access as marketplace apps (edited)

PR #9102, opened 2024-10-16
feat: use functions-types as source of type definitions

msieroslawska:
We have created type definitions for NER events in `functions-types` package, and now we want to consolidate all NER-related types to be imported from that one source of truth. This PR removes the dependency to `node-apps-toolkit` (it will also be pulling type definitions from `functions-types`).

### DIFF
--- a/examples/native-external-references/functions/helpers.spec.ts
+++ b/examples/native-external-references/functions/helpers.spec.ts
@@ -51,7 +51,9 @@ describe('Getting the URLs', () => {
       lookupUrls: [
         'https://api.themoviedb.org/3/movie/urn:123?language=en-US',
         'https://api.themoviedb.org/3/movie/urn:456?language=en-US'
-      ]
+      ],
+      trendingUrl:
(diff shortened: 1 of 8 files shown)


examples/native-external-references/functions/helpers.spec.ts
```diff
@@ -51,7 +51,9 @@ describe('Getting the URLs', () => {
       lookupUrls: [
         'https://api.themoviedb.org/3/movie/urn:123?language=en-US',
         'https://api.themoviedb.org/3/movie/urn:456?language=en-US'
-      ]
+      ],
+      trendingUrl:
+        'https://api.themoviedb.org/3/trending/movie/day?language=en-US'
```
screamcha, line 56: Why is this addition needed?

msieroslawska: By default first search query is empty and TMDB responds with an empty array. To have _something_ displayed at first render of entity picker (before user starts typing the query), we've decided to pull results from the trending endpoint. We forgot to cover this in the example tests.

screamcha (approved):

PR #9162, opened 2024-10-30
Chore: salesforce remediate vulnerable deps [integ-2287]

sarahlessner:

### DIFF
--- a/apps/salesforce-commerce-cloud/package.json
+++ b/apps/salesforce-commerce-cloud/package.json
@@ -28,7 +28,7 @@
     "build": "tsc && vite build",
     "test": "vitest",
     "deploy": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEFINITIONS_ORG_ID} --definition-id 1O8njPYTGO44VueSJNBrdZ --token ${CONTENTFUL_CMA_TOKEN}",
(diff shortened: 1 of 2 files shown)


david-shibley-contentful (approved):

apps/salesforce-commerce-cloud/package.json
```diff
@@ -14,8 +15,8 @@
     "@tanstack/react-query": "^4.29.7",
     "@vitejs/plugin-react": "^4.3.1",
     "axios": "^1.7.7",
-    "commerce-sdk": "^2.10.1",
-    "contentful-management": "10.7.0",
+    "commerce-sdk": "^2.15.0",
+    "contentful-management": "^11.35.1",
```
mgoudy91, line 19: might cause a problem again, we'll see

mgoudy91 (commented):
code looks totally fine, only concern is the cma v11 update, but thats why we have staging envs eventually

apps/salesforce-commerce-cloud/package.json
sarahlessner: Curious do we know what specific problem did this library update cause in the past? I think I've merged in a few with the update to 11 and they were still working in prod. 

I am working on getting staging setup for this one. I got a staging environment set up and deployed the bundle but ran out of time to go through the config. I'll make sure that's done and tested before I merge.

sarahlessner: I can also update this to the earliest version with a fix (assuming that's even an option) if it's safer.

david-shibley-contentful (approved):

PR #9214, opened 2024-11-18
Chore: Smartling app frontend vulns

sarahlessner:


apps/smartling/lambda/src/index.ts
```diff
@@ -121,6 +121,18 @@ export function makeApp(fetchFn: any, issuer: any) {
     );
   });
 
+  const computeLastModifiedTime = () => {
```
sarahlessner, line 124: We were having some issues with the cache not being invalidated when deploying and copied this from Typeform's setup.

apps/smartling/frontend/vite.config.ts
```diff
@@ -10,6 +12,13 @@ export default defineConfig(() => ({
     // the parent project that combines both the frontend and the hosted app action backend needs to be able to override
     // the default location to its own build path
     outDir: process.env.BUILD_PATH || './build',
+    rollupOptions: {
```
sarahlessner, line 15: @mgoudy91  I found a suggestion on stack overflow to set up some hashing for the build in Vite so that you don't have to hard reload to refresh the cache. it seemed to resolve the issue you and I were running into last week. It's possible that the introduction of Vite into our build process caused this new behavior we saw and we didn't necessarily uncover some long standing caching issue.

david-shibley-contentful (approved):
LGTM

PR #9224, opened 2024-11-19
Chore: update vulns in frontend root and app actions package json files [INTEG-2284]

sarahlessner:


david-shibley-contentful (approved):
🚀

PR #9226, opened 2024-11-19
Chore: remediate vulnerabilities in apps examples folder [INTEG-2303]

sarahlessner:


examples/autotagger/src/locations/Sidebar.spec.tsx
```diff
@@ -2,16 +2,18 @@ import React from 'react';
...
   useSDK: () => mockSdk,
   useCMA: () => mockCma,
+  useAutoResizer: vi.fn(),
 }));
 
 describe('Sidebar component', () => {
   it('Component text exists', () => {
     const { getByText } = render(<Sidebar />);
 
-    expect(getByText('Hello Sidebar Component (AppId: test-app)')).toBeInTheDocument();
```
sarahlessner, line 15: It's weird that this test wasn't failing before (or maybe these tests just aren't a part of the CI checks for the examples?) but the component doesn't actually render this test, looks like this test was just copy/pasted from other tests for their Sidebar, but this test wasn't actually a reflection of the Sidebar content for this particular example. I updated it it to find something that the component actually renders.

david-shibley-contentful (approved):

PR #9228, opened 2024-11-19
Chore: Smartling app lambda package.json vuln remediation [INTEG-2183]

sarahlessner:


apps/smartling/frontend/vite.config.ts
```diff
@@ -12,13 +12,6 @@ export default defineConfig(() => ({
     // the parent project that combines both the frontend and the hosted app action backend needs to be able to override
     // the default location to its own build path
     outDir: process.env.BUILD_PATH || './build',
-    rollupOptions: {
```
sarahlessner, line 15: turns out this didn't really fix the caching issue so I removed it here, still need to determine why the cache is temporarily stale on staging until you clear cache/hard reload but not a blocking issue for the work in this PR.

apps/smartling/lambda/src/index.ts
```diff
@@ -148,6 +148,9 @@ export function makeApp(fetchFn: any, issuer: any) {
       '/frontend',
       express.static(path.dirname(require.resolve('@contentful/smartling-frontend')), {
         lastModified,
+        setHeaders: (res: any) => {
```
sarahlessner, line 151: lastModified was passed in as a boolean option but this didn't ultimately set the lastModified date in the header, causing an issue with serving up the assets on the frontend. This change explicitly sets the last modified header with the last modified date as a fix.

mgoudy91 (approved):
Looks great to me!

I was looking at these deps and I think for what we're trying to do now, it's not worth the effort of doing something like forcing resolutions (not even sure if that would fix the problem anyway).

In my investigation though I did see we could probably move a lot of these into dev deps at the least https://github.com/contentful/apps/pull/9255 feel free to either use or ignore the change, I think this is good to go as is.

mgoudy91 (approved):

PR #9283, opened 2024-12-03
Chore: Set up staging app for GA4 [INTEG-2296]

sarahlessner:


apps/google-analytics-4/frontend/src/config.ts
```diff
@@ -12,8 +12,13 @@ function getEnvironmentVariable(
   return environmentVariableValue;
 }
 
+const backendApiUrl =
+  getEnvironmentVariable('REACT_APP_ENV', 'prod') === 'test'
```
sarahlessner, line 16: REACT_APP_ENV is set to `test` in circle CI in the apps-staging context so that we can determine which URL to use for API calls.

apps/google-analytics-4/lambda/src/middlewares/verifySignedRequests.ts
```diff
@@ -38,7 +38,7 @@ const makeCanonicalReq = (req: Request) => {
   });
 
   // TODO: make this stage prefixing logic better? (yuck)
-  const pathPrefix = config.stage !== 'prd' ? `/${config.stage}` : '';
+  const pathPrefix = config.stage !== 'prd' && config.stage !== 'test' ? `/${config.stage}` : '';
```
sarahlessner, line 41: I noticed this logic when trying to troubleshoot early on, and I figure it must be designed to handle something different between prod/local and so the logic for prod should apply to staging as well.

mgoudy91: Have to agree with that previous line TODO, ha. Wdyt:
```suggestion
  const pathPrefix = ['prd', 'test'].includes(config.stage) ? '' : `/${config.stage}`;
```

mgoudy91 (approved):
looks great, amazing work digging that up

left one quick suggestion to consider, up to you if you want it

mgoudy91 (approved):

PR #9301, opened 2024-12-09
Chore: GA-4 App - Vite Migration [INTEG-2201]

sarahlessner:


apps/google-analytics-4/frontend/.eslintrc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
sarahlessner, line 1: eslint was extending a library from react scripts, so I added a basic config instead

apps/google-analytics-4/frontend/src/components/config-screen/GoogleAnalyticsConfigPage/GoogleAnalyticsConfigPage.spec.tsx
```diff
@@ -106,15 +121,26 @@ describe('Installed Service Account Key', () => {
       },
     });
     mockSdk.app.isInstalled.mockReturnValue(true);
+    process.on('unhandledRejection', () => {
```
sarahlessner, line 124: This is essentially the equivalent I found for how to ignore that unhandled rejection error the Vitest way, similar to what is in the `setupTests` file for Jest. But this is a bit better since it only swallows that error for this particular test, so we don't risk our test not catching legitimate errors that could crop up in future development.

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.spec.tsx
```diff
@@ -1,14 +1,16 @@
+// @ts-nocheck
```
mgoudy91, line 1: do we still need this? or maybe more line-specific rules?

apps/google-analytics-4/frontend/test/mocks/mockCma.ts
```diff
@@ -1,6 +1,11 @@
 const mockCma: any = {
   appSignedRequest: {
-    create: () => ({}),
+    create: () =>
+      Promise.resolve({
+        additionalHeaders: 'lol',
+        'X-Contentful-ServiceAccountKeyId':
+          'eyJpZCI6IlBSSVZBVEVfS0VZX0lEIiwiY2xpZW50SWQiOiJDTElFTlRfSUQiLCJjbGllbnRFbWFpbCI6IkNMSUVOVF9JRCIsInByb2plY3RJZCI6IlBST0pFQ1RfSUQifQ==',
```
mgoudy91, line 7: thinking about this more, I wonder if we should just use a string of the same length? not because the header representation of this service test key is so sensitive, but just so that anyone looking at this in the future doesn't wonder if we added an actually in-use key?

apps/google-analytics-4/frontend/src/hooks/useSidebarSlug/useSidebarSlug.spec.tsx
sarahlessner: Yeah I meant to revisit that good call out. Vite's `importActual` method that replaces the jest version returns a promise, which had its typing incompatible with our SDK mocking. This test file had that error in like 5 places. I typed the block as `any` type and left a comment about it because I'm not totally sure how to resolve without fully rewriting the tests and I don't understand what we're trying to test well enough to do that with any confidence.

apps/google-analytics-4/frontend/test/mocks/mockCma.ts
sarahlessner: I think we could just remove this change entirely now that I went about resolving the test failures differently.

mgoudy91 (approved):

PR #9344, opened 2024-12-17
Remove unused CONTENTFUL_HOST environment variable

twogood:
## Purpose

Solve problem for Windows user that posted in Contentful Discord:

https://discord.com/channels/1030176999471321129/1317987388567457835

## Approach

Use `cross-env` so that the scripts works on Windows too.

Note: `CONTENTFUL_HOST` is read by the `host` variable in http.ts but then the `host` variable is not actually used in http.ts.

## Testing steps

Run affected scripts at command line.

## Breaking Changes

No.

No.

## Deployment

No, this is example code.


examples/native-external-references/package.json
```diff
@@ -25,8 +25,8 @@
     "prettier": "prettier --write functions",
     "open-settings": "contentful-app-scripts open-settings",
     "install-app": "contentful-app-scripts install",
-    "create-resource-entities": "CONTENTFUL_HOST=api.contentful.com ts-node -T -r dotenv/config ./src/tools/create-resource-entities.ts",
-    "show-resource-entities": "CONTENTFUL_HOST=api.contentful.com ts-node -T -r dotenv/config ./src/tools/show-resource-entities.ts"
+    "create-resource-entities": "ts-node -T -r dotenv/config ./src/tools/create-resource-entities.ts",
```
Seth-Carter, line 28:
```suggestion
    "create-resource-entities": "cross-env CONTENTFUL_HOST=api.contentful.com ts-node -T -r dotenv/config ./src/tools/create-resource-entities.ts",
```

examples/native-external-references/package.json
```diff
@@ -25,8 +25,8 @@
...
+    "show-resource-entities": "ts-node -T -r dotenv/config ./src/tools/show-resource-entities.ts"
```
Seth-Carter, line 29:
```suggestion
    "show-resource-entities": "cross-env CONTENTFUL_HOST=api.contentful.com ts-node -T -r dotenv/config ./src/tools/show-resource-entities.ts"
```

examples/native-external-references/src/tools/http.ts
```diff
@@ -19,8 +19,6 @@ import {
 type ResourceProviderResult = APIError | APIResourceProvider;
 type ResourceTypeResult = APIError | APIResourceType;
```
Seth-Carter, line 21:
```suggestion

const host = process.env.CONTENTFUL_HOST ?? 'api.contentful.com';

```

Seth-Carter (changes requested):
Let's just leave it as is and go head and use `cross-env`.

twogood:
> Let's just leave it as is and go head and use `cross-env`.

PR updated based on feedback. Commit history got weird though.

useyourillusions:
Hi!
Cross-env could be a great solution for cases where we truly need this environment variable, but it's entirely unusable within the source code. Wouldn't it make more sense to remove it?

PR #9390, opened 2025-01-10
fix: focal point app bug with focal point field validation [ZEND-5811]

sarahlessner:
A customer reported a bug in the validation logic within the image focal point app.

image

This image shows the state of the editor when you open up an entry that has a focal point selected, and then you click `reset focal point`. 

Focal point is a required field in the app, so the expected behavior is that there should be a validation error on that field, and the `publish` button should be disabled. 

This PR fixes the validation logic to reset the focal point app field to undefined instead of an object with {focalPoint: null} so that the editor form validation works as expected. 

I also made a CSS improvement after noticing there is no space between elements for the Focal point field. 

Fix shown:

image


apps/image-focal-point/src/components/AppView/__snapshots__/AppView.spec.jsx.snap
sarahlessner, line 1: I had a failing snapshot that I updated, but it looked like there were some other obsolete snapshots that came up in the test run, so when I pressed `u` to update my failing snap, these ones were removed.

mgoudy91 (approved):
Good catch and good fix, everything looks sensible to me!

As always, appreciate the great notes on your PR

PR #9423, opened 2025-01-23
feat: add new app parameter example app [EXT-6096]

whitelisab:
## Purpose

This new example app demonstrates how app developers can utilize the three app parameter types (installation, instance, and invocation). We have received feedback and support requests that our documentation around app parameters could be improved, so this is a step to provide a better example to help developers utilize this important part of the app framework.

## Approach

Developers can follow along in the README to learn more about how to define, set, and read parameter values. A setup script is also included to get set up quickly with this app.


BobHemphill76:
This looks great.  You did a great job breaking down the differences of the three types of params and where they are used.  My only question is if theres a way to run through cca with this example and follow the readme to ensure the instructions are accurate.

whitelisab:
> This looks great. You did a great job breaking down the differences of the three types of params and where they are used. My only question is if theres a way to run through cca with this example and follow the readme to ensure the instructions are accurate.

Thanks for reviewing! Yes, I was able to test the app with the CCA. This was a great suggestion because it actually helped me to identify a bug on the app config page once I ran through the newly created app again!

jjolton-contentful (approved):

PR #9437, opened 2025-01-30
chore(slack-app): replace ngrok with localtunnel []

denkristoffer:
## Purpose

I couldn't get the dev script to work so I had to modernise a bit. The programmatic npm API was removed in version 8, so it's been gone for a long time.

ngrok v1 is no longer in development and later versions cost money to use. [`localtunnel`](https://github.com/localtunnel/localtunnel) seems like a good replacement.

Overall, I think the question is if this script is even in use? If not, maybe we should remove it and update the README on how to deploy or test this app.


denkristoffer:
I will leave this here in case anyone needs it, but I'll close the PR.

PR #9491, opened 2025-02-20
feat: functions templates and examples [EXT-6022]

michaelpineirocontentful:
## Purpose

Moving function templates and examples to it's own folder will:
A. Prevent cluttering in the `examples` folder
B. Designate a spot for `create-app-function` to pull from
C. Create a more extensible experience by separating app templates from function templates

## Approach

This is a copy-and-paste of existing function templates and the app-event-handler example.


BobHemphill76:
This looks like a no-op as far as create-contentful-app is concerned?  If that's true I'm good and this is a great starting point to enriching how we manage apps

BobHemphill76 (approved):

PR #9497, opened 2025-02-25
fix: correct banner date in graphql app [INTEG-2425]

sarahlessner:
The banner saying we're sunsetting the app has a date of 2-31-25 instead of 3-31-25


primeinteger (approved):
🚀

mgoudy91 (approved):

PR #9543, opened 2025-03-10
chore(native-external-reference): updated the folder name for native external reference example[FUS-414]

vikaskumr:
## Purpose

We will be having 2 examples now for native external references one for REST and another for GraphQL, so we are now changing the name of the folder according to the api using in example.
Renaming the native-external-reference example app to `native-external-reference-tmdb`

> Do not merge until documentation is updated.


michaelpineirocontentful (approved):
🚀🚀🚀

michaelpineirocontentful (approved):

PR #9545, opened 2025-03-10
Braze integ 2403

joaquincasal:
Add 3rd code block: json response example


apps/braze/src/helpers/assembleQuery.ts
```diff
@@ -82,6 +76,25 @@ export function assembleQuery(
...
+  const response = await fetch(`https://graphql.contentful.com/content/v1/spaces/${spaceId}`, {
+    method: 'POST',
+    headers: {
+      Authorization: `Bearer ${token}`,
+      'Content-Type': 'application/json',
+    },
+    body: query,
+  });
+  if (!response.ok) {
+    return null;
```
JuliRossi, line 89: 💭 Should we handle errors ?

PR #9557, opened 2025-03-12
Setup dialog steps

joaquincasal:
Add the ability to move forward and backwards in the dialog.

https://github.com/user-attachments/assets/60b0aaa7-8112-445a-b9e1-e9f59b8a28dd

We'll probably want to split the code in this component into multiple, but I'd rather do it when we know what to put into each

### DIFF
--- a/apps/braze/src/locations/Dialog.tsx
+++ b/apps/braze/src/locations/Dialog.tsx
@@ -8,20 +8,11 @@ import {
 } from '../helpers/assembleQuery';
 import generateLiquidTags from '../helpers/generateLiquidTags';
 
-import {
-  Box,
-  Button,
-  Flex,
-  List,
-  ListItem,
-  Paragraph,
-  Subheading,
-  TextLink,
-} from '@contentful/f36-components';
-import { ExternalLinkIcon } from '@contentful/f36-icons';
+import { Box, Button, Flex } from '@contentful/f36-components';
 import { useEffect, useState } from 'react';
-import tokens from '@contentful/f36-tokens';
-import Splitter from '../components/Splitter';
+import FieldsSelectionStep from '../components/FieldsSelectionStep';
+import CodeBlocksStep from '../components/CodeBlocksStep';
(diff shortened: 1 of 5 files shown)


apps/braze/src/locations/Dialog.tsx
```diff
@@ -37,25 +66,110 @@ const Dialog = () => {
     fetchEntry();
   }, []);
 
+  let dialogBody;
+  if (step === 'fields') {
+    dialogBody = (
```
JuliRossi, line 71: 💭 Could we extract the steps into different components?

joaquincasal: done!

apps/braze/src/locations/Dialog.tsx
```diff
@@ -38,24 +58,52 @@ const Dialog = () => {
...
+      paddingTop="spacingM"
+      paddingLeft="spacingL"
+      paddingRight="spacingL">
+      {step === 'fields' && <FieldsSelectionStep></FieldsSelectionStep>}
+      {step === 'locales' && <LocalesSelectionStep></LocalesSelectionStep>}
+      {step === 'codeBlocks' && (
+        <CodeBlocksStep
+          connectedContentCall={connectedContentCall}
+          liquidTags={liquidTags}
+          graphqlResponse={graphqlResponse}></CodeBlocksStep>
```
joaquincasal, line 72: I thought about moving the logic to generate these props to inside of the component. I'm convinced it's better, so I left it as it is, but we can consider moving that logic

apps/braze/src/locations/Dialog.tsx
```diff
@@ -38,24 +58,52 @@ const Dialog = () => {
...
+      )}
+
+      <Flex
```
joaquincasal, line 75: I don't really like this part...I'd like each component to display the buttons, but I want to avoid having the <Flex> logic repeated in each of the steps. So I left it here for the time being

JuliRossi (approved):

PR #9564, opened 2025-03-17
Braze app: allow to select locales

joaquincasal:
## Purpose

Users are now able to select the locales so they can be used when generating the liquid tags

## Approach

The locales are retrieved from the space. If the entry has no localized fields, the step is skipped.

## Testing steps

Use the app with an entry that has localized fields

## Breaking Changes

N/A

N/A

## Deployment

N/A

### DIFF
--- a/apps/braze/src/components/FieldsSelectionStep.tsx
+++ b/apps/braze/src/components/FieldsSelectionStep.tsx
@@ -1,21 +1,41 @@
-import { Paragraph, TextLink } from '@contentful/f36-components';
+import { Button, Flex, Paragraph, TextLink } from '@contentful/f36-components';
 import { ExternalLinkIcon } from '@contentful/f36-icons';
 
-const FieldsSelectionStep = () => {
+type FieldsSelectionStepProps = {
+  handleNextStep: () => void;
+};
+const FieldsSelectionStep = (props: FieldsSelectionStepProps) => {
+  const { handleNextStep } = props;
   return (
-    <Paragraph fontColor="gray700" lineHeight="lineHeightCondensed">
-      Select which fields you would like to include in your Connected Content call. Selecting fields
-      from referenced entries is limited to 5 nested references. For more information on Braze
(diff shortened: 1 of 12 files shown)


apps/braze/src/components/LocalesSelectionStep.tsx
```diff
@@ -1,10 +1,74 @@
...
   return (
-    <Paragraph fontColor="gray700" lineHeight="lineHeightCondensed">
-      Select the locales you want to reference in Braze messages.
-    </Paragraph>
+    <>
+      <Paragraph fontColor="gray700" lineHeight="lineHeightCondensed">
+        Select the locales you want to reference in Braze messages.
+      </Paragraph>
+
+      <FormControl isRequired isInvalid={selectedLocales.length === 0}>
```
JuliRossi, line 32: 💭 This may be a question for Erin but I think we may want to use as default value the default local instead of making the user choose

joaquincasal: We are doing that now! In the Dialog the selected locales are initialized with the default locale as the only value

apps/braze/src/components/FieldsSelectionStep.tsx
```diff
@@ -1,21 +1,41 @@
...
+
+      <Flex
+        padding="spacingM"
+        gap="spacingM"
+        justifyContent="end"
+        style={{
+          position: 'sticky',
+          bottom: 0,
+          background: 'white',
+        }}>
```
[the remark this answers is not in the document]
joaquincasal: done!

JuliRossi (approved):

apps/braze/src/components/LocalesSelectionStep.tsx
joaquincasal: Exactly! They shouldn't be able to move forward if they deselect all of them

PR #9565, opened 2025-03-17
Braze refactor

joaquincasal:
## Purpose

Replace field types with classes for easier usage.

## Approach

There's a class for each field type. Each class knows how to generate their part of the graphql query and also the liquid tags.

## Testing steps

Automated steps were added.

## Breaking Changes

Types are no longer available, the new classes should be used instead.

N/A

## Deployment

N/A


apps/braze/src/fields/BasicField.ts
```diff
@@ -0,0 +1,44 @@
...
+  ) {
+    super(id, entryContentTypeId, localized);
+    this.type = type;
+  }
+
+  generateQuery(): string {
+    switch (this.type) {
+      case 'RichText':
+        return '';
+      case 'Location':
```
joaquincasal, line 24: Move to another class

apps/braze/src/helpers/assembleQuery.ts
```diff
@@ -107,69 +34,5 @@ export function assembleQuery(contentTypeId: string, entryId: string, entryField
 }
 
 function assembleFieldsQuery(entryFields: Field[]): string {
```
joaquincasal, line 36: Maybe move this logic into a Entry class?

apps/braze/src/helpers/assembleQuery.ts
```diff
@@ -96,9 +26,6 @@ export async function getGraphQLResponse(spaceId: string, token: string, query:
     },
     body: query,
   });
-  if (!response.ok) {
-    return null;
-  }
   return response.json();
```
joaquincasal, line 29: Maybe add the indication that the request failed, so we can show it accordingly in the third code block

apps/braze/src/fields/BasicField.ts
```diff
@@ -0,0 +1,15 @@
+import { Field } from './Field';
+
+export class BasicField extends Field {
```
JuliRossi, line 3: nit: I think we should add a comment specifying which are the "BasicFields" (Symbol, number, decimal, date, bool, json, Text, etc) for future reference.

apps/braze/src/fields/AssetArrayField.ts
```diff
@@ -0,0 +1,26 @@
+import { ASSET_FIELDS, ASSET_FIELDS_QUERY } from '../utils';
+import { Field } from './Field';
+
+export class AssetArrayField extends Field {
+  constructor(id: string, entryContentTypeId: string, localized: boolean) {
+    super(id, entryContentTypeId, localized);
+    this.id = id;
+    this.localized = localized;
```
JuliRossi, line 8: nit: I think these aren't needed

apps/braze/src/fields/RichTextField.ts
```diff
@@ -0,0 +1,15 @@
+import { Field } from './Field';
+
+export class RichTextField extends Field {
```
JuliRossi, line 3: :nit: I'd add a comment here saying, we are not handling rich text for now.

JuliRossi (approved):
Incredible work Joa! The code really needed this refactor. Added a few no-blocking comments.

PR #9573, opened 2025-03-18
Fix queries case

joaquincasal:
This PR includes two main changes:

- Save the id of the content type instead of the name
- Save the content type setting the first letter to lowercase

### DIFF
--- a/apps/braze/src/utils.ts
+++ b/apps/braze/src/utils.ts
@@ -18,5 +18,5 @@ export function capitalize(s: string): string {
 }
 
 export function firstLetterToLowercase(s: string): string {
-  return s.charAt(0).toLocaleLowerCase() + s.slice(1);
+  return s.charAt(0).toLowerCase() + s.slice(1);
 }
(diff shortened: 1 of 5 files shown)


apps/braze/src/utils.ts
```diff
@@ -16,3 +16,7 @@ export const SIDEBAR_BUTTON_TEXT = 'Generate Braze Connected Content';
 export function capitalize(s: string): string {
   return s.charAt(0).toUpperCase() + s.slice(1);
 }
+
+export function firstLetterToLowercase(s: string): string {
+  return s.charAt(0).toLocaleLowerCase() + s.slice(1);
```
JuliRossi, line 21: This should be `toLowerCase()` instead of `.toLocaleLowerCase()`

JuliRossi, line 21:
```suggestion
  return s.charAt(0).toLowerCase() + s.slice(1);
```

joaquincasal: Good catch!

JuliRossi (approved):

PR #9576, opened 2025-03-19
Braze app generate localize content

JuliRossi:

### DIFF
--- a/apps/braze/src/fields/Entry.ts
+++ b/apps/braze/src/fields/Entry.ts
@@ -7,7 +7,6 @@ export class Entry {
   public fields: Field[];
   private spaceId: string;
   private contentfulToken: string;
-  private query: string;
   constructor(
     id: string,
     contentType: string,
@@ -20,12 +19,11 @@ export class Entry {
     this.fields = fields;
     this.spaceId = spaceId;
     this.contentfulToken = contentfulToken;
-    this.query = this.assembleQuery();
   }
 
   generateConnectedContentCall(locales: string[]) {
     return `{% capture body %}
-    ${locales.length > 0 ? this.localizeQuery(locales) : this.query}
+    ${this.assembleQuery(locales)}
   {% endcapture %}
   
   {% connected_content
@@ -56,7 +54,7 @@ export class Entry {
           Authorization: `Bearer ${this.contentfulToken}`,
           'Content-Type': 'application/json',
         },
-        body: locales?.length > 0 ? this.localizeQuery(locales) : this.query,
+        body: this.assembleQuery(locales),
       }
     );
     return response.json();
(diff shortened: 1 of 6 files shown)


apps/braze/src/fields/Entry.ts
```diff
@@ -38,11 +36,16 @@ export class Entry {
   %}`;
   }
 
-  generateLiquidTags() {
+  generateLiquidTags(locales: string[]) {
+    if (locales.length > 0) {
```
joaquincasal, line 40: Should this be `> 1`?

apps/braze/src/fields/Entry.ts
```diff
@@ -61,8 +64,23 @@ export class Entry {
     return this.fields.some((field) => field.localized);
   }
 
-  private assembleQuery() {
-    return `{"query":"{${this.contentType}(id:\\"${this.id}\\"){${this.assembleFieldsQuery()}}}"}`;
+  assembleQuery(locales: string[]) {
+    const body = locales.length > 0 ? this.localizedQueryBody(locales) : this.queryBody();
```
joaquincasal, line 68: Should this be `> 1`?

joaquincasal (commented):
Really nice! Left two questions

apps/braze/src/fields/Entry.ts
```diff
@@ -38,11 +36,16 @@ export class Entry {
   %}`;
   }
 
-  generateLiquidTags() {
+  generateLiquidTags(locales: string[]) {
+    if (locales.length > 0) {
+      return locales.flatMap((locale) =>
+        this.fields.flatMap((field) => field.generateLiquidTag(locale))
+      );
```
FBanfi, line 43: 💭 Perhaps we could extract this flat maps in 2 different methods, one for localized liquid tags generation and other for those which are not localized. I dont think it is necessary or complex enough to justify it, but it would be easier to read and understand.

FBanfi (commented):
Really nice codeee! 🙌 I just added one silly comment that came to my mind.

apps/braze/src/fields/Entry.ts
JuliRossi: It shouldn't. If the user selects a locale different to the default, it should be only that one that shows. Only if we are skipping the locale step, then we should show the unlocalized version. But this made me realize we need to change the initialization for the SelectedLocales in the Dialog.

apps/braze/src/fields/Entry.ts
JuliRossi: Same as above

apps/braze/src/locations/Dialog.tsx
```diff
@@ -25,7 +25,7 @@ const Dialog = () => {
   const locales = sdk.locales.available;
   const entryInfo = sdk.parameters.invocation as EntryInfo;
   const [step, setStep] = useState('fields');
-  const [selectedLocales, setSelectedLocales] = useState<string[]>([sdk.locales.default]);
+  const [selectedLocales, setSelectedLocales] = useState<string[]>([]);
```
joaquincasal, line 28: With this change the checkbox for the default locale won't be selected if the user has to go through the locale selection step. Maybe we can set it to contain the defualt locale if `shouldChooseLocales` is true?

apps/braze/src/fields/Entry.ts
JuliRossi: Done!

PR #9524, opened 2025-03-05
Feat/upgrade example app to vite [INTEG-2434]

david-shibley-contentful:

PR #9585, opened 2025-03-20
fix: update Vite config for delivery function examples [EXT-6273]

whitelisab:
## Purpose

We updated the example apps to use Vite instead of CRA/react-scripts in [this PR](https://github.com/contentful/apps/pull/9524). However, a few things were missed, so this PR aims to clean those up.

## Approach

Here's an overview of the changes:

### Vite updates

- Ensure Vite, Vitest, @vitejs/plugin-react and any other dependencies are up to date
- Ensure that in the Vite config, the build directory is set to `./build` instead of the default `./dist`. This is because the `build-functions` script from the `@contentful/app-scripts` package expects this to be the name for the build directory
- Add `"base": ''" to the Vite config to support relative paths, more info [here](https://www.contentful.com/developers/docs/extensibility/app-framework/app-bundle/#limitations)
- Ensure Vitest is configured properly; add `/// <reference types="vitest/config" />` when using Vitest, adding the triple slash command as documented [here](https://vitest.dev/config/)
- Rename the Vite config file to address [this warning](https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated)
- Add `start` script in package.json in addition to to `dev`

### `tsconfig.json` updates

- Remove the CRA `tsconfig` and the associated `react-app-env.d.ts` file; add any necessary configuration after removing that
- Ensure that the functions directory has its own `tsconfig.json` like our function templates

### Contentful package dependency updates

- Update `@contentful/app-scripts` and `@contentful/node-apps-toolkit`
- In the functions code, use the `FunctionTypeEnum` instead of string values

### Documentation updates

- Remove references to Create React App
- Remove references to the legacy GraphQL Playground app and instead link to the GraphiQL app

### Example-specific changes

The `function-mock-shop` app had some mistakes copied from another app in the `package.json file`, so I've addressed those.

The `native-external-references-tmdb` example needed some additional updates in the test files to move from Jest to Vitest.

## Testing steps

I tested these locally in the apps repo and also tested them using CCA according [these instructions](https://github.com/contentful/create-contentful-app/blob/main/README.md#local-testing-of-an-examples-in-the-apps-repo)


examples/function-mock-shop/package.json
```diff
@@ -41,19 +39,19 @@
     ]
   },
   "devDependencies": {
-    "@contentful/app-scripts": "1.33.2",
+    "@contentful/app-scripts": "^2.3.0",
```
BobHemphill76, line 42: Nice!  I believe that's add-locales release!

examples/native-external-references-tmdb/tsconfig.json
```diff
@@ -1,11 +1,11 @@
...
   "compilerOptions": {
     "target": "es5",
     "lib": ["dom", "dom.iterable", "esnext"],
     "moduleResolution": "node",
     "resolveJsonModule": true,
+    "esModuleInterop": true,
     "jsx": "react"
   },
-  "include": ["./src/**/*.ts", "./setupTests.ts", "./functions/**/*.ts"]
+  "exclude": ["functions"]
```
BobHemphill76, line 10: This is because build:functions uses a different tsc?

BobHemphill76 (approved):

examples/function-mock-shop/package.json
whitelisab: Yes, this should bump all of these examples so they'll have that available!

examples/native-external-references-tmdb/tsconfig.json
whitelisab: Exactly! I ended up updating this example to follow the convention we had in some examples where we have separate tsconfigs for the frontend app and then the functions directory

PR #9590, opened 2025-03-20
Braze app

joaquincasal:
## Purpose

This is the first version of the Braze app.

## Approach

The main logic of the app is in the Dialog location, which consists of a three-step wizard: field selection, locale selection and code blocks. There's one component for each of the steps of the wizard to separate their logic.
Under the `fields` folder there are typescript classes that represent the different types of contentful fields. The idea is to separate the logic of generating the graphql query and the liquid tags for each field, since each field has different information.
The `FieldsFactory` class retrieves the necessary information from the entry fields and content type and creates one instance of the correspoding class for each field of the entry.
Reference fields can contain other nested fields or references up to 5 levels.

## Testing steps

- Automated tests can be run with `npm test`.
- For testing the app in a Contentful entry, install the app in a space. In the app registration set the locations for `App configuration` and `Entry sidebar`. Then add the app to the sidebar for the content type that will be used for testing. Then start the app with `npm start` and go to an entry of that content type, a button should appear in the sidebar, which will trigger the dialog to appear. The second step of the wizard only appears if there are localized fields in the entry, otherwise it's skipped.
  - Once in the third step, three code blocks will be shown: connected content call, liquid tags and example response. The generated connected content call and the liquid tags can be pasted into a content block in Braze to verify that they work.

## Breaking Changes

N/A

N/A

## Deployment

N/A


apps/braze/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "react-dom": "18.3.1",
+    "react-syntax-highlighter": "^15.6.1"
+  },
+  "scripts": {
+    "start": "vite",
+    "build": "tsc && vite build",
+    "test": "vitest",
+    "create-app-definition": "contentful-app-scripts create-app-definition",
+    "upload": "contentful-app-scripts upload --bundle-dir ./dist",
+    "upload-ci": "contentful-app-scripts upload --ci --bundle-dir ./dist  --organization-id $CONTENTFUL_ORG_ID --definition-id $CONTENTFUL_APP_DEF_ID --token $CONTENTFUL_ACCESS_TOKEN"
```
mgoudy91, line 26:
```suggestion
```
We can remove this two for now, since they have implications if we merge this PR. I'll re-add them with the correct values post-merge

apps/braze/src/fields/RichTextField.ts
mgoudy91, line 1: 👍

apps/braze/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,144 @@
...
+import Splitter from '../components/Splitter';
+
+export interface AppInstallationParameters {
+  apiKey: string;
+}
+
+export const BRAZE_DOCUMENTATION =
+  'https://braze.com/docs/user_guide/personalization_and_dynamic_content/connected_content';
+
+export async function callToContentful(url: string, newApiKey: string) {
```
mgoudy91, line 26: I really like this pattern and will probably use it for other apps we maintain 😄

mgoudy91 (commented):
I really cannot find anything meaningful to be concerned with here. This is a really well-organized and written app. Great work 🎉 

Two procedural points:
1. We'll need to remove the two deploy commands for now. I left a suggestion to do so. Happy to approve once that's done!
2. Looks like the rebase can't be automatic, which is totally fine. Assuming no one on the marketplace team has any concerns, I'll go ahead and squash the history and merge this in on Monday.

apps/braze/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "happy-dom": "^15.11.6",
+    "react": "18.3.1",
+    "react-dom": "18.3.1",
+    "react-syntax-highlighter": "^15.6.1"
+  },
+  "scripts": {
+    "start": "vite",
+    "build": "tsc && vite build",
+    "test": "vitest",
+    "create-app-definition": "contentful-app-scripts create-app-definition",
```
joaquincasal, line 24:
```suggestion
    "create-app-definition": "contentful-app-scripts create-app-definition"
```

mgoudy91 (approved):

PR #9593, opened 2025-03-24
fix: update Vite config for example apps [EXT-6273]

whitelisab:
## Purpose

We updated the example apps to use Vite instead of CRA/react-scripts in [this PR](https://github.com/contentful/apps/pull/9524). However, a few things were missed, so this PR aims to clean those up. We fixed up the delivery function examples in [this other PR](https://github.com/contentful/apps/pull/9585), so this current PR updates the rest of the examples that were migrated to Vite.

## Approach

Here's an overview of the changes:

### Vite updates

- Ensure Vite, Vitest, @vitejs/plugin-react and any other dependencies are up to date
- Ensure that in the Vite config, the build directory is set to `./build` instead of the default `./dist`. We are standardizing on this for our examples because the `build-functions` script from the `@contentful/app-scripts` package expects this to be the name for the build directory. Even for apps that don't have functions, we want to be consistent in case the user adds functions to their app.
- Add `"base": ''" to the Vite config to support relative paths, more info [here](https://www.contentful.com/developers/docs/extensibility/app-framework/app-bundle/#limitations)
- Ensure Vitest is configured properly; add `/// <reference types="vitest/config" />` when using Vitest, adding the triple slash command as documented [here](https://vitest.dev/config/)
- Rename the Vite config file to address [this warning](https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated)
- Add `start` script in package.json in addition to to `dev`

### `tsconfig.json` updates

- Remove the CRA `tsconfig` and the associated `react-app-env.d.ts` file; add any necessary configuration after removing that
- Ensure that the functions directory has its own `tsconfig.json` like our function templates

### Documentation updates

- Remove references to Create React App in `README` files

### Example-specific changes

The `app-actions` app did not need to be updated to Vite since it already uses `next`. So I reverted the changes that were previously made.

## Testing steps

I tested these locally in the apps repo and also tested them using CCA according [these instructions](https://github.com/contentful/create-contentful-app/blob/main/README.md#local-testing-of-an-examples-in-the-apps-repo). I ran the apps locally and also built and uploaded them.


examples/blog-post-metrics/vite.config.mts
```diff
@@ -1,3 +1,4 @@
+/// <reference types="vitest/config" />
```
david-shibley-contentful, line 1: is this needed?

examples/editor-assignment/vite.config.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference types="vitest/config" />
```
david-shibley-contentful, line 1: what does this do?

david-shibley-contentful (approved):

PR #9610, opened 2025-03-27
fix: update appevent-transformation example [EXT-6286]

whitelisab:
## Purpose

Since creating the new function templates, some of our function examples were not working properly due to the sub-directory setup. This PR changes the App Event Transformation Function example to be a single frontendless app example that can be bootstrapped via the CCA using `npx create-contentful-app@latest --example function-appevent-transformation`.

## Approach

The example originally used a google maps api for geocoding. I changed this to a free opensource  geocoding api ([Nominatim](https://nominatim.org/)) so that users wouldn't need to set up an API Key.

Here is a summary of the changes:
- Removed the TS/JS sub-directories in favor of a singular app example
- This example has no frontend; the only code is for functions and some setup scripts
- Add more details to the `INSTRUCTIONS` file, including how to use the scripts to easily get setup 
- Added a script for creating an app event subscription that links to the transformation function and a script to create a content type with the correct latitude and longitude fields and publish an entry
- Bumped `@contentful/node-apps-toolkit` and use the `FunctionTypeEnum`

I also made a small update to the comments in the filter example 😄 

## Testing steps

I tested this app locally and also via the CCA by pointing the local CCA to this branch. I ran through all the scripts in the instructions to ensure that they work properly.


examples/function-appevent-transformation/functions/appevent-transformation-example.ts
```diff
@@ -0,0 +1,66 @@
...
+// Use the Nominatim API to get an address from geocode coordinates
+// https://nominatim.org/release-docs/develop/api/Reverse/
+const geocodeApi = 'https://nominatim.openstreetmap.org/reverse';
+
+async function geocode(lat: number, long: number) {
+  const url = `${geocodeApi}?lat=${lat}&lon=${long}&format=json`;
+  try {
+    const response = await fetch(url, {
+      headers: {
+        'User-Agent': 'Contentful-App-Event-Transformation/1.0',
```
whitelisab, line 18: The Nominatim API requires a valid HTTP Referer or User-Agent identifying the application in their usage policy: https://operations.osmfoundation.org/policies/nominatim/, so I added this. Definitely open to other options here if you have suggestions!

michaelpineirocontentful (approved):

examples/function-appevent-transformation/functions/appevent-transformation-example.ts
michaelpineirocontentful: +1 this! Looks good

PR #9611, opened 2025-03-27
fix: braze checkbox fields height [INTEG-2499]

FBanfi:
## Purpose

- The fields that contain nested fields have a bigger height than the ones without nested fields, which is due to the height of the checkbox container. 
- They should all be the same height. 
- Using spacingXl would work fine.

## Approach

Configuring the container component's style to set its height to `spacingXL`, extracting a new component to eliminate duplicated logic, and simplifying the code.

Now, it looks like this:

https://github.com/user-attachments/assets/72971ead-0319-4124-8bee-59e8682b4069

Link to [INTEG-2499](https://contentful.atlassian.net/browse/INTEG-2499) ticket

[INTEG-2499]: https://contentful.atlassian.net/browse/INTEG-2499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/braze/src/components/CheckboxCard.styles.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const checkboxCard: CSSProperties = {
+  border: `1px solid ${tokens.gray200}`,
+  borderRadius: tokens.borderRadiusSmall,
+  height: tokens.spacingXl,
+};
+
+export const iconButton: CSSProperties = {
+  minHeight: '0',
+};
```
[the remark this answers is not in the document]
FBanfi: Oh yes, I missed that.

Deleted! ✓

FBanfi (commented):
Correction of comments done!

JuliRossi (approved):

apps/braze/src/components/CheckboxCard.styles.ts
```diff
@@ -0,0 +1,8 @@
+import { CSSProperties } from 'react';
+import tokens from '@contentful/f36-tokens';
+
+export const checkboxCard: CSSProperties = {
+  border: `1px solid ${tokens.gray200}`,
+  borderRadius: tokens.borderRadiusSmall,
+  height: tokens.spacingXl,
+};
```
primeinteger, line 8: It might be nice to switch these styles over to using `@emotion/css` since that's used in many other apps in this repo to define css for components. Not required but just a thought.

primeinteger (approved):
Nice work on this! I like that you separate the CheckboxCard into a separate component and that you are using lots of design tokens in the css. Just one thought below about possibly using `@emotion/css` but that's not a big deal so Approving

apps/braze/src/components/CheckboxCard.styles.ts
primeinteger: For example like this:
https://github.com/contentful/apps/blob/39de3b3a1964827c37565f50f0e5b9ec3967ec73/apps/ai-image-generator/frontend/src/components/config/CostSection/CostSection.styles.ts

primeinteger (approved):

PR #9612, opened 2025-03-28
fix: update comment-bot example to be a frontendless app [EXT-6286]

whitelisab:
## Purpose

Since creating the new function templates, some of our function examples were not working properly due to the sub-directory setup. This PR changes the Comment Bot Function example to be a single frontendless app example that can be bootstrapped via the CCA using `npx create-contentful-app@latest --example function-comment-bot`.

## Approach

Here is a summary of the changes:
- Removed the TS/JS sub-directories in favor of a singular app example
- This example has no frontend; the only code is for functions and some setup scripts
- Add more details to the `INSTRUCTIONS` file, including how to use the scripts to easily get setup 
- Added a script for creating an app event subscription that links to the handler function, a script to create a content type/entry, and two scripts for adding comments for `/hide-publish` and `/show-publish`
- Bumped `@contentful/node-apps-toolkit` and use the `FunctionTypeEnum`
- Added the default editor interface in the cases where no sidebar value is in the editor interface so that the defaults remain

## Testing steps

I tested this app locally and also via the CCA by pointing the local CCA to this branch. I ran through all the scripts in the instructions to ensure that they work properly.


examples/function-comment-bot/functions/comment-bot.ts
```diff
@@ -16,6 +17,8 @@ export const handler: EventHandler<'appevent.handler'> = async (
   const { parentEntity } = createCommentBody.sys.newComment.sys;
   const commentBody = createCommentBody.sys.newComment.body;
 
+  console.log(`Received comment: ${commentBody}`);
```
whitelisab, line 20: This log is intentional, I wanted to add a log that users could see when inspecting the function logs

examples/function-comment-bot/INSTRUCTIONS.md
```diff
@@ -0,0 +1,187 @@
...
+
+To run the scripts:
+
+```bash
+npm run create-content-type-and-entry
+npm run hide-publish
+npm run show-publish
+```
+
+Since the publish button by default is added to the sidebar, run `npm run hide-publish` first. Then check the Contentful web app to see that the publish button is no longer visible in the web app. The run `npm run show-publish` to make it visible again.
```
t-col, line 143:
```suggestion
Since the publish button by default is added to the sidebar, run `npm run hide-publish` first. Then check the Contentful web app to see that the publish button is no longer visible in the web app. Then run `npm run show-publish` to make it visible again.
```

examples/function-comment-bot/src/tools/show-publish.ts
```diff
@@ -0,0 +1,29 @@
...
+        {
+          body: '/show-publish',
+          status: 'active',
+        }
+      );
+
+      console.log('Comment created');
+      console.dir(comment, { depth: 5 });
+    } catch (error) {
+      console.error(error);
```
t-col, line 24: should this throw, or do we want to just swallow the error?

examples/function-comment-bot/INSTRUCTIONS.md
whitelisab: Good catch, thanks! I also added a small note about needing to refresh the web app to see the changes to the publish button

examples/function-comment-bot/src/tools/show-publish.ts
whitelisab: I think it's ok to swallow the error here, since it's just a script that you'll run using `npm run show-publish` and you see the output of the error in the terminal when you are running the command. But happy to hear other suggestions/thoughts on this!

michaelpineirocontentful (approved):

PR #9616, opened 2025-03-31
Braze app: Adjust checkbox 'select all' behaviour [INTEG-2507]

joaquincasal:
## Purpose

We're updating the checkbox behavior per the designs.
Before: if at least one nested checkbox was selected, the parent checkbox was selected as well
Now: all nested checkboxes need to be selected in order for the parent to be automatically selected

## Approach

We've separated the "internal" `selected` logic (used for assembling the graphql query and the liquid tags) from the view logic (what is shown to the user).
The internal logic needs that the parent checkbox if at least one of the nested ones are selected (which is different from the view logic).

## Testing steps

Use the app with a content type that has references or array of references.

## Breaking Changes

N/A

The base of this PR is not master, so we may need to update it if the base branch gets squashed

## Deployment

N/A

https://github.com/user-attachments/assets/4d8c54ad-6f02-4d8d-9562-77d10e4bee53


apps/braze/src/components/FieldsSelectionStep.tsx
```diff
@@ -61,17 +99,29 @@ const FieldsSelectionStep = (props: FieldsSelectionStepProps) => {
         </TextLink>
       </Paragraph>
 
-      <CheckboxCard
```
joaquincasal, line 64: I'm going back to use a specific checkbox here because the logic for the `checked` state is different from the logic from fields (which is what is used in the `CheckboxCard` component)

apps/braze/src/components/FieldsSelectionStep.tsx
```diff
@@ -1,47 +1,85 @@
...
-      setEntrySelected(false);
+    const field = allFields.find((field) => field.uniqueId() === id);
+    if (!field) {
+      return;
+    }
+
+    field.toggle(checked);
+    const newSelectedFields = new Set(selectedFields);
+
+    if (checked) {
```
FBanfi, line 34: 💭 This logic is repeated in `toggleNestedFields`. Perhaps it could be extracted in a function such as:
```
const toggleAction = (checked: boolean, selectedSet: Set<string>, id: string): void => {
    checked ? selectedSet.add(id) : selectedSet.delete(id);
};
```

apps/braze/src/components/FieldsSelectionStep.tsx
```diff
@@ -1,47 +1,85 @@
...
+    toggleNestedFields(field, checked, newSelectedFields);
+    toggleParentField(field, newSelectedFields, allFields);
+
+    setSelectedFields(newSelectedFields);
+    setEntrySelected(newSelectedFields.size === allFields.length);
+  };
+
+  const toggleNestedFields = (field: any, checked: boolean, selectedFields: Set<string>): void => {
+    for (const child of field.getChildren()) {
+      if (checked) {
```
FBanfi, line 49: The [previous comment](https://github.com/contentful/apps/pull/9616/files#r2021473973) refers to this part.

apps/braze/src/components/FieldsSelectionStep.tsx
```diff
@@ -1,47 +1,85 @@
...
+      if (child.getChildren().length > 0) {
+        toggleNestedFields(child, checked, selectedFields);
+      }
     }
   };
 
+  const toggleParentField = (field: Field, selectedSet: Set<string>, allFields: any[]): void => {
+    if (!field.parent) return;
+
+    if (field.parent.getChildren().every((child) => selectedSet.has(child.uniqueId()))) {
```
FBanfi, line 64: 💭 We could extract this conditions too, I was thinking that it could be more readable or simpler.

FBanfi (dismissed):
Nice changes!! 💯 🙌 . I commented two minimal things.

PR #9621, opened 2025-04-01
Braze app: Add margin to locale dropdown [INTEG-2498]

joaquincasal:
If the space had several locales, the dropdown was not enterely visible, now it will be independently of the amount of locales

https://github.com/user-attachments/assets/05e70a75-b4e3-4e3f-a96b-da96f7c2d679


apps/braze/src/components/LocalesSelectionStep.tsx
```diff
@@ -30,12 +30,12 @@ const LocalesSelectionStep = (props: LocalesSelectionStepProps) => {
...
       </Paragraph>
 
-      <FormControl isRequired isInvalid={selectedLocales.length === 0}>
+      <FormControl isRequired isInvalid={selectedLocales.length === 0} marginBottom="spacing4Xl">
         <FormControl.Label>Locales</FormControl.Label>
 
         <Multiselect
           currentSelection={selectedLocales}
-          popoverProps={{ isFullWidth: true }}
+          popoverProps={{ isFullWidth: true, listMaxHeight: 108 }}
```
joaquincasal, line 38: Each Multiselect.Option is 36px height, that's where this 108 comes from (I'm not a fun of the harcoded height, but couldn't find a way to reuse the height of the option)

JuliRossi (commented):
Looks good to me, but I would check if Erin is okey with hiding the buttons when the multiselect is open

JuliRossi (approved):

FBanfi (approved):

mgoudy91 (approved):

PR #9677, opened 2025-04-10
Braze app: Specify environment in graphql request [INTEG-2564]

joaquincasal:
Graphql calls would only work if the default environment was used. We now specify the current environment so the request calls in any environment

https://github.com/user-attachments/assets/26fc4046-5e2e-4996-9970-a7d4d7436331


JuliRossi (approved):
Nice fix!

FBanfi (approved):

primeinteger (approved):
👍

PR #9698, opened 2025-04-14
Braze app: FieldsFactory refactor [INTEG-2495]

joaquincasal:
## Purpose

This inverts the way in which the fields from the entry are obtained, this way if a field is empty it can still be selected

## Approach

We now first get the content type from the entry (which is enough for most fields) and then the entry itself is fetched, including the references (this is used for reference fields only)

## Testing steps

Automated tests were updated.
For manual test, verify that the displayed fields in the first step are the ones in the entry, even if they are empty. A 'complex' content type is recommended.

## Breaking Changes

N/A

N/A

## Deployment

N/A

https://github.com/user-attachments/assets/1123910e-1b85-42ed-a6a7-93659d7c5dbf

> [!NOTE]
> This also solves INTEG-2560

https://github.com/user-attachments/assets/fb66e355-7e0b-4308-a3e8-6ad3bf9ac9fe


FBanfi (approved):
Awesome refactor! 💯 And I really liked how you extracted some of the logic and conditions from `createFields`. It’s much easier to follow now!

mgoudy91 (approved):

PR #9717, opened 2025-04-22
Braze app: create and generate sidebar buttons [INTEG-2535]

FBanfi:
## Purpose

This update adds a Create button to the sidebar, renames the existing one to Generate, and includes descriptions for each option.

## Approach

In addition of the changes mentioned above, this update also adds a new component to reuse some repeated logic between the config screen and the sidebar.

With these changes now the sidebar looks like this:
Captura de pantalla 2025-04-22 a la(s) 12 47 40 p  m

## Testing steps

Automated tests were added for the new component and the new sidebar button. Additionally, some updates were made to existing ones.

Link to the [ticket](https://contentful.atlassian.net/browse/INTEG-2535).

**Important note:** This PR doesn’t include the logic in the dialog to distinguish which button was clicked. That will be handled in a separate ticket.

### DIFF
--- a/apps/braze/src/locations/Sidebar.tsx
+++ b/apps/braze/src/locations/Sidebar.tsx
@@ -9,7 +9,6 @@ import {
   SIDEBAR_CREATE_BUTTON_TEXT,
   SIDEBAR_GENERATE_BUTTON_TEXT,
 } from '../utils';
-import tokens from '@contentful/f36-tokens';
 import { styles } from './Sidebar.styles';
 import { KeyValueMap } from 'contentful-management';
 import InformationWithLink from '../components/InformationWithLink';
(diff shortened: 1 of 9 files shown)


apps/braze/src/locations/Sidebar.tsx
```diff
@@ -15,18 +40,46 @@ const Sidebar = () => {
...
+        <InformationWithLink
+          url={CONNECTED_CONTENT_DOCUMENTATION}
+          linkText={'Learn more'}
+          fontColor="gray500"
+          marginTop="spacing2Xs"
+          marginBottom="spacingS">
+          Generate a Connected Content call to copy and paste into Braze.
+        </InformationWithLink>
+        <Button
+          color={tokens.gray500}
```
joaquincasal, line 55: Should we use the `variant` prop here? Also in the other button

FBanfi: Yes, that's right, done! ✅

joaquincasal (approved):

PR #9721, opened 2025-04-24
Braze app: change width of Dialog for last step [INTEG-2506]

joaquincasal:
## Purpose

Until now we were using `fullWidth` for all three steps of the app, but we needed to use the `large` width for the first two steps.

## Approach

Since the width of the Dialog cannot be changed once open, it has to be closed and reopened before the last step. So we had to serialize and deserialize all information from the entry and the fields.

## Testing steps

Automated tests were added.
To test manually, just use the app and get to the last step, the Dialog should close and reopen in the last step.

https://github.com/user-attachments/assets/7683965b-75a3-48f3-8daa-fb37b5255f5e

## Breaking Changes

N/A

N/A

## Deployment

 N/A

### DIFF
--- a/apps/braze/src/locations/Dialog.tsx
+++ b/apps/braze/src/locations/Dialog.tsx
@@ -10,10 +10,13 @@ import { FieldsFactory } from '../fields/FieldsFactory';
 import { Entry } from '../fields/Entry';
 import { Field } from '../fields/Field';
 
-export type EntryInfo = {
-  id: string;
-  contentTypeId: string;
-  title: string;
+export type InvocationParams = {
+  step: string;
+  entryId?: string;
+  contentTypeId?: string;
+  title?: string;
+  selectedLocales?: string[];
+  serializedEntry?: {};
 };
 
 const FIELDS_STEP = 'fields';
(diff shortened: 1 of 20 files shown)


apps/braze/src/locations/Dialog.tsx
```diff
@@ -23,31 +27,49 @@ const CODE_BLOCKS_STEP = 'codeBlocks';
...
+      ).createFields();
       fieldsRef.current = fields;
-      const entry = new Entry(
-        entryInfo.id,
-        entryInfo.contentTypeId,
-        entryInfo.title,
+      const entry = new Entry( // TODO: check if we can remove the !
+        invocationParams.entryId!,
+        invocationParams.contentTypeId!,
+        invocationParams.title!,
```
primeinteger, line 72: It seems like the `?` could be removed from these props in the type definition to make them required fields, and then the `!` wont be needed here.

primeinteger (approved):
This looks great! The inheritance pattern for different field types makes this nice and clean. Just one small comment since I noticed there's a TODO item

apps/braze/src/locations/Dialog.tsx
joaquincasal: Oh we totally missed that comment, good catch!
I just made those params required

PR #9724, opened 2025-04-26
feat: add klavio app [INTEG-2598]

david-shibley-contentful:


apps/klaviyo/frontend/README.md
primeinteger, line 1: Should the "frontend" folder be removed? It seems like an extra layer that isn't needed, and could be confusing since there are backend functions in this app.

apps/klaviyo/frontend/functions/checkStatus.ts
```diff
@@ -0,0 +1,50 @@
...
+} from '@contentful/node-apps-toolkit';
+interface CompleteOAuthParams {
+  code: string;
+  state: string;
+}
+
+export async function checkStatus(sdk: OAuthSDK): Promise<boolean> {
+  try {
+    const token = await sdk.token();
+    console.log('Token:', token);
```
primeinteger: Should any of the console.log statements be removed before the release?

apps/klaviyo/frontend/src/App.tsx
```diff
@@ -0,0 +1,63 @@
...
+      logger.error('Error loading mappings:', error);
+    }
+  }, [sdk.parameters.installation]);
+
+  // Get the component for the current location
+  const getCurrentComponent = () => {
+    // Find which location we're in and get the corresponding component
+    for (const [location, Component] of Object.entries(ComponentLocationSettings)) {
+      if (sdk.location.is(location)) {
+        const entry = (sdk as any).entry || {};
```
primeinteger: It doesn't look like this `entry` variable is used anywhere so it could be removed

apps/klaviyo/frontend/src/@types/index.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module '../utils/field-mappings';
+declare module './ErrorList';
```
primeinteger: What is the purpose of this file? I thought type definitions files are used to provide type hints for libraries that get used by javascript projects. I wonder if this can be removed.

primeinteger (approved):
This looks pretty good to me. Just a few comments about some minor details

apps/klaviyo/frontend/README.md
david-shibley-contentful: Bob asked me to add it for something he was doing, I will check with him if its still needed

apps/klaviyo/frontend/functions/checkStatus.ts
david-shibley-contentful: yeah that's a good call out, there shouldn't be console logs left around

apps/klaviyo/frontend/src/@types/index.d.ts
david-shibley-contentful: I bet it can be

PR #9730, opened 2025-04-29
Braze app: Add functions and action skeleton [INTEG-2590]

joaquincasal:
## Purpose

Adding functions and actions to the app

## Approach

- The functions were added by running `npx --no-install @contentful/app-scripts generate-function`
- That created the functions and the manifest file. It also updated the package.json file
- I created a new app definition and then uploaded a bundle by running `npm run build` and `npm run upload`
  - Here I had to update the `build` script in the package.json: I had to use the `outDir` param so the result ends up in the `build` directory instead of the `dist` directory (the `upload` script expects the build to be in the `build` folder)
- I did a quick test using `cma.appActionCall.createWithResponse` to test the function that gets called from an action and I also subscribed the app to some events to test the other function. Both functions worked correctly

Relevant documentation: https://www.contentful.com/developers/docs/extensibility/app-framework/working-with-functions/

### DIFF
--- a/apps/braze/contentful-app-manifest.json
+++ b/apps/braze/contentful-app-manifest.json
@@ -1,0 +1,36 @@
+{
+  "functions": [
+    {
+      "id": "createContentBlocksFunction",
+      "name": "Create content blocks function",
+      "description": "Function to create content blocks from App Action.",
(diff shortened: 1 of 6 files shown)


FBanfi (approved):
Nice update! 🏆

PR #9738, opened 2025-04-30
Braze app: Add first step for v2 wizard [INTEG-2537] 

FBanfi:
## Purpose

This update includes the first step of the braze app v2 wizard.

## Approach

A `MultiSelect` forma component is used to display all the fields of the entry excepts from `Asset`, `AssetArray`, `Reference` and `ReferenceArray` fields that are not supported in braze v2.

https://github.com/user-attachments/assets/642b758b-c4b1-481e-bfab-912b964b156c

**Important**: I couldn't find a way to make this work exactly like the Figma design.

Captura de pantalla 2025-04-30 a la(s) 3 12 40 p  m

I found that there is a `popoverProps` prop you could use in Multiselect that's called usePortal. 
> Boolean property to control whether or not to render the Popover in a React Portal. Rendering content inside a Portal allows the Popover to escape the bounds of its parent while still being positioned correctly.

But it doesn't seem to work either. For now, it's behaving like the second step of the Braze App v1, where the listMaxHeight is fixed to a specific size and the Box containing the step is expanded.

## Testing steps

Update in an existing automated test. You can manually test this by clicking the `create` button in the content block section of the sidebar.

## Breaking Changes

N/A

Lint to [INTEG-2537](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?selectedIssue=INTEG-2537) ticket.

## Deployment

N/A


apps/braze/package-lock.json
joaquincasal, line 1: We don't want this one gone 😅

apps/braze/src/components/create/CreateFlow.tsx
```diff
@@ -20,12 +20,23 @@ type CreateFlowProps = {
...
   const { sdk, entry, initialSelectedFields = [] } = props;
-
   const [step, setStep] = useState(FIELDS_STEP);
   const [selectedFields, setSelectedFields] = useState<Set<string>>(new Set(initialSelectedFields));
   const [contentBlockName, setContentBlockName] = useState('');
   const [isSubmitting, setIsSubmitting] = useState(false);
 
+  const allowedFields = entry.fields.filter((field) => field.isEnabledForCreate());
+
+  const filteredEntry = new Entry(
```
joaquincasal, line 30: If everything is the same except for the fields, we could just set the fields in the original entry
```suggestion
  entry.fields = entry.fields.filter((field) => field.isEnabledForCreate());
```

apps/braze/src/components/create/FieldsStep.tsx
```diff
@@ -15,37 +17,68 @@ const FieldsStep = ({
...
+        <Multiselect
+          currentSelection={currentSelection}
+          popoverProps={{ isFullWidth: true, listMaxHeight: 108 }}
+          placeholder="Select one or more">
+          <Multiselect.SelectAll onSelectItem={handleSelectAll} isChecked={allSelected} />
+          {allFields.map((field) => (
+            <Multiselect.Option
+              key={field.uniqueId()}
+              itemId={field.displayNameForCreate()}
+              value={field.displayNameForCreate()}
```
joaquincasal, line 72: I think the value should be the id of the field, so we can use it to identify the field. This way we probably won't need the `labelToId` function

apps/braze/src/components/create/FieldsStep.tsx
```diff
@@ -15,37 +17,68 @@ const FieldsStep = ({
...
-          <Option value="">Select a field</Option>
-          {entry.fields.map((field) => (
-            <Option key={field.uniqueId()} value={field.uniqueId()}>
-              {field.displayName()}
-            </Option>
+        <Multiselect
+          currentSelection={currentSelection}
+          popoverProps={{ isFullWidth: true, listMaxHeight: 108 }}
+          placeholder="Select one or more">
+          <Multiselect.SelectAll onSelectItem={handleSelectAll} isChecked={allSelected} />
```
joaquincasal, line 67: I would inline the allSelected boolean since it's not used in other places
```suggestion
          <Multiselect.SelectAll onSelectItem={handleSelectAll} isChecked={selectedFields.size === allFieldIds.length} />
```

joaquincasal (commented):
Overall looks good! I left some questions/comments. If we merge this without solving the popover issue, let's create a separate ticket for that

apps/braze/src/components/create/FieldsStep.tsx
FBanfi: Done! ✅

apps/braze/src/components/create/FieldsStep.tsx
FBanfi: Yes that's better! ✅

apps/braze/src/components/create/CreateFlow.tsx
FBanfi: I wasn't sure if we wanted the entry to be filtered in others steps, so I didn't want to change the original entry. But it looks like we don't need the rest of the fields then. ✅

apps/braze/package-lock.json
FBanfi: Oh yes absolutely 😆. I deleted it so I could generate it again and I ended up missing it! Thanks! ✅

apps/braze/src/components/create/CreateFlow.tsx
```diff
@@ -20,12 +20,13 @@ type CreateFlowProps = {
 
 const CreateFlow = (props: CreateFlowProps) => {
   const { sdk, entry, initialSelectedFields = [] } = props;
-
   const [step, setStep] = useState(FIELDS_STEP);
   const [selectedFields, setSelectedFields] = useState<Set<string>>(new Set(initialSelectedFields));
   const [contentBlockName, setContentBlockName] = useState('');
   const [isSubmitting, setIsSubmitting] = useState(false);
 
+  // entry.fields = entry.fields.filter((field) => field.isEnabledForCreate());
```
JuliRossi, line 28: 💭 Do we need this comment?

JuliRossi (approved):

PR #9749, opened 2025-04-30
[WIP] Braze app: create content blocks [INTEG-2541]

joaquincasal:
## Purpose

Implement content block creation

## Approach

Add implementation to existing function and call it from the frontend using an action.
For each connected field, this saves the id of the field and the newly created content block id in an installation parameters called "brazeConnectedFields".

## Testing steps

Configure the app with all required params and follow the 'create' flow. A content block should be created in Braze. Note that the name of the content block can't be repeated. Content blocks cannot be deleted but the name can be changed

## Breaking Changes

N/A

N/A

## Deployment

N/A

### DIFF
--- a/apps/braze/functions/createContentBlocks.ts
+++ b/apps/braze/functions/createContentBlocks.ts
@@ -5,21 +5,16 @@ import type {
   AppActionRequest,
 } from '@contentful/node-apps-toolkit';
 import { type PlainClientAPI, createClient } from 'contentful-management';
+import { documentToHtmlString } from '@contentful/rich-text-html-renderer';
 
-/**
- * App Action Function Template
- *
- * This template provides a starting point for creating an App Action function.
- * You'll need to implement your custom logic in the handler function below.
- */
+type AppInstallationParameters = {
+  brazeApiKey: string;
+  brazeEndpoint: string;
+};
 
-// TODO: Define your App Action parameters here
-// These parameters should match what you configured in your App Action definition
-type AppActionParameters = {
-  // Example:
-  // paramName: string;
-  // numberParam: number;
-  // booleanParam: boolean;
+export type AppActionParameters = {
+  entryId: string;
+  fieldIds: string;
 };
 
 function initContentfulManagementClient(context: FunctionEventContext): PlainClientAPI {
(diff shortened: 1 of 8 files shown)


apps/braze/functions/createContentBlocks.ts
```diff
@@ -37,38 +32,92 @@ function initContentfulManagementClient(context: FunctionEventContext): PlainCli
...
+  const cma = initContentfulManagementClient(context);
 
-  // Extract parameters from the event body
-  // const { paramName } = event.body;
+  const { entryId, fieldIds } = event.body;
+  const entry = await cma.entry.get({ entryId });
+  const contentType = await cma.contentType.get({
+    contentTypeId: entry.sys.contentType.sys.id,
+  });
+  const locale = entry.sys.locale || 'en-US'; // TODO: define what to do here
```
joaquincasal, line 46: We are discussing if users should be able to select locales or not. If not, we would use the default one

apps/braze/functions/createContentBlocks.ts
```diff
@@ -37,38 +32,92 @@ function initContentfulManagementClient(context: FunctionEventContext): PlainCli
...
+      }
+
+      const response = await fetch(`${brazeEndpoint}/content_blocks/create`, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${brazeApiKey}`,
+        },
+        body: JSON.stringify({
+          name: `${entryTitle}-${fieldId}`,
```
joaquincasal, line 89: entryTitle may have whitespaces, is that allowed in braze?

apps/braze/src/components/create/CreateFlow.tsx
```diff
@@ -26,17 +27,49 @@ const CreateFlow = (props: CreateFlowProps) => {
...
-      // TODO: Implement the actual creation of content blocks in Braze
-      // This would involve making an API call to Braze with the selected fields
-      setStep(SUCCESS_STEP);
-    } catch (error) {
-      console.error('Failed to create content block:', error);
-    } finally {
-      setIsSubmitting(false);
-    }
+
+    const connectedFields = JSON.parse(sdk.parameters.installation.brazeConnectedFields || '{}');
```
joaquincasal, line 44: The format of this param is:
```
{
'contentfulEntryId1': [['contentfulFieldId1', 'brazeContentBlockId1'], ['contentfulFieldId2', 'brazeContentBlockId2'], ...],
'contentfulEntryId2': ...
}
```

apps/braze/src/components/create/CreateFlow.tsx
```diff
@@ -26,17 +27,49 @@ const CreateFlow = (props: CreateFlowProps) => {
...
+    {
+      type: 'plain',
+      defaults: {
+        environmentId: sdk.ids.environment,
+        spaceId: sdk.ids.space,
+      },
+    }
+  );
+
   const handleCreate = async () => {
```
joaquincasal, line 41: Not sure if the definition of this handler needs to be here, maybe it could be in a more specific component

apps/braze/src/components/create/CreateFlow.tsx
```diff
@@ -26,17 +27,49 @@ const CreateFlow = (props: CreateFlowProps) => {
...
+      {
+        spaceId: sdk.ids.space,
+        environmentId: sdk.ids.environmentAlias ?? sdk.ids.environment,
+        appDefinitionId: '3xrrkR4QQr5YqMvOOoIt4y', //sdk.ids.app!,
+        appActionId: 'createContentBlocksAction',
+      },
+      {
+        parameters: {
+          entryId: entry.id,
+          fieldIds: Array.from(selectedFields).join(','),
```
joaquincasal, line 57: This is a string like this: 'fieldId1,fieldId2,...'. We could also stringify a json object if that helps in any way

apps/braze/functions/createContentBlocks.ts
JuliRossi: We are allowing users select locales, but I think we can handle localization in another ticket

apps/braze/functions/createContentBlocks.ts
JuliRossi: It isn't. We'll need to remove whitespaces

apps/braze/src/components/create/CreateFlow.tsx
JuliRossi: Let's leave it this way. Perhaps a good refactor for the future

apps/braze/functions/createContentBlocks.ts
JuliRossi: I think we are missing the case with no localization fields. I'll add it

FBanfi (approved):
It looks great! I liked how you handled testing and mocking the Braze API calls! 💯

apps/braze/src/components/create/CreateFlow.tsx
JuliRossi: Let's leave it this way. Perhaps a good refactor for the future

JuliRossi (approved):

PR #9750, opened 2025-05-02
Mux app: select video quality

R-Delfino95:
## Purpose

We are currently working on integrating the latest features provided by Mux into the plugin. This PR adds support for selecting the video quality level of an asset, which was not previously possible.
More information can be found [here](https://www.mux.com/docs/guides/use-video-quality-levels#introducing-video-quality-levels).

## Approach

Currently, the asset is uploaded as soon as it is loaded into the container. Adding options outside of this flow felt unintuitive, so we opted to introduce a modal that allows users to select options when uploading an asset to Mux.
We implemented a pre-upload modal to let users choose the video quality level. This approach also sets the groundwork for supporting additional options in the future, such as MP4 renditions or captions.

A new drag-and-drop component had to be created because Mux provides its own implementation, which doesn't allow modifying the upload behavior to show the modal beforehand. The design closely follows Mux's original style.

### Screenshots

image
image
image

## Testing steps

Create a content model with the Mux JSON Object. Then, create a new entry for that content type, and you’ll be able to test uploading assets while selecting the desired video quality.

Another case to test is by providing a URL of an existing video, for example [this one](http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4). In this case, the modal should appear to allow the user to select the desired quality.

If an Asset ID is provided instead, the modal **should not** appear, as this flow is intended for associating an existing Mux video.

## Breaking Changes

None.

More information can be found [here](https://www.mux.com/docs/guides/use-video-quality-levels#introducing-video-quality-levels).

## Deployment

None.

### DIFF
--- a/apps/mux/frontend/src/components/UploadArea/UploadArea.module.css
+++ b/apps/mux/frontend/src/components/UploadArea/UploadArea.module.css
@@ -1,0 +1,40 @@
+.dropzone {
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  padding: 2em;
+  min-height: 250px;
+  border: 2px dashed #aaa;
+  border-radius: 12px;
+  background: #fff;
+}
+
+.title {
+  font-size: 2.2rem;
+  font-weight: 500;
+  margin: 0;
+  text-align: center;
+}
+
+.or {
+  margin: 1.5em 0 1em 0;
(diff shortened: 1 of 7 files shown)


R-Delfino95:
Based on internal feedback, I updated the style of the video quality selection options.
Also updated the original message with a new screenshot reflecting the changes.

PR #9751, opened 2025-05-02
fix: update ai-image-generator error for insufficient DALL-E credits [INTEG-2609]

primeinteger:
## Purpose

For the AI Image Generator app, the OpenAI API returns a vague `400 BadRequestError` when the OpenAI account has insufficient DALL-E credits to generate an image. 

Since the AI Image Generator app exposes the OpenAI API error details directly to the user in an toast notification, the error details in this case were not very helpful and made it seem like our app is sending a bad request.

## Approach

I updated the AI Image Generator error handling to return a more descriptive error message in this scenario, so the user will have better details to go on.

Screenshot 2025-05-02 at 2 38 18 PM

### Side Quests

I fixed the link for "Learn more about how DALL-E credits work" in the App Config instructions as the page it was pointing to before no longer exists. 

When I opened this PR I noticed the apps-test workflow was failing with this error:
```
npm error code EBADENGINE
npm error engine Unsupported engine
npm error engine Not compatible with your version of node/npm: glob@11.0.2
npm error notsup Not compatible with your version of node/npm: glob@11.0.2
npm error notsup Required: {"node":"20 || >=22"}
npm error notsup Actual:   {"npm":"10.8.2","node":"v18.20.8"}
```

I changed the circleCI config to use node v22.15 instead ([currently the latest LTS version available](https://circleci.com/developer/images/image/cimg/node)) and updated the node engines config in the top-level package.json to allow node versions 16+ instead of only v16 and v18.


sarahlessner (approved):
Way to make it better all around outside of just the fix for better error handling

ryunsong-contentful (approved):

PR #9770, opened 2025-05-06
Braze app: show connected fields [INTEG-2551]

FBanfi:
## Purpose

This update adds the `Connected Content Block entries` section to the Sidebar.

## Approach

This section is only rendered when there are some existing connected content block entries. It shows all the fields connected to `that specific entry`. There is also a button that redirects the users to the page location in which they will be able to see `all` the connected content block entries in the future (detailed modeling of the page location in other ticket).
- if we have more than five fields in the entry, a scroll bar is added and the container fixed so it doesn't take up a lot of space.

With less than five fields:

https://github.com/user-attachments/assets/0a76b095-ff8a-4813-9282-b3ebbd130014

With more than five fields:

https://github.com/user-attachments/assets/e353143a-d2ba-41b7-b533-402c5b8b448d

## Testing steps

Two automated tests were added. To test it locally you could change the `connectedFields` in the `Sidebar.tsx` code, to the following:
```
const connectedFields = JSON.parse(sdk.parameters.installation.brazeConnectedFields ||
      '{"contentfulEntryId1": [["contentfulFieldId1", "brazeContentBlockId1"], ["contentfulFieldId2", "brazeContentBlockId2"]]}'
);
```
And then changing the `connectedFieldsForCurrentEntry` for:
```
const connectedFieldsForCurrentEntry: [string, string][] = connectedFields['contentfulEntryId1'] || [];
```

## Breaking Changes

N/A

Link to [INTEG-2551](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?selectedIssue=INTEG-2551) ticket.

## Deployment

N/A

[INTEG-2551]: https://contentful.atlassian.net/browse/INTEG-2551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/braze/src/locations/Page.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { PageAppSDK } from '@contentful/app-sdk';
+
+const Page = () => {
+  const sdk = useSDK<PageAppSDK>();
+  /*
+     To use the cma, inject it as follows.
+     If it is not needed, you can remove the next line.
+  */
+  // const cma = useCMA();
+
```
JuliRossi, line 13: 💭 I'd remove these comments

apps/braze/src/locations/Sidebar.styles.ts
```diff
@@ -1,10 +1,29 @@
...
+};
+
 export const styles = {
   subheading: css({
-    margin: 0,
+    ...baseSubheadingStyle,
+  }),
+  subheadingCard: css({
+    ...baseSubheadingStyle,
+    fontWeight: tokens.fontWeightMedium,
```
JuliRossi, line 16: ❤️ Love this!

apps/braze/src/locations/Sidebar.tsx
```diff
@@ -84,6 +90,35 @@ const Sidebar = () => {
...
+            <Card className={styles.card}>
+              <Subheading className={styles.subheadingCard}>
+                Connected Content Block entries
+              </Subheading>
+              <Splitter />
+              <Stack
+                flexDirection="column"
+                spacing="spacingXs"
+                alignItems="initial"
+                className={connectedFieldsForCurrentEntry.length > 5 ? styles.stack : undefined}>
```
JuliRossi, line 105: 💭 is the if here necessary if we are using overflowY auto?

JuliRossi (approved):
Leave a few non blocking comments

apps/braze/src/utils.ts
```diff
@@ -14,6 +14,7 @@ export const GENERATE_DIALOG_TITLE = 'Generate Braze Connected Content Call';
 export const CREATE_DIALOG_TITLE = 'Generate Braze Content Blocks';
 export const SIDEBAR_GENERATE_BUTTON_TEXT = 'Generate';
 export const SIDEBAR_CREATE_BUTTON_TEXT = 'Create';
+export const SIDEBAR_CONNECTED_ENTRIES_BUTTON_TEXT = 'View all connected entries';
```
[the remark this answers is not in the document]
FBanfi: I was keeping the same convention as the other buttons of the sidebar. I added there as a const in order to be able to reuse it in the component logic and also in the tests.

apps/braze/src/locations/Sidebar.tsx
FBanfi: Yes, you are right it isn't necessary! 😁 

The if is removed now ✅

apps/braze/src/locations/Page.tsx
FBanfi: Done ✅

PR #9799, opened 2025-05-07
Braze app: entry Draft state warning [INTEG-2559]

FBanfi:
## Purpose

This update adds a new modal to warn the user when the entry is in a “**Draft**” state.

## Approach

Some changes were made in the Entry class so we can get it's state. In addition, the `create flow` changed a bit so we can check the state of the entry between `create step` and `success step` and decide if we want to render the warning or not.

When the entry is in draft state:

https://github.com/user-attachments/assets/993f07eb-cdfc-44bf-9212-d5988c9cb421

When the entry is published:

https://github.com/user-attachments/assets/301a4f92-3e1f-4676-bcb8-776797ac1382

## Testing steps

A bunch of the existing automated tests for `Dialog.spec` and `FieldsFactory.spec` changed. On the other hand, if you wanted to check ir locally you could follow the videos steps.

## Breaking Changes

N/A

Link to [INTEG-2559](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?selectedIssue=INTEG-2559) ticket.

## Deployment

N/A

[INTEG-2559]: https://contentful.atlassian.net/browse/INTEG-2559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/braze/src/fields/Entry.ts
```diff
@@ -10,14 +10,18 @@ export class Entry {
...
     id: string,
     contentType: string,
     title: string = 'Untitled',
     fields: Field[],
     spaceId: string,
     environment: string,
-    contentfulToken: string
+    contentfulToken: string,
+    publishedAt?: string,
+    updatedAt?: string
```
JuliRossi, line 24: ⚔️  I think it makes more sense to store the state of the entry instead of this 2 values. That way we only calculate the state of the entry once in the constructor. Unless we are using the values of _publishedAt_ and _updatedAt_ for something else I'm missing

FBanfi: Yes that sounds better! 💯 
Done ✅

apps/braze/src/fields/Entry.ts
```diff
@@ -139,4 +145,16 @@ export class Entry {
   private selectedFields() {
     return this.fields.filter((field) => field.selected);
   }
+
+  public getEntryState = (publishedAt: string | undefined, updatedAt: string | undefined) => {
```
JuliRossi, line 149: 💭 This could be private now

apps/braze/test/locations/Dialog.spec.tsx
```diff
@@ -21,13 +21,53 @@ vi.mock('contentful-management', () => ({
...
+      'en-US': {
+        sys: {
+          id: 'referencedEntryId',
+          linkType: 'Entry',
+          type: 'Link',
+        },
+      },
+    },
+  },
+};
```
JuliRossi, line 53: 💭  you could use the createEntry() mock function for this

apps/braze/test/locations/Dialog.spec.tsx
```diff
@@ -21,13 +21,53 @@ vi.mock('contentful-management', () => ({
...
+  name: 'Example Content Type',
+  fields: [
+    {
+      id: 'fieldId',
+      name: 'Field Name',
+      type: 'Link',
+      linkType: 'Entry',
+    },
+  ],
+};
```
JuliRossi, line 69: 💭  you can use the createContentType() mock function for this

apps/braze/src/locations/Dialog.tsx
```diff
@@ -53,20 +53,23 @@ const Dialog = () => {
...
-      const fields = await new FieldsFactory(
+      const fieldsFactory = new FieldsFactory(
         invocationParams.entryId!,
         invocationParams.contentTypeId!,
         cma
-      ).createFields();
-      fieldsRef.current = fields;
+      );
+      const [cmaEntry, cmaContentType] = await fieldsFactory.getEntryAndContentType();
+      fieldsRef.current = await fieldsFactory.createFieldsForEntry(cmaEntry.fields, cmaContentType);
```
JuliRossi, line 62: 💭  Got an idea for this part. We can pair it tomorrow

apps/braze/src/fields/Entry.ts
JuliRossi: Great!

JuliRossi (approved):

apps/braze/src/locations/Dialog.tsx
JuliRossi: We'll revisit this idea in he future

apps/braze/test/locations/Dialog.spec.tsx
FBanfi: Done ✅

apps/braze/test/locations/Dialog.spec.tsx
FBanfi: Done ✅

apps/braze/src/fields/Entry.ts
FBanfi: Done ✅

apps/braze/src/locations/Dialog.tsx
FBanfi: After pairing:
The idea was to **only** get the `entry` to obtain the updateAt and publishAt, and have the contentType inside the createFieldsForEntry.

Now it is refactored that way ✅

apps/braze/src/components/create/CreateStep.tsx
```diff
@@ -19,8 +19,8 @@ import { editButton } from './CreateStep.styles';
 type CreateStepProps = {
   entry: Entry;
   selectedFields: Set<string>;
-  contentBlockName: string;
-  setContentBlockName: (name: string) => void;
+  contentBlockNames: Record<string, string>;
+  setContentBlockNames: any;
```
JuliRossi, line 23:
```suggestion
  setContentBlockNames: (names:Record<string, string>) => void
```

PR #9805, opened 2025-05-08
Braze app: Layout and card component [INTEG-2538]

JuliRossi:
## Purpose

- Card component for wizard step
- Refactor the way we calculated names so we can edit them. Now we do it on the FE

## Testing steps
- Added test to make sure we aren't breaking anything

https://github.com/user-attachments/assets/444fd9b2-fba3-44a5-b3f9-23331fec8ce8

## References
INTEG-2538

### DIFF
--- a/apps/braze/src/components/create/CreateStep.tsx
+++ b/apps/braze/src/components/create/CreateStep.tsx
@@ -8,6 +8,7 @@ import {
   IconButton,
   Flex,
   TextInput,
+  FormControl,
 } from '@contentful/f36-components';
 import { EditIcon } from '@contentful/f36-icons';
 import { Entry } from '../../fields/Entry';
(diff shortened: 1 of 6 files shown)


apps/braze/src/components/create/CreateStep.tsx
```diff
@@ -9,23 +23,91 @@ type CreateStepProps = {
...
+  handleNextStep,
+}: CreateStepProps) => {
+  const [contentBlockNames, setContentBlockNames] = useState<Record<string, string>>({});
+  const [editingField, setEditingField] = useState<string | null>(null);
+
+  useEffect(() => {
+    // Initialize content block names with defaults
+    const initialNames: Record<string, string> = {};
+    Array.from(selectedFields).forEach((fieldId) => {
+      initialNames[fieldId] = getDefaultContentBlockName(entry, fieldId);
```
FBanfi, line 48: 💭 Could we remove the `Array.from()`?

Perhaps we could use directly `selectedFields.forEach()`

apps/braze/src/components/create/CreateStep.tsx
```diff
@@ -9,23 +23,91 @@ type CreateStepProps = {
...
+                        autoFocus
+                      />
+                      <FormControl.HelpText>Name should be unique.</FormControl.HelpText>
+                    </FormControl>
+                  ) : (
+                    <>
+                      <Text>Name</Text>
+                      <Text fontWeight="fontWeightDemiBold">{contentBlockNames[fieldId]}</Text>
+                    </>
+                  )}
```
FBanfi, line 91: 💭 I was thinking that maybe we could extract this ternary options into two separated components to make it more readable. Or perhaps just delegating all this in one hole separated component. But I'm not sure if it's worth it.

FBanfi: ⚔️ Checking the figma design, I noticed that the label should was slightly different:
It should be named `Content Block name`. 

💭 And the pencil in the design is white, but I can't find a white one in forma.

FBanfi (commented):
Awesome and nice tests! 💯. I only added one blocking comment related to the design.

FBanfi (approved):

apps/braze/src/components/create/CreateStep.tsx
JuliRossi: I'll ask Erin about the names and address all of these issues on my next ticket

apps/braze/src/components/create/CreateStep.tsx
JuliRossi: Great catch!

PR #9827, opened 2025-05-09
Braze config screen fix: [INTEG-2643]

JuliRossi:
## Purpose

- Fix a bug where a user attemps to config braze with a environment different to master in braze.


JuliRossi:
Fixed on https://github.com/contentful/apps/pull/9828

PR #9828, opened 2025-05-12
 Braze config screen fix: [INTEG-2643]

joaquincasal:
## Purpose

Fix a bug where a user attemps to install the braze app in an environment different to master.

## Approach

When we install the app, we make a request to contentful to check that the API key is valid, the previous call wasn't using the environment in which the app is being installed. Now it does to make sure that the API key has permissions on that environment.

## Testing steps

Install/Configure the app in an environment that isn't master with an API key that has permissions on the environment.

https://github.com/user-attachments/assets/869e7c14-7357-40f8-8f6d-7a7f0f1deb66


JuliRossi (approved):
👍

sam-is-content (approved):
looks good, thanks for getting this done so quickly!

PR #9829, opened 2025-05-12
Braze app: Error modal for 4XX errors [INTEG-2544]

FBanfi:
## Purpose

This update was needed to display the user the 4xx class errors in the creation of the content blocks.

## Approach

The following changes were introduced:
- We added status codes in the app function to identify the types of the errors.
- We added a new modal step that is showed when the creation of the content blocks fails due to a 4xx class error.
- Testing to validate this changes.

Here is an example of the result when the creation fails and when it is successfull:

https://github.com/user-attachments/assets/bbeb186c-96da-42c2-b697-b16ff346b3df

## Testing steps

- `createContentBlocks.ts` function tests were updated and one more automated test was added into `Dialog.spec.tsx`

## Breaking Changes

N/A

Link to [INTEG-2544](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?selectedIssue=INTEG-2544) ticket.

## Deployment

N/A

[INTEG-2544]: https://contentful.atlassian.net/browse/INTEG-2544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/braze/test/fields/FieldsFactory.spec.ts
+++ b/apps/braze/test/fields/FieldsFactory.spec.ts
@@ -311,7 +311,7 @@ describe('FieldsFactory', () => {
     expect(fieldInstance.items[1].fields[0].id).toBe('name');
   });
 
-  it('should limit the recursion depth for nested references based on NESTED_DEPTH constant', async () => {
+  it('qences based on NESTED_DEPTH constant', async () => {
     const max_depth = 5;
 
     // Create mock entries with deep nesting structure
--- a/apps/braze/test/fields/FieldsFactory.spec.ts
+++ b/apps/braze/test/fields/FieldsFactory.spec.ts
@@ -311,7 +311,7 @@ describe('FieldsFactory', () => {
     expect(fieldInstance.items[1].fields[0].id).toBe('name');
   });
 
-  it('qences based on NESTED_DEPTH constant', async () => {
+  it('should limit the recursion depth for nested references based on NESTED_DEPTH constant', async () => {
     const max_depth = 5;
 
     // Create mock entries with deep nesting structure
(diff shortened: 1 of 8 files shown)


apps/braze/test/fields/FieldsFactory.spec.ts
```diff
@@ -311,7 +311,7 @@ describe('FieldsFactory', () => {
     expect(fieldInstance.items[1].fields[0].id).toBe('name');
   });
 
-  it('should limit the recursion depth for nested references based on NESTED_DEPTH constant', async () => {
+  it('qences based on NESTED_DEPTH constant', async () => {
```
JuliRossi, line 314: 💭 Nitpick: I think we shouldn't change this

apps/braze/test/functions/createContentBlocks.test.ts
```diff
@@ -189,8 +192,12 @@ describe('createContentBlocks', () => {
     vi.mocked(mockCma.contentType.get).mockResolvedValue(contentType);
     vi.mocked(global.fetch).mockResolvedValue({
       ok: false,
-      statusText: 'Unauthorized',
-    } as Response);
+      status: 500,
+      json: () =>
+        Promise.resolve({
+          message: 'Unauthorized',
+        }),
```
JuliRossi, line 199: 💭  Nitpick: 5XX error, so it should be something like "Internal server error"

JuliRossi (approved):
Incredible work!

apps/braze/test/fields/FieldsFactory.spec.ts
FBanfi: Oh, I don't know why this was changed! Thanks 💯

apps/braze/test/functions/createContentBlocks.test.ts
FBanfi: Yes, I missed that one! I left the unauthorized but changed the status code for a 401, because the initial intend for that test was to check that scenario. 
Done ✅

PR #9834, opened 2025-05-14
Braze app: Edit card component [INTEG-2539]

JuliRossi:
## Purpose

We want users to be allowed to edit name and description of a contentblock when creating a new one
## Approach
The following changes were introduced:
- Introduce description to card component and to create content block action
- Add validations for name
- Enhance UI for editing name and description
- Updated package.json to better test in the dev environment

## Testing steps

Updated automatic testing.
Tested using my own app:

https://github.com/user-attachments/assets/88112a47-466c-4a4a-933e-ba2ddc545cae

### DIFF
--- a/apps/braze/src/components/create/DraftStep.tsx
+++ b/apps/braze/src/components/create/DraftStep.tsx
@@ -4,9 +4,9 @@ import { ContentBlockState } from './CreateFlow';
 
 type DraftStepProps = {
   isSubmitting: boolean;
-  contentBlockStates: Record<string, ContentBlockState>;
+  contentBlockStates: ContentBlockState;
   handlePreviousStep: () => void;
-  handleNextStep: (data: Record<string, ContentBlockState>) => void;
+  handleNextStep: (data: ContentBlockState) => void;
 };
 
 const DraftStep = ({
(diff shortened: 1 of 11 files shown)


apps/braze/src/components/create/DraftStep.tsx
```diff
@@ -1,23 +1,24 @@
...
+  contentBlockStates,
   handlePreviousStep,
   handleNextStep,
-}: CreateStepProps) => {
+}: DraftStepProps) => {
   return (
     <>
       <Subheading fontWeight="fontWeightDemiBold" fontSize="fontSizeXl" lineHeight="lineHeightL">
-        This entry is in a “Draft” state.
+        This entry is in a "Draft" state.
```
FBanfi, line 21: 💭 Nitpick: I'm not a big fan of the `contentBlockStates` useState name. It makes me think it's holding status-related info (like draft, published, etc.), but it's actually storing names and descriptions.
In the `createContentBlocks` function, this data is referred to as `ContentBlockData`, which I think is a much better name!

FBanfi (approved):
Amazing work! It seems really cool! 💯

apps/braze/src/components/create/DraftStep.tsx
JuliRossi: Agree, not a fan of that name either. I'll use ContentBlockData or simply contentBlock for this. Thanks!

PR #9838, opened 2025-05-15
Braze app : Update config screen edge case [INTEG-2610]

JuliRossi:
## Purpose

Handle edge case in which we don't have available the installation parameters. This can happen if the user was already using the v1 version of the app.

## Approach
In this PR:
- Added warning note
- Added link to config page 
- Disable buttons of the app, until the user re-configures the page
## Testing steps
Added test to handle edge cases. Tested using the v1

https://github.com/user-attachments/assets/0b33892a-a7ba-47a2-a267-8af24bb184a5


joaquincasal (approved):
🚀

PR #9839, opened 2025-05-15
Braze app: Adjusting create button[]

JuliRossi:
## Purpose

Change it to use the animation from forma

## Approach

- Added the isLoading property
- Refactor to use the same button in Draft and create

## Testing steps

- updated tests

With an entry published:
https://github.com/user-attachments/assets/89b78bbe-6aa1-4b10-ad1d-7307756b5a8b
With entry in draft state:

https://github.com/user-attachments/assets/f43fe7a0-12f2-4681-a875-c82fb88bbd13

### DIFF
--- a/apps/braze/src/components/create/CreateButton.tsx
+++ b/apps/braze/src/components/create/CreateButton.tsx
@@ -1,0 +1,33 @@
+import { Button } from '@contentful/f36-components';
+
+type CreateButtonProps = {
+  isLoading: boolean;
+  onClick: () => void;
+  isDisabled?: boolean;
+  variant?: 'primary' | 'secondary';
+  size?: 'small' | 'medium' | 'large';
+  dataTestId?: string;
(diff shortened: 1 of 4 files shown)


apps/braze/src/components/create/CreateButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+  isDisabled?: boolean;
+  size?: 'small' | 'medium' | 'large';
+};
+
+const CreateButton = ({
+  isLoading,
+  onClick,
+  isDisabled = false,
+  size = 'small',
+  dataTestId,
```
joaquincasal, line 15:
```suggestion
```

joaquincasal (commented):
Looks like tests are failing because of a non-existing prop, other than that looks good"

joaquincasal (approved):

PR #9842, opened 2025-05-19
Braze: Allow to retry when name already in use [INTEG-2540]

joaquincasal:
## Purpose

If the chosen content block name for a field is already in use in Braze, the creation will fail. We now allow to edit the name and retry.

## Approach

Instead of going directly into the ErrorStep, we first check if any of the errors were caused by the name of the content block already existing in Braze. If that's the case, we return to the CreateStep showing the errors and allowing to edit the names for the content blocks.
The main change in the implementation is that now the `creationResultFields` **accumulates** all results instead of having only the results for the last call. That way we can now the status of all the selected fields.

If there are errors but none of them are from the name already existing, we do display the ErrorStep as we used to do.

## Testing steps

Try to create a content block using a name that already exists in Braze.

> [!WARNING]
> Some visual adjustements were donde after the recording of the video, see screenshots below for the final versions 

https://github.com/user-attachments/assets/b126ff73-a152-49e3-b794-61b762d9ad00

image
image

### DIFF
--- a/apps/braze/src/components/create/CreateStep.tsx
+++ b/apps/braze/src/components/create/CreateStep.tsx
@@ -16,10 +16,12 @@ import { Entry } from '../../fields/Entry';
 import WizardFooter from '../WizardFooter';
 import { useState, useEffect, Dispatch, SetStateAction } from 'react';
 import { editButton } from './CreateStep.styles';
-import { PencilSimple } from '@phosphor-icons/react';
+import { PencilSimple, Check } from '@phosphor-icons/react';
 import tokens from '@contentful/f36-tokens';
 import CreateButton from './CreateButton';
+import { CreationResultField } from './CreateFlow';
 
+const MAX_NAME_LENGTH = 100;
 const MAX_DESCRIPTION_LENGTH = 250;
 
 // Types
@@ -43,6 +45,8 @@ type ContentBlockViewProps = {
   name: string;
   description: string;
   onEdit: () => void;
+  isDisabled?: boolean;
+  error?: string;
 };
 
 type ContentBlockCardProps = {
@@ -57,6 +61,8 @@ type ContentBlockCardProps = {
   onCancel: () => void;
   onSave: (fieldId: string) => void;
   isNameValid: boolean;
+  isDisabled?: boolean;
+  error?: string;
 };
 
 type CreateStepProps = {
@@ -67,6 +73,7 @@ type CreateStepProps = {
   isSubmitting: boolean;
   handlePreviousStep: () => void;
   handleNextStep: (contentBlocksData: ContentBlockData) => void;
+  creationResultFields: CreationResultField[];
 };
 
 // Utils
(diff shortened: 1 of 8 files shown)


FBanfi (approved):
Really nice update! 💯

apps/braze/src/components/create/CreateStep.tsx
```diff
@@ -161,16 +183,28 @@ const ContentBlockView = ({ name, description, onEdit }: ContentBlockViewProps)
...
-        onClick={onEdit}
-        className={editButton}
-        withTooltip
-        tooltipProps={{ content: 'Edit name or add a description' }}
-      />
+      {isDisabled ? (
+        <IconButton
+          size="small"
+          variant="positive"
+          icon={<Check size={16} />}
```
JuliRossi, line 190: 💭  I think we should use this one instead
```suggestion
          icon={<CheckCircle size={16} />}
```

apps/braze/src/components/create/CreateStep.tsx
```diff
@@ -16,10 +16,12 @@ import { Entry } from '../../fields/Entry';
 import WizardFooter from '../WizardFooter';
 import { useState, useEffect, Dispatch, SetStateAction } from 'react';
 import { editButton } from './CreateStep.styles';
-import { PencilSimple } from '@phosphor-icons/react';
+import { PencilSimple, Check } from '@phosphor-icons/react';
```
JuliRossi, line 19:
```suggestion
import { PencilSimple, CheckCircle } from '@phosphor-icons/react';
```

JuliRossi (commented):
Left a few comments, as I realised there were some differences with the figma

apps/braze/src/components/create/CreateStep.tsx
```diff
@@ -153,6 +170,11 @@ const ContentBlockView = ({ name, description, onEdit }: ContentBlockViewProps)
         <Stack spacing="spacingXs" flexDirection="column" alignItems="flex-start">
           <Text>Content Block name</Text>
           <Text fontWeight="fontWeightDemiBold">{name}</Text>
+          {error && (
+            <Text fontColor="red600" fontSize="fontSizeS">
+              {error}
+            </Text>
+          )}
```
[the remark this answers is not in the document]
joaquincasal: Thanks!
Just updated this, this is how it looks now

Screenshot 2025-05-19 at 13 48 09

apps/braze/src/components/create/CreateStep.tsx
```diff
@@ -328,6 +362,10 @@ const CreateStep = ({
               onCancel={handleCancel}
               onSave={handleSave}
               isNameValid={isNameValid}
+              isCreated={creationResultFields.some(
+                (result) => result.fieldId === fieldId && result.success
+              )}
+              error={creationResultFields.find((result) => result.fieldId === fieldId)?.message}
```
JuliRossi, line 368: 💭  Perhaps the createFlow should handle this 2 states. But unsure if it makes a huge difference

JuliRossi (approved):
Thanks for revising this with Erin! It looks great!

apps/braze/src/components/create/CreateStep.tsx
joaquincasal: Oh I actually tried to have everything in the `creationResultFields` so child components can derive the data they need from that themselves. Does that sound reasonable?

JuliRossi: Yes! Sorry I didn't pick up on that.

PR #9843, opened 2025-05-19
Braze app: Adding dropdown to config screen for Braze Endpoint [INTEG-2655]

JuliRossi:
## Purpose

Add dropdown to braze endpoint to avoid user error in config screen

## Approach

- Added Autocomplete component to Config screen
- Added list of braze endpoints
- Refactor tests to use the endpoint

## Testing steps
When an app is installed and has to update

https://github.com/user-attachments/assets/8d22cd7e-d49b-40ce-83ed-5915ce1c4271

When app is not installed

https://github.com/user-attachments/assets/a02480f8-a804-426d-a093-1ced826f8fec

When app is missing every config parameter

https://github.com/user-attachments/assets/ad1622ec-a456-41c3-b9f2-cc086419ead0

### DIFF
--- a/apps/braze/src/locations/ConfigScreen.tsx
+++ b/apps/braze/src/locations/ConfigScreen.tsx
@@ -94,7 +94,7 @@ const ConfigScreen = () => {
 
     const isContentfulKeyValid = await checkContentfulApiKey(parameters.contentfulApiKey);
     const isBrazeKeyValid = checkIfHasValue(parameters.brazeApiKey, setBrazeApiKeyIsValid);
-    console.log('parameters.brazeEndpoint', parameters.brazeEndpoint);
+
     const isBrazeEndpointValid = checkIfHasValue(parameters.brazeEndpoint, setBrazeEndpointIsValid);
 
     if (!isContentfulKeyValid || !isBrazeKeyValid || !isBrazeEndpointValid) {
@@ -123,7 +123,6 @@ const ConfigScreen = () => {
   useEffect(() => {
     (async () => {
       const currentParameters: AppInstallationParameters | null = await sdk.app.getParameters();
-      console.log('currentParameters', currentParameters);
 
       if (currentParameters) {
         setParameters(currentParameters);
--- a/apps/braze/src/locations/ConfigScreen.tsx
+++ b/apps/braze/src/locations/ConfigScreen.tsx
@@ -8,6 +8,7 @@ import {
   TextInput,
   Text,
   Subheading,
+  Autocomplete,
 } from '@contentful/f36-components';
 import { useSDK } from '@contentful/react-apps-toolkit';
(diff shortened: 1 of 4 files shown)


apps/braze/src/locations/ConfigScreen.tsx
```diff
@@ -245,12 +249,50 @@ function ContentTypeSection() {
...
+
+  // Initialize selectedBrazeEndpoint when parameters change
+  useEffect(() => {
+    if (parameters.brazeEndpoint) {
+      const matchingEndpoint = BRAZE_ENDPOINTS.find((end) => end.url === parameters.brazeEndpoint);
+      if (matchingEndpoint) {
+        setSelectedBrazeEndpoint(matchingEndpoint);
+      }
+    }
+  }, [parameters.brazeEndpoint]);
```
joaquincasal, line 280: Do we need this useEffect? 🤔
I think we can have this logic and pass the result to the default value in the useState

apps/braze/src/locations/ConfigScreen.tsx
```diff
@@ -245,12 +249,50 @@ function ContentTypeSection() {
...
+
+  // We filter the "spaces" array by the inputValue
+  // we use 'toLowerCase()' to make the search case insensitive
+  const handleInputValueChange = (nameValue: string) => {
+    const newFilteredItems = BRAZE_ENDPOINTS.filter(({ name }) =>
+      name.toLowerCase().includes(nameValue.toLowerCase())
+    );
+    setFilteredItems(newFilteredItems);
+  };
```
joaquincasal, line 289: I'm not sure I follow this logic, what's this for?

joaquincasal (commented):
Left some questions!

apps/braze/src/locations/ConfigScreen.tsx
JuliRossi: Autocomplete works as a Search bar When filtering results that match, you want to make sure it is case insensitive (thus the lowerCasing).

apps/braze/src/locations/ConfigScreen.tsx
JuliRossi: I'll try to do it that way. But I had issues initializing the selectedEndpoint as parameters.brazeEndpoint wasn't available.

apps/braze/src/locations/ConfigScreen.tsx
JuliRossi: Exactly. Unfortunately, it is a required prop.  https://f36.contentful.com/components/autocomplete

JuliRossi: Something we can so is use a Menu instead of an autocomplete. Let me ask Erin about it

PR #9853, opened 2025-05-21
feat: update jira config page to be more functional [INTEG-2672]

ryunsong-contentful:
## Purpose
**Customer Issue Video Link:** https://share.birdie.so/GBJHPKL3
**What I see in production with the live app now**: 
Screenshot 2025-05-21 at 12 43 47 PM
https://github.com/user-attachments/assets/d60b03fb-a149-49d7-97f0-a95e9290af24

**FIXED DEMO VID**
https://github.com/user-attachments/assets/0276d689-71ad-441e-89aa-7d49c1c1c508

### DIFF
--- a/apps/jira/jira-app/src/components/App/Steps/ContentTypeStep.tsx
+++ b/apps/jira/jira-app/src/components/App/Steps/ContentTypeStep.tsx
@@ -61,9 +61,9 @@ const ContentTypeStep = ({
     );
   } else {
     contentToRender = (
-      <Flex>
+      <Flex flexDirection="column" gap="spacingM">
         {sortedContentTypes.map((ct) => (
-          <FormControl id={ct.name} key={ct.id}>
+          <FormControl id={ct.name} key={ct.id} style={{ marginBottom: 0 }}>
             <Checkbox
               testId={`ct-item-${ct.id}`}
               name={ct.name}
--- a/apps/jira/jira-app/src/components/App/Steps/ContentTypeStep.tsx
+++ b/apps/jira/jira-app/src/components/App/Steps/ContentTypeStep.tsx
@@ -63,7 +63,7 @@ const ContentTypeStep = ({
     contentToRender = (
       <Flex flexDirection="column" gap="spacingM">
         {sortedContentTypes.map((ct) => (
-          <FormControl id={ct.name} key={ct.id} style={{ marginBottom: 0 }}>
+          <FormControl id={ct.name} key={ct.id} marginBottom="none">
             <Checkbox
               testId={`ct-item-${ct.id}`}
               name={ct.name}
--- a/apps/jira/jira-app/src/components/App/Steps/InstanceStep.tsx
+++ b/apps/jira/jira-app/src/components/App/Steps/InstanceStep.tsx
@@ -1,18 +1,6 @@
 import React from 'react';
-import { Heading, Paragraph, Select, Option, TextInput, Card } from '@contentful/f36-components';
+import { Heading, Paragraph, Select, Option, Autocomplete } from '@contentful/f36-components';
 import { JiraCloudResource, CloudProject } from '../../../interfaces';
-
-// using lodash.debouce basically breaks test with infinite timers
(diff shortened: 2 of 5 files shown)


apps/jira/jira-app/src/index.scss
```diff
@@ -234,3 +234,20 @@ header img {
...
+  width: 100% !important;
+  min-width: 320px !important;
+  max-width: 400px !important;
+}
+.project-autocomplete [role='listbox'] {
+  width: 100% !important;
+  min-width: 320px !important;
+  max-width: 400px !important;
+  max-height: 600px !important;
+}
```
primeinteger, line 253: It's a bit heavy handed to add all these rules with `!important`. With scss we can combine selectors and use nesting to try and remove some of the duplication, for example below. Did you already try something like this?

```scss
.project-autocomplete {
  &, input, [role='listbox'] {
    width: 100%;
    min-width: 320px;
    max-width: 400px;
  }
  [role='listbox'] {
    max-height: 600px;
  }
}
```

apps/jira/jira-app/src/components/App/Steps/InstanceStep.tsx
```diff
@@ -61,51 +38,52 @@ export default class InstanceStep extends React.Component<Props, State> {
...
-              width="full"
-              placeholder={selectedProject ? selectedProject.name : 'Search for a project'}
-              value={this.state.inputValue}
-              onChange={(ev) => {
-                ev.persist();
-                this.handleInputChange(ev);
-              }}
-              onFocus={() => {
-                this.setState({ inputValue: '' });
+            <Autocomplete<CloudProject>
```
primeinteger, line 57: Nice update to add Autocomplete. I thought this `<CloudProject>` looked out of place at first, but learned from the docs that's how we can set the type for items 🔥

apps/jira/jira-app/src/components/App/Steps/ContentTypeStep.tsx
```diff
@@ -61,9 +61,9 @@ const ContentTypeStep = ({
     );
   } else {
     contentToRender = (
-      <Flex>
+      <Flex flexDirection="column" gap="spacingM">
         {sortedContentTypes.map((ct) => (
-          <FormControl id={ct.name} key={ct.id}>
+          <FormControl id={ct.name} key={ct.id} style={{ marginBottom: 0 }}>
```
primeinteger, line 66: We could use the marginBottom prop here to avoid adding inline styles with the style prop.  I think using the style prop would generally be a last ditch option.

```suggestion
          <FormControl id={ct.name} key={ct.id} marginBottom="none">
```

apps/jira/jira-app/src/components/App/Steps/InstanceStep.tsx
```diff
@@ -61,51 +38,52 @@ export default class InstanceStep extends React.Component<Props, State> {
...
-            isDisabled={resources.length === 1}
-            value={selectedResource || ''}>
-            <Option value="">Select a site</Option>
-            {resources.map((r) => (
-              <Option key={r.id} value={r.id}>
-                {r.url.replace('https://', '')}
-              </Option>
-            ))}
-          </Select>
+          <div style={{ display: 'flex', gap: 16, alignItems: 'center', width: '100%' }}>
```
primeinteger, line 41: I see lots of new inline styles being added with the style prop. I think to keep consistency with the existing app code it could be better to add `className` here props instead, and then add styling based on classes to the SCSS file instead.

primeinteger (approved):
The result of the updates looks so much better than before when there were issues with the dropdown and broken styling on overlapping elements!

I have a few nit picks with how the styling was implemented though.  I'll approve either way because I think those are minor details but I think there are some opportunities for improvement if you think it's worth it.

apps/jira/jira-app/src/index.scss
ryunsong-contentful: Nice suggestion. Clean af

primeinteger (approved):
🫶

PR #9854, opened 2025-05-21
Braze: Update content blocks through app events [INTEG-2591]

joaquincasal:
## Purpose

Add function for responding to app events:

- If a field gets updated, update the corresponding content block
- If an entry gets deleted, remove the entry from the brazeConfig entry
- If the app gets uninstalled, remove the custom content type and entry

## Approach

Branch the code using the header sent by the App Event.
A retry mechanism has been included in case the request fails.

## Testing steps

1. Create a content block
2. Update the field in Contentful
3. Check the content block has been updated

### DIFF
--- a/apps/braze/contentful-app-manifest.json
+++ b/apps/braze/contentful-app-manifest.json
@@ -10,11 +10,11 @@
       "accepts": ["appaction.call"]
     },
     {
-      "id": "updateContentBlocksFunction",
-      "name": "Update content blocks function",
-      "description": "Function to update content blocks from App Events.",
-      "path": "functions/updateContentBlocks.js",
-      "entryFile": "functions/updateContentBlocks.ts",
+      "id": "appEventHandler",
+      "name": "App event handler function",
+      "description": "Function to handle App Events.",
+      "path": "functions/appEventHandler.js",
+      "entryFile": "functions/appEventHandler.ts",
       "allowNetworks": ["https://*.braze.com"],
       "accepts": ["appevent.handler"]
     }
--- a/apps/braze/functions/appEventHandler.ts
+++ b/apps/braze/functions/appEventHandler.ts
@@ -12,6 +12,7 @@ import {
   updateConfig,
   ConnectedFields,
   AppInstallationParameters,
+  EntryConnectedFields,
 } from '../src/utils';
 import { EntryProps, KeyValueMap, PlainClientAPI } from 'contentful-management';
 import { documentToHtmlString } from '@contentful/rich-text-html-renderer';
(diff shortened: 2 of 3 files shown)


apps/braze/src/utils.ts
```diff
@@ -61,15 +61,15 @@ export function removeHypens(str: string) {
...
 export async function updateConfig(
   configEntry: EntryProps<KeyValueMap>,
-  locale: string,
   connectedFields: EntryConnectedFields,
   cma: PlainClientAPI
 ) {
   if (!configEntry.fields[CONFIG_FIELD_ID]) {
     configEntry.fields[CONFIG_FIELD_ID] = {};
   }
+  const locale = 'en-US'; // TODO: get default locale
```
JuliRossi, line 70: ❓  Why are we not including in the signature the locale?

apps/braze/functions/updateContentBlocks.ts
```diff
@@ -4,33 +4,143 @@ import {
...
+  }
+
+  const body = event.body as any;
+  const entryId = body.sys.id;
+
+  const configEntry: EntryProps<KeyValueMap> = await getConfigEntry(cma);
+  const configField = configEntry.fields[CONFIG_FIELD_ID];
+  if (!configField) {
+    return;
+  }
```
JuliRossi, line 64: 💭 I'd add a console.error("Not able to find {CONFIG_FIELD_ID}") here so if it fails, we can see it in the function's logs

apps/braze/functions/updateContentBlocks.ts
```diff
@@ -4,33 +4,143 @@ import {
...
+  const connectedFields = Object.values(configField)[0] as {
+    [entryId: string]: EntryConnectedFields;
+  };
+  const entryConnectedFields = connectedFields[entryId];
+  if (!entryConnectedFields) {
+    return;
+  }
```
JuliRossi, line 71: 💭  Same here

apps/braze/src/utils.ts
```diff
@@ -61,15 +61,15 @@ export function removeHypens(str: string) {
 
 export async function updateConfig(
```
JuliRossi, line 62: 💭 Nitpick: I think it could make sense to move this to updateContentBlock, even if the dialog is using it

apps/braze/functions/updateContentBlocks.ts
```diff
@@ -4,33 +4,143 @@ import {
...
+      contentTypeId: body.sys.contentType.sys.id,
+    });
+
+    for (const fieldId of Object.keys(body.fields)) {
+      const field = entryConnectedFields.find(
+        (connectedField) => connectedField.fieldId === fieldId
+      );
+      if (!field) {
+        continue;
+      }
```
JuliRossi, line 93: 💭  Same as other errors here

apps/braze/functions/updateContentBlocks.ts
JuliRossi, line 1: If we are going to handle all events including uninstalling in the same handler, perhaps a better name would be eventHandler.

apps/braze/src/utils.ts
joaquincasal: Because the locale will always be the same since this entry is managed by us. We would always use the default locale. And if it's going to be always the same, I think it's better to solve it inside of the function instead of solve it every time we call it

apps/braze/functions/updateContentBlocks.ts
joaquincasal: Hmm but this one is not an error, we're just filtering events that do not apply. Because we'll receive events for all updated entries within the environment in which the app is installed in, but we're only interested in the ones that are part of the config

apps/braze/functions/updateContentBlocks.ts
joaquincasal: Same as above, the entry may have some fields that are not connected and we want to ignore those, but it's not an error

apps/braze/functions/updateContentBlocks.ts
```diff
@@ -4,33 +4,143 @@ import {
...
+} from '../src/utils';
+import { createClient, EntryProps, KeyValueMap, PlainClientAPI } from 'contentful-management';
+import { documentToHtmlString } from '@contentful/rich-text-html-renderer';
+
+const WAIT_TIMES = [0, 5000, 10000];
+
+type EntryConnectedFields = {
+  fieldId: string;
+  contentBlockId: string;
+}[];
```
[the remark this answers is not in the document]
joaquincasal: Totally. This is addressed in the second commit of this PR

apps/braze/functions/updateContentBlocks.ts
```diff
@@ -4,33 +4,143 @@ import {
...
+
+type AppInstallationParameters = {
+  brazeApiKey: string;
+  brazeEndpoint: string;
+};
+
+function initContentfulManagementClient(context: FunctionEventContext): PlainClientAPI {
```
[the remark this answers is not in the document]
joaquincasal: Totally. This is addressed in the third commit of this PR

JuliRossi (approved):
👍

PR #9858, opened 2025-05-22
Braze: Page view connected entries table [INTEG-2547] 

FBanfi:
## Purpose

This updates includes the creation of the table which shows the connected entries in the Page location.

## Approach

We added the new structure in the Page.tsx location and a fetchBrazeConnectedEntries.ts file which contains the logic to get the connected entries from the config entry of the user.

if the user has connected entries, it should look like the following video (taking into account that it's only possible to have 25 connected entries):

https://github.com/user-attachments/assets/2305992c-5a93-44e8-ab14-ce79c26f87fc

if the user doesn't have connected entries, the following message will be displayed:

Captura de pantalla 2025-05-22 a la(s) 2 59 42 p  m

If there is an error, then: 

Captura de pantalla 2025-05-22 a la(s) 3 21 28 p  m

## Testing steps

- An automated test was added.
- You can test it manually by creating a new content block and going to the page location, there, a connected entry should be displayed.

## Breaking Changes

N/A

Link to [INTEG-2547](https://contentful.atlassian.net/browse/INTEG-2547) ticket.

## Deployment

N/A

### DIFF
--- a/apps/braze/src/locations/Page.tsx
+++ b/apps/braze/src/locations/Page.tsx
@@ -30,7 +30,10 @@ const getStatusBadge = (status: string) => {
   return <Badge variant="warning">Draft</Badge>;
 };
 
-const getUpdatedLabel = (dateString: string) => {
+const getUpdatedLabel = (dateString: string | undefined) => {
+  if (!dateString) {
+    return '-';
+  }
   const date = new Date(dateString);
   const now = new Date();
   const diffMs = now.getTime() - date.getTime();
@@ -69,7 +72,12 @@ const Page = () => {
   useEffect(() => {
     setLoading(true);
     setError(null);
-    fetchBrazeConnectedEntries(cma, sdk.ids.space, sdk.ids.environment || 'master')
+    fetchBrazeConnectedEntries(
+      cma,
+      sdk.parameters?.installation?.contentfulApiKey,
+      sdk.ids.space,
+      sdk.ids.environment || 'master'
+    )
       .then((entries) => {
         setEntries(entries);
         setLoading(false);
@@ -107,13 +115,13 @@ const Page = () => {
                 <Text marginLeft="spacingM">Loading...</Text>
               </Flex>
             ) : error ? (
-              <Box marginBottom="spacingL">
+              <Box marginTop="spacingL">
                 <Note variant="negative" title="Error">
                   {error}
                 </Note>
               </Box>
             ) : entries.length === 0 ? (
-              <Box marginBottom="spacingL">
+              <Box marginTop="spacingL">
                 <Note variant="warning" title="No connected entries">
                   There are no entries connected to Braze Content Blocks.
                 </Note>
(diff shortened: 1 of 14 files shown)


apps/braze/src/locations/Page.tsx
```diff
@@ -1,15 +1,219 @@
...
+      defaults: {
+        environmentId: sdk.ids.environment,
+        spaceId: sdk.ids.space,
+      },
+    }
+  );
+
+  useEffect(() => {
+    setLoading(true);
+    setError(null);
```
FBanfi, line 151: Delete this since it's only rendered once and we are already initializing it in null.

apps/braze/src/utils/fetchBrazeConnectedEntries.ts
```diff
@@ -0,0 +1,58 @@
...
+        entryTitle,
+        fieldsAndTitle.fields,
+        spaceId,
+        environmentId,
+        contentfulApiKey,
+        rawEntry.sys.publishedAt,
+        rawEntry.sys.updatedAt
+      );
+    })
+  );
```
FBanfi, line 57: 💭 We were talking that perhaps in another ticket we could create a builder for this connected entry, creating the fields and being able to return the entry entity.

apps/braze/src/utils/fetchBrazeConnectedEntries.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  return Promise.all(
+    rawEntries.map(async (rawEntry: any) => {
+      const entryId = rawEntry.sys.id;
+      const entryContentTypeId = rawEntry.sys.contentType?.sys?.id || '';
+      const entryConnectedFields: EntryConnectedField[] = entries[entryId];
+      const connectedFieldIds = entryConnectedFields.map((f) => f.fieldId);
+
+      const fieldsFactory = new FieldsFactory(entryId, entryContentTypeId, cma);
+      const fieldsAndTitle = await fieldsFactory.createFieldsForConnectedEntries(connectedFieldIds);
```
FBanfi, line 42: 💭 Shouldn't this name be singular? `createFieldsForConnectedEntry`, since we are creating the fieldsAndTitle for a single entry.

apps/braze/src/locations/Page.tsx
```diff
@@ -1,15 +1,218 @@
...
+  );
+
+  useEffect(() => {
+    setLoading(true);
+    fetchBrazeConnectedEntries(
+      cma,
+      sdk.parameters?.installation?.contentfulApiKey,
+      sdk.ids.space,
+      sdk.ids.environment || 'master',
+      sdk.locales.default || 'en-US'
```
joaquincasal, line 157: Do we need this hardcoded defaults? 🤔

apps/braze/src/locations/Page.tsx
```diff
@@ -1,15 +1,218 @@
...
+    return '-';
+  }
+  const date = new Date(dateString);
+  const now = new Date();
+  const diffMs = now.getTime() - date.getTime();
+  const diffHours = Math.floor(diffMs / (1000 * 60 * 60));
+  if (diffHours < 24) {
+    return `${diffHours} hour${diffHours === 1 ? '' : 's'} ago`;
+  }
+  return date.toLocaleDateString('en-GB', {
```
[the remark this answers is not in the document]
JuliRossi: This function given a date ie 26/05/2025 returns the string "26 May 2025" , which is the same as the figma. I'm not sure if we really want to localize this text because the rest of the texts in this page aren't localized.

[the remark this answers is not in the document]
joaquincasal: Got it, sounds good!

joaquincasal (approved):

PR #9864, opened 2025-05-27
Braze: fix skeleton hanging [INTEG-2680]

joaquincasal:
## Purpose

- There was a bug in which the skeleton in the creation step wouldn't finish loading
- We've also removed any special characters from the entry title since they are not allowed in content blocks names

image


JuliRossi (approved):
Thanks for looking at this!

FBanfi (approved):

PR #9870, opened 2025-05-29
Braze: Config screen issues [INTEG-2685]

joaquincasal:
- If a content type already has the app in the sidebar and we configure the app again, the app gets removed: we weren't keeping the current state, now we do
- We should display the content types that already have the app in the sidebar: done
- Do not display the “brazeConfig” content type as an option: done

https://github.com/user-attachments/assets/5a7191e0-e32c-4d79-9c1a-39aa580bdba7


JuliRossi (commented):
Nice finds!

JuliRossi (approved):

PR #9889, opened 2025-06-03
Braze: Handle empty sidebar []

joaquincasal:
## Purpose

An error was being thrown in the Sidebar if the entry did not have any connected fields. We now return an empty array since it's a valid scenario.


JuliRossi (approved):
🚀

PR #9892, opened 2025-06-04
General page [INTEG-2662]

JuliRossi:
## Purpose
Decided to break this ticket into 2 parts because it was getting to big to review. In this PR will take care of:
- Show table for entries and different types
- Show content types navigation

Part 2 will include:
- Disabling fields
- Adding checkboxes
- Pagination and sorting
## Approach
- Deleted all unnecessary locations
- Added table for showing content
- Added links for display names
- Added tests

## Testing steps

Automated tests are working

https://github.com/user-attachments/assets/7742b3b2-176f-459b-9cd7-8436d220a8b0


apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+  const [entries, setEntries] = useState<Entry[]>([]);
+  const [entriesLoading, setEntriesLoading] = useState(false);
+  const [fields, setFields] = useState<ContentTypeField[]>([]);
+  const LOCALE = sdk.locales.default;
+
+  useEffect(() => {
+    const fetchContentTypes = async (): Promise<void> => {
+      setLoading(true);
+      try {
+        const res = await sdk.cma.contentType.getMany({});
```
joaquincasal: I think this request has a limit on how many content types get returned, we may want to make several requests until we've received all of them. We do something like that in Braze in the config screen

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+    if (typeof value === 'object' && value !== null) {
+      return '';
+    }
+
+    return value !== undefined && value !== null ? truncate(String(value)) : '-';
+  };
+
+  const getEntryTitle = (entry: Entry, fields: ContentTypeField[]): string => {
+    if (!fields.length) return 'Untitled';
+    const value = entry.fields[fields[0].id]?.[LOCALE];
```
joaquincasal: The title of the entry is not always from the first field. There's a property `displayField` in the content type that indicates the field id for the field used for the title of the entry

apps/bulk-edit/src/locations/Page/index.tsx
JuliRossi: Good catch!

ryunsong-contentful:
Just double checking something, in the figma file https://www.figma.com/design/bHX4i2exnrRIyG98J1b6fi/SE-PS-Marketplace-App-Contribution?node-id=566-5024&p=f&t=6bgIWmSO30hNO2A9-0 there's a Display Name and Name which have the same value. The reason is because while the user is editing the name they can see what the original name was. But in this PR, there is only the display name. Let me know if this is something that changed.

Screenshot 2025-06-05 at 9 53 45 AM

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+        query: { skip, limit },
+      });
+      const items = response.items as ContentTypeProps[];
+      allContentTypes.push(...items);
+      fetched = items.length;
+      skip += limit;
+    } while (fetched === limit);
+
+    return allContentTypes;
+  };
```
FBanfi, line 52: Cool way to handle the limit! 💯

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+    if (sys.version >= sys.publishedVersion + 2) {
+      return { label: 'Changed', color: 'primary' };
+    }
+    if (sys.version === sys.publishedVersion + 1) {
+      return { label: 'Published', color: 'positive' };
+    }
+    return { label: 'Unknown', color: 'negative' };
+  };
+
+  const isLocationValue = (value: unknown): value is { lat: number; lon: number } => {
```
FBanfi, line 130: 💭 I'm not sure if I like having the `isLocationValue, isLinkValue, renderFieldValue, getEntryTitle, getEntryUrl ` functions in this file, since it involves adding logic of the `content type fields` types, which the page doesn't need to know.

apps/bulk-edit/test/locations/Page/PageNavBar.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { mockSdk } from '../../mocks/mockSdk';
+import { createMockCma, getManyContentTypes } from '../../mocks/mockCma';
+import {
+  condoAContentType,
+  condoBContentType,
+  condoCContentType,
+} from '../../mocks/mockContentTypes';
+
+// Correctly mock Forma 36 NavList component
+// Mock useSDK to return mockSdk
```
FBanfi: 💭 Nitpick: We could kill this ones 🔫

apps/bulk-edit/src/locations/Page/index.tsx
JuliRossi: Will make refactors in another PR

ryunsong-contentful (approved):
You're welcome to await the merge based on your 10Pines teammates approval, but this is good to ship for me. Tested the app and generally works like I expect

PR #9904, opened 2025-06-09
General page part 2 [INTEG-2662]

JuliRossi:
## Purpose
First part of this ticket here: https://github.com/contentful/apps/pull/9892
Part 2 includes

- Disabling fields
- Adding checkboxes
- Pagination and sorting

Disclaimer: decided to add localization in a future PR as this one was looking pretty large
## Approach
- Refactor to have several components instead of a huge one
- Added the Pagination Component and capability
- Added sorting menu and capability
- Added tooltips for disable fields
- Added disable styles
- Added tests

## Testing steps

https://github.com/user-attachments/assets/94a008f6-76cf-4366-b6b3-552179d5914f


ryunsong-contentful:
Two things 
1. Double check with Erin about the sort dropdown if we want to update the text or not based on what the user selects (designs have it the way you built it)
2. There's a permanent hover highlight for the first element in that sort dropdown which should not be there.
https://github.com/user-attachments/assets/f47a55e2-552f-48f1-9360-6b0b0aa267a8

apps/bulk-edit/src/locations/Page/components/ContentTypeSidebar.tsx
```diff
@@ -0,0 +1,35 @@
...
+  onContentTypeSelect: (id: string) => void;
+}
+
+export const ContentTypeSidebar: React.FC<ContentTypeSidebarProps> = ({
+  contentTypes,
+  selectedContentTypeId,
+  onContentTypeSelect,
+}) => {
+  return (
+    <Flex style={styles.sidebar} padding="spacingM" flexDirection="column" gap="spacingXs">
```
ryunsong-contentful, line 19: Check with Erin about the empty case and if there should be some kind of message displayed in the sidebar. Like an italicized `No Content Types Available` for example

JuliRossi:
Will take care of comment in other PRs! Thanks for the review!

PR #9906, opened 2025-06-10
feat: initialize iterable project [INTEG-2748]

ryunsong-contentful:


apps/iterable/src/locations/Dialog.tsx
```diff
@@ -0,0 +1,16 @@
...
+  /*
+     To use the cma, inject it as follows.
+     If it is not needed, you can remove the next line.
+  */
+  // const cma = useCMA();
+
+  return <Paragraph>Hello Dialog Component (AppId: {sdk.ids.app})</Paragraph>;
+};
+
+export default Dialog;
```
david-shibley-contentful, line 16: Do we need the boiler plate code?

ryunsong-contentful: No, I could probably remove the location entirely, but for the first commit of a new app, I like having the freshest fullest starting point which is what the CLI tool gives you for a new app. Then for follow up PRs, we delete locations, and start the actual implementation

primeinteger (approved):
hello world

PR #9908, opened 2025-06-11
Bulk Edit App: Pagination buttons adjustment [INTEG-2737] 

FBanfi:
## Purpose

The pagination buttons won't show while scrolling horizontally. This update fix this situation and now the user can always see the pagination buttons.

## Approach

The pagination is now sticky, following the same idea of the first column:

https://github.com/user-attachments/assets/228066df-5412-4079-a27a-6356e65ca2d5

## Testing steps

N/A

## Breaking Changes

N/A

Link to [INTEG-2737](https://contentful.atlassian.net/browse/INTEG-2737) ticket.

## Deployment

N/A

### DIFF
--- a/apps/bulk-edit/src/locations/Page/styles.ts
+++ b/apps/bulk-edit/src/locations/Page/styles.ts
@@ -85,4 +85,14 @@ export const styles = {
     zIndex: 1,
     width: 'fit-content',
   },
+  paginationContainer: {
+    position: 'sticky',
+    left: '18%',
+    zIndex: 1,
+    background: tokens.colorWhite,
+    paddingBottom: tokens.spacingM,
+    maxWidth: `80vw`,
+    marginTop: '20px',
+    marginRight: '10px',
+  },
 } as const;
--- a/apps/bulk-edit/src/locations/Page/components/EntryTable.tsx
+++ b/apps/bulk-edit/src/locations/Page/components/EntryTable.tsx
@@ -156,7 +156,7 @@ export const EntryTable: React.FC<EntryTableProps> = ({
           ))}
         </Table.Body>
       </Table>
-      <Box marginTop="spacingM">
+      <Box style={styles.paginationContainer}>
         <Pagination
           activePage={activePage}
           onPageChange={onPageChange}


apps/bulk-edit/src/locations/Page/styles.ts
```diff
@@ -85,4 +85,14 @@ export const styles = {
...
   },
+  paginationContainer: {
+    position: 'sticky',
+    left: '18%',
+    zIndex: 1,
+    background: tokens.colorWhite,
+    paddingBottom: tokens.spacingM,
+    maxWidth: `80vw`,
+    marginTop: '20px',
+    marginRight: '10px',
```
JuliRossi: 💭 Nitpick: Can we use spacingL or spacingM for the top and spacingS for the right? https://f36.contentful.com/tokens/spacing

apps/bulk-edit/src/locations/Page/styles.ts
```diff
@@ -85,4 +85,14 @@ export const styles = {
     zIndex: 1,
     width: 'fit-content',
   },
+  paginationContainer: {
+    position: 'sticky',
+    left: '18%',
```
JuliRossi: 💭  can we use SIDEBAR_WIDTH + STICKY_SPACER_SPACING so they are align with the other sticky components? Another thing that could work is using that 18% in the other places

JuliRossi (approved):
Left a few non-blocking comments. Amazing work figuring out how to make it sticky!

apps/bulk-edit/src/locations/Page/styles.ts
FBanfi: Yes, totally, I missed that!
Done ✅

apps/bulk-edit/src/locations/Page/styles.ts
FBanfi: We could use the `SIDEBAR_WIDTH + STICKY_SPACER_SPACING` in the left property instead of the `18%`. Now it's changed that way, but I'm not sure if that was what you meant.

JuliRossi: Yep exactly like that. Thanks!

PR #9909, opened 2025-06-12
Bulk Edit App: Empty content type and entries handling [INTEG-2728]

FBanfi:
## Purpose

There wasn't a handling when there weren't any content types or entries.

## Approach

Now a text is rendered depending if the fetch's of the content type and entries are done correctly:
- if there is no content types - a message is rendered
- if there is no entries of a content type - a message is rendered

## Testing steps

You can install the app on an empty state and you will see this messages:
Captura de pantalla 2025-06-12 a la(s) 10 01 41 a  m

Then if you add a content type but no entries, you will see the following:
Captura de pantalla 2025-06-12 a la(s) 10 02 15 a  m

If you add entries then you will see them:
Captura de pantalla 2025-06-12 a la(s) 10 02 33 a  m

## Breaking Changes

N/A

Link to [INTEG-2728](https://contentful.atlassian.net/browse/INTEG-2728) ticket

## Deployment

N/A

[INTEG-2728]: https://contentful.atlassian.net/browse/INTEG-2728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bulk-edit/src/locations/Page/index.tsx
+++ b/apps/bulk-edit/src/locations/Page/index.tsx
@@ -153,38 +153,38 @@ const Page = () => {
             />
             <div style={styles.stickySpacer} />
             <Box>
-              {!selectedContentType ? (
-                <Spinner />
-              ) : (
-                <>
-                  <Heading style={styles.stickyPageHeader}>
-                    {selectedContentType
-                      ? `Bulk edit ${selectedContentType.name}`
-                      : 'Bulk Edit App'}
-                  </Heading>
+              <>
+                <Heading style={styles.stickyPageHeader}>
+                  {selectedContentType ? `Bulk edit ${selectedContentType.name}` : 'Bulk Edit App'}
+                </Heading>
+                {(entries.length === 0 && !entriesLoading) || !selectedContentType ? (
+                  <Box style={styles.noEntriesText}>No entries found.</Box>
+                ) : (
                   <>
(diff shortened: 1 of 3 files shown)


apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -153,38 +153,38 @@ const Page = () => {
...
+                  {selectedContentType ? `Bulk edit ${selectedContentType.name}` : 'Bulk Edit App'}
+                </Heading>
+                {(entries.length === 0 && !entriesLoading) || !selectedContentType ? (
+                  <Box style={styles.noEntriesText}>No entries found.</Box>
+                ) : (
                   <>
-                    <SortMenu sortOption={sortOption} onSortChange={setSortOption} />
                     {entriesLoading ? (
                       <Spinner />
                     ) : (
```
JuliRossi: 💭 Nitpick: Perhaps to avoid nested ifs, this spinner could be an early return instead.

JuliRossi (approved):

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: Yes! Nice call, I like that refactor!
Done ✅

PR #9917, opened 2025-06-17
Fix null displayname bug- [INTEG-2708]

JuliRossi:
## Purpose

Franco found an edge case: the content type can have null displayName. In the table shows as if the user doesn't have entries:
Captura de pantalla 2025-06-17 a la(s) 12 38 58 p  m

Captura de pantalla 2025-06-17 a la(s) 12 38 53 p  m

## Approach
 - Added new tests
 - Updated query for order 
 - Allow for displayName to be null

## Testing steps
- Created a type with only a json field, this causes for the display name to be null
- Now it loads the entry

https://github.com/user-attachments/assets/78e8e423-85cf-4aef-9fa4-e84cb2c2387a

### DIFF
--- a/apps/bulk-edit/src/locations/Page/index.tsx
+++ b/apps/bulk-edit/src/locations/Page/index.tsx
@@ -65,12 +65,13 @@ const Page = () => {
     return allContentTypes;
   };
 
-  const buildQuery = (sortOption: string, displayField: string) => {
+  const buildQuery = (sortOption: string, displayField: string | null) => {
     const getOrder = (sortOption: string) => {
-      if (sortOption === 'displayName_asc') return `fields.${displayField}`;
-      else if (sortOption === 'displayName_desc') return `-fields.${displayField}`;
-      else if (sortOption === 'updatedAt_desc') return '-sys.updatedAt';
+      if (sortOption === 'updatedAt_desc') return '-sys.updatedAt';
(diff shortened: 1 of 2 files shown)


apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -65,12 +65,13 @@ const Page = () => {
...
 
-  const buildQuery = (sortOption: string, displayField: string) => {
+  const buildQuery = (sortOption: string, displayField: string | null) => {
     const getOrder = (sortOption: string) => {
-      if (sortOption === 'displayName_asc') return `fields.${displayField}`;
-      else if (sortOption === 'displayName_desc') return `-fields.${displayField}`;
-      else if (sortOption === 'updatedAt_desc') return '-sys.updatedAt';
+      if (sortOption === 'updatedAt_desc') return '-sys.updatedAt';
       else if (sortOption === 'updatedAt_asc') return 'sys.updatedAt';
+      else if (displayField === null) return undefined;
```
ryunsong-contentful, line 72: nit: `else if (!displayField) return undefined` - would check if displayField is undefined as well. But maybe you had a specific reason for this specific check

JuliRossi: We are setting displayField to null in 143 if it isn't present. We changed the default value to null

PR #9918, opened 2025-06-17
Bulk edit: Use display field when showing update errors [INTEG-2801]

joaquincasal:
When an error occurs after an update, we were displaying the previous value of the updated field for the failed entry. We now display the title of the entry so help the user recognize the entry that failed

image


JuliRossi (approved):
✅

PR #9920, opened 2025-06-18
Fix no string for success notif and modal issue [INTEG-2708]

JuliRossi:
## Purpose

Found 2 issue while creating the videos for the main feature branch in which using:
- using 0 made it as empty value
- we want to handle empty fields as well

## Approach

Asked erin about this case, and fix it

## Testing steps

https://github.com/user-attachments/assets/f3722fb2-34c0-45da-8ee8-2d3810203631


FBanfi (approved):
💯 !

PR #9924, opened 2025-06-19
Hubspot: Config screen base structure [INTEG-2769]

FBanfi:
## Purpose

This update introduce the first version of the config screen with the base structure:
- the hubspot access token (with empty check validation)
- the instructions expandable section

Note: Deeper validations are handled in another ticket.

## Approach

The config screen location was modified so the UI matches the figma design.

## Testing steps

Automated tests added. Manual check:

https://github.com/user-attachments/assets/ae323ced-623a-43db-a34a-5981f2ebcc17

## Breaking Changes

N/A

Link to [INTEG-2769](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?issueParent=367108%2C365033%2C346554&selectedIssue=INTEG-2769) ticket

## Deployment

N/A

[INTEG-2769]: https://contentful.atlassian.net/browse/INTEG-2769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/hubspot/src/locations/ConfigScreen.tsx
+++ b/apps/hubspot/src/locations/ConfigScreen.tsx
@@ -12,6 +12,7 @@ import {
   IconButton,
   Subheading,
   TextLink,
+  List,
 } from '@contentful/f36-components';
 import { ChevronDownIcon, ChevronUpIcon, ExternalLinkIcon } from '@contentful/f36-icons';
 import { useSDK } from '@contentful/react-apps-toolkit';
@@ -26,7 +27,7 @@ import {
 
 const ConfigScreen = () => {
   const sdk = useSDK<ConfigAppSDK>();
-  const [isExpanded, setIsExpanded] = useState(true);
+  const [isExpanded, setIsExpanded] = useState(false);
   const [isHubspotTokenInvalid, setIsHubspotTokenInvalid] = useState(false);
   const [parameters, setParameters] = useState<AppInstallationParameters>({
     hubspotAccessToken: '',
@@ -66,9 +67,7 @@ const ConfigScreen = () => {
       const currentParameters: AppInstallationParameters | null = await sdk.app.getParameters();
 
       if (currentParameters) {
-        setParameters({
-          hubspotAccessToken: currentParameters.hubspotAccessToken ?? '',
-        });
+        setParameters(currentParameters);
       }
       sdk.app.setReady();
     })();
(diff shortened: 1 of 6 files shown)


apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -1,67 +1,152 @@
...
 
 const ConfigScreen = () => {
-  const [parameters, setParameters] = useState<AppInstallationParameters>({});
   const sdk = useSDK<ConfigAppSDK>();
-  /*
-     To use the cma, inject it as follows.
-     If it is not needed, you can remove the next line.
-  */
-  // const cma = useCMA();
+  const [isExpanded, setIsExpanded] = useState(true);
```
JuliRossi: ❓ Is this okey to be expanded when the user enters the config screen? Mark it as resolve if yes.

apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -1,67 +1,152 @@
...
     (async () => {
-      // Get current parameters of the app.
-      // If the app is not installed yet, `parameters` will be `null`.
       const currentParameters: AppInstallationParameters | null = await sdk.app.getParameters();
 
       if (currentParameters) {
-        setParameters(currentParameters);
+        setParameters({
+          hubspotAccessToken: currentParameters.hubspotAccessToken ?? '',
+        });
```
JuliRossi: ⚔️ this should be setParameters(currentParameters); because currentParameters is going to have all of the parameters we need for installation and so we don't have to indicate each one individually. Also this should never be empty

apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -1,67 +1,152 @@
...
+            <Paragraph marginBottom="none" fontColor="gray500">
+              To create a private app access token:
+            </Paragraph>
+            <ol className={styles.orderList}>
+              {CONFIG_SCREEN_INSTRUCTIONS.map((step, i) => (
+                <li key={i} className={styles.listItem}>
+                  {step}
+                </li>
+              ))}
+            </ol>
```
JuliRossi: 💭 Nitpick: We could use here the List component and List Item from forma. Use with as=ol prop

FBanfi: Yes, sounds better!

Done ✅

apps/hubspot/src/locations/ConfigScreen.tsx
FBanfi: Yes totally, I did it that way because I had a empty warning in the UI console. But I double checked it now with `setParameters(currentParameters)` and now it doesnt show me the warning. Perhaps another change fixed it, anyway good call, thanks!

Done ✅

apps/hubspot/src/locations/ConfigScreen.tsx
FBanfi: Nope, you are right!
Corrected ✅

JuliRossi (approved):
Looks great!

PR #9925, opened 2025-06-19
Fix errors and add test [INTEG-2819]

JuliRossi:
## Purpose

We wanted to fix the following:
"When changed the name, the error doesn't go until we click again the create"

## Approach
- Clear errors after save instead of after onCreate
- Added more tests

## Testing steps

https://github.com/user-attachments/assets/250377dc-33c6-4a78-b602-1b17de556149

### DIFF
--- a/apps/braze/src/components/create/CreateStep.tsx
+++ b/apps/braze/src/components/create/CreateStep.tsx
@@ -280,6 +280,8 @@ const CreateStep = ({
     description: '',
   });
   const [isNameValid, setIsNameValid] = useState(true);
+  // Track which errors should be cleared after save
+  const [clearedErrors, setClearedErrors] = useState<Set<string>>(new Set());
 
   const localizedFieldsIds = Array.from(selectedFields).flatMap((fieldId) => {
     const isLocalized = entry.fields.find((f) => f.id === fieldId)?.localized || false;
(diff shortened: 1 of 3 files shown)


apps/braze/src/components/create/CreateStep.tsx
```diff
@@ -360,20 +362,55 @@ const CreateStep = ({
...
+      setClearedErrors((prev) => new Set(prev).add(fieldId));
+    }
+
     setEditingField(null);
     setEditDraft({ name: '', description: '' });
   };
 
+  const shouldShowError = (fieldId: string): boolean => {
+    const hasError = creationResultFields.some(
+      (result) => fieldId === localizeFieldId(result.fieldId, result.locale)
```
FBanfi: 💭 I'm a bit confused by just checking if it's in `creationResultFields`, since that might include both errors and successful creations.
Shouldn't we also check if success === false?

FBanfi (approved):
Great fix!! 😃

apps/braze/src/components/create/CreateStep.tsx
JuliRossi: You are corect! I'll fix it

JuliRossi: Captura de pantalla 2025-06-23 a la(s) 12 19 59 p  m

PR #9929, opened 2025-06-23
Hubspot: adding app to content type through config screen [INTEG-2770]

FBanfi:
## Purpose

This update includes the content type configuration section in the config screen. Through this section we will be able to add the app into multiple content types when it is installed.

## Approach

A new ContentTypeMultiSelect was created, which allows the users to check in which content types they want to install the app. In addition, modifications in the config screen to use this component and section were made.

## Testing steps

Automated tests were added in the config screen location. A new suit test was created for the new component.

https://github.com/user-attachments/assets/b28cdeea-eafd-4752-83c4-e5b1d8236c86

## Breaking Changes

N/A

N/A

## Deployment

N/A


apps/hubspot/src/components/ContentTypeMultiSelect.tsx
```diff
@@ -0,0 +1,141 @@
...
+  setSelectedContentTypes,
+  sdk,
+  cma,
+}) => {
+  const [availableContentTypes, setAvailableContentTypes] = useState<ContentType[]>([]);
+  const getPlaceholderText = useCallback(() => {
+    if (selectedContentTypes.length === 0) return 'Select one or more';
+    if (selectedContentTypes.length === 1) return selectedContentTypes[0].name;
+    return `${selectedContentTypes[0].name} and ${selectedContentTypes.length - 1} more`;
+  }, [selectedContentTypes]);
```
JuliRossi: 💭 Nitpick: is there a reason we are using a callback for this?

FBanfi: Nou, it is not necessary! Thanks

Removed ✅

PR #9932, opened 2025-06-24
Hubspot app:Add sidebar []

joaquincasal:
## Purpose

Adding the sidebar location for the Hubspot app.
It consists of two buttons: one opens the dialog location and the other one redirects to the page location.

## Testing steps

- Add app to the sidebar of a content type
- Go to an entry of that content type
- Use both buttons

https://github.com/user-attachments/assets/53648082-6fca-4ca7-8189-0c5820dba948


JuliRossi (approved):
👍

PR #9933, opened 2025-06-24
Braze app: Fix long loading time [INTEG-2822]

JuliRossi:
## Purpose

We want to fix a bug that causes an infinite loader.

### Bug cause
The way we were getting the displayName(title) for references and arrays of references with no displayName(title) was undefined when doing` Object.values(f.fields[fieldContentType.displayField] as { [key: string]: any })[0] `and that cause it to infinite load.

### What was done to fix this
- Fix the logic for getting a name for references and array references
- Encapsulate logic
- Added tests

## Testing steps
Automated testing

Use a duplicated entry for the one that failed. This reference is untitle and has no content

https://github.com/user-attachments/assets/ea05fe52-cccf-44ad-9c7f-509618bb8ba1

Testing everything still works

https://github.com/user-attachments/assets/75af5758-3b5e-47ee-a970-3a3739bc5b10

### DIFF
--- a/apps/braze/src/fields/FieldsFactory.ts
+++ b/apps/braze/src/fields/FieldsFactory.ts
@@ -21,12 +21,19 @@ export class FieldsFactory {
   private entryId: string;
   private entryContentTypeId: string;
   private cma: PlainClientAPI;
+  private defaultLocale: string;
   NESTED_DEPTH = 5;
 
-  public constructor(entryId: string, entryContentTypeId: string, cma: PlainClientAPI) {
+  public constructor(
+    entryId: string,
+    entryContentTypeId: string,
+    cma: PlainClientAPI,
+    defaultLocale: string
+  ) {
     this.entryId = entryId;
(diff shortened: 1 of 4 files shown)


apps/braze/src/fields/FieldsFactory.ts
```diff
@@ -186,4 +184,15 @@ export class FieldsFactory {
       items
     );
   }
+
+  private getDisplayFieldValue(fieldValue: any, displayField: string | undefined): string {
+    if (!displayField) return 'Untitled';
+    const displayFieldValue = fieldValue.fields?.[displayField];
+    if (!displayFieldValue || displayFieldValue === null) return 'Untitled';
+    if (typeof displayFieldValue === 'object') {
+      return Object.values(displayFieldValue)[0] as string;
```
joaquincasal: I think ideally here we would use the defaultLocale instead of using the first value, which may not be the value for the default locale

joaquincasal (approved):

PR #9940, opened 2025-06-26
Hubspot: Add fields selection step [INTEG-2774]

joaquincasal:
## Purpose

The first step in the Dialog is being able to select the fields that will be synced to Hubspot

## Approach

The entry fields are obtained from the sdk in the Sidebar and sent as parameters to the Dialog. This provides the relevant information for displaying (field name, if it's supported or not, etc.).

Special attention was given to ensuring the FieldsSelection component is highly reusable, we should be able to use it in similar apps.

## Testing steps

Open the Dialog from the Sidebar button and check fields are displayed as expected.
Automateed tests were also added.

https://github.com/user-attachments/assets/1b671a95-3d24-4d08-99fd-cf577f118107


apps/hubspot/src/components/FieldSelection.tsx
```diff
@@ -0,0 +1,121 @@
...
+        Select fields that you want to sync. Each field will map to a custom module that can be used
+        in Hubspot emails. If edits are made in Contentful, content will update automatically.
+      </Paragraph>
+
+      <Box
+        className={css({
+          border: `1px solid ${tokens.gray300}`,
+          borderRadius: tokens.borderRadiusSmall,
+        })}>
+        <Box
```
JuliRossi, line 38: 💭  Nit: we could move this to a style file

apps/hubspot/src/components/FieldSelection.tsx
```diff
@@ -0,0 +1,121 @@
...
+          paddingLeft="spacingS"
+          paddingTop="spacingXs"
+          paddingBottom="spacingXs"
+          className={css({
+            borderBottom: `1px solid ${tokens.gray300}`,
+          })}>
```
JuliRossi, line 44: ⚔️ The line dividing select all fields and the fields seems thicker. It may be because of this

apps/hubspot/src/locations/Dialog.tsx
```diff
@@ -1,16 +1,62 @@
...
+  uniqueId: string;
+  name: string;
+  locale?: string;
+  linkType?: string; // FieldLinkType
+  items?: {
+    type: string;
+    linkType: string;
+  }; // Items
+  supported: boolean;
+};
```
JuliRossi, line 19: 💭  You should be able to use import { FieldType } from 'contentful-management';  and avoid defining all types
```suggestion
export type SdkField = {
  id: string;
  uniqueId: string;
  name: string;
  locale?: string;
 supported: boolean;
} & FieldType;
```

JuliRossi (approved):
👍

PR #9944, opened 2025-06-27
Hubspot: Create modules [INTEG-2776]

joaquincasal:
## Purpose

Being able to create custom modules in Hubspot from Contentful fields

## Approach

A Function was added to make the calls to Hubspot for the module creation. 
This PR only supports text fields, the rest will be added in a separate PR.

It's necessary to create three different files to create a module in hubspot: module.html, fields.json and meta.json. Each of those has their own template from which we generate the final file. The value of the contentful field is set to the default value of the hubspot field in the fields.json file.

## Testing steps

Open the dialog from the sidebar, select one or multiple text fields, click next and wait (there's no feedback for the user after clicking next but the request is triggered).
Automated tests were also added.

### DIFF
--- a/apps/hubspot/test/functions/createModules.spec.ts
+++ b/apps/hubspot/test/functions/createModules.spec.ts
@@ -10,7 +10,7 @@ describe('createModules', () => {
     vi.clearAllMocks();
   });
 
-  test('should successfully create a module for a text field', async () => {
+  it('should successfully create a module for a text field', async () => {
     // Mock successful fetch responses
     const mockFetch = vi.mocked(fetch);
     mockFetch.mockResolvedValue({


apps/hubspot/src/locations/Dialog.tsx
```diff
@@ -1,34 +1,59 @@
...
+      },
+    }
+  );
   useAutoResizer();
   const invocationParams = sdk.parameters.invocation as InvocationParams;
   const fields = invocationParams.fields;
   const [selectedFields, setSelectedFields] = useState<string[]>([]);
 
+  const callAction = async () => {
+    console.log('Starting call action');
```
JuliRossi, line 31: ❌ Remove this console.log

apps/hubspot/src/locations/Dialog.tsx
```diff
@@ -1,34 +1,59 @@
...
+          appActionId: 'createModulesAction',
+        },
+        {
+          parameters: {
+            fields: JSON.stringify(fieldsToSend),
+          },
+        }
+      );
+      const responseData = JSON.parse(response.response.body);
+      console.log('Response data: ', responseData);
```
JuliRossi, line 51: ❌ Remove this console.log

apps/hubspot/src/locations/Sidebar.tsx
```diff
@@ -58,11 +59,14 @@ const Sidebar = () => {
           ...linkType,
           ...items,
           supported: supported,
+          value: field.getValue(locale),
         });
       }
     }
   }
 
+  console.log('FIELDS: ', fields);
```
JuliRossi, line 68: ❌ Remove this console.log

apps/hubspot/test/functions/createModules.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+      expect.objectContaining({
+        method: 'PUT',
+        headers: {
+          Authorization: 'Bearer test-token',
+        },
+      })
+    );
+  });
+
+  test('should handle API errors and return failed fields', async () => {
```
JuliRossi: 💭 Nit: To keep it consistent, I think this should be:
```suggestion
  it('should handle API errors and return failed fields', async () => {
```

apps/hubspot/test/functions/createModules.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+
+// Mock fetch globally
+global.fetch = vi.fn();
+
+describe('createModules', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  test('should successfully create a module for a text field', async () => {
```
JuliRossi: 💭 Nit: To keep it consistent, I think this should be:
```suggestion
  it('should successfully create a module for a text field', async () => {
```

apps/hubspot/src/locations/Dialog.tsx
joaquincasal: The ticket is not finished yet, we still need to add support for other field types, so those console logs will still be useful and we can remove it in the next PR

FBanfi (approved):
Awesome!

PR #9945, opened 2025-06-27
Hubspot: Config screen validations with a function [INTEG-2771]

FBanfi:
## Purpose

We needed to check that the access token provided is correct and has the appropiate scopes.

## Approach

Since Hubspot API doesn't allow us to make calls in the frontend, we created a function to validate:
- If the access token is valid
- If the access token has the necessary scope (content)

We make 2 different calls to differentiate between this two cases, the first endpoint doesn't require any scope, while the second one need the "content" scope.

## Testing steps

Test to validate the proper config screen behavior were added.
Manual testing:

https://github.com/user-attachments/assets/9c1ad74c-7a6c-4f42-979b-6d3d1c475f62

## Breaking Changes

N/A

N/A

## Deployment

N/A

### DIFF
--- a/apps/hubspot/functions/validateHubspotToken.ts
+++ b/apps/hubspot/functions/validateHubspotToken.ts
@@ -14,41 +14,51 @@ type AppActionParameters = {
   token: string;
 };
 
+const ENDPOINT_WITHOUT_SCOPE_CHECK = 'https://api.hubapi.com/integrations/v1/me';
+const ENDPOINT_WITH_SCOPE_CHECK =
+  'https://api.hubapi.com/cms/v3/source-code/draft/metadata/@hubspot';
+
+const INVALID_ACCESS_TOKEN = 'Invalid HubSpot access token';
+const INVALID_ACCESS_TOKEN_SCOPE = 'The HubSpot token is missing the required "content" scope.';
+const UNKNOWN_ACCESS_TOKEN_ERROR = 'Unknown error';
+
 export const handler: FunctionEventHandler<FunctionTypeEnum.AppActionCall> = async (
   event: AppActionRequest<'Custom', AppActionParameters>
 ) => {
   const token = event.body.token;
 
   // Step 1: Basic token validation
-  const basicCheck = await fetch('https://api.hubapi.com/integrations/v1/me', {
+  const basicCheck = await fetch(ENDPOINT_WITHOUT_SCOPE_CHECK, {
     headers: {
       Authorization: `Bearer ${token}`,
     },
   });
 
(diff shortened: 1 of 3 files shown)


apps/hubspot/functions/validateHubspotToken.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (contentCheck.status === 200) {
+    return { valid: true, hasContentScope: true };
+  } else {
+    const error = await contentCheck.json();
+
+    if (error?.category === 'MISSING_SCOPES') {
+      return { valid: true, hasContentScope: false };
+    }
+
+    return { valid: true, hasContentScope: false };
```
joaquincasal: We're checking if the error is caused by missing scopes but we return the same even if it isn't. I think I'd remove the if or change what we return so in the frontend we can display "Failed for an unkown reason" or something like that

apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -65,6 +69,47 @@ const ConfigScreen = () => {
...
+        setIsHubspotTokenInvalid(true);
+        sdk.notifier.error('Invalid HubSpot access token');
+        return false;
+      }
+
+      if (!responseData.hasContentScope) {
+        setHubspotTokenScopeError('The HubSpot token is missing the required "content" scope.');
+        sdk.notifier.error('The HubSpot token is missing the required "content" scope.');
+        return false;
+      }
```
joaquincasal: Maybe we can only one state instead of two. Instead of having `isHubspotTokenInvalid` and `hubspotTokenScopeError` we could have something more generic like `hubspotTokenError` and in this part of the code set the appropriate error string. And if it's null we know there are no errors. And if there is, we can use the state to display the error to the user

joaquincasal (approved):
Left two minor comments, overall looks good!

apps/hubspot/functions/validateHubspotToken.ts
FBanfi: Done ✅ 

Now the function returns an object with the `valid: boolean`, `hasContentScope: boolean` and `error: string` properties. And the error property stores the reason of the error, which will be stored in a useState in the config screen. This allow us to remove some ifs and conditional rendering from the config screen, and also to differentiate the three different error states (invalid access token, invalid scope, unknown error).

apps/hubspot/src/locations/ConfigScreen.tsx
FBanfi: Done ✅

PR #9947, opened 2025-06-30
Fix undefined for localized fields bulk edit app []

JuliRossi:
## Purpose

Emily found a bug while working with bulk edit.  

https://github.com/user-attachments/assets/571feaf3-2514-416c-a250-2f98e3a74dc5

### Cause

After investigating a bit, realized the error had to do with empty fields for localized fields

## Approach

- Changed the logic for stringifying undefined values
- Added tests
## Testing steps

https://github.com/user-attachments/assets/b09716ce-8c32-4d55-986d-0993580f5450

### DIFF
--- a/apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
+++ b/apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
@@ -137,5 +137,8 @@ export function getEntryFieldValue(
   defaultLocale: string
 ): string {
   if (!entry || !field || !field.id) return '';
-  return entry.fields[field.id]?.[field.locale || defaultLocale].toString() || 'empty field';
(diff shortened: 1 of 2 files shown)


apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
```diff
@@ -136,6 +136,9 @@ export function getEntryFieldValue(
   field: { id: string; locale?: string } | null | undefined,
   defaultLocale: string
 ): string {
-  if (!entry || !field || !field.id) return '';
-  return entry.fields[field.id]?.[field.locale || defaultLocale].toString() || 'empty field';
+  if (!entry || !field || !field.id) return 'empty field';
```
ryunsong-contentful, line 139: Small nit: Could make these constants in the future (like an object that holds these constants)

PR #9948, opened 2025-06-30
Hubspot: Add support for other field types [INTEG-2776]

joaquincasal:
## Purpose

Add support for the remaining field types to the Hubspot app (which now only support text fields).
This is the second part of #9944 

## Approach

We followed the same approach as with the text field: there are templates for each type that get filled with the field value and then sent to Hubspot.

This PR also solves some unrelated warnings we were getting when running tests. They were related to state changes that trigger a re-render not being wait for in the tests.

## Testing steps

Select the fields to send to Hubspot, click next (there is no feedback right now after clicking) and then check that the modules exist in Hubspot.
Automated tests were also added.

https://github.com/user-attachments/assets/44e20c34-1354-4f77-9a40-b36d43d27b51


apps/hubspot/functions/createModules.ts
```diff
@@ -76,3 +90,84 @@ const createModuleFile = async (
     );
   }
 };
+
+const getFields = (field: SdkField): string => {
+  const { type } = field;
+  let fields;
```
FBanfi: 💭 The `fields` variable gives the impression that this is an array, but it isn't (not sure).
I think this variable is used to create the `fields.json` file and you name it that way to match the file but it is a bit confusing 😆

apps/hubspot/functions/createModules.ts
```diff
@@ -76,3 +90,84 @@ const createModuleFile = async (
...
+    case 'Integer':
+      fields = structuredClone(NUMBER_FIELD_TEMPLATE);
+      if (field.value) fields[0].default = field.value;
+      break;
+    case 'Date':
+      const value = field.value as string;
+      if (!value) {
+        fields = structuredClone(DATETIME_FIELD_TEMPLATE);
+      } else if (value.includes('T')) {
+        fields = structuredClone(DATETIME_FIELD_TEMPLATE);
```
JuliRossi: 💭 duplicated logic

apps/hubspot/functions/createModules.ts
```diff
@@ -76,3 +90,84 @@ const createModuleFile = async (
...
+        fields[0].default = new Date(value).getTime();
+      } else {
+        fields = structuredClone(DATE_FIELD_TEMPLATE);
```
JuliRossi: 💭 same here

apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -153,7 +153,7 @@ const ConfigScreen = () => {
               <TextInput
                 name="hubspotAccessToken"
                 placeholder="Enter your access token"
-                value={parameters.hubspotAccessToken}
+                value={parameters.hubspotAccessToken || ''}
```
FBanfi: 😅

apps/hubspot/src/locations/Sidebar.tsx
```diff
@@ -1,85 +1,39 @@
...
-    const linkType = (field as any).linkType && { linkType: (field as any).linkType };
-    const items = (field as any).items && {
-      items: {
-        type: (field as any).items.type,
-        linkType: (field as any).items.linkType,
+  const dialogParams = async () => {
+    return {
+      title: 'Sync entry fields to Hubspot',
+      parameters: {
+        entryTitle: sdk.entry.fields[sdk.contentType.displayField].getValue(),
```
FBanfi: 💭 Should we handle the case in which there isn't a value? Or it isn't necessary?
Like this: `sdk.entry.fields[sdk.contentType.displayField].getValue() || 'Untitled'`

apps/hubspot/test/components/ContentTypeMultiSelect.spec.tsx
```diff
@@ -31,7 +30,6 @@ describe('ContentTypeMultiSelect', () => {
       limit: 100,
       sys: { type: 'Array' },
     });
-    render(<TestWrapper />);
```
FBanfi, line 34: 💭 I don't know if this was necessary, I usually try to leave it in the `beforeEach` to avoid rendering multiple times

apps/hubspot/test/components/ContentTypeMultiSelect.spec.tsx
```diff
@@ -62,41 +60,67 @@ describe('ContentTypeMultiSelect', () => {
...
+    });
   });
 
   it('renders available content types in the dropdown', async () => {
-    expect(await screen.findByText('Blog Post')).toBeTruthy();
+    render(<TestWrapper />);
+    // Wait for content types to be loaded and rendered
+    await waitFor(() => {
+      expect(screen.getByText('Blog Post')).toBeTruthy();
+    });
```
FBanfi: 💭 I know you probably refactor this to have the same consistency in all the file using `waitFor`, but we could have used `expect(await screen.findByText('Blog Post')).toBeTruthy();` since it still wait for the element to be rendered and is a little simpler

apps/hubspot/test/locations/ConfigScreen.spec.tsx
```diff
@@ -102,6 +126,12 @@ describe('Hubspot Config Screen ', () => {
     };
 
     it('adds and removes app from sidebar for each content type', async () => {
+      render(<ConfigScreen />);
+
+      await waitFor(() => {
+        expect(screen.getByText('Select one or more')).toBeTruthy();
+      });
```
FBanfi: 💭 Is this necessary? If we had the previous version we could have avoid to write this multiple times.

apps/hubspot/test/locations/ConfigScreen.spec.tsx
```diff
@@ -125,7 +157,9 @@ describe('Hubspot Config Screen ', () => {
       // removing the app from the content type
       await user.click(closeButton);
 
-      const saveRemovingContentType = await saveAppInstallation();
+      const saveRemovingContentType = await act(async () => {
+        return await saveAppInstallation();
+      });
```
FBanfi, line 149: This was necessary to avoid the test warning right? Nice catch and thanks for fixing it, I didn't pay attention to it while doing these tests. ❤️ 🚀

apps/hubspot/src/locations/ConfigScreen.tsx
JuliRossi: 💭 This isn't needed. The default is '' already

apps/hubspot/functions/createModules.ts
```diff
@@ -76,3 +90,84 @@ const createModuleFile = async (
...
+      if (field.value) {
+        fields[0].default.src = field.value.url;
+        fields[0].default.width = field.value.width;
+        fields[0].default.height = field.value.height;
+      }
+      break;
+    default:
+      fields = structuredClone(TEXT_FIELD_TEMPLATE);
+      if (field.value) fields[0].default = field.value;
+      break;
```
JuliRossi: 💭 If you reach here, you should probably fail because we are not supporting that kind of field

apps/hubspot/src/locations/ConfigScreen.tsx
joaquincasal: Oh a bit of context around this one: when running one test (I don't remember which one) there was a warning of one input moving from controlled to uncontrolled and this was the fix. I don't think it makes much sense to be honest since the parameters is initially '' as well

apps/hubspot/functions/createModules.ts
```diff
@@ -76,3 +90,84 @@ const createModuleFile = async (
     );
   }
 };
+
+const getFields = (field: SdkField): string => {
+  const { type } = field;
+  let fields;
+  switch (type) {
+    case 'Symbol':
+      fields = structuredClone(TEXT_FIELD_TEMPLATE);
```
JuliRossi, line 99: 💭 I feel this could be in the createModule, because we already are calculating the module in line 59. Something like
```suggestion
      fields = structuredClone(module);
```

apps/hubspot/functions/createModules.ts
```diff
@@ -76,3 +90,84 @@ const createModuleFile = async (
...
+      if (field.value) fields[0].default = field.value;
```
JuliRossi: 💭  If we do what I suggested above, this could be outside the switch case

apps/hubspot/functions/createModules.ts
```diff
@@ -37,16 +54,13 @@ export const handler: FunctionEventHandler<FunctionTypeEnum.AppActionCall> = asy
...
-  await createModuleFile(JSON.stringify(META_JSON_TEMPLATE), 'meta.json', field.uniqueId, token);
-
-  const fields = TEXT_FIELD_TEMPLATE;
-  fields[0].default = field.value;
-  await createModuleFile(JSON.stringify(fields), 'fields.json', field.uniqueId, token);
-
-  await createModuleFile(TEXT_MODULES_TEMPLATE, 'module.html', field.uniqueId, token);
+const createModule = async (field: SdkField, token: string, entryTitle: string) => {
+  const fields = getFields(field);
+  const module = getModule(field);
```
JuliRossi: 💭  getHTMLFile?

apps/hubspot/functions/createModules.ts
```diff
@@ -76,3 +90,84 @@ const createModuleFile = async (
...
+        fields[0].default.width = field.value.width;
+        fields[0].default.height = field.value.height;
+      }
+      break;
+    default:
+      fields = structuredClone(TEXT_FIELD_TEMPLATE);
+      if (field.value) fields[0].default = field.value;
+      break;
+  }
+  return JSON.stringify(fields);
```
JuliRossi: ⚔️ {module: module, fields: JSON.stringify(fields);}

apps/hubspot/src/utils/fieldsProcessing.ts
```diff
@@ -0,0 +1,99 @@
...
+  const processedFields: SdkField[] = [];
+
+  for (const field of fields) {
+    const linkType = (field as any).linkType && { linkType: (field as any).linkType };
+    const items = (field as any).items && {
+      items: {
+        type: (field as any).items.type,
+        linkType: (field as any).items.linkType,
+      },
+    };
```
JuliRossi: ⚔️  Use typeOf to avoid casting to any

apps/hubspot/test/functions/createModules.spec.ts
```diff
@@ -140,4 +157,558 @@ describe('createModules', () => {
...
+        entryTitle: 'test-entry-title',
+        fields: JSON.stringify([mockField]),
+      },
+    };
+
+    const mockContext = {
+      appInstallationParameters: {
+        hubspotAccessToken: 'test-token',
+      },
+    };
```
JuliRossi: 💭  We could move this outside

apps/hubspot/test/functions/createModules.spec.ts
```diff
@@ -140,4 +157,558 @@ describe('createModules', () => {
...
     );
   });
+
+  it('should successfully create a module for a text field', async () => {
+    // Mock successful fetch responses
+    const mockFetch = vi.mocked(fetch);
+    mockFetch.mockResolvedValue({
+      ok: true,
+      json: () => Promise.resolve({ success: true }),
+    } as Response);
```
JuliRossi: 💭 We can move this outside

apps/hubspot/test/components/ContentTypeMultiSelect.spec.tsx
joaquincasal: Oh I usually like having all the context of what's being executed in the test itself. It also allows to render the component in different ways if needed in new tests

apps/hubspot/test/components/ContentTypeMultiSelect.spec.tsx
joaquincasal: I actually like that better, I'll change it

apps/hubspot/src/locations/ConfigScreen.tsx
joaquincasal: Found the issue: in line 129 we were checking if `currentParameters` is null, but we didn't check if `currentParameters.hubspotAccessToken` is null, so that's how it was getting set as undefined

apps/hubspot/test/locations/ConfigScreen.spec.tsx
joaquincasal: Yep! Because there are state changes that are caused by this call

apps/hubspot/src/utils/fieldsProcessing.ts
joaquincasal: Let me know if the current solution is what you had in mind!

apps/hubspot/functions/createModules.ts
joaquincasal: I'm not sure this is possible, since the action taken is different in each case depending on the field type

apps/hubspot/src/utils/fieldsProcessing.ts
JuliRossi: Looks good!

PR #9949, opened 2025-07-01
Hubspot: update copy []

joaquincasal:
Update wording in ConfigScreen and Dialog

https://github.com/user-attachments/assets/9feb5fb8-a94d-40d7-a36c-8fabd5aad007

### DIFF
--- a/apps/hubspot/src/locations/ConfigScreen.tsx
+++ b/apps/hubspot/src/locations/ConfigScreen.tsx
@@ -5,7 +5,6 @@ import {
   Heading,
   Text,
   Form,
-  Image,
   FormControl,
   TextInput,
   Paragraph,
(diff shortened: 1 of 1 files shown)


FBanfi (approved):
Great!

FBanfi (approved):
🚀

apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -202,13 +202,16 @@ const ConfigScreen = () => {
                 Read about creating private apps in Hubspot
               </TextLink>
             </Box>
+            <video controls width="100%">
+              <source src={demoVideo} type="video/mp4" />
+            </video>
```
JuliRossi, line 207: 💭 Should we add the ability to pause and continue?

joaquincasal: That's what the `controls` attribute does! Not sure why it doesn't show up in the video, but the controls are there

PR #9950, opened 2025-07-01
Integ 2837 setup environment and install app

harikakondur:
## Purpose

scripts to help with space setup:
- creates app definitions
- creates spaces
- installs apps to spaces
- invites team to spaces

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,82 @@
+# Contentful Space Setup Script
+
+This script automates the complete setup of Contentful spaces for your app, including app definition creation, space creation, team assignments, and app installations.
+
+## What This Script Does
+
+1. **Creates an App Definition** in your Contentful organization
+2. **Creates Two Spaces**: Production and Staging environments
+3. **Assigns Teams** to both spaces with admin privileges
+4. **Installs the App** in both spaces
+5. **Provides Configuration Link** to finish app setup
+
+
+## ⚙️ Configuration
+
+Before running the script, update these variables in `scripts/scripts.js`:
+
+```javascript
+const accessToken = 'YOUR_CONTENTFUL_ACCESS_TOKEN';
+const organizationId = 'YOUR_ORGANIZATION_ID';
+const appName = 'Your App Name';
+const teamId = 'YOUR_TEAM_ID'; 
+const environmentId = 'master'; 
+```
+
+### 🔑 How to Get Required IDs
+
+#### Access Token
+1. Go to [Contentful Web App](https://app.contentful.com)
+2. Navigate to Settings → API keys → Content management tokens
+3. Generate a new personal access token
+
+#### Organization ID
+1. In Contentful, go to your organization settings
+2. The organization ID is in the URL: `app.contentful.com/account/organizations/{ORGANIZATION_ID}`
+
+#### Team ID
+1. Go to Settings → Teams
+2. Click on the team you want to assign
+3. The team ID is in the URL: `app.contentful.com/account/organizations/{ORG_ID}/teams/{TEAM_ID}`
+
+## How to Run
+
+1. **Navigate to the scripts directory:**
+   ```bash
+   cd scripts
+   ```
+
+2. **Install dependencies** (if not already installed):
+   ```bash
+   npm install contentful-management
+   ```
+
+3. **Configure the script** with your values (see Configuration section above)
+
+4. **Run the script:**
+   ```bash
+   node scripts.js
+   ```
+
+## What Gets Created
+
+### App Definition
+- **Name**: `{appName}`
+- **Location**: Your Contentful organization
+- **Status**: Created but needs configuration
+
+### Spaces
+- **Production**: `{appName} (production)`
+- **Staging**: `{appName} (staging)`
+- **Team Access**: Admin privileges for specified team
+- **App Installation**: Your app installed in both spaces
+
+
+## 📝 Next Steps
+
+After running the script successfully:
+
+1. **Configure your app definition** using the provided link
+2. **Set up app locations** (entry field, page, etc.)
+3. **Configure app parameters** if needed
+
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -1,0 +1,44 @@
+import contentful from 'contentful-management';
+import { createSpace,  } from './actions/createSpace.js';
+import { createAppDefinition } from './actions/createAppDefinition.js';
+import { addTeamToSpace } from './actions/addTeamToSpace.js';
+import { installApp } from './actions/installApp.js';
+import { deleteSpace } from './actions/deleteSpace.js';
+import { deleteAppDefinition } from './actions/deleteAppDefinition.js';
+
+
+// Fill out before running the script
+const accessToken = 'YOUR_CONTENTFUL_ACCESS_TOKEN';
+const organizationId = 'YOUR_ORGANIZATION_ID';
+const appName = 'Your App Name';
+const teamId = 'YOUR_TEAM_ID'; 
+const environmentId = 'master'; // Usually 'master'
+
+
+const client = contentful.createClient({
+  accessToken: accessToken,
+});
+
+const appDefinition = await createAppDefinition(client, organizationId, appName).catch(console.error);
+
+const productionSpace = await createSpace(client, organizationId, appName + ' (production)').catch(console.error);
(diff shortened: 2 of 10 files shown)


scripts/scripts.js
```diff
@@ -0,0 +1,45 @@
...
+import { createSpace,  } from './actions/createSpace.js';
+import { createAppDefinition } from './actions/createAppDefinition.js';
+import { addTeamToSpace } from './actions/addTeamToSpace.js';
+import { installApp } from './actions/installApp.js';
+import { deleteSpace } from './actions/deleteSpace.js';
+import { deleteAppDefinition } from './actions/deleteAppDefinition.js';
+
+
+// Fill out before running the script
+const accessToken = 'YOUR_CONTENTFUL_ACCESS_TOKEN';
```
mgoudy91: could maybe set these to be process.env / command args. not needed for v1

sarahlessner (approved):

scripts/actions/addTeamToSpace.ts
```diff
@@ -0,0 +1,45 @@
+import contentful from 'contentful-management';
```
ryunsong-contentful: unused import

scripts/actions/addTeamToSpace.ts
```diff
@@ -0,0 +1,45 @@
...
+    const space = await client.getSpace(spaceId);
+
+    // Check if team already has access
+    const existingMemberships = await space.getTeamSpaceMemberships();
+    const existingMembership = existingMemberships.items.find(
+      (membership) => membership.sys.team.sys.id === teamId
+    );
+
+    if (existingMembership) {
+      console.log('👥 Team already has access to space:', existingMembership.sys.id);
```
ryunsong-contentful: Update this message so that it's clear to the user that no further action has been taken

scripts/actions/addTeamToSpace.ts
```diff
@@ -0,0 +1,45 @@
...
+export async function addTeamToSpace({ client, spaceId, teamId }: AddTeamToSpaceProps) {
+  try {
+    if (!client) {
+      client = await createCMAClient();
+    }
+    const space = await client.getSpace(spaceId);
+
+    // Check if team already has access
+    const existingMemberships = await space.getTeamSpaceMemberships();
+    const existingMembership = existingMemberships.items.find(
```
ryunsong-contentful: since this is used as a boolean, make this .some() and have it return a boolean instead

scripts/actions/addTeamToSpace.ts
```diff
@@ -0,0 +1,45 @@
...
+      (membership) => membership.sys.team.sys.id === teamId
+    );
+
+    if (existingMembership) {
+      console.log('👥 Team already has access to space:', existingMembership.sys.id);
+      return existingMembership;
+    }
+
+    // Create team space membership with admin privileges
+    const teamSpaceMembership = await space.createTeamSpaceMembership(teamId, {
```
ryunsong-contentful, line 28: Create a ticket to make these admin and roles optional down the line

scripts/actions/addTeamToSpace.ts
```diff
@@ -0,0 +1,45 @@
...
+      admin: true,
+      roles: [],
+    });
+
+    console.log(
```
ryunsong-contentful, line 33: It's more preferred to use back ticks ` rather than string concatenation

scripts/actions/createAppDefinition.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!client) {
+    client = await createCMAClient();
+  }
+
+  const organization = await client.getOrganization(organizationId);
+  const appDefinition = await organization.createAppDefinition({
+    name: appName,
+  });
+
+  console.log(
```
ryunsong-contentful, line 22: same comment about concatenation

scripts/actions/deleteTeamSpaceMembership.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    const organization = await client.getOrganization(organizationId);
+    const team = await organization.getTeam(teamId);
+    const space = await client.getSpace(spaceId);
+
+    // Get team space memberships for this specific space
+    const teamSpaceMemberships = await space.getTeamSpaceMemberships();
+
+    const team_membership = teamSpaceMemberships.items.find(
```
ryunsong-contentful: `teamMembership` why the underscore?

scripts/actions/deleteTeamSpaceMembership.ts
```diff
@@ -0,0 +1,55 @@
...
+          team.sys.id +
+          ')' +
+          ' in space: ' +
+          space.name +
+          ' (' +
+          space.sys.id +
+          ')'
+      );
+      await team_membership.delete();
+      console.log('deleted team space membership successfully');
```
ryunsong-contentful: Capitalize `Deleted`

scripts/setup.ts
```diff
@@ -0,0 +1,130 @@
...
+  console.log(
+    '\n ⚙️  Production: https://app.contentful.com/account/organizations/' +
+      organizationId +
+      '/apps/definitions/' +
+      appDefinitionProduction.sys.id +
+      '/general'
+  );
+
+  // UNDO - comment out for testing
+  try {
```
ryunsong-contentful: why do you have a try block nested inside another try block

You can probably separate the try-catch blocks separately with one following after the other is complete

scripts/actions/addTeamToSpace.ts
harikakondur: do you want specific roles for different members on the team?

scripts/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Before running the script, update these variables in `scripts/scripts.js`:
+
+```javascript
+const accessToken = 'YOUR_CONTENTFUL_ACCESS_TOKEN';
+const organizationId = 'YOUR_ORGANIZATION_ID';
+const appName = 'Your App Name';
+const teamId = 'YOUR_TEAM_ID';
+const environmentId = 'master';
+```
```
primeinteger: Should this be updated? It looks like you changed these over to using env vars

scripts/.env-example
primeinteger, line 1: This file is empty. Could you add the env vars that are used for an example?

scripts/README.md
```diff
@@ -0,0 +1,86 @@
...
+   ```bash
+   npm install contentful-management
+   ```
+
+3. **Configure the script** with your values (see Configuration section above)
+
+4. **Run the script:**
+   ```bash
+   node scripts.js
+   ```
```
primeinteger: Is this still the right way to run the script, or should this be updated now that the scripts are written in typescript?

scripts/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+   ```bash
+   cd scripts
+   ```
+
+2. **Install dependencies** (if not already installed):
+
+   ```bash
+   npm install contentful-management
+   ```
```
primeinteger: [nit] It might be better to create a package.json file in the scripts directory that includes this dependency, so that you only need to run `npm install` in the scripts folder, instead of installing dependencies individually.

Having a package.json file could also be useful for the "scripts" section, which would allow you to define different "npm run" commands such as `npm run setup` or `npm run teardown` (as examples)

package.json
```diff
@@ -31,6 +33,8 @@
     "sentry-release": "SENTRY_RELEASE=$(git rev-parse --short HEAD); sentry-cli releases --log-level=debug new -p marketplace-apps $SENTRY_RELEASE && sentry-cli releases --log-level=debug set-commits --auto --ignore-missing $SENTRY_RELEASE && sentry-cli releases --log-level=debug finalize $SENTRY_RELEASE && sentry-cli releases --log-level=debug deploys $SENTRY_RELEASE new -e production"
   },
   "dependencies": {
+    "contentful-management": "^11.52.0",
+    "dotenv": "^16.0.0",
```
primeinteger: If these are dependencies intended for the scripts folder, adding them here at the top-level project package.json probably isn't the right place. These should probably be added to `scripts/package.json` instead

scripts/README.md
primeinteger: I noticed there is no `tsconfig.json` file, is that intentional?

primeinteger (approved):
👏

PR #9952, opened 2025-07-02
Hubspot: Setup custom content type for configuration [INTEG-2784]

FBanfi:
## Purpose

We needed to store which fields are being synced to Hubspot. For that we create a custom content type with a single json field and one single entry to store that information. This is setup when configuring the app.

## Approach

When we install the app, we create the config content type and entry. We catch and ignore the "already existing" content type or entry error, but the other kind of errors are thrown and we make the installation fail. By doing this, we are able to do one less call since we can avoid making the get (to check if it exists or not).

## Testing steps

Automated test were added. To validate this manually you should delete the config entry and content type (if existent) and save or install the app, you will see that the config entry and content type are created again.

https://github.com/user-attachments/assets/e5a20a37-88dc-4e80-9025-9b86232f6252

## Breaking Changes

N/A

N/A

## Deployment

N/A

### DIFF
--- a/apps/hubspot/src/locations/ConfigScreen.tsx
+++ b/apps/hubspot/src/locations/ConfigScreen.tsx
@@ -1,18 +1,18 @@
-import { useState, useEffect, useCallback } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import {
   Box,
+  Collapse,
   Flex,
-  Heading,
-  Text,
   Form,
   FormControl,
-  TextInput,
-  Paragraph,
-  Collapse,
+  Heading,
   IconButton,
-  Subheading,
-  TextLink,
   List,
+  Paragraph,
+  Subheading,
+  Text,
+  TextInput,
+  TextLink,
 } from '@contentful/f36-components';
 import { ChevronDownIcon, ChevronUpIcon, ExternalLinkIcon } from '@contentful/f36-icons';
 import { useSDK } from '@contentful/react-apps-toolkit';
@@ -23,6 +23,7 @@ import {
   AppInstallationParameters,
   CONFIG_SCREEN_INSTRUCTIONS,
   ContentType,
+  createConfig,
   HUBSPOT_PRIVATE_APPS_URL,
 } from '../utils/utils';
 import { createClient } from 'contentful-management';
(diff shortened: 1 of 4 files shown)


apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -103,6 +111,14 @@ const ConfigScreen = () => {
       return false;
     }
 
+    try {
+      await createConfig(cma);
+    } catch (e) {
+      console.error(e);
+      sdk.notifier.error('Error creating config entry');
```
joaquincasal: This message is probably not very clear for users...I don't really know what's a good name for "config entry". Maybe we could ask in Slack to see if anyone comes up a good name? I'd call it "Configuration entry" at least I think

apps/hubspot/src/utils/utils.ts
```diff
@@ -17,4 +24,64 @@ export type AppInstallationParameters = {
   hubspotAccessToken: string;
 };
 
-export const CONFIG_CONTENT_TYPE_ID = 'hubspotConfig';
+export type SdkField = {
```
joaquincasal: This should be in the processingFields file already 🤔

apps/hubspot/src/utils/utils.ts
```diff
@@ -17,4 +24,64 @@ export type AppInstallationParameters = {
...
+
+export async function createConfig(cma: PlainClientAPI) {
+  await createContentType(cma);
+  await createEntry(cma);
+}
+
+export async function createContentType(cma: PlainClientAPI) {
+  const contentTypeBody = {
+    name: CONFIG_CONTENT_TYPE_ID,
+    description: 'Content Type used by the Braze app. Do not delete or modify manually.',
```
joaquincasal:
```suggestion
    description: 'Content Type used by the Hubspot app. Do not delete or modify manually.',
```
😛

apps/hubspot/src/locations/ConfigScreen.tsx
FBanfi: Sure, I'll check with them. For now I'll leave the error as 'Error creating configuration entry' and probably leave this change to the next PR with the service refactor.

apps/hubspot/src/utils/utils.ts
FBanfi: Yes it does 😄 .
Removed ✅

FBanfi:
> Left some minor comments! I was thinking maybe we can include some helper methods to fetch the config entry and also update it (maybe we could try the 'Service' idea that Juli had). That way we have everything we'll need to interact with the config entry. But that can also be done in a separate PR, as this one makes sense on it's own.

This other changes and refactors will be handled in another PR

joaquincasal (approved):
🚀

PR #9953, opened 2025-07-02
Hubspot: adding config entry service [INTEG-2785]

FBanfi:
## Purpose

This update includes a refactor to have a configuration entry service so we can fetch, update or create the config entry in a simpler way and more reusable.

## Approach

A ConfigEntryService was created and we now use that service in the config screen to create the config entry when the app is installed.

The methods of the service:
- getConfig
- updateConfig

Are not being used for the moment but are going to be needed later so they were also included to simplify the following tickets implementations.

## Testing steps

This was the use case:

https://github.com/user-attachments/assets/ac8e2848-5393-4a93-bf76-043f7dcd68f2

## Breaking Changes

N/A

This PR is a refactor for this ticket: [App Events - Setup custom content type for configuration](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?issueParent=367108&label=10-Pines&selectedIssue=INTEG-2784)

## Deployment

N/A

### DIFF
--- a/apps/hubspot/src/locations/ConfigScreen.tsx
+++ b/apps/hubspot/src/locations/ConfigScreen.tsx
@@ -112,7 +112,7 @@ const ConfigScreen = () => {
     }
 
     try {
-      const configService = new ConfigEntryService(cma);
+      const configService = new ConfigEntryService(cma, sdk.locales.default);
       await configService.createConfig();
     } catch (e) {
       sdk.notifier.error('The app configuration was not saved. Please try again.');
--- a/apps/hubspot/src/locations/ConfigScreen.tsx
+++ b/apps/hubspot/src/locations/ConfigScreen.tsx
@@ -23,11 +23,11 @@ import {
   AppInstallationParameters,
   CONFIG_SCREEN_INSTRUCTIONS,
   ContentType,
-  createConfig,
   HUBSPOT_PRIVATE_APPS_URL,
 } from '../utils/utils';
 import { createClient } from 'contentful-management';
 import ContentTypeMultiSelect from '../components/ContentTypeMultiSelect';
+import ConfigEntryService from '../utils/ConfigEntryService';
 
 export const EMPTY_MESSAGE = 'Some fields are missing';
 
@@ -112,10 +112,10 @@ const ConfigScreen = () => {
     }
 
     try {
-      await createConfig(cma);
+      const configService = new ConfigEntryService(cma);
+      await configService.createConfig();
     } catch (e) {
-      console.error(e);
-      sdk.notifier.error('Error creating configuration entry');
+      sdk.notifier.error('The app configuration was not saved. Please try again.');
       return false;
     }
 
--- a/apps/hubspot/src/utils/ConfigEntryService.ts
+++ b/apps/hubspot/src/utils/ConfigEntryService.ts
@@ -1,0 +1,86 @@
+import { EntryProps, KeyValueMap, PlainClientAPI } from 'contentful-management';
+import {
+  CONFIG_CONTENT_TYPE_ID,
+  CONFIG_ENTRY_ID,
+  CONFIG_FIELD_ID,
+  ConnectedFields,
+  getDefaultLocale,
+} from './utils';
+
+class ConfigEntryService {
+  private cma: PlainClientAPI;
+
+  constructor(cma: PlainClientAPI) {
+    this.cma = cma;
+  }
+
+  async createConfig() {
+    await this.createContentType();
+    await this.createEntry();
+  }
+
+  async createContentType() {
+    const contentTypeBody = {
+      name: CONFIG_CONTENT_TYPE_ID,
+      description: 'Content Type used by the Hubspot app. Do not delete or modify manually.',
+      fields: [
+        {
+          id: CONFIG_FIELD_ID,
+          name: CONFIG_FIELD_ID,
+          required: false,
+          localized: false,
+          type: 'Object',
+        },
+      ],
+    };
+    try {
+      const contentTypeProps = await this.cma.contentType.createWithId(
+        { contentTypeId: CONFIG_CONTENT_TYPE_ID },
+        contentTypeBody
+      );
+      await this.cma.contentType.publish(
+        { contentTypeId: CONFIG_CONTENT_TYPE_ID },
+        contentTypeProps
+      );
+    } catch (e: any) {
+      // Only ignore error if content type already exists
+      if (e?.code !== 'VersionMismatch') {
+        throw e;
+      }
+    }
+  }
+
+  async createEntry() {
+    try {
+      await this.cma.entry.createWithId(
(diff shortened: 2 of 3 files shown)


apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -112,10 +112,10 @@ const ConfigScreen = () => {
     }
 
     try {
-      await createConfig(cma);
+      const configService = new ConfigEntryService(cma);
+      await configService.createConfig();
     } catch (e) {
-      console.error(e);
-      sdk.notifier.error('Error creating configuration entry');
+      sdk.notifier.error('The app configuration was not saved. Please try again.');
```
FBanfi, line 118: This was a pending change from the previous [PR of this ticket](https://github.com/contentful/apps/pull/9952#discussion_r2180199346).

After talking through slack this is what they would like to show in this case, more context below:
> Generally speaking, being specific with errors is optimal, but in this case there's nothing the user can do other than try again, so I'm thinking we can stick with binary options: Worked / didn't work.

apps/hubspot/src/utils/ConfigEntryService.ts
JuliRossi, line 1: 💭 Should we test this new service?

JuliRossi (approved):
Really liked the refactor! Left a comment regarding tests, if you feel we are already covered with what we have, you can let it be.

apps/hubspot/src/utils/utils.ts
```diff
@@ -23,49 +23,27 @@ export type AppInstallationParameters = {
...
 };
 
-export async function createConfig(cma: PlainClientAPI) {
-  await createContentType(cma);
-  await createEntry(cma);
-}
+export type ConnectedField = {
+  fieldId: string;
+  locale: string;
+  contentBlockId: string;
```
joaquincasal:
```suggestion
  moduleId: string;
```

apps/hubspot/src/utils/utils.ts
```diff
@@ -23,49 +23,27 @@ export type AppInstallationParameters = {
   hubspotAccessToken: string;
 };
 
-export async function createConfig(cma: PlainClientAPI) {
-  await createContentType(cma);
-  await createEntry(cma);
-}
+export type ConnectedField = {
```
joaquincasal, line 24: I think we'll need a "updatedAt" property as well, to display in the sidebar

apps/hubspot/src/utils/ConfigEntryService.ts
```diff
@@ -0,0 +1,86 @@
...
+  async updateConfig(
+    configEntry: EntryProps<KeyValueMap>,
+    connectedFields: ConnectedFields,
+    cma: PlainClientAPI,
+    defaultLocale?: string
+  ) {
+    if (!configEntry.fields[CONFIG_FIELD_ID]) {
+      configEntry.fields[CONFIG_FIELD_ID] = {};
+    }
+    defaultLocale ||= await getDefaultLocale(cma);
```
joaquincasal: I'm almost sure we'll be able to require the defaultLocale, we can make it a required param and then if we need it to be optoinal we can change it

apps/hubspot/src/utils/ConfigEntryService.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }
+  }
+
+  async getConfigEntry(): Promise<EntryProps<KeyValueMap>> {
+    return this.cma.entry.get({ entryId: CONFIG_ENTRY_ID });
+  }
+
+  async updateConfig(
+    configEntry: EntryProps<KeyValueMap>,
```
joaquincasal: I think it'd be better if we don't receive this as a param, so this method is easier to use. Maybe we can store it as a property of the class in the `getConfigEntry` and fetch it here only if the propert is not set

joaquincasal (commented):
Left some comments! Let me know if you want to discuss any of them

apps/hubspot/src/utils/ConfigEntryService.ts
```diff
@@ -0,0 +1,86 @@
...
+  async createConfig() {
+    await this.createContentType();
+    await this.createEntry();
+  }
+
+  async createContentType() {
+    const contentTypeBody = {
+      name: CONFIG_CONTENT_TYPE_ID,
+      description: 'Content Type used by the Hubspot app. Do not delete or modify manually.',
+      fields: [
```
joaquincasal, line 45: I was thinking maybe we can add a text field that acts as the title of the entry so it's not called "Untitled"

apps/hubspot/src/utils/ConfigEntryService.ts
```diff
@@ -0,0 +1,95 @@
...
+  private cma: PlainClientAPI;
+  private configEntry?: EntryProps<KeyValueMap>;
+
+  constructor(cma: PlainClientAPI) {
+    this.cma = cma;
+  }
+
+  async createConfig() {
+    await this.createContentType();
+    await this.createEntry();
```
joaquincasal, line 16: I think these two could be private

apps/hubspot/src/utils/ConfigEntryService.ts
FBanfi: Ohh yes, I really like this change!

Done ✅

apps/hubspot/src/utils/utils.ts
FBanfi: Good call ✅

apps/hubspot/src/utils/ConfigEntryService.ts
FBanfi: Ok 👌 
Changed ✅

apps/hubspot/src/utils/ConfigEntryService.ts
FBanfi: Now the entry is created with this name:
Captura de pantalla 2025-07-03 a la(s) 11 34 57 a  m

apps/hubspot/src/utils/ConfigEntryService.ts
FBanfi: Done ✅

apps/hubspot/src/utils/ConfigEntryService.ts
FBanfi: At the end I decided that we could have a variable in the ConfigEntryService class, to define the default locale, and initialize it when the service is constructed.
I think it is a good approach since it is very simple to get the default locale through the cms and pass it to the constructor.
For the moment I left it this way and used it to create the text field that displays the title of the entry. But let me know if you have an opposite opinion about this!

apps/hubspot/src/utils/ConfigEntryService.ts
FBanfi: For the history of this comment: We discussed this later in a meeting and concluded that we won't test the service since it would be used and tested in some other locations later.

PR #9956, opened 2025-07-02
Hubspot: Allow to change modules name [INTEG-2775]

joaquincasal:
## Purpose

Add a second step to the Dialog to be able to change the module names in Hubspot

## Approach

Each field gets a default module name from the entry title and the field name that can be updated in the second step of the Dialog.
States are kept in the Dialog component and not in the specific step component to keep the step component simple and without much behavior.

## Testing steps

Open the Dialog from the Sidebar, select some fields, click next and update the module name for one of them, click save and check that the modules have the expected name in Hubspot.

https://github.com/user-attachments/assets/2ebf9295-b3f9-4e10-ac73-c2645cd18ace

Notifier:

https://github.com/user-attachments/assets/c62f2110-507f-46b1-897c-35ed28ecb94f


apps/hubspot/src/locations/Dialog.tsx
```diff
@@ -24,14 +31,37 @@ const Dialog = () => {
...
+  );
 
-  const callAction = async () => {
-    const fieldsToSend = selectedFields.map((field) => {
-      return fields.find((f) => f.uniqueId === field);
+  const handleNext = () => {
+    const initialNameMapping: { [fieldId: string]: string } = {};
+    selectedFieldObjects.forEach((field) => {
+      initialNameMapping[field.uniqueId] =
+        moduleNameMapping[field.uniqueId] ?? `${entryTitle} - ${field.name}`;
```
JuliRossi: 💭 do we have restrictions for symbols or spaces for module names?

joaquincasal: That's a good question, it looks like it only accepts letters, numbers, hyphens, underscores and whitespaces. I'll check how to handle that

apps/hubspot/src/components/FieldModuleNameMapping.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Text as="p" marginBottom="spacingM">
+        Optionally, name the Hubspot custom modules that will be synced to entry field content.
+      </Text>
+      {selectedFields.map((field) => (
+        <Box
+          key={field.uniqueId}
+          marginBottom="spacingM"
+          padding="spacingM"
+          style={{
+            border: `1px solid ${tokens.gray300}`,
```
FBanfi: 💭 I was thinking that perhaps we could move this style and the one from below to a separated styles file, same way as you did for the dialog styles. But perhaps you didn't because this component is not a location.

FBanfi (approved):
💯

apps/hubspot/src/components/FieldModuleNameMapping.tsx
joaquincasal: I didn't do it because the styles are only used in one place, so the styles files wouldn't provide reusability. But it's not a strong opinion so I'm open to adding it

FBanfi: Yes, the main advantage I saw was that we could encapsulate all the style logic in a single file, even though the styles themselves aren’t reused. In other style files, we usually don’t reuse styles either, we just separate the styling logic from the component or location for clarity.

But I don’t have a strong opinion neither, I just wanted to bring it up for discussion and hear your thoughts for future cases 😛 . In this case I since there is very little styling logic, I agree that's better to leave it that way.

FBanfi (approved):
🚀

PR #9962, opened 2025-07-04
Hubspot: validations update []

joaquincasal:
## Purpose

We learned we cannot functions until an app is installed, so we moved the token validations we were doing in the Config Screen to the Dialog.

## Approach

Instead of validating the token, we now catch the specific invalid_authentication or missing_scopes errors when trying to create modules. If those errors happen, we show an appropiate error message in the notification.

## Testing steps

Set an invalid token in the config screen (e.g. 'asdf') and attempt to create modules from the sidebar. Then add a valid token to the sidebar but remove the 'content' scope from Husbpot configuration. Both cases should fail and show a different error message.

https://github.com/user-attachments/assets/4d4b8644-9be4-44e2-b6ed-8e975fdf68db


apps/hubspot/src/assets/hubspot.mp4
JuliRossi, line 1: 💭  Did we update the video? Or move it?

joaquincasal: Yep, it got updated to include the scopes

FBanfi (approved):
Nice catch and refactor! 💯

PR #9963, opened 2025-07-04
Hubspot: add sync info to sidebar [INTEG-2773]

joaquincasal:
## Purpose

Add sync information in the Sidebar.

## Approach

- If there are errors in synced fields, we display an error note.
- If there are synced fields, display how long ago they were sinced.

## Testing steps

- Update the config entry with one connected field (make sure to include the `updatedAt` property), go to the entry and check the "Synced x time ago" appears
- Update the config entry with one connected field that includes the "error" property and check the warning appears in the sidebar

Screenshot 2025-07-04 at 16 26 32

### DIFF
--- a/apps/hubspot/src/locations/Sidebar.tsx
+++ b/apps/hubspot/src/locations/Sidebar.tsx
@@ -1,8 +1,11 @@
-import { Button, Flex } from '@contentful/f36-components';
+import { Button, Flex, Text, RelativeDateTime } from '@contentful/f36-components';
 import { SidebarAppSDK } from '@contentful/app-sdk';
 import { useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
 import { processFields } from '../utils/fieldsProcessing';
 import { createClient } from 'contentful-management';
+import { useEffect, useState } from 'react';
+import ConfigEntryService from '../utils/ConfigEntryService';
+import { EntryConnectedFields } from '../utils/utils';
 
 const Sidebar = () => {
   const sdk = useSDK<SidebarAppSDK>();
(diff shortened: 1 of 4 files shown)


apps/hubspot/src/locations/Sidebar.tsx
```diff
@@ -39,14 +43,51 @@ const Sidebar = () => {
...
     <Flex gap="spacingM" flexDirection="column">
-      <Button
-        variant="secondary"
-        isFullWidth={true}
-        onClick={async () => sdk.dialogs.openCurrentApp(await dialogParams())}>
-        Sync entry fields to Hubspot
-      </Button>
+      {connectedFields &&
+        connectedFields.length > 0 &&
+        connectedFields.some((field) => field.error) && (
```
FBanfi:
```suggestion
      {connectedFields?.some(field => field.error) && (
```
💭 perhaps we could replace this conditions for this other version. It should have the same effect since `.some` returns false if the array is empty.

apps/hubspot/src/locations/Sidebar.tsx
```diff
@@ -39,14 +43,51 @@ const Sidebar = () => {
...
+            </Text>
+          </Note>
+        )}
+
+      <Flex gap="spacingXs" flexDirection="column">
+        <Button
+          variant="secondary"
+          isFullWidth={true}
+          onClick={async () => sdk.dialogs.openCurrentApp(await dialogParams())}>
+          Sync entry fields to Hubspot
```
FBanfi, line 79: 💭 Do we need to disable this button if there is an error?

FBanfi (commented):
Nice one! 🚀 

Left some simple comments

apps/hubspot/src/locations/Sidebar.tsx
joaquincasal: Hmm I guess it would depend on the error. Maybe there was a sync error but creating new fields is still possible, so I think we should let this button enabled and if there's an error they'll see it when attempting to create the new fields

FBanfi: Roger that 🤝😁

PR #9964, opened 2025-07-07
Iterable: Sidebar [INTEG-2842]

JuliRossi:
## Purpose
Add the sidebar part to Iterable 
Out of scope: Add link with information, as we need the tech doc

## Approach

- Add sidebar
- Add error handling
- Fix some whitespace error in the config screen (when the api key had spaces at the end or beginning, it affected the link creation)
-  Updated error handling in config screen to make it looks like the design
## Testing steps
Sidebar

Captura de pantalla 2025-07-07 a la(s) 12 24 11 p  m

Using it to create a data feed with a 50 field content type

https://github.com/user-attachments/assets/d95f1886-660f-4357-b293-864906b2a97b

Using data feed to create email templates

https://github.com/user-attachments/assets/1b134ee0-1200-4756-9f4b-d2c09b781fa6

Using a reference

https://github.com/user-attachments/assets/354f0c9b-ac6d-4ec5-9d0f-496870040711

### DIFF
--- a/apps/iterable/src/locations/Sidebar.tsx
+++ b/apps/iterable/src/locations/Sidebar.tsx
@@ -1,27 +1,23 @@
 import { CopyButton, Flex, Paragraph, TextInput, TextLink, Text } from '@contentful/f36-components';
 import { SidebarAppSDK } from '@contentful/app-sdk';
-import { /* useCMA, */ useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
-import { CopyIcon, ExternalLinkIcon } from '@contentful/f36-icons';
+import { useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
+import { ExternalLinkIcon } from '@contentful/f36-icons';
 import tokens from '@contentful/f36-tokens';
 
 const Sidebar = () => {
   const sdk = useSDK<SidebarAppSDK>();
   useAutoResizer();
-  /*
-     To use the cma, inject it as follows.
-     If it is not needed, you can remove the next line.
-  */
-  // const cma = useCMA();
 
-  const value = 'a link';
+  const link = `https://cdn.contentful.com/spaces/${sdk.ids.space}/environments/master/entries?access_token=${sdk.parameters.installation.contentfulApiKey}&sys.id=${sdk.ids.entry}&include=2`;
+
   return (
     <Flex flexDirection="column">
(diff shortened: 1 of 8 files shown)


apps/iterable/src/locations/Sidebar.tsx
```diff
@@ -1,16 +1,65 @@
...
 
-  return <Paragraph>Hello Sidebar Component (AppId: {sdk.ids.app})</Paragraph>;
+  const space = sdk.ids.space;
+  const apiKey = sdk.parameters.installation.contentfulApiKey;
+  const entryId = sdk.ids.entry;
+
+  const hasError = !space || !apiKey || !entryId;
+  const link = hasError
+    ? ''
+    : `https://cdn.contentful.com/spaces/${space}/environments/master/entries?access_token=${apiKey}&sys.id=${entryId}&include=2`;
```
joaquincasal, line 28: Is there a reason to set the include param to 2? The maximum value is 10

joaquincasal: I'm also a little hesitant about using the /entries endpoint because then you need to do `items[0]` to access the entry data. But we can go ahead with this PR and then discuss which endpoint to use

apps/iterable/test/mocks/locations/Sidebar.spec.tsx
```diff
@@ -1,18 +1,67 @@
...
+  ids: {
+    space: 'space-id',
+    entry: 'entry-id',
+  },
+  parameters: {
+    installation: {
+      contentfulApiKey: 'api-key',
+    },
+  },
+};
```
FBanfi: 💭 Could we have this properties in the existent `mockSdk.ts` file and then just import that file?

FBanfi (approved):
Great update! 🚀

joaquincasal (approved):
🚀

apps/iterable/src/locations/Sidebar.tsx
JuliRossi: We had[ this conversation about this on slack](https://contentful.slack.com/archives/C08V6QFHKJ7/p1751396708785459). From my understanding there isn't a technical limitation. Tested it with an entry within an entry(10 times) and Iterable allows it. My concern is ordering as it isn't doing what at least I would expect. I'll ask more about it tomorrow.  Captura de pantalla 2025-07-07 a la(s) 3 23 22 p  m

PR #9965, opened 2025-07-07
Trim whitespaces for access token in Config Screen []

JuliRossi:
## Purpose

Found this can happen in Iterable

## Approach
- Trim leading and trailing whitespaces

## Tests 

First add white spaces => Add trim => User no longer can leave white spaces as lead or trail:

https://github.com/user-attachments/assets/cc8f6825-a015-416f-bf6b-95e21d14a48c


FBanfi (approved):
Brilliant 😁

PR #9966, opened 2025-07-07
Hubspot: handle other events [INTEG-2786]

FBanfi:
## Purpose

- We needed to update the configuration entry if a synced entry got deleted: the entry should be removed from the configuration
- ~If the app gets uninstalled, the custom content type and entry should be deleted~
  - This won't be contemplated in this PR since it doesn't seem to be possible to trigger a function on an uninstall event. We will look further and try to add it in another PR.
- Also this update includes tests for the ConfigEntryService logic.

## Approach

A function was created to use it in the app definition as an event.

## Testing steps

Automated tests were added. To check this manually you could add an entry in the `hubspotConfig` json and then delete the entry connected in the json:  it will delete the entry from the json

https://github.com/user-attachments/assets/39cc9b8f-3e09-4ed5-818b-98411f0793c7

## Breaking Changes

N/A

Link to [INTEG-2786](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?issueParent=367108&label=10-Pines&selectedIssue=INTEG-2786) ticket.

## Deployment

N/A

[INTEG-2786]: https://contentful.atlassian.net/browse/INTEG-2786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/hubspot/functions/appEventHandler.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const configService = new ConfigEntryService(cma);
+  let connectedFields: ConnectedFields;
+  let entryConnectedFields: EntryConnectedFields;
+  try {
+    connectedFields = await configService.getConnectedFields();
+    entryConnectedFields = await configService.getEntryConnectedFields(entryId);
+  } catch (error) {
+    return;
+  }
```
joaquincasal: If this fails we probably want the function to throw the error instead of failing silently. And since this can throw, it may be useful to wrap it in the callAndRetry call (ideally we don't have two separate calls to callAndRetry and we find a way to wrap everything toghether)

apps/hubspot/src/utils/ConfigEntryService.ts
```diff
@@ -23,25 +29,30 @@ class ConfigEntryService {
...
 
     return updatedEntry;
   }
 
+  async getConnectedFields(): Promise<ConnectedFields> {
+    const configEntry = await this.getConfigEntry();
+    return configEntry.fields[CONFIG_FIELD_ID][this.getDefaultLocale()];
+  }
+
+  async getEntryConnectedFields(entryId: string): Promise<EntryConnectedFields> {
```
joaquincasal, line 51: Now that we're introducing this method we could replace the usage of `getConnectedFields` in the sidebar

apps/hubspot/src/utils/ConfigEntryService.ts
```diff
@@ -1,14 +1,20 @@
...
 class ConfigEntryService {
   private cma: PlainClientAPI;
   private configEntry?: EntryProps<KeyValueMap>;
   private defaultLocale: string;
 
-  constructor(cma: PlainClientAPI, defaultLocale: string) {
+  constructor(cma: PlainClientAPI, defaultLocale?: string) {
     this.cma = cma;
-    this.defaultLocale = defaultLocale;
+    this.defaultLocale = defaultLocale || 'en-US';
```
joaquincasal: I had forgotten about the use case in which we need to call this service from an app event function 😅 there's no way of getting the default locale without the cma in that case and we now need the defaultLocale to be optional.
We could bring back the function to obtain the default locale through the cma so we don't have to hardcode the 'en-US' locale

joaquincasal (commented):
Left some comments!

apps/hubspot/src/utils/utils.ts
```diff
@@ -37,3 +40,18 @@ export type EntryConnectedFields = ConnectedField[];
 export type ConnectedFields = {
   [entryId: string]: EntryConnectedFields;
 };
+
+export function initContentfulManagementClient(context: FunctionEventContext): PlainClientAPI {
```
joaquincasal: I think I'd put this in the functions folder (maybe a common.ts file) because this can only be used in the context of functions

FBanfi:
> Left some comments!

Implemented! Thanks 💯

apps/hubspot/src/locations/Sidebar.tsx
```diff
@@ -38,7 +38,11 @@ const Sidebar = () => {
       title: 'Sync entry fields to Hubspot',
       parameters: {
         entryTitle: getEntryTitle(),
-        fields: await processFields(Object.values(sdk.entry.fields), cma, sdk.locales.default),
+        fields: JSON.parse(
+          JSON.stringify(
```
joaquincasal, line 42: I think these two cancel each other out

joaquincasal, line 42: I think these two cancel each other out

apps/hubspot/src/utils/ConfigEntryService.ts
```diff
@@ -10,11 +10,9 @@ import {
 class ConfigEntryService {
   private cma: PlainClientAPI;
   private configEntry?: EntryProps<KeyValueMap>;
-  private defaultLocale: string;
```
joaquincasal, line 7: We could keep this property and set it when we fetch the locale the first time. Then if we need it again we can first check if it's already set so we avoid making additional requests

apps/hubspot/src/locations/Sidebar.tsx
FBanfi: I added this to fix a compiling type error I saw in the sidebar while doing the corrections of the PR. In this part of the code (`sdk.dialogs.openCurrentApp(await dialogParams())`) a compiling error was thrown because not all the data we were sending through `dialogParams` was serializable. In this case we needed this because:
- `JSON.stringify` converts a JavaScript value into a JSON string. This process removes any functions, undefined values, or non-serializable properties.
- `JSON.parse` then takes that string and turns it back into a plain JavaScript object/array, but now it is guaranteed to be fully serializable (only containing data that can be represented in JSON)

So they combined ensure that the data is serializable and that fixed the compiling type error.

apps/hubspot/src/utils/ConfigEntryService.ts
FBanfi: Yes, I missed that one! Thanks

Done ✅

apps/hubspot/src/locations/Sidebar.tsx
joaquincasal: Hmm everything in the dialogParams should be serializable. Let's merge it and then we can check that error in particular

FBanfi: Sure, I'll write this down for later!

PR #9967, opened 2025-07-08
Hubspot: Persist error in sidebar []

joaquincasal:
## Purpose

Give more details and a CTA if there's an error with the hubspot token when creating new modules.

## Approach

The Dialog returns a boolean indicating if there was an error related to the token so a note can be displayed in the sidebar.

## Testing steps

Paste an invalid token in the config screen and try to create new modules. An error should be displayed in the sidebar.

image

### DIFF
--- a/apps/hubspot/src/locations/Dialog.tsx
+++ b/apps/hubspot/src/locations/Dialog.tsx
@@ -100,25 +100,30 @@ const Dialog = () => {
     invalidToken: boolean,
     missingScopes: boolean
   ) => {
-    const successfullFields = `${success.length} entry field${success.length === 1 ? '' : 's'}`;
-    const failedFields = `${failed.length} entry field${failed.length === 1 ? '' : 's'}`;
+    const resultMessage = (fields: SelectedSdkField[]): string => {
+      return `${fields.length} entry field${fields.length === 1 ? '' : 's'}`;
+    };
+
+    const successMessage = `${resultMessage(success)} successfully synced.`;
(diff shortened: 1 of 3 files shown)


JuliRossi (approved):
👍 Looks good to me!

apps/hubspot/src/locations/Dialog.tsx
```diff
@@ -101,25 +100,25 @@ const Dialog = () => {
...
   ) => {
-    const successMessage = `${success.length} entry field${
-      success.length === 1 ? '' : 's'
-    } successfully synced`;
-    const failedMessage = `${failed.length} entry field${
-      failed.length === 1 ? '' : 's'
-    } did not sync, please try again`;
-    const mixedMessage = `${successMessage} but ${failedMessage}`;
+    const successfullFields = `${success.length} entry field${success.length === 1 ? '' : 's'}`;
+    const failedFields = `${failed.length} entry field${failed.length === 1 ? '' : 's'}`;
```
FBanfi: 💭 Nitpick 😛 : we could make a function outside the `handleSaveAndSync` to avoid this little repeated logic
```
    const resultMessage: string = (fields) => {
        return `${fields.length} entry field${fields.length === 1 ? '' : 's'}`;
    } 
```
```suggestion    
    const successfullFields = resultMessage(success);
    const failedFields = resultMessage(failed);
```

FBanfi (approved):
Great! 🚀  Added only a nitpick

PR #9971, opened 2025-07-09
feat: use new Forma 36 Layout component [UFO-1876]

veu:
## Purpose

We’re replacing the old Workbench component with the new Layout to prepare for the next Forma 36 release. For the Wistia and Salesforce apps this affects the config screens only while for GraphQL Playground it changes the main screen a bit but only for the better.

### Wistia config screen before / after

image

image

### Salesforce config screen before / after

image

image

### GraphQL Playground main view before / after

image

image

Look at all that space!

https://contentful.atlassian.net/browse/UFO-1876

## Deployment

The Wistia app config screen looks a bit broken because of [changes introduced last year](https://github.com/contentful/apps/pull/8942) that were never deployed. I’m not going to fix them here but this should be looked at before deploying the app again.


apps/graphql-playground-v2/package.json
```diff
@@ -5,8 +5,8 @@
   "dependencies": {
     "@contentful/app-sdk": "^4.23.1",
     "@contentful/f36-components": "^4.54.0",
+    "@contentful/f36-layout-alpha": "npm:@contentful/f36-layout@5.0.0-alpha.18",
```
Lelith: when we no longer have double-layouts included we can just import the layout component directly on its alpha version and don't need to name it alpha. less cleanup after releasing v5 :)

```suggestion
    "@contentful/f36-layout": "5.0.0-alpha.18",
```

Lelith (approved):
just one nit-pick about the way we import the layout package, to minimize future cleanup.

mgoudy91 (approved):

PR #9976, opened 2025-07-11
Iterable: Use single entry endpoint []

joaquincasal:
## Purpose

Change the link generation so it uses the single entry endpoint instead of the entries collection endpoint.
The single entry endpoint is simpler and the usage of the entries collection will be documented in the app documentation.


JuliRossi (approved):
👍

PR #9977, opened 2025-07-14
Hubspot: Fix long text issue [INTEG-2907]

JuliRossi:
## Purpose
There was a bug for long texts which didn't showcase the newlines in Hubspots modules

## Approach
We decided to use the rich text type for long texts.

## Testing steps

https://github.com/user-attachments/assets/1ef29a03-b5eb-45ee-b6ac-e1432e3ca6d1

### DIFF
--- a/apps/hubspot/functions/createModules.ts
+++ b/apps/hubspot/functions/createModules.ts
@@ -87,3 +87,117 @@ const createModule = async (field: SelectedSdkField, token: string) => {
   await createModuleFile(fieldsFile, 'fields.json', moduleName, token);
   await createModuleFile(moduleFile, 'module.html', moduleName, token);
 };
+
+const createModuleFile = async (
+  file: string,
+  fileName: string,
+  moduleName: string,
+  token: string
+) => {
+  const fileBuffer = Buffer.from(file);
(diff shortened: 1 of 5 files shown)


apps/hubspot/functions/createModules.ts
```diff
@@ -143,11 +143,18 @@ const getFiles = (field: SelectedSdkField): { fieldsFile: string; moduleFile: st
...
       if (field.value) fieldsFile[0].default = field.value;
       moduleFile = TEXT_MODULE_TEMPLATE;
       break;
+    case 'Text':
+      fieldsFile = structuredClone(RICH_TEXT_FIELD_TEMPLATE);
+      if (field.value) {
+        const lines: string[] = field.value.split('\n');
+        fieldsFile[0].default = lines.map((line) => `<p>${line}</p>`).join('');
+      }
+      moduleFile = RICH_TEXT_MODULE_TEMPLATE;
```
JuliRossi: if there's another way to handle these files, let me know!

joaquincasal: I was thinking about this and I don't think there's a better way to handle this

joaquincasal (approved):

PR #9990, opened 2025-07-18
Hubspot: Fix localization default names for modules issue [INTEG-2936]

JuliRossi:
## Purpose
Localize and non-localized fields have the same default name when creating a module. We want to change that.

## Approach

- Improve naming convention to include localization. 
- Every whitespace is now a hypen

## Testing steps

https://github.com/user-attachments/assets/ed2cffee-d1cd-4375-b6c7-6b26824fce8f


FBanfi (approved):
🚀

PR #9994, opened 2025-07-21
Iterable app [INTEG-2811]

JuliRossi:
## Purpose

We wanted to create an app for Iterable

## Approach
The following features were implemented:

### In the config screen
- Add Contentful API key
- Add content Types
- Link  to Iterable documentation

### In the sidebar
- Allows user to create a link to create Data Feeds in Iterable using the single entry endpoint from the CDA*
-  Link to Iterable docs

*If the user wants to access References o Assets, it should use the endpoint and technics specified in the Iterable docs
## Testing steps

https://github.com/user-attachments/assets/f96d6cdd-b859-4173-8374-4cf09afda872

## References

[Epic](https://contentful.atlassian.net/browse/INTEG-2811)


apps/iterable/src/components/ContentTypeMultiSelect.tsx
```diff
@@ -0,0 +1,143 @@
+import React, { useEffect, useState } from 'react';
+import { Box, Stack, Pill } from '@contentful/f36-components';
+import { Multiselect } from '@contentful/f36-multiselect';
+import { ContentType } from '../utils.ts';
```
ryunsong-contentful: `import { ContentType } from '../utils';`

apps/iterable/src/components/ContentTypeMultiSelect.tsx
```diff
@@ -0,0 +1,143 @@
...
+    let areMoreContentTypes = true;
+
+    while (areMoreContentTypes) {
+      const response = await cma.contentType.getMany({
+        spaceId: sdk.ids.space,
+        environmentId: sdk.ids.environment,
+        query: { skip, limit },
+      });
+      if (response.items) {
+        allContentTypes = allContentTypes.concat(response.items as ContentTypeProps[]);
```
ryunsong-contentful: Don't need the `as ContentTypeProps[]);`  since `allContentTypes` is already that type in `allContentTypes = allContentTypes.concat(response.items);`

apps/iterable/src/locations/ConfigScreen.tsx
```diff
@@ -1,67 +1,147 @@
...
 
   const onConfigure = useCallback(async () => {
-    // This method will be called when a user clicks on "Install"
-    // or "Save" in the configuration screen.
-    // for more details see https://www.contentful.com/developers/docs/extensibility/ui-extensions/sdk-reference/#register-an-app-configuration-hook
+    const isContentfulKeyValid = await validateContentfulApiKey(parameters.contentfulApiKey, sdk);
+    setContentfulApiKeyIsValid(isContentfulKeyValid);
+
+    if (!isContentfulKeyValid) {
+      sdk.notifier.error('Some fields are missing or invalid');
```
ryunsong-contentful: We have this specific check `isContentfulKeyValid` but the message is more generic. Is there a reason why we can't just make it more specific? I'm guessing this is just following the figma docs?

JuliRossi: We actually don't have the notifier in the figma. We could do something like what we are doing in Hubspot and show "The app configuration was not saved. Please try again. " However, the user knows that the API Key is wrong because we are showing the Helper text below the input like this:
Captura de pantalla 2025-07-21 a la(s) 2 05 14 p  m

ryunsong-contentful: Nice, that will do

PR #10000, opened 2025-07-22
Hubspot disconnect - [INTEG-2782]

JuliRossi:
## Purpose

Add disconnect functionality

## Approach

- Updated modal design (Consulted with Emily)
- Add disconnect functionality
- Remove unused types
- Add test

## Testing steps

https://github.com/user-attachments/assets/8b8769f7-79a7-43d5-8b7b-6856bdc5456b

### DIFF
--- a/apps/hubspot/src/components/ConnectedFieldsModal.tsx
+++ b/apps/hubspot/src/components/ConnectedFieldsModal.tsx
@@ -114,17 +114,17 @@ const ConnectedFieldsModal: React.FC<ConnectedFieldsModalProps> = ({
                 </Button>
               </Box>
               {fieldsWithErrors.length > 0 && (
-            <Note
-              variant="negative"
-              icon={<WarningOctagonIcon className={styles.warningIconNote} />}
-              className={styles.modalErrorBanner}>
-              <Text lineHeight="lineHeightCondensed" fontColor="gray800">
-                Unable to sync content. Review your connected fields or{' '}
-                <TextLink onClick={() => onAppConfigRedirect()}>app configuration</TextLink>.
-              </Text>
-            </Note>
-          )}
-          {selectedFields.size > 0 && (
+                <Note
+                  variant="negative"
+                  icon={<WarningOctagonIcon className={styles.warningIconNote} />}
+                  className={styles.modalErrorBanner}>
+                  <Text lineHeight="lineHeightCondensed" fontColor="gray800">
+                    Unable to sync content. Review your connected fields or{' '}
+                    <TextLink onClick={() => onAppConfigRedirect()}>app configuration</TextLink>.
+                  </Text>
+                </Note>
+              )}
+              {selectedFields.size > 0 && (
                 <Stack
                   flexDirection="column"
                   spacing="spacing2Xs"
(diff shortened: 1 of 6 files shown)


apps/hubspot/src/components/ConnectedFieldsModal.tsx
```diff
@@ -67,104 +70,102 @@ const ConnectedFieldsModal: React.FC<ConnectedFieldsModalProps> = ({
...
-                </Table.Cell>
-              </Table.Row>
-            </Table.Head>
-            <Table.Body>
-              {entryConnectedFields.map((field, index) => {
-                const uniqueFieldId = getUniqueFieldId(field.fieldId, field.locale);
-                return (
-                  <Table.Row key={`${uniqueFieldId}-${index}`}>
+      {() => (
+        <>
```
FBanfi, line 92: 💭 I think we could maintain the previous version without this bracket and fragmente here.

apps/hubspot/src/locations/Sidebar.tsx
```diff
@@ -5,7 +5,7 @@ import { processFields } from '../utils/fieldsProcessing';
 import { createClient } from 'contentful-management';
 import { useEffect, useState } from 'react';
 import ConfigEntryService from '../utils/ConfigEntryService';
-import { EntryConnectedFields } from '../utils/utils';
+import { AppInstallationParameters, EntryConnectedFields } from '../utils/utils';
```
FBanfi: 💭 Are we using this somewhere?

apps/hubspot/test/locations/Page.spec.tsx
```diff
@@ -54,39 +65,6 @@ describe('Page Location', () => {
...
-    });
-
-    render(<Page />);
-
-    await waitFor(() => {
-      expect(screen.getByText('The app cannot load content.')).toBeTruthy();
-      expect(screen.getByRole('link', { name: /app configuration/i })).toBeTruthy();
-    });
-  });
-
```
FBanfi: ⚔️ I don't think this tests should be removed since they are for the page location table

FBanfi (commented):
Nice update! Left some comments!

apps/hubspot/src/components/ConnectedFieldsModal.tsx
JuliRossi: This way it keeps up with the design we have. This is the correct use according to forma, that is why we were having issues with the modal header splitter styling. https://f36.contentful.com/components/modal

apps/hubspot/src/locations/Sidebar.tsx
JuliRossi: Nope we can remove them

apps/hubspot/test/locations/Page.spec.tsx
JuliRossi: Must have remove them while rebasing. Good catch!

apps/hubspot/src/components/ConnectedFieldsModal.tsx
FBanfi: Ohh nice catch, forget my comment then! 👍

FBanfi (approved):
🚀

PR #10013, opened 2025-07-23
Hubspot: Avoid duplicated module creation [INTEG-2935]

JuliRossi:
## Purpose

We want to avoid creating the same module for a field in an entry twice

## Approach

- Call the Config entry service to get Connected fields for that entry 
- Block selection from Field Selection
- Add tooltip to warn the user that field is already Connected 

## Testing steps

https://github.com/user-attachments/assets/8e94bc57-b172-41ea-b8de-1d47a5696447

### DIFF
--- a/apps/hubspot/src/components/FieldSelection.tsx
+++ b/apps/hubspot/src/components/FieldSelection.tsx
@@ -10,10 +10,11 @@ type FieldsSelectionProps = {
   fields: SdkField[];
   selectedFields: string[];
   setSelectedFields: React.Dispatch<React.SetStateAction<string[]>>;
+  connectedFieldIds: string[];
 };
 
 const FieldSelection = (props: FieldsSelectionProps) => {
-  const { fields, selectedFields, setSelectedFields } = props;
+  const { fields, selectedFields, setSelectedFields, connectedFieldIds } = props;
   const supportedFields = fields.filter((field) => field.supported);
 
   const allSelected = useMemo(() => {
@@ -48,6 +49,7 @@ const FieldSelection = (props: FieldsSelectionProps) => {
               key={field.uniqueId}
               selectedFields={selectedFields}
               setSelectedFields={setSelectedFields}
+              isConnected={connectedFieldIds.includes(field.uniqueId)}
             />
           );
         })}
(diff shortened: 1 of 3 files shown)


apps/hubspot/test/locations/Dialog.spec.tsx
```diff
@@ -35,28 +51,30 @@ describe('Dialog component', () => {
     render(<Dialog />);
 
     // Check that the select all checkbox text is displayed correctly
-    expect(screen.getByText('Select all fields (5)')).toBeInTheDocument();
+    waitFor(() => {
```
JuliRossi, line 54: Fix this as it was tossing a warning

apps/hubspot/src/components/FieldSelection.tsx
```diff
@@ -88,9 +101,14 @@ const FieldCheckbox = (props: FieldCheckboxProps) => {
         id={field.uniqueId}
         isChecked={checked}
         onChange={onFieldSelected}
-        isDisabled={!field.supported}>
-        <Tooltip isDisabled={field.supported} content="Syncing not supported" placement="right">
-          <Text fontColor={field.supported ? 'gray900' : 'gray500'} fontWeight="fontWeightMedium">
+        isDisabled={!field.supported || isConnected}>
+        <Tooltip
+          isDisabled={field.supported && !isConnected}
+          content={<span style={{ textWrap: 'nowrap' }}>{getTooltipContent()}</span>}
```
JuliRossi, line 107: Use span instead of Text to keep the tooltip's styling

apps/hubspot/src/locations/Dialog.tsx
```diff
@@ -40,6 +41,23 @@ const Dialog = () => {
...
+      );
+
+      const ids = entryConnectedFields.map((f) =>
+        f.locale ? `${f.fieldId}-${f.locale}` : f.fieldId
+      );
+
+      setConnectedFieldIds(ids);
+    };
+    fetchConnectedFields();
+  }, [invocationParams.entryId]);
```
joaquincasal: Nit: I think this is not needed as the id won't change but it probably won't hurt either

apps/hubspot/test/locations/Dialog.spec.tsx
```diff
@@ -35,28 +51,30 @@ describe('Dialog component', () => {
...
+      expect(screen.getByText('Select all fields (5)')).toBeInTheDocument();
 
-    // Check that individual field checkboxes display correctly
-    expect(screen.getByText('Title')).toBeInTheDocument();
-    expect(screen.getByText('(Short text)')).toBeInTheDocument();
+      // Check that individual field checkboxes display correctly
+      expect(screen.getByText('Title')).toBeInTheDocument();
```
joaquincasal, line 58: Just checking if these changes were intended

joaquincasal (approved):
🚀

apps/hubspot/test/locations/Dialog.spec.tsx
JuliRossi: Yes, they were tossing an act warning, so I wrap the expects in a waitFor

PR #9826, opened 2025-05-09
Mux app: new features

R-Delfino95:
## Purpose

We are currently working on integrating the latest features provided by Mux into the plugin.
This PR introduces some of the new functionalities, including:
- Video quality selection (related to PR #9750).
- Asset visibility controls.
- Captions support.

The PR is currently in draft status as we are still actively working on adding more features and fixing bugs.

## Approach

Based on PR #9750, we refactored the modal by breaking it down into smaller components and updating the UI to make it clearer and more intuitive.

Two new components were created to handle captions and asset visibility. The configuration selected in the modal is passed to index.tsx, which uses it to make the appropriate API calls to Mux.
- For asset visibility, users can choose between private, public, or both.
- For captions, users can select between two options: auto-generated or custom (uploading their own captions).

### Screenshots
WIP

## Testing steps

Create a content model with the Mux JSON Object. Then, create a new entry for that content type, and you’ll be able to test uploading assets while selecting the desired video quality, privacy and captions.

Another case to test is by providing a URL of an existing video, for example [this one](http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4). In this case, the modal should appear to allow the user to select the desired options.

If an Asset ID is provided instead, the modal **should not** appear, as this flow is intended for associating an existing Mux video.

## Breaking Changes

None.

## Dependencies and/or References

More information can be found [here](https://www.mux.com/docs).

## Deployment

None.

PR #10034, opened 2025-07-29
Mux: Fixed signed tokens [INTEG-00]

R-Delfino95:
## Purpose

When changing the visibility of an Asset between public and protected, the plugin would end up in a broken state where its content could no longer be viewed unless the page was refreshed.  
These changes prevent that issue. Additionally, we removed the use of Hosted App Actions for obtaining signed tokens and replaced it with a Function + Action setup.  
This was necessary to properly test the plugin in a development environment.

## Approach

We removed the use of Hosted App Actions and deleted all related logic. Instead, we introduced a new Function (`getSignedUrlTokens.ts`) to obtain the signed tokens.  
We also noticed issues when re-rendering the player while using a protected playback.  
To address this, the player is now always rendered but hidden via CSS. When it needs to be shown, we simply adjust its styles accordingly.

## Testing steps

The main issues were related to changes in the `playbackId` and how the Player responded to them.

You can test the following scenarios:
- Upload protected and public videos.
- Switch their visibility between public and protected.
- Click the **Resync** button to verify playback updates correctly without needing a page refresh.

## Breaking Changes

There are no breaking changes for end users.

However, there **are changes in the app deployment process**, which are detailed in the **Deployment** section.

None.

## Deployment

The Hosted App Action previously used to fetch tokens no longer exists and should be removed from the deployment process.

Deployment should follow the same steps defined prior to [this PR](https://github.com/contentful/apps/pull/9826), with the addition of the following changes:

- Add the new Function: `getSignedUrlTokens.ts`
- Create a new Action named `getSignedUrlTokens` that calls this Function. I left some screenshots.
  - It's important that the Action have that name because I used it to call it later in the frontend side.

image
image


apps/mux/functions/src/getSignedUrlTokens.ts
```diff
@@ -0,0 +1,78 @@
...
+    },
+  } = context;
+  const mux = new Mux({ tokenId: muxAccessTokenId, tokenSecret: muxAccessTokenSecret });
+
+  if (typeof muxSigningKeyId !== 'string' || typeof muxSigningKeyPrivate !== 'string') {
+    throw new TypeError('missing required mux signing key id or signing key private');
+  }
+
+  const signedTokens = await sign(mux, playbackId, muxSigningKeyId, muxSigningKeyPrivate);
+  console.log('Tokens:', signedTokens);
```
primeinteger: This console log may have been left in from development. Could you please remove this console log from the function before this gets merged? It would probably be better to not expose these tokens in the backend function logs, but let me know if you disagree.

R-Delfino95: You're right! I removed the console.log 👍 
Thanks for reviewing it

PR #10035, opened 2025-07-29
Hubspot app [INTEG-2766]

joaquincasal:
## Purpose

Seamlessly sync Contentful entry content to email campaigns in Hubspot. Map entry fields to custom email modules in Hubspot to automatically keep content consistent at scale. 

## Approach

The app allows to create custom modules in Hubspot that can be used in emails. The modules are kept in sync with Contentful's content.

https://github.com/user-attachments/assets/d279a01b-943b-4db1-8d36-86ad9a33295d


apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -1,67 +1,268 @@
...
+    }
+  );
+
+  function checkIfHasValue(
+    value: string,
+    setError: (error: string | null) => void,
+    errorMessage: string
+  ) {
+    const hasValue = !!value?.trim();
+    setError(hasValue ? null : errorMessage);
```
ryunsong-contentful, line 66: Checkers shouldn't set other params like this, violates SRP (single-responsbility-principal)

apps/hubspot/src/locations/ConfigScreen.tsx
```diff
@@ -1,67 +1,268 @@
...
+    setHubspotTokenError(null);
+
+    const hubspotTokenHasValue = checkIfHasValue(
+      parameters.hubspotAccessToken,
+      setHubspotTokenError,
+      EMPTY_MESSAGE
+    );
+
+    if (!hubspotTokenHasValue) {
+      sdk.notifier.error(EMPTY_MESSAGE);
```
ryunsong-contentful, line 80: Related to the above comment - this is now slightly awkward where one has to understand why `setHubspotTokenError` isn't called here (because it's somehow already handled in the checker).

apps/hubspot/functions/appEventHandler.ts
```diff
@@ -0,0 +1,140 @@
...
+      }
+    }
+  }
+  throw lastError;
+}
+
+async function deleteConfigEntry(cma: PlainClientAPI) {
+  try {
+    await cma.entry.unpublish({ entryId: CONFIG_ENTRY_ID });
+  } catch (e) {}
```
ryunsong-contentful, line 130: Should we at least not log the errors if these things fail to at least make debugging easier

apps/hubspot/src/components/Splitter.tsx
```diff
@@ -0,0 +1,26 @@
+import { Box, CommonProps, MarginProps, PaddingProps } from '@contentful/f36-components';
+import { cx, css } from 'emotion';
+import tokens from '@contentful/f36-tokens';
+
+interface SplitterProps extends CommonProps, MarginProps, PaddingProps {}
+
+const Splitter = (props: SplitterProps) => {
```
ryunsong-contentful, line 7: Add this to the discussion with Sarah about adding this to the component library

apps/hubspot/src/locations/Page.tsx
```diff
@@ -1,16 +1,196 @@
...
+  const [entriesWithContentType, setEntriesWithContentType] = useState<EntryWithContentType[]>([]);
+  const [loading, setLoading] = useState<boolean>(true);
+  const [error, setError] = useState<string | null>(null);
+  const [connectedFields, setConnectedFields] = useState<ConnectedFields>({});
+  const defaultLocale = sdk.locales.default;
+  const [modalEntry, setModalEntry] = useState<EntryWithContentType | null>(null);
+  const [modalOpen, setModalOpen] = useState<boolean>(false);
 
-  return <Paragraph>Hello Page Component (AppId: {sdk.ids.app})</Paragraph>;
+  const cma = createClient(
```
ryunsong-contentful, line 22: Does this need to be created on every re-render? Could insert a `useMemo` here. I think I saw something similar in ConfigScreen.tsx

apps/hubspot/src/locations/Page.tsx
```diff
@@ -1,16 +1,196 @@
...
+        const entriesResponse = await cma.entry.getMany({ query: { 'sys.id[in]': entryIds } });
+        const fetchEntriesWithContentType = await Promise.all(
+          entriesResponse.items.map(async (entry) => {
+            try {
+              const contentType = await cma.contentType.get({
+                contentTypeId: entry.sys.contentType.sys.id,
+              });
+              return { entry, contentType };
+            } catch (err) {
+              return null;
```
ryunsong-contentful, line 57: We can probably do something better here no?

apps/hubspot/functions/appEventHandler.ts
joaquincasal: Some of these calls are expected to fail in some cases (when the entry is not published). And this runs when the app gets uninstalled, so logs would be lost immediately after

PR #10040, opened 2025-07-31
Hubspot improvements []

joaquincasal:
## Purpose

Address PR comments for the Hubspot app

## Approach

- Config screen: set the error outside of the `checkIfHasValue` function
- Replace CMA usages for `sdk.cma` instead of creating it with `createClient`
- Remove an unnecessary try/catch block in the page screen (as it was already surrounded by another try/catch)
- Add deploy script


ryunsong-contentful (approved):
Nice

PR #10041, opened 2025-07-31
Rich text setup project - [INTEG-3001]

JuliRossi:
## Purpose

First PR for setting up the project

## Approach
Add  the following:
- App locations
- Tests
- Dependencies
- Config files

## Testing steps

https://github.com/user-attachments/assets/1ef30063-8c87-43eb-9cfd-500847d07c81

### DIFF
--- a/apps/rich-text-versioning/package.json
+++ b/apps/rich-text-versioning/package.json
@@ -1,0 +1,38 @@
+{
+  "name": "rich-text-versioning",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@contentful/app-sdk": "4.36.0",
+    "@contentful/f36-components": "4.81.1",
+    "@contentful/f36-tokens": "4.2.0",
+    "@contentful/react-apps-toolkit": "1.2.16",
+    "contentful-management": "11.54.4",
+    "emotion": "10.0.0",
+    "happy-dom": "^18.0.1",
+    "react": "18.3.1",
+    "react-dom": "18.3.1"
+  },
+  "scripts": {
+    "start": "vite",
+    "build": "tsc && vite build",
+    "test": "vitest",
+    "preview": "vite preview",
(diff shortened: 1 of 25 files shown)


apps/rich-text-versioning/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "devDependencies": {
+    "@contentful/app-scripts": "^2.5.5",
+    "@testing-library/jest-dom": "^6.6.4",
+    "@testing-library/react": "16.3.0",
+    "@types/node": "24.1.0",
+    "@types/react": "19.1.9",
+    "@types/react-dom": "19.1.7",
+    "@vitejs/plugin-react": "4.7.0",
+    "jsdom": "^26.0.0",
```
FBanfi, line 35: 💭 I saw that jest is appearing in some places, should we remove this? Since we are using vite

apps/rich-text-versioning/src/setupTests.ts
```diff
@@ -0,0 +1,10 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import '@testing-library/jest-dom';
```
FBanfi, line 5: 💭 Same concern as the above one

FBanfi (approved):
🚀 🚀

apps/rich-text-versioning/package.json
JuliRossi: We now can use jest matchers with vitest!

apps/rich-text-versioning/src/setupTests.ts
JuliRossi: We now can use jest matchers with vitest!

PR #10046, opened 2025-08-05
Config screen rich text versioning - [INTEG-3008]

JuliRossi:
## Purpose
Add config screen for rich text versioning 
Out of scope: 
- Content types selector

## Approach
We decided to split this into 2 so we don't block with the modal UI stuff
- Add components
- Add labels
- Add tests

## Testing steps

https://github.com/user-attachments/assets/f40abff8-274d-4fc9-8dda-db6d64713eb7


FBanfi (approved):
Great work!!

PR #10047, opened 2025-08-05
Braze: Disable resource link fields

joaquincasal:
## Purpose

The ResourceLink field type is not supported in the Braze app. However, users were able to select them. This PR disables those fields.

## Approach

The FieldFactory wasn't aware of the existance of the ResourceLink type, so it created a BasicField. Now it creates a ExternalResourceField, which is disabled for both versions of the app.

## Testing steps

Try to use both versions of the app with an entry that has a ResourceLink field.

### DIFF
--- a/apps/braze/src/fields/ExternalResourceField.ts
+++ b/apps/braze/src/fields/ExternalResourceField.ts
@@ -1,22 +1,21 @@
 import { Field } from './Field';
 import { FieldRegistry } from './fieldRegistry';
 
-// We are not supporting rich text fields for now
-export class RichTextField extends Field {
+export class ExternalResourceField extends Field {
   constructor(id: string, name: string, entryContentTypeId: string, localized: boolean) {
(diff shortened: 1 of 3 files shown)


apps/braze/src/fields/ExternalResourceField.ts
JuliRossi, line 1: 💭  I feel like this class and richText have a lot in common. Perhaps there is a way to avoid duplication. On the other hand, I like that it is explicit on how we handle ExternalResourceField.  So I leave it up to you if you want to remove the duplicates

JuliRossi (approved):
LGTM! Left a non-blocking comment

apps/braze/src/fields/ExternalResourceField.ts
joaquincasal: The thing is that rich text is supported in one of the versions, that's why I kept them separate

PR #10048, opened 2025-08-05
fix: Update package.json [INTEG-00]

david-shibley-contentful:
## Purpose

To include the build functions script in the package.json


apps/mux/package.json
```diff
@@ -4,7 +4,8 @@
   "private": true,
   "scripts": {
     "create-app-definition": "contentful-app-scripts create-app-definition",
-    "build": "rimraf ./build && npm run build-frontend",
+    "build:functions": "contentful-app-scripts build-functions --ci",
+    "build": "rimraf ./build && npm run build-frontend && npm run build:functions",
```
primeinteger: Looks like the build is failing with `error: unknown command 'build-functions'`

david-shibley-contentful: yeah looks like they are using an outdated version of contentful app scripts

PR #10053, opened 2025-08-07
Config screen - Add field content type multi selector [INTEG-3009]

JuliRossi:
## Purpose

Add contentype's specific fields for an app

## Approach

- Add field content type multi selector 
- Adjust styles
- Updated tests
- Updated emotion dependency for multiselect
- Updated some dependencies that were causing issues when deploying ci

## Testing steps

video

https://github.com/user-attachments/assets/62a47ad4-a471-4d37-9b89-b1f0f15005d1

### DIFF
--- a/apps/rich-text-versioning/package-lock.json
+++ b/apps/rich-text-versioning/package-lock.json
@@ -1,0 +1,7988 @@
+{
+  "name": "rich-text-versioning",
+  "version": "0.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "rich-text-versioning",
+      "version": "0.1.0",
+      "dependencies": {
+        "@ali-tas/htmldiff-js": "^2.0.1",
+        "@contentful/app-sdk": "4.40.0",
+        "@contentful/f36-components": "4.81.1",
+        "@contentful/f36-multiselect": "^4.81.1",
+        "@contentful/f36-tokens": "4.2.0",
+        "@contentful/field-editor-rich-text": "3.16.14",
+        "@contentful/react-apps-toolkit": "1.2.16",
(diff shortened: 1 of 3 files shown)


apps/rich-text-versioning/src/utils.ts
```diff
@@ -0,0 +1,46 @@
+import { AppState, ContentTypeField } from '@contentful/app-sdk';
+import { ContentTypeProps } from 'contentful-management';
+
+export interface FieldWithContext {
```
JuliRossi: 💭  Hate this name but I think changing it to ContentTypeField may be confunsing because of the app sdk type. If you have any suggestions, please leave them

FBanfi: Perhaps `RichTextFieldInfo` I couldn't find a better name too.

apps/rich-text-versioning/test/locations/ConfigScreen.spec.tsx
```diff
@@ -1,36 +1,72 @@
...
 vi.mock('@contentful/react-apps-toolkit', () => ({
   useSDK: () => mockSdk,
   useCMA: () => mockCma,
 }));
 
 async function saveAppInstallation() {
-  return await mockSdk.app.onConfigure.mock.calls.at(-1)[0]();
+  const result = await mockSdk.app.onConfigure.mock.calls.at(-1)[0]();
+  return result;
 }
```
FBanfi: Nitpick 😝: We could inline this, and maintain this previous version

FBanfi (approved):
Great work!!! 💯

apps/rich-text-versioning/src/utils.ts
JuliRossi: Yeah, that could be it! Thanks.

PR #10054, opened 2025-08-07
Rich-Text-Versioning: UI adjustments and diff component [INTEG-2976]

FBanfi:
## Purpose

This update includes the UI adjustments and logic from the Field and Dialog locations. The error handling, the references and the media cases will be contemplated in another PR.

## Approach

A diff component was created to handle the difference between the two html. In addition all the logic for the Field and Dialog location was added. Finally, the call to get the published version and the current state was contemplated in this update too.

## Testing steps

Automated test added + manual test:

https://github.com/user-attachments/assets/961cabfa-ff29-43ea-abc6-311396e18df2

## Breaking Changes

N/A

Link to [INTEG-2976](https://contentful.atlassian.net/browse/INTEG-2976)
- Specifically subtasks 3006 and 3007!

## Deployment

N/A

[INTEG-2976]: https://contentful.atlassian.net/browse/INTEG-2976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


FBanfi:
One comment:
- The `rich text editor` was mocked in the tests because it needed several and complex properties mocks to work there. Since that component is being used from a library and to simplify the code, I decided to leave it as a mock.

apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -1,16 +1,75 @@
...
+  const [changeCount, setChangeCount] = useState(0);
+
+  useAutoResizer();
+
+  const currentField = invocationParams?.currentField;
+  const publishedField = invocationParams?.publishedField;
 
-  return <Paragraph>Hello Dialog Component (AppId: {sdk.ids.app})</Paragraph>;
+  return (
+    <Flex flexDirection="column" className={styles}>
```
JuliRossi, line 26: 💭 I think the original has a box shadow as well

apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -1,16 +1,75 @@
...
+      <Grid
+        columns="1fr 1fr"
+        rows="1"
+        marginTop="spacingM"
+        marginLeft="spacingM"
+        marginRight="spacingM">
+        <GridItem className="grid-item">
+          <Flex alignItems="space-between">
+            <SectionHeading className="section-heading-margin">Current version</SectionHeading>
```
JuliRossi: ⚔️  Use subheading

apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -1,16 +1,75 @@
...
+          {publishedField && (
+            <HtmlDiffViewer
+              currentField={currentField}
+              publishedField={publishedField}
+              onChangeCount={setChangeCount}
+            />
+          )}
+        </GridItem>
+        <GridItem className="grid-item">
+          <SectionHeading className="section-heading-base">Published version</SectionHeading>
```
JuliRossi: ⚔️ Use subheading

apps/rich-text-versioning/src/locations/Field.tsx
```diff
@@ -1,18 +1,114 @@
...
+      return createClient({
+        space: sdk.ids.space,
+        environment: sdk.ids.environment,
+        accessToken: accessToken,
+      });
+    } catch (error) {
+      console.error('Error creating CDA client:', error);
+      return null;
+    }
+  }, [sdk.ids.space, sdk.ids.environment, sdk.parameters]);
```
JuliRossi: 💭 I think this is not needed as the parameters won't change over time

JuliRossi (commented):
Left a few comments mainly on styling! but nice

apps/rich-text-versioning/src/locations/Dialog.tsx
FBanfi: The design that has the shadow was the initial design, the new one doesn't include it. But let me know if something else comes up!! 🥇

PR #10055, opened 2025-08-07
fix: use correct orgId for app actions [INTEG-3018]

sam-is-content:
the sdk.ids.orgId refers to the current orgId but when we search for app actions, we need to search for the org where the app action lives. This code makes sure we use the correct org - if the app is the marketplace version, use the marketplace apps org, if this is a custom version, use the current orgId.

I was worried about exposing the orgId but Tyler said they are already exposed in our api calls so whatevs

### DIFF
--- a/apps/mux/frontend/src/index.tsx
+++ b/apps/mux/frontend/src/index.tsx
@@ -53,6 +53,7 @@ import {
   generateAutoCaptions,
 } from './util/muxApi';
 import Sidebar from './locations/Sidebar';
+import { APP_ORGANIZATION_ID, isMarketplaceVersion } from './util/constants';
 
 const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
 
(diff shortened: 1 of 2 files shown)


apps/mux/frontend/src/index.tsx
```diff
@@ -503,7 +506,9 @@ export class App extends React.Component<AppProps, AppState> {
         response: { body },
       } = await this.cmaClient.appActionCall.createWithResponse(
         {
-          organizationId: this.props.sdk.ids.organization,
+          organizationId: isMarketplaceVersion({ appId: this.props.sdk.ids.app })
+            ? APP_ORGANIZATION_ID
+            : this.props.sdk.ids.organization,
```
primeinteger: Since this organization ID will be the same for each running instance of the app, this logic could be done once in the `constructor()` instead of every time the `componentDidMount()` and `fetchSignedTokens()` methods run.

For example, we can add a new class property `orgId` that lives next to `apiClient` and `cmaClient` properties, and move this logic to set the value in the `constructor()`

```ts
export class App extends React.Component<AppProps, AppState> {
  apiClient: ApiClient;
  cmaClient: PlainClientAPI;
  // ... other existing props ...
  orgId: string; // <-- new property
  
  constructor(props: AppProps) {
    super(props);
    
    // ... other existing code ...
    
    this.orgId = isMarketplaceVersion({ appId: this.props.sdk.ids.app })
      ? APP_ORGANIZATION_ID
      : this.props.sdk.ids.organization;
  }
```

Then we can use `this.orgId` in the `componentDidMount()` and `fetchSignedTokens()` methods.

primeinteger (approved):
Thanks for the revision! 🔥

PR #10056, opened 2025-08-07
fix: klaviyo actions orgIds [INTEG-3018]

sam-is-content:
the sdk.ids.orgId refers to the current orgId but when we search for app actions, we need to search for the org where the app action lives. This code makes sure we use the correct org - if the app is the marketplace version, use the marketplace apps org, if this is a custom version, use the current orgId.

I was worried about exposing the orgId but Tyler said they are already exposed in our api calls so whatevs

### DIFF
--- a/apps/klaviyo/src/utils/sdk-helpers.ts
+++ b/apps/klaviyo/src/utils/sdk-helpers.ts
@@ -1,3 +1,4 @@
+import { ConfigAppSDK } from '@contentful/app-sdk';
 import { logger } from './logger';
 
 // The key used to store the app definition ID in localStorage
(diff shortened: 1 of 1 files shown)


apps/klaviyo/src/utils/sdk-helpers.ts
```diff
@@ -34,3 +34,11 @@ export const getGlobalSDK = async (retries = MAX_RETRIES): Promise<any | null> =
     }, RETRY_DELAY);
   });
 };
+
+const APP_ORGANIZATION_ID = '5EJGHo8tYJcjnEhYWDxivp';
+const APP_DEFINITION_ID = '2x4lsvuxPAGS2UUHwWVYee';
+
+export const getOrgIdForSdk = (sdk: any) => {
```
primeinteger: The sdk param could use a proper type

```suggestion
export const getOrgIdForSdk = (sdk: ConfigAppSDK) => {
```

primeinteger (approved):
Approving with one small type nit

PR #10059, opened 2025-08-11
Rich-Text-Versioning: Error handling and cma call refactor [INTEG-3013]

FBanfi:
## Purpose

This update includes the error handling in the Field and Dialog location. In addition, some other changes were introduced:
- Refactor to use the `sdk.cma` instead of creating a `cma client`, this remove the necessity to have an access token and simplify the solution.
- Making the `view diff` button visible but disabled when there isn't any value in the rich text field.

## Approach

- The Field location now stores the error state, which is passed to the Dialog location. The Dialog renders an error note when necessary.
- The error state is set to true when there is an error fetching the published value.
- The `cma client` was removed since now we use the `sdk.cma`

## Testing steps

Automated tests were added in the Field and Dialog location to:
- validate the params passed from the field location to the dialog.
- check if the error note is rendered correctly in the dialog.

https://github.com/user-attachments/assets/6784cb48-97e4-415e-ac10-5d4938f91b76

## Breaking Changes

N/A

Link to [INTEG-3013](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?selectedIssue=INTEG-3013)

## Deployment

N/A

[INTEG-3013]: https://contentful.atlassian.net/browse/INTEG-3013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -21,19 +33,47 @@ const Dialog = () => {
...
         rows="1"
         marginTop="spacingM"
         marginLeft="spacingM"
         marginRight="spacingM">
-        <GridItem className="grid-item">
+        <GridItem className={styles.gridItem}>
           <Flex alignItems="space-between">
-            <Subheading>Current version</Subheading>
-            <Badge variant="secondary" className="change-badge">
+            <SectionHeading className={styles.sectionHeadingMargin}>Current version</SectionHeading>
```
JuliRossi: 💭 I think the class here shouldn't exist as the styling is different from figma

apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -45,12 +85,12 @@ const Dialog = () => {
             />
           )}
         </GridItem>
-        <GridItem className="grid-item">
-          <Subheading>Published version</Subheading>
-          <Box className="diff">{documentToReactComponents(publishedField)}</Box>
+        <GridItem className={styles.gridItem}>
+          <SectionHeading className={styles.sectionHeadingBase}>Published version</SectionHeading>
```
JuliRossi: 💭  Same here

apps/rich-text-versioning/src/locations/Field.tsx
```diff
@@ -62,51 +43,66 @@ const Field = () => {
...
-      if (!client) {
-        console.error('CDA client not available');
-        return;
-      }
-
-      const publishedEntry = await client.getEntry(sdk.ids.entry);
+      const publishedEntries = await sdk.cma.entry.getPublished({
+        query: { 'sys.id': sdk.ids.entry },
+      });
+      const publishedEntry = publishedEntries.items[0];
```
JuliRossi, line 58: ⭐  Great discovery!

apps/rich-text-versioning/src/locations/Field.tsx
```diff
@@ -62,51 +43,66 @@ const Field = () => {
...
+      switch (Object.values(error)[0]) {
+        default:
+          currentErrorInfo = {
+            hasError: true,
+            errorCode: '500',
+            errorMessage: 'Error loading content',
+          };
+      }
+
+      setErrorInfo(currentErrorInfo);
```
JuliRossi: ⚔️  This logic seems complex for what is currently doing. We don't need a switch.

apps/rich-text-versioning/vitest.config.mts
JuliRossi, line 1: 💭 Nit: Unsure on what change in this file, perhaps prettier? My 2 cents is that we shouldn't upload the changes for a file if that file only has linter changes. 

There are 2 exceptions I can think would make sense:
-  If CI is breaking because of prettier for that branch (If this is your case, disregard this message)
- If we are only doing linter changes for that PR.

FBanfi: Yes, it was probably a linter change that was added while formating the whole proyect with prettier. I will follow that lead then, thanks!
As for this change, I removed this! ✅

apps/rich-text-versioning/src/locations/Field.tsx
FBanfi: Oh, yess! That switch was from a previous version in which we had different types of errors. Nice catch, corrected ✅

apps/rich-text-versioning/src/locations/Dialog.tsx
FBanfi: Yes! Actually this change may have come from the rebase but I think we want the previous version, the one with the subheading component! Corrected ✅

JuliRossi (approved):

PR #10063, opened 2025-08-14
References rich text [INTEG-2977]

JuliRossi:
## Purpose

Add references to rich text comparison
Disclamer: Diff library will be changed in another PR, as it's causing issues while diffing

## Approach

- Add logic for adding references
- Add error handling
- Tests

## Testing steps

https://github.com/user-attachments/assets/df5f201d-3568-4da0-ad1d-d02664d0543c

https://github.com/user-attachments/assets/e908bc6a-813d-488a-a29f-8b6f0ad6ca6f

## Collaborators
Thanks @joaquincasal for pairing this and @FBanfi for fixing the margin issues :D

### DIFF
--- a/apps/rich-text-versioning/package.json
+++ b/apps/rich-text-versioning/package.json
@@ -9,7 +9,6 @@
     "@contentful/f36-components": "4.81.1",
     "@contentful/f36-multiselect": "^4.81.1",
     "@contentful/f36-tokens": "4.2.0",
-    "@contentful/field-editor-reference": "^6.10.2",
     "@contentful/field-editor-rich-text": "3.16.14",
     "@contentful/react-apps-toolkit": "1.2.16",
     "@contentful/rich-text-html-renderer": "^17.1.0",
--- a/apps/rich-text-versioning/src/components/HtmlDiffViewer.tsx
+++ b/apps/rich-text-versioning/src/components/HtmlDiffViewer.tsx
@@ -1,7 +1,6 @@
-import { Block, Document, Inline } from '@contentful/rich-text-types';
-import { documentToHtmlString } from '@contentful/rich-text-html-renderer';
+import { Document } from '@contentful/rich-text-types';
 import { Diff } from '@ali-tas/htmldiff-js';
-import { useEffect, useState, useMemo } from 'react';
+import { useEffect, useState } from 'react';
 import { styles } from './HtmlDiffViewer.styles';
 import { documentToReactComponents } from '@contentful/rich-text-react-renderer';
 import { Skeleton } from '@contentful/f36-components';
@@ -28,11 +27,6 @@ const HtmlDiffViewer = ({
 }: HtmlDiffViewerProps) => {
   const [diffHtml, setDiffHtml] = useState<string>('');
 
-  // Helper function to convert React component to HTML string
-  const componentToHtml = (component: React.ReactElement): string => {
-    return renderToString(component);
-  };
-
   useEffect(() => {
     const processDiff = async () => {
       // Convert current field to React components with embedded entry renderers
@@ -46,8 +40,8 @@ const HtmlDiffViewer = ({
       );
 
       // Convert React components to HTML strings
-      const currentHtml = componentToHtml(<>{currentComponents}</>);
-      const publishedHtml = componentToHtml(<>{publishedComponents}</>);
+      const currentHtml = renderToString(<>{currentComponents}</>);
+      const publishedHtml = renderToString(<>{publishedComponents}</>);
 
       // Perform diff
       const diff = Diff.execute(publishedHtml, currentHtml);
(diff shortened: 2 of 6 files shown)


apps/rich-text-versioning/src/components/HtmlDiffViewer.tsx
```diff
@@ -1,41 +1,86 @@
...
+  entries,
+  entryContentTypes,
+  defaultLocale,
+}: HtmlDiffViewerProps) => {
   const [diffHtml, setDiffHtml] = useState<string>('');
 
+  // Helper function to convert React component to HTML string
+  const componentToHtml = (component: React.ReactElement): string => {
+    return renderToString(component);
+  };
```
FBanfi: 💭 Perhaps we could remove the `componentToHtml` method and directly use the `renderToString` in the component.

apps/rich-text-versioning/src/components/HtmlDiffViewer.tsx
```diff
@@ -1,41 +1,86 @@
-import { Document } from '@contentful/rich-text-types';
+import { Block, Document, Inline } from '@contentful/rich-text-types';
 import { documentToHtmlString } from '@contentful/rich-text-html-renderer';
 import { Diff } from '@ali-tas/htmldiff-js';
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useMemo } from 'react';
```
FBanfi: We could remove this imports that are not being used anymore:
- `Block`, `Inline`, `documentToHtmlString`, `useMemo`

FBanfi (approved):
Really cool reference handling!! Lgtm
Left two minor comments

apps/rich-text-versioning/package.json
```diff
@@ -5,9 +5,11 @@
   "dependencies": {
     "@ali-tas/htmldiff-js": "^2.0.1",
     "@contentful/app-sdk": "4.40.0",
+    "@contentful/rich-text-types": "^16.1.0",
     "@contentful/f36-components": "4.81.1",
     "@contentful/f36-multiselect": "^4.81.1",
     "@contentful/f36-tokens": "4.2.0",
+    "@contentful/field-editor-reference": "^6.10.2",
```
joaquincasal: I think we can remove this one

apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -28,11 +31,89 @@ const Dialog = () => {
...
+      }
+      if ('content' in node) {
+        node.content.forEach(getEntryIdsFromNode);
+      }
+    };
+    doc.content?.forEach(getEntryIdsFromNode);
+    return ids;
+  };
+
+  useEffect(() => {
```
joaquincasal, line 60: What do you think about moving this logic into the `createOptions` helper? If the renderer function can be made async, we wouldn't need to fetch every entry in advance

apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -77,17 +158,25 @@ const Dialog = () => {
               {changeCount} change{changeCount !== 1 ? 's' : ''}
             </Badge>
           </Flex>
-          {(publishedField || currentField) && (
+          {(publishedField || currentField) && !loading && (
```
joaquincasal: I don't think we need the `!loading` here since we have the early return above in line 108

apps/rich-text-versioning/src/utils.ts
```diff
@@ -56,3 +57,32 @@ export const restoreSelectedFields = (
...
+  if (!displayFieldId) return 'Untitled';
+
+  const value = entry.fields[displayFieldId]?.[locale];
+  if (value === undefined || value === null || value === '') {
+    return 'Untitled';
+  }
+  return String(value);
+};
+
+export const getEntryStatus = (entry: EntryProps): EntityStatus | undefined => {
```
joaquincasal, line 76: There's a `fieldStatus` field in the entry response that we can use for this instead of calculating ourselves. This way we can also differentiate each locale

apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -77,17 +158,25 @@ const Dialog = () => {
...
             <HtmlDiffViewer
               currentField={currentField}
               publishedField={publishedField}
               onChangeCount={setChangeCount}
+              entries={entries}
+              entryContentTypes={entryContentTypes}
+              defaultLocale={sdk.locales.default}
```
joaquincasal, line 169: I think we want to use the locale of the field that was clicked, we probably need to share it when opening the dialog from the field

joaquincasal (commented):
Left some comments!

apps/rich-text-versioning/src/locations/Dialog.tsx
joaquincasal: Never mind, this is not possible, the renderer cannot be async

PR #10069, opened 2025-08-18
Rich text versioning: Asset support [INTEG-3019]

joaquincasal:
## Purpose

Display added or removed assets in the diff

## Approach

Same approach as with references: add options to the renderer so it displays an AssetCard when it finds an asset node

## Testing steps

Add or remove assets from a rich text and verify the diff

image

### DIFF
--- a/apps/rich-text-versioning/src/components/createOptions.tsx
+++ b/apps/rich-text-versioning/src/components/createOptions.tsx
@@ -1,16 +1,19 @@
-import { Box, EntryCard, InlineEntryCard } from '@contentful/f36-components';
-import { EntryProps, ContentTypeProps } from 'contentful-management';
-import { getEntryStatus, getEntryTitle } from '../utils';
-import { BLOCKS, INLINES } from '@contentful/rich-text-types';
+import { AssetCard, Box, EntryCard, InlineEntryCard } from '@contentful/f36-components';
+import { EntryProps, ContentTypeProps, AssetProps } from 'contentful-management';
+import { getEntryTitle } from '../utils';
+import { Block, BLOCKS, Inline, INLINES } from '@contentful/rich-text-types';
 import { Options } from '@contentful/rich-text-react-renderer';
+import { css } from 'emotion';
 
 const UNKNOWN = 'Unknown';
-const NOT_FOUND = 'Entry not found';
+const ENTRY_NOT_FOUND = 'Entry not found';
+const ASSET_NOT_FOUND = 'Asset not found';
 const DELETED = 'deleted';
 
 export const createOptions = (
   entries: EntryProps[],
   entryContentTypes: Record<string, ContentTypeProps>,
+  assets: AssetProps[],
   locale: string
 ): Options => ({
   renderNode: {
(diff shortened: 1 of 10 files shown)


apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -43,10 +43,10 @@ const Dialog = () => {
   const publishedField = invocationParams?.publishedField || { content: [] };
   const locale = invocationParams?.locale;
 
-  const getEntryIdsFromDocument = (doc: Document): string[] => {
+  const getReferenceIdsFromDocument = (doc: Document, types: string[]): string[] => {
```
JuliRossi, line 46: ⭐ Love the refactor and reuse of this function

apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -98,6 +110,19 @@ const Dialog = () => {
...
+            query: {
+              'sys.id[in]': allAssetIds.join(','),
+            },
+          });
+          setAssets(fetchedAssets.items);
+        } catch (error) {
+          console.error('Error fetching assets:', error);
+        }
+      }
+      setEntryContentTypes(contentTypes);
```
JuliRossi: 💭 I think we don't need this _setEntryContentTypes_ because we are setting it up on line 111

apps/rich-text-versioning/src/components/createOptions.tsx
```diff
@@ -1,16 +1,18 @@
-import { Box, EntryCard, InlineEntryCard } from '@contentful/f36-components';
-import { EntryProps, ContentTypeProps } from 'contentful-management';
-import { getEntryStatus, getEntryTitle } from '../utils';
-import { BLOCKS, INLINES } from '@contentful/rich-text-types';
+import { AssetCard, Box, EntryCard, InlineEntryCard } from '@contentful/f36-components';
+import { EntryProps, ContentTypeProps, AssetProps } from 'contentful-management';
+import { getEntryTitle } from '../utils';
+import { Block, BLOCKS, Inline, INLINES } from '@contentful/rich-text-types';
```
JuliRossi: 💭 I think we don't need some of these anymore

JuliRossi (approved):
LGTM! Left a few non-blocking comments

apps/rich-text-versioning/src/components/createOptions.tsx
```diff
@@ -19,7 +21,7 @@ export const createOptions = (
       if (!entry) {
         return (
           <Box marginBottom="spacingM">
-            <EntryCard contentType={UNKNOWN} title={NOT_FOUND} status={DELETED} />
+            <EntryCard contentType={UNKNOWN} title={ENTRY_NOT_FOUND} status={DELETED} />
           </Box>
         );
```
FBanfi, line 26: Even though that there is no entry, I don't know if we want to have the deleted status in the Published version. Because there should not be reference to the changes, just the published content in that side of the grid.

Same concern in the other parts (inline reference and assets)

PR #10070, opened 2025-08-18
fix: salesforce commerce cloud when there is a untitled product listing from salesforce [INTEG-3036]

ryunsong-contentful:
## Purpose
JIRA Link: https://contentful.atlassian.net/browse/INTEG-3036
Zendesk link: https://contentful.atlassian.net/browse/ZEND-6790

User noted an issue in zendesk
Screenshot 2025-08-18 at 3 03 26 PM

With the fix it now does this
Screenshot 2025-08-18 at 3 12 14 PM

### Root Cause
The error was caused by unsafe property access patterns where `.default` properties were accessed without proper null/undefined checks:

```typescript
// PROBLEMATIC CODE
{product.name.default}                    // Crashes if product.name is undefined
{product.image.alt.default}               // Crashes if image or alt is missing
if (product.shortDescription.default)     // Crashes if shortDescription is undefined
```

### Solution

**1. Implemented Safe Property Access**
Replaced unsafe property access with optional chaining (`?.`) and nullish coalescing (`||`) operators:

```typescript
// FIXED CODE
{product.name?.default || 'Untitled Product'}
{product.image?.alt?.default || 'Product image'}  
if (product.shortDescription?.default) { ... }
```

**2. Fixed Modal UI Structure**
- Removed conflicting wrapper causing header overlap
- Fixed sticky positioning
- Improved spacing and alignment
- Added proper scrolling for search results

### DIFF
--- a/apps/salesforce-commerce-cloud/src/components/dialog/CategoryPreview.tsx
+++ b/apps/salesforce-commerce-cloud/src/components/dialog/CategoryPreview.tsx
@@ -13,7 +13,7 @@ const CategoryPreview = (props: { category: any }) => {
           <Text as="span" fontWeight="fontWeightDemiBold">
             {category.name?.default}
           </Text>{' '}
-          (Catalog: {catalogInfo?.name?.default ? catalogInfo.name.default : category.catalogId})
+          (Catalog: {catalogInfo?.name?.default || category.catalogId})
         </Text>
         {category.pageDescription?.default && (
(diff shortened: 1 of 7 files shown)


primeinteger (approved):
Approving so this doesn’t block the main fix. 

One small note from the screenshots: the modal design could be tightened up a bit. There seems to be extra padding in the modal content area that slightly throws off the overall layout. Not a blocker, but maybe worth refining later.

PR #10073, opened 2025-08-21
Rich Text Versioning app [INTEG-2763]

joaquincasal:
## Purpose

This app allows content creators to visually compare changes in a rich text field against the last published version. A two-column view highlights added, removed, and modified content, including referenced entries and assets.

## Approach

- Both the current and published content of the rich text field are obtained
- Then the contents are transformed from their json representation to React nodes. Here we choose how to display references to entries and assets are displayed, since the library that makes the transformation ignores them by default
- Then the content gets rendered to html and both html codes get compared to mark the differences.

## Testing steps

- Install the app
- Choose a content type with a rich text field and change the appearance of the field to be the one from the app
- Edit an entry from the chosen content type and click "Compare versions"

https://github.com/user-attachments/assets/a447c8b5-236a-44fe-96b7-a1decced74ac


apps/rich-text-versioning/src/components/HtmlDiffViewer.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <Skeleton.BodyText numberOfLines={4} />
+      </Skeleton.Container>
+    );
+  }
+
+  return (
+    <>
+      <div
+        className={styles}
+        dangerouslySetInnerHTML={{
```
ryunsong-contentful, line 83: See the last commit I added here to add more security around XSS attacks. 

I did a quick double check to make sure it didn't affect the final product but I encourage you to also do a check as well and make sure my commit didn't change anything other than adding more security

ryunsong-contentful (approved):

PR #10074, opened 2025-08-21
Bulk edit [INTEG-2847]

JuliRossi:
## Purpose

We want to allow viewing and editing more than 100 entries at a time

## Approach

- Added scripts for generating entries and deleting for testing purposes
- Added more options for viewing
- Fix several truncating issues
- Fix some performance issues when entries are large
- Disable buttons while saving - undoing

## Testing steps
Generated entries > Go to app > They load in aprox 5seg 
Selected a row > Edited > It takes up to 1:30 to edit 500 entries
Save

https://github.com/user-attachments/assets/1b4d8273-c2db-49c8-b11b-476dde828f0f

Load (The notifier is now properly truncated)

https://github.com/user-attachments/assets/ca24a974-d2e4-4173-92ed-2ded370135c7

Undo 

https://github.com/user-attachments/assets/d32a0505-f815-4c85-b40f-b447d62325d7


apps/bulk-edit/src/locations/Page/components/BulkEditModal.tsx
```diff
@@ -48,7 +48,7 @@ export const BulkEditModal: React.FC<BulkEditModalProps> = ({
           </Text>
           <Flex>
             <Text>
-              <Text fontWeight="fontWeightDemiBold">{firstValueToUpdate}</Text>{' '}
+              <Text fontWeight="fontWeightDemiBold">{truncate(firstValueToUpdate, 100)}</Text>{' '}
```
FBanfi, line 51: 💭 Nitpick: Since we are truncating everything to `100` characters, we could have that number hardcoded in the utils instead of hardcoding it every time we use the truncate function. So if something changes in the future, we just change that single place.

In the utils:
```
export const truncate = (str: string, max: number = 100) =>
  str.length > max ? str.slice(0, max) + ' ...' : str;
```
  
  And in the code:
  `truncate(firstValueToUpdate)`

apps/bulk-edit/src/locations/Page/components/UndoBulkEditModal.tsx
```diff
@@ -28,19 +29,23 @@ export const UndoBulkEditModal: React.FC<UndoBulkEditModalProps> = ({
...
-                The update for <b>{firstEntryFieldValue}</b> will be reverted.
+                The update for <b>{truncate(firstEntryFieldValue, 100)}</b> will be reverted.
               </>
             ) : (
               <>
-                The update for <b>{firstEntryFieldValue}</b> and <b>{entryCount - 1}</b> more entry
-                fields will be reverted.
+                The update for <b>{truncate(firstEntryFieldValue, 100)}</b> and{' '}
+                <b>{entryCount - 1}</b> more entry fields will be reverted.
               </>
```
FBanfi, line 38: 💭 Same as the comment from above

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -77,9 +83,9 @@ const Page = () => {
 
     return {
       content_type: selectedContentTypeId,
+      order: getOrder(sortOption),
       skip: activePage * itemsPerPage,
       limit: itemsPerPage,
-      order: getOrder(sortOption),
     };
   };
```
FBanfi, line 95: Placing the `order` property in the right **_order_** 😆 😆 🔥

apps/bulk-edit/.gitignore
```diff
@@ -23,3 +23,10 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# dotenv environment variables file
+.env
+.env.*
+!.env*.example
```
joaquincasal: There 3 were already included above

apps/bulk-edit/vitest.config.mts
```diff
@@ -0,0 +1,13 @@
+import { loadEnv } from 'vite';
+import { defineConfig } from 'vitest/config';
+import react from '@vitejs/plugin-react';
+
+export default defineConfig({
+  plugins: [react()],
+  test: {
```
joaquincasal, line 7: Maybe we want to delete the `test` section from the `vite.config` file if we're including a separate file

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -145,13 +151,12 @@ const Page = () => {
...
-        const { items, total } = await sdk.cma.entry.getMany({
-          spaceId: sdk.ids.space,
-          environmentId: sdk.ids.environment,
-          query: buildQuery(sortOption, displayField),
-        });
-        setEntries(items || []);
-        setTotalEntries(total || 0);
+        const baseQuery = buildQuery(sortOption, displayField);
+
+        const { entries, total } = await fetchEntriesWithBatching(sdk, baseQuery, baseQuery.limit);
```
FBanfi: We could remove the third param because we are passing the whole `baseQuery` in the second one

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
```diff
@@ -148,3 +149,109 @@ export function getEntryFieldValue(
...
+
+/**
+ * Processes entries in batches with configurable delays to avoid rate limiting.
+ * @param entries - Array of entries to process
+ * @param updateFunction - Function to call for each entry
+ * @param batchSize - Number of entries to process in each batch
+ * @param delayMs - Delay in milliseconds between batches
+ * @returns Promise that resolves to array of results from updateFunction
+ */
+//TODO: Remove generic type and batchSize and delayMs as parameters
```
joaquincasal: Do we want to address this before merging?

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
```diff
@@ -148,3 +149,109 @@ export function getEntryFieldValue(
...
+export async function processEntriesInBatches<T, R>(
```
joaquincasal: Do we need this to be generic? Not sure if there's an advantage

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
```diff
@@ -148,3 +149,109 @@ export function getEntryFieldValue(
...
+      hasMore =
+        items.length === batchSize &&
+        (limit === undefined || allEntries.length < limit) &&
+        allEntries.length < total;
+
+      batchSkip += batchSize;
+
+      if (hasMore) {
+        await new Promise((resolve) => setTimeout(resolve, BATCH_FETCHING.BATCH_DELAY_MS));
+      }
```
joaquincasal: Do we need a delay here? Would we hit the rate limit otherwise? I'd think a few requests should be enough and that won't hit the limit

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -198,14 +206,39 @@ const Page = () => {
   }
 
   const fetchLatestEntries = async (entryIds: string[]) => {
```
joaquincasal: Would it make sense to use `fetchEntriesWithBatching` here?

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -293,12 +331,14 @@ const Page = () => {
...
     setIsSaving(true);
+    setFailedUpdates([]);
 
     try {
       const entryIds = Object.keys(backupToUse);
       const currentEntries = await fetchLatestEntries(entryIds);
 
-      const restorePromises = currentEntries.map(async (currentEntry) => {
+      // Create restore function for batch processing
+      const restoreEntry = async (currentEntry: EntryProps) => {
```
FBanfi, line 324: We fetch the latests entries in case some of them has been manually modified since the edition?

If that's the case, perhaps we could remove this fetch... 🤔  
Because the `undo` is only available during notification message time. It is a really small time and it would be strange that someone modifies an entry manually in that time. Moreover, even without removing this fetch, some user could still do a manual change in the middle of the bulk edit and we would be in the same scenario.

So perhaps we could remove that fetch, simplifying the code and gaining a little more of performance 🤔

apps/bulk-edit/src/locations/Page/utils/constants.ts
```diff
@@ -3,7 +3,32 @@ export const ENTRY_STATUS_COLUMN = 'notfield_status';
 export const PAGE_TITLE = 'Bulk Edit';
 export const PAGE_DESCRIPTION = 'Bulk edit your content entries';
 
-export const ERROR_MESSAGES = {
-  LOADING_ERROR: 'Failed to load data',
-  SAVE_ERROR: 'Failed to save changes',
+// Batch fetching configuration - used in entryUtils.ts
```
FBanfi: 💭 Just a head's up to see if there is value in maintaining all this logs or if they are a bit redundant.

apps/bulk-edit/test/entryUtils.test.ts
FBanfi, line 1: 💭 I noticed that there are two suits of tests with the **entryUtils** name, should it be just one with the content of both of them?
- `apps/bulk-edit/test/entryUtils.test.ts`
- `apps/bulk-edit/test/locations/Page/entryUtils.test.tsx`

joaquincasal (commented):
Left some comments/questions!

apps/bulk-edit/.gitignore
JuliRossi: Good catch. Yes, I must leave only the test one

apps/bulk-edit/vitest.config.mts
JuliRossi: sure

apps/bulk-edit/test/entryUtils.test.ts
JuliRossi: Good catch. It might have happen when moving files

apps/bulk-edit/src/locations/Page/components/BulkEditModal.tsx
JuliRossi: We don't truncate everything to 100. Table rows are truncated to 20 and in the notifier to 30

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
JuliRossi: Yep, meant to change that but forgot . Thanks!

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
JuliRossi: Yep, meant to change that but forgot . Thanks!

apps/bulk-edit/src/locations/Page/utils/constants.ts
JuliRossi: some yes, others no

apps/bulk-edit/src/locations/Page/index.tsx
JuliRossi: Unfortunatelly, we need to pass the batchSize because we are using recursion. If the batch is too big, then the batch size is split in half

apps/bulk-edit/src/locations/Page/components/UndoBulkEditModal.tsx
JuliRossi: Same as above

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
JuliRossi: Yes, I add it to be safe. Entries would need to be insanely big and batch small for this to make sense.

apps/bulk-edit/src/locations/Page/index.tsx
JuliRossi: yes, I need to modify it a bit to not pass the request limit

apps/bulk-edit/src/locations/Page/index.tsx
JuliRossi: Tried it but using stale entries  to update brings up issues with the version of the entry. The following error arises:
 {
    "sys": {
        "type": "Error",
        "id": "VersionMismatch"
    },
}

FBanfi: Oh nevermind then! 👍

PR #10085, opened 2025-08-27
Bulk-Edit-App: Show the progress of the edition [INTEG-3106]

FBanfi:
## Purpose

This update includes a message in the modal showing the user the progress of the edition. Introduced to let the user know  that the edition isn't stuck.

## Approach

Added 2 states `totalUpdateCount` (the total of entries that are going to be edited) and `editionCount` (the current number of entries edited). This state is passed to the `BulkEditModal` component to display and change a message each time a entry is edited.

## Testing steps

Automated tests added. You could check this update by bulk editing entries:

https://github.com/user-attachments/assets/242d3d8f-2e87-4b69-b5bd-74cac3c7077e

## Breaking Changes

N/A

Link to [INTEG-3106](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?selectedIssue=INTEG-3106)

## Deployment

N/A


apps/bulk-edit/src/locations/Page/components/ProgressMessage.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { ClockIcon } from '@contentful/f36-icons';
+import { css } from 'emotion';
+
+interface ProgressMessageProps {
+  totalCount: number;
+  currentCount: number;
+  styles: Record<string, string | number>;
+}
+
+export const ProgressMessage: React.FC<ProgressMessageProps> = ({
```
JuliRossi: 💭 I think we should be using this forma component https://f36.contentful.com/components/note instead of creating a custom one

FBanfi: I didn't see that the note component could take the same same form as the one needed in this case. That's much better!!
Thanks, I removed the custom one and it's tests. ✅

PR #10088, opened 2025-08-28
Bulk Edit v1.1 []

JuliRossi:
## Purpose
 Version 1.1 for bulk edit

## Approach

In this iteration:
- Freeze top headers and status column
- Editing more than 100+ at a time. 
- Viewing more than 100+ at a time
- Skeletons for loading
- Modal adjustments:
    - Loading waiting times.
    - Disable cancel button
- Fix values for fields using truncation

Main tickets:
2847, 2953, 3106 y 3104

### DIFF
--- a/apps/bulk-edit/.env.example
+++ b/apps/bulk-edit/.env.example
@@ -1,0 +1,25 @@
+# Content Management API token
+# Get this from: https://app.contentful.com/spaces/{SPACE_ID}/api/keys
+CONTENTFUL_ACCESS_TOKEN=your_content_management_api_token_here
+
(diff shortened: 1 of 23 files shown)


ryunsong-contentful (approved):
Nice, good refactoring and abstractions.

Definitely not a blocker but view 300 at a time is an option still just fyi.
Also not a blocker would be nice to find a way to create some space so that the button just appears rather than the entire table moving to make space for the button
Screenshot 2025-08-29 at 10 30 22 AM

Screenshot 2025-08-29 at 10 30 08 AM

PR #10097, opened 2025-09-04
Bulk-Edit-App: Adding status filter + refactors + some fixes [INTEG-3116]

FBanfi:
## Purpose

This update allow the user to filter the entries table by status.

## Approach

In addition to the new status filter logic some other changes were made, some of the most important ones would be:
- A `GenericMultiselect` component was created to avoid having two multiselects (one for the `ColumnMultiselect` and another for the `StatusMultiselect`). This component use the F36 Multiselect inside. This allow us to have a simpler code and also test better the functionality.
- A `CustomMultiselectAll` component was created since the `Multiselect.MultiselectAll` from F36 has a bug bubbling the event when we use 2 or more multiselect at a time.
- Changing the sorting button and the filter field multiselect to be visible but disabled when there are no entries.

## Testing steps

Tests were modify to match the new generic multiselect component. Manual testing:

https://github.com/user-attachments/assets/dfebffe1-4cb9-438b-b081-0a4c5ec2b74e

## Breaking Changes

N/A

Link to [INTEG-3116](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?selectedIssue=INTEG-3116)

## Deployment

N/A


apps/bulk-edit/src/locations/Page/components/CustomMultiselectAll.styles.ts
JuliRossi, line 1: ⭐  Thanks for starting the process of unraveling the style file

apps/bulk-edit/src/locations/Page/components/CustomMultiselectAll.styles.ts
```diff
@@ -0,0 +1,22 @@
...
+export const selectAllCheckboxStyles = (areAllSelected: boolean) => css`
+  background-color: ${areAllSelected ? tokens.gray200 : 'transparent'};
+  border-radius: ${tokens.borderRadiusSmall};
+  padding-left: ${tokens.spacingXs};
+  padding-right: ${tokens.spacing2Xs};
+  padding-bottom: ${tokens.spacing2Xs};
+  padding-top: ${tokens.spacing2Xs};
+  &:hover {
+    background-color: ${areAllSelected ? tokens.gray200 : tokens.gray100};
+  }
```
JuliRossi: ❓ Just curious, is there a reason we aren't using camelCase for these styles?

apps/bulk-edit/src/locations/Page/components/CustomMultiselectAll.tsx
```diff
@@ -0,0 +1,32 @@
...
+  onChange,
+}: CustomMultiselectAllProps) => {
+  return (
+    <Box>
+      <Checkbox
+        className={selectAllCheckboxStyles(areAllSelected)}
+        isDisabled={disabled}
+        isChecked={areAllSelected}
+        onChange={onChange}>
+        <Text className={selectAllTextStyles}>
```
JuliRossi: 💭  I think this should de semibold, but not sure

apps/bulk-edit/src/locations/Page/components/GenericMultiselect.tsx
```diff
@@ -0,0 +1,101 @@
...
+        placeholder={getPlaceholderText()}
+        triggerButtonProps={{ size: 'small' }}
+        popoverProps={{ isFullWidth: true }}>
+        <CustomMultiselectAll
+          areAllSelected={areAllSelected}
+          disabled={disabled}
+          onChange={toggleAll}
+        />
+        {options.map((option) => (
+          <F36MultiselectOption
```
JuliRossi: 💭  Shouldn't this be Multiselect.Option ?

apps/bulk-edit/src/locations/Page/components/SortMenu.tsx
```diff
@@ -35,6 +36,7 @@ export const SortMenu: React.FC<SortMenuProps> = ({ sortOption, onSortChange })
         <Menu.List style={styles.sortMenuList}>
           {SORT_OPTIONS.map((option) => (
             <Menu.Item
+              isDisabled={disabled}
```
JuliRossi, line 39: 💭 Shouldn't we disable the button here instead of the options?

apps/bulk-edit/test/locations/Page/ColumnMultiSelect.test.tsx
JuliRossi, line 1: ⚔️  This is only testing the column filter and not the status one

apps/bulk-edit/src/locations/Page/components/GenericMultiselect.tsx
```diff
@@ -0,0 +1,101 @@
...
+    singleSelected: string;
+    multipleSelected: string;
+  };
+  truncateLength?: number;
+  getItemKey: (item: T) => string;
+  getItemValue: (item: T) => string;
+  isItemSelected: (item: T, selectedItems: T[]) => boolean;
+}
+
+const GenericMultiselect = <T extends BaseMultiselectOption>({
```
JuliRossi: ⚔️  I like the idea of reusing code, but the usage of this component is complex and it makes index handles most of the logic, when we could have smaller components that handle the filters.  We don't want index to be a "God" component. I think we can reach a middle ground in which take some things from this version so we don't have a lot of duplicated code but we reduce complexity.  Let's pair this a bit.

JuliRossi (commented):
Left a few comments. Let's chat a bit.

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -84,20 +87,74 @@ const Page = () => {
...
+
+  const getStatusFilter = (statuses: Status[]) => {
+    // If no statuses selected, return a filter that matches nothing
+    if (statuses.length === 0) {
+      return { 'sys.id[in]': 'nonexistent-id' };
+    }
+
+    const hasDraft = statuses.some((s) => s.label === 'Draft');
+    const hasPublished = statuses.some((s) => s.label === 'Published');
+    const hasChanged = statuses.some((s) => s.label === 'Changed');
```
JuliRossi: 💭 
```suggestion
    const hasDraft = statuses.includes('Draft');
    const hasPublished = statuses.includes( 'Published');
    const hasChanged = statuses.includes('Changed');
```

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -84,20 +87,74 @@ const Page = () => {
...
+
+    // Single status filtering
+    if (hasDraft && statuses.length === 1) {
+      return { 'sys.publishedAt[exists]': false };
+    }
+
+    if (!hasDraft && (hasPublished || hasChanged)) {
+      return { 'sys.publishedAt[exists]': true, 'sys.archivedAt[exists]': false };
```
JuliRossi, line 125: 💭  I feel we might not need this

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -84,20 +87,74 @@ const Page = () => {
...
+  const filterEntriesByStatus = (entries: EntryProps[], statuses: Status[]): EntryProps[] => {
+    // If no statuses selected, return empty array
+    if (statuses.length === 0) return [];
+
+    // If all statuses selected, return all entries
+    if (statuses.length === 3) return entries;
 
+    return entries.filter((entry) => {
+      const entryStatus = getStatus(entry);
+      return statuses.some((status) => status.label === entryStatus.label);
```
JuliRossi, line 142:
```suggestion
      return statuses.includes(entryStatus.label);
```

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -84,20 +87,74 @@ const Page = () => {
...
+    });
+  };
+
+  function needsClientFiltering() {
+    const hasDraft = selectedStatuses.some((s) => s.label === 'Draft');
+    const hasPublished = selectedStatuses.some((s) => s.label === 'Published');
+    const hasChanged = selectedStatuses.some((s) => s.label === 'Changed');
```
JuliRossi:
```suggestion
    const hasDraft = selectedStatuses.includes('Draft');
    const hasPublished = selectedStatuses.includes('Published');
    const hasChanged = selectedStatuses.includes('Changed');
```

apps/bulk-edit/src/locations/Page/components/CustomMultiselectAll.styles.ts
FBanfi: Because in this case I was using Emotion's css template literal, and it seems to need pure css as the input.
But I now changed it to follow the camel case format so we maintain consistency ✅

apps/bulk-edit/src/locations/Page/components/CustomMultiselectAll.tsx
FBanfi: You are right! ✅

apps/bulk-edit/src/locations/Page/components/SortMenu.tsx
FBanfi: This was one of the things I wasn't sure, but after asking in the weekly, Erin said that she will think about it and then she will update the design. So this will probably change in another PR.

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: I cannot make this change because selectedStatuses has a `Status` type (an object composed by label and color) and "Draft", "Published" or "Changed" are `Strings`.

I could do something like this, to use the `includes`:
`const hasDraft = selectedStatuses.map((s) => s.label).includes('Draft');`

But I personally prefer the previous version:
`const hasDraft = selectedStatuses.some((s) => s.label === 'Draft');
`

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: The same thing from [below](https://github.com/contentful/apps/pull/10097#discussion_r2325835294) happens here.

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: We need this because in this case we will fetch all the ones except the ones that are on draft because we don't want them, and that will save us some performance.
If we remove this and the filter has "Changed and Published", we will create a query that fetch all statuses when we just want two of them.

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: responded here: https://github.com/contentful/apps/pull/10097#discussion_r2331121170

apps/bulk-edit/src/locations/Page/components/GenericMultiselect.tsx
FBanfi: I understand the concern about complexity, and I appreciate the feedback! ❤️ 

I think the Page component was already a god object before this change. Without this `GenericMultiselect`, we'd have the same god object problem plus duplicated logic and two separate components to test that work identically (except for 3 or 4 oneliner changes).

I love the idea of pairing to make the index component less "god" 😆, but I was thinking that perhaps those refactors could be made in another PR, since this would involve a big structural change and I don't want to be stuck on this ticket much longer.
But let me know what you think! 🤝

apps/bulk-edit/test/locations/Page/ColumnMultiSelect.test.tsx
FBanfi: Oh yes, thanks! ✅ Added!

apps/bulk-edit/src/locations/Page/index.tsx
JuliRossi: But why do we need the keep color in the selectedStatuses as it is only used in one place ?  Perhaps we could change that.

If you still want to keep it as is, we could do something like the following in utils:
```
const DRAFT_STATUS = {label: 'Draft', color: 'warning'}
```
In index: 
```
const hasDraft = selectedStatuses.includes(DRAFT_STATUS)
```

apps/bulk-edit/src/locations/Page/components/GenericMultiselect.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { Multiselect } from '@contentful/f36-multiselect';
+import { useMemo } from 'react';
+import { truncate } from '../utils/entryUtils';
+import CustomMultiselectAll from './CustomMultiselectAll';
+import { styles } from '../styles';
+import { optionStyles } from './GenericMultiselect.styles';
+
+export interface BaseMultiselectOption {
+  label: string;
+}
```
JuliRossi: 💭 Use label + value

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -443,42 +515,79 @@ const Page = () => {
...
+                    }
+                  />
+                  <GenericMultiselect<ColumnOption>
+                    options={getFieldsMapped(fields)}
+                    selectedItems={selectedColumns}
+                    setSelectedItems={(selectedColumns) => {
+                      setSelectedColumns(selectedColumns);
+                      setActivePage(0);
+                    }}
+                    disabled={(entries.length === 0 && !entriesLoading) || !selectedContentType}
```
JuliRossi, line 555: 💭 We could use the same disable for both filters

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -443,42 +515,79 @@ const Page = () => {
...
+                    placeholderConfig={{
+                      noneSelected: 'No fields selected',
+                      allSelected: 'Filter fields',
+                      singleSelected: '',
+                      multipleSelected: '',
+                    }}
+                    truncateLength={30}
```
JuliRossi: 💭  We can use same truncate

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -443,42 +515,79 @@ const Page = () => {
...
+                      noneSelected: 'No statuses selected',
+                      allSelected: 'Filter by status',
+                      singleSelected: '',
+                      multipleSelected: '',
+                    }}
+                    truncateLength={20}
+                    getItemKey={(item) => item.label}
+                    getItemValue={(item) => item.label}
+                    isItemSelected={(item, selectedItems) =>
+                      selectedItems.some((status) => status.label === item.label)
```
JuliRossi: 💭 We could change this using value for both

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -443,42 +515,79 @@ const Page = () => {
...
+                <Flex gap="spacingS" alignItems="center">
+                  <SortMenu
+                    sortOption={sortOption}
+                    onSortChange={(newSort) => {
+                      setSortOption(newSort);
+                      setActivePage(0);
+                    }}
+                    disabled={(entries.length === 0 && !entriesLoading) || !selectedContentType}
+                  />
+                  <GenericMultiselect<Status>
```
JuliRossi: 💭 FilterMultiselect

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
```diff
@@ -4,6 +4,14 @@ import { Entry, ContentTypeField, Status, Fields } from '../types';
 import { ContentTypeProps, EntryProps, QueryOptions } from 'contentful-management';
 import { BATCH_FETCHING } from './constants';
 
+export const getAllStatuses = (): Status[] => {
+  return [
+    { label: 'Draft', color: 'warning' },
+    { label: 'Changed', color: 'primary' },
+    { label: 'Published', color: 'positive' },
+  ];
```
JuliRossi: ['Draft', 'Changed', 'Warning']
```suggestion
export const getAllStatuses = (): Status[] => {
  return [
    'Draft'
    'Changed'
   'Published'
  ];
```

JuliRossi: 💭 GetStatusColor to get color status

apps/bulk-edit/src/locations/Page/components/SortMenu.tsx
FBanfi: Erin updated the design, since this PR is getting too long. I will implement that in another PR.

apps/bulk-edit/src/locations/Page/components/GenericMultiselect.tsx
FBanfi: The update of this comment:
- We met and paired this, now we introduced some refactors but some others are going to be introduced on another ticket

apps/bulk-edit/src/locations/Page/components/GenericMultiselect.tsx
FBanfi: Done ✅

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: Done ✅

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: Done ✅

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: Done ✅

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: Done ✅

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: I used this approach:
`const hasDraft = selectedStatuses.includes(DRAFT_STATUS)`

And in another PR, we will add the color refactor.

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
FBanfi: responded here: https://github.com/contentful/apps/pull/10097#discussion_r2331121170

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: responded here: https://github.com/contentful/apps/pull/10097#discussion_r2331121170

apps/bulk-edit/src/locations/Page/components/GenericMultiselect.styles.ts
JuliRossi, line 1: 💭  We should rename this as well. Unsure if we need this file, tho

JuliRossi (approved):
👍

apps/bulk-edit/src/locations/Page/components/GenericMultiselect.styles.ts
FBanfi: Oh yeah! Changed ✅

PR #10108, opened 2025-09-10
feat: MS-teams-new-app-action-specs [EXT-6729]

michaelpineirocontentful:
## Purpose

- Update app actions to utilize the new (in-development) specifications. This will give us useful feedback on the user experience for upgrading.

## Breaking Changes

None

Updated contentful-management


apps/microsoft-teams/app-actions/src/actions/list-channels.ts
```diff
@@ -1,30 +1,39 @@
 import { AppActionCallContext } from '@contentful/node-apps-toolkit';
-import { AppActionCallResponse, Channel } from '../types';
+import { Channel, AppActionCallResponse } from '../types';
 import { fetchTenantId } from '../utils';
 import helpers from '../helpers';
-import { withAsyncAppActionErrorHandling } from '../helpers/error-handling';
+import { ApiError } from '../errors';
 
-export const handler = withAsyncAppActionErrorHandling(
```
jsdalton, line 7: I'm _somewhat_ partial to this handler wrapper (probably because I wrote it 😬 ) but that's mainly because it "guarantees" a result type response to be returned from the handler (ie since all unhandled exceptions are caught and recast to that object.

With the refactor we lose that guarentee -- eg if fetchTenantId fails for unknown reasons then an unexpected result is returned.

This isn't necessarily a bad thing -- because now we have this amazing new result type behavior baked into our API. However, that implies two things:

1. We would need to adapt the upstream consumer(s) of this app action to work with that new result type.
2. The addition of the try/catch and the unknown error we raise there would not be needed. We'd just want to let the unhandled exception bubble up.

apps/microsoft-teams/frontend/src/components/config/NotificationEditMode/NotificationEditMode.tsx
```diff
@@ -60,24 +61,22 @@ const NotificationEditMode = (props: Props) => {
...
         }
       );
-      const body = JSON.parse(response.body);
 
-      if (body.ok) {
+      if (result && isSendTestMessageResult(result) && result.ok) {
         sdk.notifier.success(editModeFooter.testSuccess);
-      } else {
-        throw new Error(`${editModeFooter.testError}: ${body.error.message}`);
+      } else if (error) {
```
jsdalton: By using an `else if` here we have a non exhaustive conditional: That is, it's possible e.g. for us to have a result but not eg result.ok, and no error, thus leading to a silent failure.

I'd recommend we either:

* include an `else` and throw a generic error if we get there (since it's quite likely an application error at that point)
* handle the other cases within the result explicitly (eg. what should we do if the test message result is nto right).

apps/microsoft-teams/frontend/src/components/config/NotificationEditMode/NotificationEditMode.tsx
```diff
@@ -60,24 +61,22 @@ const NotificationEditMode = (props: Props) => {
...
-          userId: sdk.ids.user,
         },
         {
           parameters,
         }
       );
-      const body = JSON.parse(response.body);
 
-      if (body.ok) {
+      if (result && isSendTestMessageResult(result) && result.ok) {
```
jsdalton: result.ok should not be needed, since presumably the parser isSendTestMessageResult should do that check already.

apps/microsoft-teams/frontend/src/components/config/NotificationEditMode/NotificationEditMode.tsx
jsdalton: Yeah thinking about this more, we should use the `sys.status` as our check here for more reliable behavior.

Also, we should refactor https://github.com/contentful/contentful-management.js/blob/64a9660487d56f2e4d8fd8b7120acb4fb8b0f5f9/lib/entities/app-action-call.ts#L37 to use a discriminating union type. Technically the type definition we've provided here is "incorrect" since it allows for invalid representations. We can fix this and clean your code way up.

jsdalton: ☝️ for this last comment we can ignore it for now. Let's tackle it in a follow up

apps/microsoft-teams/frontend/src/hooks/useGetTeamsChannels.spec.ts
```diff
@@ -10,13 +10,12 @@ vi.mock('@contentful/react-apps-toolkit', () => ({
...
-        body: JSON.stringify({
-          ok: true,
-          data: mockChannels,
-        }),
+    mockSdk.cma.appActionCall.createWithResult = vi.fn().mockReturnValueOnce({
+      result: {
+        ok: true,
+        data: mockChannels,
       },
+      status: 'processing',
```
jsdalton: Hm, this is an invalid state (we cannot have a result while app action is processing) so we should change the status here to "succeeded" lest someone get confused.

apps/microsoft-teams/frontend/src/types/appActionResults.ts
```diff
@@ -0,0 +1,74 @@
...
+  };
+  error?: {
+    type: string;
+    message: string;
+    details?: Record<string, unknown>;
+  };
+}
+
+// Type guards for runtime validation
+export function isListChannelsResult(value: unknown): value is ListChannelsResult {
```
jsdalton: Note: I think we can address this in a future refactor, but I'm going to (then) suggest we convert these into assertion functions rather than type predicates.

apps/microsoft-teams/frontend/src/types/appActionResults.ts
```diff
@@ -0,0 +1,74 @@
...
+  const obj = value as Record<string, unknown>;
+
+  return (
+    typeof obj.ok === 'boolean' &&
+    ((obj.data === undefined && obj.error !== undefined) ||
+      (obj.data !== undefined && obj.error === undefined))
+  );
+}
+
+export function isChannelResult(value: unknown): value is ChannelResult {
```
jsdalton: Shower thought: As the app developer here I'm seeing a missed opportunity to align my front end and back end 🤔 .

It might not be realistic but it would be interesting to imagine:

* Declare my result schemas as constants somewhere. Share the code between app action and front end.
* For app action, use that result schema for the app action result schema that's bundled and pushed
* For front end, use typeschema and pass the same result schema to it to validate instead of handrolling these predicate functions.

apps/microsoft-teams/scripts/call-app-action.ts
```diff
@@ -51,6 +51,11 @@ class AppActionRunner {
         parameters: this.params,
       }
     );
+    if (result.status === 'succeeded') {
+      return result.result as AppActionCallResponse;
+    }
+
+    return undefined;
```
jsdalton: I think if i'm using this as a utiltiy and i get an error, I'd want to see the error?

apps/microsoft-teams/scripts/call-app-action.ts
```diff
@@ -51,6 +51,11 @@ class AppActionRunner {
         parameters: this.params,
       }
     );
+    if (result.status === 'succeeded') {
+      return result.result as AppActionCallResponse;
```
jsdalton: This was kind of nice just to see what I got back, the whole thing. So maybe just return the result?

apps/microsoft-teams/scripts/call-app-action.ts
```diff
@@ -38,8 +38,8 @@ class AppActionRunner {
 
   private async callAppAction(
     appAction: AppActionProps
-  ): Promise<WebhookCallDetailsProps | undefined> {
-    return await this.client.appActionCall.createWithResponse(
+  ): Promise<AppActionCallResponse | undefined> {
+    const result = await this.client.appActionCall.createWithResult(
```
jsdalton: I think calling this `result` and having to do `result.result` might melt some brains. Maybe it's just the `appActionCall`?

jsdalton (changes requested):
A few suggested changes from me, but let's definitely defer the ones related to refactoring the `result` vs. `error` stuff, since I think we can handle that in the CMA.

There are also some optional suggestions in there as well.

apps/microsoft-teams/frontend/src/components/config/NotificationEditMode/NotificationEditMode.tsx
michaelpineirocontentful: Resolving as the assertion logic has been overhauled

apps/microsoft-teams/frontend/src/types/appActionResults.ts
michaelpineirocontentful: Did this with `types.ts` in the root level of the project. Only a few types in there as of now, but a good structure for future "duplicate" work between the frontend and backend. I did not define additional schemas there, as they are not used in the app-action calls themselves. Instead I added `frontend/src/types/listChannelsAppActionResults.ts` and `frontend/src/types/sendTestMessageAppActionResults.ts` and their corresponding tests.

apps/microsoft-teams/frontend/package.json
```diff
@@ -6,16 +6,17 @@
     "@azure/msal-browser": "^3.7.1",
     "@azure/msal-react": "^2.0.10",
     "@contentful/app-scripts": "^1.30.1",
-    "@contentful/app-sdk": "^4.23.1",
+    "@contentful/app-sdk": "^4.42.0",
     "@contentful/f36-components": "4.67.0",
     "@contentful/f36-icons": "^4.27.0",
     "@contentful/f36-tokens": "4.0.5",
```
jjolton-contentful, line 12: curious if frontend needs an explicit type-fest dep

apps/microsoft-teams/frontend/src/types/listChannelsAppActionsResults.spec.ts
jjolton-contentful, line 1: mega nit spec filename uses plural "Actions" while source file uses singular "Action"

apps/microsoft-teams/frontend/package.json
jjolton-contentful: likely not causing any issues now because the dep gets hoisted but might cause issues in the future, probably worth adding

apps/microsoft-teams/frontend/src/hooks/useGetTeamsChannels.ts
```diff
@@ -12,24 +18,44 @@ const useGetTeamsChannels = () => {
...
         {
           parameters: {},
         }
       );
-      const body = JSON.parse(response.body);
-      if (body.ok) {
-        setChannels(body.data);
-        setError(undefined);
+
+      if (appActionCallSys.status === 'succeeded') {
```
jsdalton, line 29: Mega dump of some thoughts via code. Let's unpack in a sync discussion:

```ts
interface ActionError {
  type: string;
  message: string;
}
interface AppActionResultSuccess<TResult> {
  ok: true;
  data: TResult;
}
interface AppActionResultError {
  ok: false;
  error: ActionError;
}
type AppActionResult<T> = AppActionResultSuccess<T> | AppActionResultError;


      function assertAppActionResult<T>(
        value: unknown
      ): asserts value is AppActionResult<T> {
        /* eslint-disable lingui/no-unlocalized-strings */
        if (!value) {
          throw new TypeError('value undefined or null');
        }
        if (typeof value !== 'object') {
          throw new TypeError('value is not an object');
        }
        if (!('ok' in value)) {
          throw new TypeError('`ok` missing in value');
        }
      }

      const appActionCall = await sdk.cma.appActionCall.createWithResult(
        {
          appActionId: 'msteamsListChannels',
          environmentId: sdk.ids.environment,
          spaceId: sdk.ids.space,
          appDefinitionId: sdk.ids.app!,
        },
        {
          parameters: {},
        }
      );

      if (appActionCall.sys.status === 'succeeded') {
        const { sys: { result } } = appActionCall
        assertAppActionResult<TeamsChannel[]>(result)

        if (result.ok) {
          const data = getListChannelsData(result);
          setChannels(data || []);
          setError(undefined);
        } else {
          const errorData = getListChannelsError(appActionCallSys.result);
          const error = new Error(`Failed to fetch Teams channels: ${errorData.message}`);
          setError(error);
          throw error;
        }
      } else if (appActionCall.sys.status === 'failed') {
        const errorMessage = new Error(
          `There was an error invoking the app action: ${appActionCall.sys.error?.message}`
        );
        setError(errorMessage);
        throw errorMessage;
      } else {
        const error = new Error('Failed to fetch Teams channels. An unknown error occurred.');
        setError(error);
        throw error;
      }
    }
```

PR #10110, opened 2025-09-11
Keyboard accesibility for bulk edit [MAPS-29]

JuliRossi:
## Purpose

Keyboard navigation for bulk edit. This is part of bulk edit v1.4

## Approach

Add the following keyboad
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-33978269-7fff-9dc5-75ca-d745f1cc5c71"><div dir="ltr" style="margin-left:0pt;" align="left">
Action | Key command
-- | --
Move focus (1 cell at a time) | Arrow keys
Select checkbox/open link | Enter/Space
Select entire column | Alt+Space
Extend selection (by one cell) | Shift+Arrow
Extend selection (to edge of region) | Alt+Shift+Arrow

</div></b>

## Testing steps

https://github.com/user-attachments/assets/50f8c365-eba1-4f04-b9db-9c60eee5be05

### DIFF
--- a/apps/bulk-edit/src/locations/Page/components/TableHeader.styles.ts
+++ b/apps/bulk-edit/src/locations/Page/components/TableHeader.styles.ts
@@ -1,5 +1,6 @@
 import tokens from '@contentful/f36-tokens';
-import { CELL_WIDTH, STICKY_SPACER_SPACING, stickyCell } from '../styles';
+import { stickyCell } from '../styles';
+import { CELL_WIDTH, STICKY_SPACER_SPACING } from '../utils/constants';
 
 export const headerStyles = {
   tableHead: {
--- a/apps/bulk-edit/src/locations/Page/components/TableRow.styles.ts
+++ b/apps/bulk-edit/src/locations/Page/components/TableRow.styles.ts
@@ -1,5 +1,6 @@
 import tokens from '@contentful/f36-tokens';
-import { CELL_WIDTH, stickyCell, STICKY_SPACER_SPACING } from '../styles';
+import { CELL_WIDTH, STICKY_SPACER_SPACING } from '../utils/constants';
+import { stickyCell } from '../styles';
 
 export const rowStyles = {
   cell: {


apps/bulk-edit/src/locations/Page/components/TableHeader.tsx
```diff
@@ -1,32 +1,85 @@
...
   fields,
   headerCheckboxes,
   onHeaderCheckboxChange,
   checkboxesDisabled,
+  focusedCell,
+  selectionRange,
+  onCellFocus,
 }) => {
+  const isCellFocused = (columnIndex: number) => {
+    return focusedCell?.row === HEADERS_ROW && focusedCell?.column === columnIndex;
```
FBanfi: 💭 We could create a utils function to check this `focusedCell?.row === HEADERS_ROW` since we are doing this header row check multiple times in other places.

apps/bulk-edit/src/locations/Page/components/TableHeader.tsx
```diff
@@ -1,32 +1,85 @@
...
+  };
+
+  const isCellSelected = (columnIndex: number) => {
+    if (!selectionRange) return false;
+    const { start, end } = selectionRange;
+    const minRow = Math.min(start.row, end.row);
+    const maxRow = Math.max(start.row, end.row);
+    const column = start.column; // Single column selection
+
+    return -1 >= minRow && -1 <= maxRow && columnIndex === column;
```
FBanfi: 💭 Just curious about what does `-1 >= minRow && -1 <= maxRow` mean? 

Is `-1` the `HEADERS_ROW` constant? If so, perhaps we could use the constant instead.

apps/bulk-edit/src/locations/Page/components/TableRow.tsx
```diff
@@ -35,63 +41,108 @@ export const TableRow: React.FC<TableRowProps> = ({
...
+  const isCellSelected = (columnIndex: number) => {
+    if (!selectionRange) return false;
+    const { start, end } = selectionRange;
+    const minRow = Math.min(start.row, end.row);
+    const maxRow = Math.max(start.row, end.row);
+
+    return rowIndex >= minRow && rowIndex <= maxRow && columnIndex === start.column;
+  };
+
+  const getColumnIndex = (field: ContentTypeField | string) => {
```
FBanfi: We could move this to a utils file too and reuse it here and in `TableHeader`
```
const getColumnIndex = (field: ContentTypeField | string) => {
    const fieldId = typeof field === 'string' ? field : field.uniqueId;
    const allColumns = [DISPLAY_NAME_COLUMN, ENTRY_STATUS_COLUMN, ...fields.map((f) => f.uniqueId)];
    return allColumns.indexOf(fieldId);
  };
```

apps/bulk-edit/src/locations/Page/components/TableHeader.tsx
FBanfi: If we create a variable above this method such as
`const rowIndex = -1;'

and move the method to a utils files we could reuse this method here and in `TableRow`
 ```
const isCellSelected = (columnIndex: number) => {
    if (!selectionRange) return false;
    const { start, end } = selectionRange;
    const minRow = Math.min(start.row, end.row);
    const maxRow = Math.max(start.row, end.row);

    return rowIndex >= minRow && rowIndex <= maxRow && columnIndex === start.column;
  };
```

File: apps/bulk-edit/src/locations/Page/styles.ts [hunk 4]
```diff
@@ -32,54 +32,12 @@ export const styles = {
     marginTop: tokens.spacingM,
     minWidth: `${CELL_WIDTH * 4}px`,
     overflowX: 'auto',
-  },
```
FBanfi, line 38: Oh this is awesome, much cleaner! 👏 💯

apps/bulk-edit/src/locations/Page/hooks/useKeyboardNavigation.ts
FBanfi, line 1: I like how you wrapped the logic into a separate custom hook, its usage in the other components looks nice. Even though this code seems a bit complex, I don't come up with a simpler approach so this looks great to me! 💯 .

apps/bulk-edit/src/locations/Page/hooks/useKeyboardNavigation.ts
```diff
@@ -0,0 +1,255 @@
+import { useState, useCallback, useEffect, useRef } from 'react';
+import { HEADERS_ROW } from '../utils/constants';
+
+export interface FocusPosition {
```
joaquincasal, line 4: nit: I'd use types whenever possible

apps/bulk-edit/src/locations/Page/components/TableHeader.tsx
JuliRossi: Def agree! Good catch. I'll work to remove duplication

joaquincasal (approved):
Great great work, I really like having the logic extracted to the custom hook so existing components get as few changes as possible

PR #10109, opened 2025-09-11
Bulk-Edit-App: search bar + reset button + pending refactors [MAPS-34]

FBanfi:
## Purpose

This update includes the following things:
- search bar filter (this should filter through the query)
- reset filter button
- pending refactors: [comment of the PR talking about the pending things](https://github.com/contentful/apps/pull/10097#issuecomment-3267603240) 
    - divide color from status label in entry utils (create a getColorStatus)
    - fix new bug detected changing multiple time from example A content type to another (Bulk Edit Development space, master enviroment)
    - update in empty states and disabled filters from figma

## Approach

- New component was added for the `SearchFilter`.
- New component was created to simplify the logic of the two different empty states
- Refactors to fix the bugs were introduced
- The entry utils was modified to divide the getStatus and get the color through another method.

## Testing steps

Some automated test were added for the `SearchFilter` component.

Search Filter component:

https://github.com/user-attachments/assets/b578af46-9c33-47db-9e96-2cc7890cd1ff

Fixed bug:

https://github.com/user-attachments/assets/e11d351f-c36c-42b3-8691-725f84c91449

Empty states:

https://github.com/user-attachments/assets/9d9df660-7e32-4006-b67a-3f739bf890ca

Reset button:

https://github.com/user-attachments/assets/5f955e3b-153f-44f5-b4f4-602c7cba5675

## Breaking Changes

N/A

## Dependencies and/or References

Link to [MAPS-34](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?selectedIssue=MAPS-34)

## Deployment

N/A

[MAPS-34]: https://contentful.atlassian.net/browse/MAPS-34?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10111, opened 2025-09-11
Bulk edit v1.3 []

JuliRossi:
## Purpose

Adding new filters for bulk edit 

## Approach

Mainly adding:
- Status filter
- SearchBar
- Filter reset
- Improvements in the empty states for bulk edit

PRs in this version:
 - https://github.com/contentful/apps/pull/10097
 - https://github.com/contentful/apps/pull/10109
 
## Testing steps
(see videos in PRs)


apps/bulk-edit/src/locations/Page/components/FilterMultiselect.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface FilterMultiselectProps {
+  id?: string;
+  options: FilterOption[];
+  selectedItems: FilterOption[];
+  setSelectedItems: (items: FilterOption[]) => void;
+  disabled?: boolean;
+  placeholderConfig: {
+    noneSelected: string;
+    allSelected: string;
+    singleSelected: string;
```
ryunsong-contentful: the single and multi selected fields are never used in this file (let me know if this is for future versions)

Instead for single selected for example we just do this 
```
    if (selectedItems.length === 1) {
      return selectedItems[0].label;
    }
```

apps/bulk-edit/src/locations/Page/components/FilterMultiselect.tsx
```diff
@@ -0,0 +1,91 @@
...
+        />
+        {options.map((option) => (
+          <Multiselect.Option
+            isDisabled={disabled}
+            className={optionStyles}
+            key={option.value}
+            label={truncate(option.label, 30)}
+            value={option.value}
+            itemId={`option-${id}-${option.value}`}
+            isChecked={selectedItems.some((item) => item.value === option.value)}
```
ryunsong-contentful: We're in the safe zone in terms of performance since in most cases the number of options would probably be in the dozens max, and the max listed items are 500. (30x500=15000). 

But I'd say it's close enough that we should make some performance optimizations since technically, a user could probably make hundreds or even thousands of options (theoretically) and that's when the app will really suffer in performance.
So I'd say created a one time converted Set so that this search is constant O(1) runtime. 

This matters even more if we end up allowing 1000s of items to appear in the list

```
const selectedValuesSet = useMemo(
  () => new Set(selectedItems.map(item => item.value)),
  [selectedItems]
);
```

FBanfi: Sounds better!
Implemented ✅

apps/bulk-edit/src/locations/Page/components/FilterMultiselect.tsx
FBanfi: That was something from a previous version! Nice catch, removed the single and multi selected placeholders ✅ Ty

ryunsong-contentful (approved):
I found this issue which I was able to consistently reproduce by hitting a CT with 0 entries and then hitting the Example A CT (see video). However, I decided to uninstall and reinstall the app and could no longer reproduce - so I will recommend two action items

1. Test it yourself and see if you cannot reproduce 
2. If you can - fix it before release. If you cannot reproduce, release v1.3 and see if it happens again down the line and fix it then.

https://github.com/user-attachments/assets/499edaa4-407d-489e-981c-7bb2cb06d775

JuliRossi:
We couldn't reproduce the issue, but we'll keep an eye on it.

PR #10123, opened 2025-09-15
Bulk-Edit-App: Fix v1.3 comments [MAPS-50]

FBanfi:
## Purpose

This update was included to fix the comments of this PR: https://github.com/contentful/apps/pull/10111.

## Approach

- The second bullet of this [comment](https://github.com/contentful/apps/pull/10111#issuecomment-3282314577), will be addressed in the v1.4 (this is the [ticket](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?selectedIssue=MAPS-51))
- The search on integer fix was due to an issue with the query: it seems to be that full-text search trough query doesn't work well with pure numeric strings. The solution was to implement client-side filtering for pure numeric searches.

## Testing steps

Search on integer fix:

https://github.com/user-attachments/assets/88ded4e4-9528-40df-a3a3-21b3fce4709a

## Breaking Changes

N/A

Link to [MAPS-50](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/3109?selectedIssue=MAPS-50)

## Deployment

N/A


apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -134,9 +136,9 @@ const Page = () => {
       return { 'sys.id[in]': 'nonexistent-id' };
     }
 
-    const hasDraft = statuses.includes(getStatusesMapped()[0]);
-    const hasPublished = statuses.includes(getStatusesMapped()[1]);
-    const hasChanged = statuses.includes(getStatusesMapped()[2]);
+    const hasDraft = statuses.map((status) => status.label).includes(getAllStatuses()[0]);
+    const hasPublished = statuses.map((status) => status.label).includes(getAllStatuses()[1]);
+    const hasChanged = statuses.map((status) => status.label).includes(getAllStatuses()[2]);
```
FBanfi: this was changed because `getMappedStatuses()` returned a new object each time it was called, and this messed the comparison introducing a bug in the status filter.

apps/bulk-edit/src/locations/Page/index.tsx
```diff
@@ -166,14 +168,17 @@ const Page = () => {
...
 
   function needsClientFiltering() {
-    const hasDraft = selectedStatuses.includes(getStatusesMapped()[0]);
-    const hasPublished = selectedStatuses.includes(getStatusesMapped()[1]);
-    const hasChanged = selectedStatuses.includes(getStatusesMapped()[2]);
+    const hasDraft = selectedStatuses.map((status) => status.label).includes(getAllStatuses()[0]);
+    const hasPublished = selectedStatuses
+      .map((status) => status.label)
+      .includes(getAllStatuses()[1]);
+    const hasChanged = selectedStatuses.map((status) => status.label).includes(getAllStatuses()[2]);
```
FBanfi: same as the previous comment

apps/bulk-edit/src/locations/Page/index.tsx
JuliRossi: Instead of doing this and mapping it multiple times, you could map the labels only once where this function is used similar and keep it similar as it was. Something like 
`const statusLabels = statuses.map(s => s.label)`
 Another thing: I rather have this 
`const hasDraft = statuses.includes ('Draft')` or `const hasDraft = statuses.includes (DRAFT_STATUS)` because is more expresive . `getStatusesMapped()[0] `doesn't give me a lot of information

apps/bulk-edit/src/locations/Page/index.tsx
JuliRossi: Same here, but the mapping has to happen here instead of outside the function

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
```diff
@@ -272,3 +272,29 @@ export async function fetchEntriesWithBatching(
...
+  return /^\d+$/.test(query.trim());
+};
+
+export const filterEntriesByNumericSearch = (
+  entries: EntryProps[],
+  query: string,
+  fields: ContentTypeField[],
+  defaultLocale: string
+): EntryProps[] => {
+  if (!query.trim() || !isNumericSearch(query)) return entries;
```
JuliRossi: 💭 Because of the name of the function, I feel this validation should happen outside the function

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
```diff
@@ -272,3 +272,29 @@ export async function fetchEntriesWithBatching(
...
+
+  return entries.filter((entry) => {
```
JuliRossi, line 294: bummer there isn't a way to fix this using the cma, but good workaround

apps/bulk-edit/src/locations/Page/index.tsx
FBanfi: Sure, I like the refactor 👌. Thanks 🤝

apps/bulk-edit/src/locations/Page/utils/entryUtils.ts
FBanfi: 👍  Done ✅

PR #10133, opened 2025-09-25
feat/EXT-6729/app-action-createWithResult

michaelpineirocontentful:
## Purpose

- Update app actions to utilize the new (in-development) specifications (`createWithResult`). This will give us useful feedback on the user experience for upgrading.

- Updated `contentful-management`


BobHemphill76 (approved):

apps/microsoft-teams/app-actions/src/actions/list-channels.ts
```diff
@@ -1,22 +1,20 @@
...
     const {
       cma,
-      appActionCallContext: { appInstallationId, userId, environmentId, spaceId },
+      appActionCallContext: { appInstallationId: appDefinitionId, userId, environmentId, spaceId },
     } = context;
 
-    const tenantId = await fetchTenantId(cma, appInstallationId);
-    const channels = await helpers.getChannelsList(tenantId, {
-      appInstallationId,
+    const tenantId = await fetchTenantId(cma, appDefinitionId);
```
jsdalton: Subtle thing but I think we should keep the current naming (appInstlaationId) instead of this change.

Concretely fetchTenantId asks for an app installation id, and even conceptually, the specific tenant id is stored as part of the installation, not the app definition.

apps/microsoft-teams/frontend/src/hooks/useGetTeamsChannels.ts
```diff
@@ -12,24 +14,42 @@ const useGetTeamsChannels = () => {
...
-      if (body.ok) {
-        setChannels(body.data);
-        setError(undefined);
+
+      if (appActionCallSys.status === 'succeeded') {
+        const { result: appActionResult } = appActionCallSys;
+        assertAppActionResult<TeamsChannel[]>(appActionResult);
+
+        if (appActionResult.ok) {
+          setChannels((appActionResult as AppActionResultSuccess<TeamsChannel[]>).data || []);
```
jsdalton: Hm, this type assertion should not be needed? TS should already have the correct type via the assertion above.

jsdalton (approved):
Looks good! Just a few super minor comments and suggestions.

PR #10139, opened 2025-09-30
Bulk edit: Insert columns in initial order after filtering [MAPS-41]

joaquincasal:
## Purpose

Right now if you deselect a column and add it again, it will be added at the end of the table instead of its original position

## Approach

Every time we insert a column into the selectedColumns state, we reorder it

## Testing steps

Remove and add a column from the columns filter

https://github.com/user-attachments/assets/00601089-e345-4e06-9b5a-6791e9c3e627


ryunsong-contentful (commented):
Looking at this PR to review the column ordering implementation...

ryunsong-contentful (approved):
Nice

PR #10145, opened 2025-10-03
Closest Preview: Add Sidebar UI [MAPS-58]

joaquincasal:
## Purpose

Add UI for the app sidebar.

## Approach

This only includes how the sidebar looks, the behaviour of showing the correct link will be handled in a separate PR.

## Testing steps

Automatic tests were added.

Screenshot 2025-10-03 at 16 01 46

We display entries id for now, when we have to fetch the actual entries we'll use entry titles.

### DIFF
--- a/apps/closest-preview/test/locations/Sidebar.spec.tsx
+++ b/apps/closest-preview/test/locations/Sidebar.spec.tsx
@@ -9,12 +9,12 @@ vi.mock('@contentful/react-apps-toolkit', () => ({
 }));
 
 describe('Sidebar component', () => {
-  it('Renders entry list with titles and relative dates', async () => {
+  it('Renders entry list with ids and relative dates', async () => {
     const { getAllByText } = render(<Sidebar />);
 
     await waitFor(() => {
-      const entryTitles = getAllByText('Entry id', { exact: false });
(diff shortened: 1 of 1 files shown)


apps/closest-preview/test/locations/Sidebar.spec.tsx
```diff
@@ -1,17 +1,23 @@
...
 
-    expect(getByText('Hello Sidebar Component (AppId: test-app)')).toBeInTheDocument();
+    await waitFor(() => {
+      const entryTitles = getAllByText('Entry id', { exact: false });
+      expect(entryTitles).toHaveLength(5);
+    });
+
+    const updatedTexts = getAllByText(/Updated/);
+    expect(updatedTexts).toHaveLength(5);
   });
```
FBanfi: **Nitpick**: Perhaps I'll change the title of this test to 'Renders entry list with ids and relative dates', and  when we change the logic to use the titles come back here and change it. 
Because currently we are rendering a entry list with ids.

Same with the
`const entryTitles = getAllByText('Entry id', { exact: false });` 
to
`const entryIds = getAllByText('Entry id', { exact: false });`

PR #10149, opened 2025-10-07
Closest Preview: Adding backlinks logic [MAPS-60]

FBanfi:
## Purpose

This update includes the backlinks logic to get the links to the last five updated entries with Live Preview configured.

## Approach

Some logic was reused from the original app but replacing the cma client for the sdk.

## Testing steps

Automated tests added.

Updating an entry with live preview so it appears in the sidebar:

https://github.com/user-attachments/assets/cec06f0d-2c2c-4152-9645-d2dffc19b1fa

Depth message and check when in higher than 10:

https://github.com/user-attachments/assets/30b67773-1492-408b-9893-491204484763

## Breaking Changes

N/A

Link to [MAPS-60](https://contentful.atlassian.net/jira/software/c/projects/MAPS/boards/4363)

## Deployment

N/A

### DIFF
--- a/apps/closest-preview/src/locations/Sidebar.tsx
+++ b/apps/closest-preview/src/locations/Sidebar.tsx
@@ -9,7 +9,7 @@ import {
 import { ArrowSquareOutIcon } from '@contentful/f36-icons';
 import { SidebarAppSDK } from '@contentful/app-sdk';
 import { useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
-import { EntryProps, QueryOptions } from 'contentful-management';
+import { EntryProps, KeyValueMap } from 'contentful-management';
 import { useCallback, useEffect, useState } from 'react';
 
 const MAX_DEPTH = 5;
@@ -43,6 +43,26 @@ const Sidebar = () => {
     [sdk.ids.space, sdk.ids.environment]
   );
 
+  const splitEntriesFromRoot = (
+    entry: EntryProps<KeyValueMap>,
+    checkedEntries: Set<string>,
+    rootEntryData: EntryProps[]
+  ) => {
+    const entryId = entry?.sys?.id;
+    if (!entryId || checkedEntries.has(entryId)) {
+      return false;
+    }
+
+    checkedEntries.add(entryId);
+    const slug = entry.fields.slug?.[defaultLocale];
+    if (slug) {
+      rootEntryData.push(entry);
+      return false;
+    }
+
+    return true;
+  };
+
   const getUpstreamEntries = useCallback(
     async (id: string): Promise<EntryProps[]> => {
       const rootEntryData: EntryProps[] = [];
@@ -70,16 +90,7 @@ const Sidebar = () => {
 
         childEntries = relatedEntries.flatMap((rEntry) =>
           rEntry.filter((item: EntryProps) => {
-            if (item && item.sys.id && !checkedEntries.has(item.sys.id)) {
-              checkedEntries.add(item.sys.id);
-              const slug = item.fields.slug?.[defaultLocale];
-              if (slug) {
-                rootEntryData.push(item);
-                return false;
-              }
-              return true;
-            }
-            return false;
+            return splitEntriesFromRoot(item, checkedEntries, rootEntryData);
           })
         );
 
--- a/apps/closest-preview/src/locations/Sidebar.tsx
+++ b/apps/closest-preview/src/locations/Sidebar.tsx
@@ -4,6 +4,7 @@ import {
   Paragraph,
   RelativeDateTime,
   Skeleton,
+  Text,
   TextLink,
 } from '@contentful/f36-components';
 import { ArrowSquareOutIcon } from '@contentful/f36-icons';
@@ -12,13 +13,14 @@ import { useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
 import { EntryProps, KeyValueMap } from 'contentful-management';
 import { useCallback, useEffect, useState } from 'react';
 
-const MAX_DEPTH = 5;
+const MAX_DEPTH = 10;
 
 const Sidebar = () => {
   const sdk = useSDK<SidebarAppSDK>();
   useAutoResizer();
   const [entries, setEntries] = useState<EntryProps[]>([]);
   const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [maxDepthReached, setMaxDepthReached] = useState<boolean>(false);
   const defaultLocale = sdk.locales.default;
 
   const getRelatedEntries = useCallback(
@@ -63,13 +65,14 @@ const Sidebar = () => {
     return true;
   };
 
-  const getUpstreamEntries = useCallback(
-    async (id: string): Promise<EntryProps[]> => {
+  const getRootEntries = useCallback(
+    async (id: string): Promise<{ entries: EntryProps[]; maxDepthReached: boolean }> => {
       const rootEntryData: EntryProps[] = [];
       let childEntries: EntryProps[] = [];
       const checkedEntries: Set<string> = new Set([id]);
       let depth = 0;
       const maxDepth = MAX_DEPTH;
+      let depthReached = false;
 
       try {
         const initialEntry = await sdk.cma.entry.get({
@@ -80,7 +83,7 @@ const Sidebar = () => {
         childEntries = [initialEntry];
       } catch (error) {
         console.error('Failed to fetch initial entry:', error);
-        return [];
+        return { entries: [], maxDepthReached: false };
       }
 
       while (childEntries.length > 0 && depth < maxDepth) {
@@ -98,10 +101,10 @@ const Sidebar = () => {
       }
 
       if (depth >= maxDepth && rootEntryData.length === 0) {
-        console.log(`Max depth of ${maxDepth} reached for entry ID: ${id}`);
+        depthReached = true;
       }
 
-      return rootEntryData;
+      return { entries: rootEntryData, maxDepthReached: depthReached };
     },
     [getRelatedEntries, sdk.ids.space, sdk.ids.environment, defaultLocale]
   );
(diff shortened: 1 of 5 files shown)


apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,23 +1,137 @@
...
+    const slug = entry.fields.slug?.[defaultLocale];
+    if (slug) {
+      rootEntryData.push(entry);
+      return false;
+    }
+
+    return true;
+  };
+
+  const getRootEntries = useCallback(
```
joaquincasal: We could move the three functions to an utils file so it's separate from the component

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,23 +1,137 @@
...
   useAutoResizer();
   const [entries, setEntries] = useState<EntryProps[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [maxDepthReached, setMaxDepthReached] = useState<boolean>(false);
+  const defaultLocale = sdk.locales.default;
+
+  const getRelatedEntries = useCallback(
+    async (id: string): Promise<EntryProps[]> => {
+      try {
+        const response = await sdk.cma.entry.getMany({
```
joaquincasal: The original app used thre Preview API, we may want to check if there was a reason for using that one in particular or if we're okay with the CMA

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -32,8 +146,7 @@ const Sidebar = () => {
                 rel="noopener noreferrer"
                 icon={<ArrowSquareOutIcon />}
                 alignIcon="end">
-                {/* TODO: get the title of the entry */}
-                {entry.sys.id.slice(0, 8)}
+                {entry.fields.title?.[defaultLocale] || entry.sys.id.slice(0, 8)}
```
joaquincasal: We should get the content type to get the displayField and access it with entry.fields[displayField] (we should probablt use a helper function)

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,23 +1,137 @@
...
+    return (
+      <Skeleton.Container>
+        <Skeleton.BodyText numberOfLines={3} />
+      </Skeleton.Container>
+    );
+  }
+
+  if (maxDepthReached) {
+    return <Text>Max depth of {MAX_DEPTH} reached for entry.</Text>;
+  }
```
joaquincasal: Couldn't we just stop searching and show the results up to that point instead of not showing anything?

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,23 +1,137 @@
...
+          environmentId: sdk.ids.environment,
+        });
+
+        return response.items;
+      } catch (error) {
+        console.error(error);
+        return [];
+      }
+    },
+    [sdk.ids.space, sdk.ids.environment]
```
joaquincasal: I don't think these will change while using the app, not sure if the useCallback is needed

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,23 +1,137 @@
...
+        depth++;
+      }
+
+      if (depth >= maxDepth && rootEntryData.length === 0) {
+        depthReached = true;
+      }
+
+      return { entries: rootEntryData, maxDepthReached: depthReached };
+    },
+    [getRelatedEntries, sdk.ids.space, sdk.ids.environment, defaultLocale]
```
joaquincasal: I don't think these will change while using the app, not sure if the useCallback is needed

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,23 +1,137 @@
...
+    return true;
+  };
+
+  const getRootEntries = useCallback(
+    async (id: string): Promise<{ entries: EntryProps[]; maxDepthReached: boolean }> => {
+      const rootEntryData: EntryProps[] = [];
+      let childEntries: EntryProps[] = [];
+      const checkedEntries: Set<string> = new Set([id]);
+      let depth = 0;
+      const maxDepth = MAX_DEPTH;
```
joaquincasal: Are we updating this constant somewhere? If not, I think we could use MAX_DEPTH directly.
Update: do we even need to enforce the max depth? Not sure if there's a reason for stopping the search

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,23 +1,137 @@
...
+  const defaultLocale = sdk.locales.default;
+
+  const getRelatedEntries = useCallback(
+    async (id: string): Promise<EntryProps[]> => {
+      try {
+        const response = await sdk.cma.entry.getMany({
+          query: {
+            links_to_entry: id,
+            order: '-sys.updatedAt',
+            limit: 5,
```
joaquincasal: At this point I think we want all the related entries.
Actually, I think with this implementation we could end up with more than five entries in the `rootEntryData` array. Filtering and ordering needs to happen as the last step

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,23 +1,137 @@
...
+        return { entries: [], maxDepthReached: false };
+      }
+
+      while (childEntries.length > 0 && depth < maxDepth) {
+        const relatedEntries = await Promise.all(
+          childEntries.map((entry) => getRelatedEntries(entry.sys.id))
+        );
+
+        childEntries = relatedEntries.flatMap((rEntry) =>
+          rEntry.filter((item: EntryProps) => {
```
joaquincasal: Maybe instead of filtering we could create two partitions of the rEntry array: the entries that have a slug and the ones that do not. And then we can handle the logic of add the ones that do into the `rootEntryData` array in this method, so the `splitEntriesFromRoot` method doesn't have that side effect

apps/closest-preview/src/locations/Sidebar.tsx
FBanfi: After asking, there wasn't a specific reason so I used the SDK CMA to avoid needing the preview api key in the config screen. ✅

apps/closest-preview/src/locations/Sidebar.tsx
FBanfi: Not needed anymore, I moved it in the helper file so I removed it. ✅

apps/closest-preview/src/locations/Sidebar.tsx
FBanfi: Not needed any more, removed it 👍 ✅

apps/closest-preview/src/locations/Sidebar.tsx
FBanfi: Really nice catch, after trying it, there is no reason to stop the search so I removed all the `depth` logic. Everything seems to work fine without that and the code looks much simpler!  💯 💯

apps/closest-preview/src/locations/Sidebar.tsx
FBanfi: Sure, now we do the filter manually at the end and the `get` to the `cma` has no limit. ✅

apps/closest-preview/src/locations/Sidebar.tsx
FBanfi: Sounds good. Implemented ✅

apps/closest-preview/src/locations/Sidebar.tsx
```diff
@@ -1,24 +1,50 @@
...
+import { getRootEntries } from '../utils/livePreviewUtils';
+import { getContentTypesForEntries, getDisplayField } from '../utils/entryUtils';
 
 const Sidebar = () => {
   const sdk = useSDK<SidebarAppSDK>();
   useAutoResizer();
   const [entries, setEntries] = useState<EntryProps[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [contentTypes, setContentTypes] = useState<Record<string, any>>({});
+  const defaultLocale = sdk.locales.default;
```
joaquincasal: We're using this just one time, I don't think we need a variable

apps/closest-preview/src/utils/livePreviewUtils.ts
```diff
@@ -40,3 +41,82 @@ export const getContentTypesWithoutLivePreview = async (
     return [];
   }
 };
+
+export const getRelatedEntries = async (sdk: SidebarAppSDK, id: string): Promise<EntryProps[]> => {
+  try {
+    const response = await sdk.cma.entry.getMany({
+      query: {
+        links_to_entry: id,
+        order: '-sys.updatedAt',
```
joaquincasal: Not sure if this changes anything, but I think it's not needed

apps/closest-preview/src/utils/livePreviewUtils.ts
```diff
@@ -40,3 +41,82 @@ export const getContentTypesWithoutLivePreview = async (
...
+    });
+
+    return response.items;
+  } catch (error) {
+    console.error(error);
+    return [];
+  }
+};
+
+export const hasSlug = (entry: EntryProps<KeyValueMap>, defaultLocale: string): boolean => {
```
joaquincasal: I'd change the name of this one to `hasLivePreview` which is what we're intersted in. The slug thing is just the implementation

apps/closest-preview/src/utils/livePreviewUtils.ts
```diff
@@ -40,3 +41,82 @@ export const getContentTypesWithoutLivePreview = async (
...
+
+  while (childEntries.length > 0) {
+    const relatedEntries = await Promise.all(
+      childEntries.map((entry) => getRelatedEntries(sdk, entry.sys.id))
+    );
+
+    const allRelatedEntries = relatedEntries.flat();
+
+    const entriesWithSlugs: EntryProps[] = [];
+    const entriesWithoutSlugs: EntryProps[] = [];
```
joaquincasal: Same here, I would name these with something related to live preview

PR #10624, opened 2026-02-26
refactor: Mux - Remove request from the frontend and use Contentful Actions instead []

R-Delfino95:
## Purpose

Mux API calls were previously made directly from the browser, exposing API credentials client-side. This refactor moves all Mux API communication to Contentful Functions — the recommended server-side execution environment for App Framework apps. Credentials now live exclusively in appInstallationParameters (server-side).

## Approach

All standard Mux REST API calls are handled by a single generic muxProxy Function that accepts method, path, and body, forwarding the request server-side with stored credentials. onPublish and getSignedUrlTokens remain as dedicated functions due to their non-trivial server-side logic.

Signing key operations (used only during app configuration) stay as direct browser calls — routing them through a Function would cause a failure on first install, since Actions aren't available until after installation completes.

A race condition was also fixed: when switching playback policy and publishing, the frontend would resync before onPublish finished updating Mux, causing version conflicts. The frontend now skips the immediate resync if pendingActions are present and lets onPublish re-publish with fresh data first.

## Testing steps

1. Install the app fresh — enter Mux credentials, enable Signed URLs and confirm signing keys are created without errors
2. Upload a video via file and via URL — confirm it processes and plays back correctly
3. Switch playback policy (public ↔ signed) and publish — confirm the playback ID updates correctly with no "entry is outdated" error
4. Add/remove captions, static renditions, and metadata — confirm all operations work

## Breaking Changes

The app now requires Contentful Functions to be deployed and active. Existing installations must be redeployed — a frontend-only deploy would break all video operations.

## Deployment

Full redeploy of frontend + functions is required. Deploy both together to avoid any window where the frontend calls actions that don't exist yet.

A new App Action must be created manually in the Contentful dashboard:

- Action name: **muxProxy**
- Parameters schema: accept any — this action acts as a generic proxy, parameter validation is intentionally omitted
- Result schema: accept any

image

image


Jerricho93 (approved):

apps/mux/functions/src/muxProxy.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const handler: FunctionEventHandler<FunctionTypeEnum.AppActionCall> = async (
+  event: AppActionRequest<'Custom', Parameters>,
+  context: FunctionEventContext
+) => {
+  const { method, path, body } = event.body;
+  const { muxAccessTokenId, muxAccessTokenSecret } = context.appInstallationParameters;
+
+  console.log(`[muxProxy] ${method} ${path}`, body ? `body length: ${body.length}` : 'no body');
```
david-shibley-contentful: Do we mean to leave these logs in?

apps/mux/functions/src/muxProxy.ts
```diff
@@ -0,0 +1,61 @@
...
+      method,
+      path,
+      body
+    );
+  } catch (err) {
+    console.error(`[muxProxy] Network error on ${method} ${path}:`, err);
+    return { ok: false, error: 'Network error calling Mux API', status: 502 };
+  }
+
+  console.log(`[muxProxy] ${method} ${path} → ${res.status}`);
```
david-shibley-contentful: same question here

david-shibley-contentful (approved):
LGTM! Just a question about logs but non-blocking

david-shibley-contentful:
Looks like we have some prettier violations:

From github.com:**********/apps
 * branch                master     -> FETCH_HEAD
apps/mux/frontend/src/index.****.tsx apps/mux/frontend/src/index.tsx apps/mux/frontend/src/util/apiClient.tsx apps/mux/frontend/src/util/muxApi.ts apps/mux/frontend/src/util/types.tsx apps/mux/functions/src/helpers/muxClient.ts apps/mux/functions/src/muxProxy.ts apps/mux/functions/src/onPublish.ts
Running 'prettier -l' against changed files to check for problems...


prettier violations found:
apps/mux/frontend/src/index.****.tsx
apps/mux/frontend/src/index.tsx
apps/mux/frontend/src/util/muxApi.ts
apps/mux/functions/src/onPublish.ts

david-shibley-contentful:
one more

prettier violations found:
apps/mux/frontend/src/index.tsx

david-shibley-contentful (approved):
LGTM

PR #10783, opened 2026-03-27
feat: add phosphor icon app []

zacharyyankiver:
## Summary
- add the Phosphor Icon app to the shared apps repo
- support content type assignment, curated icon installation rules, and updated field dialog behavior
- align the config and entry flows with the reviewed UX and stronger Forma 36 patterns

## What changed
- added the app scaffold, icon catalog/search, field dialog, and config screen
- added config-driven content type assignment and cleanup when a content type is removed
- added curated icon selection, position option management, and installation parameter syncing into the field app
- improved entry-level picker behavior so selected icons stay visible while style and position are updated
- tightened Contentful UI alignment across the content type picker, preview cards, icon grid, and dialog layout

## Validation
- npm run test:ci
- local manual verification in Contentful across config, install, and field editing flows


apps/phosphor-icon/src/components/ContentTypePicker.tsx
```diff
@@ -0,0 +1,245 @@
...
+          setIsOpen((open) => {
+            const nextOpen = !open;
+            if (!nextOpen) {
+              setQuery('');
+            }
+            return nextOpen;
+          });
+        }}
+        style={{
+          width: '100%',
```
ryunsong-contentful, line 108: Try to tell the AI to use F36 token values. Like `token.small` for padding or `tokens.white100` for backgroundColor

ryunsong-contentful: Another thing to tell the AI/Codex is to put all the styles in a separate file and then import it into the component. Makes things a little cleaner and organized

apps/phosphor-icon/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,763 @@
...
+    error instanceof Error
+      ? error.message
+      : typeof error === 'object' &&
+        error !== null &&
+        'message' in error &&
+        typeof (error as { message?: unknown }).message === 'string'
+      ? (error as { message: string }).message
+      : '';
+
+  return message.toLowerCase().includes('version');
```
ryunsong-contentful, line 127: Tell the AI to return true for every possible error, and then if it passes all the checks, just return false by default. Makes this function cleaner

apps/phosphor-icon/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,763 @@
...
+      setSelectedIconNames(result);
+    }
+  }, [enabledWeights, normalizedPositionOptions, sdk.dialogs, selectedIconNames]);
+
+  const applyContentTypeAssignments = useCallback(
+    async (configParameters: AppInstallationParameters) => {
+      const assignedFields = await Promise.all(
+        contentTypes.map(async (contentType) => {
+          const assignApp = configParameters.selectedContentTypeIds?.includes(contentType.id);
+          const contentTypeDetails = (await sdk.cma.contentType.get({
```
ryunsong-contentful: This part here, since you're doing a `.map()` in line 289, this will fire off a cma get request for every content type selected. Tell Codex to do a `promise.all` on this part so that all the requests are made in parallel instead of one after the other to make things more efficient.

ryunsong-contentful (approved):
App works, left some comments for you/codex to address

mgoudy91 (approved):

PR #10865, opened 2026-04-14
feat: add configurable default for reference include depth to JSON Viewer app []

lambardi:
## Purpose

Mainly testing/proving Zach's LLM IDE skill, TBH. So I just tried to find a simple improvement I could make to an existing app, that I could use to test it out. 

No customer's have requested this, but I figure it is intuitively a subtle quality of life improvement for users of this app.

## Approach

I added a new optional config field to the apps config screen for the default value of the include reference depth. This will avoid a user having to change the value of the dropdown every time they open the tab, if they are just always setting it to something other than 0. (the default before this PR) 

## Testing steps

Set the 5th optional param in the app install/config screen to a value other than 0. 
Open the JSON Viewer tab in an entry, confirm that the default set in the config screen is used. 

## Breaking Changes

None

None

## Deployment

Nothing specific.

### DIFF
--- a/apps/json-viewer/src/locations/ConfigScreen.spec.tsx
+++ b/apps/json-viewer/src/locations/ConfigScreen.spec.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import ConfigScreen from './ConfigScreen';
-import { render } from '@testing-library/react';
+import { fireEvent, render, waitFor } from '@testing-library/react';
 import { mockCma, mockSdk } from '../../test/mocks';
 import { vi } from 'vitest';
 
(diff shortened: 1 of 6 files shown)


mgoudy91:
**Bug: type mismatch in `Select.Option` value**

In `ConfigScreen.tsx`, the new dropdown renders option values as numbers:

```tsx
{Array(11)
  .fill(0)
  .map((_, i) => (
    <Select.Option key={i} value={i}>
      Include: {i}
    </Select.Option>
  ))}
```

But `defaultIncludeDepth` is typed as `string` in `ReactJsonConfig`, and the `<Select>` `value` prop is also a string. This means the controlled select's value comparison (`'3' !== 3`) will fail — the dropdown won't reflect the correct selected option after a change.

Fix: use `value={String(i)}` on `Select.Option`.

throw codex at this and see what it thinks!

lambardi:
Thanks, Mitch! I (well Codex) updated the new config-screen select to use `value={String(i)}` and made the same change in the entry-editor include-depth select as well so both stay consistent with the controlled string value.

mgoudy91 (approved):

PR #10867, opened 2026-04-14
Google Docs: Frontend error handling adjustment [INTEG-3785]

FBanfi:
## Purpose

This PR improves error handling around Google Docs workflow runs in the app.

During investigation, we found a gap in the current backend behavior: when a Google Docs workflow step fails, the agent appears to take the thrown error and convert it into a normal assistant response/result instead of propagating it as a thrown exception. Because of that, the backend does not reliably return an explicit error message that the app can use to clearly mark the run as failed. The run looks completed from the client perspective even though a workflow step actually threw.

Because changing that backend contract would require coordination and approval from another team, I chose to keep the backend behavior as-is for now and make minimal frontend-side adjustments so the app handles these runs more safely and predictably.

## Approach

- Removed unused error state in `useWorkflowAgent.ts` hook.
- Remove "Try Again" button since the steps are not retryable.
- Changed GENERIC_ERROR message.

## Testing steps

If you want to check that a step run doesn't bubble the error to the workflow and doesn't change it to FAILED:
1. Start the API locally.
2. Trigger `google-docs-workflow-agent` through the `generate` endpoint.
3. Change a step adding something  forces a failure.
4. Verify the backend does not provide a useful explicit error message and that it finished with COMPLETED.
5. Verify the frontend still can notice that there was an error. This is because in `useWorkflowAgent.ts` we still check if the completed run has a payload and if not, we throw and error.

## Breaking Changes

No

[INTEG-3785](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?issueParent=466416&selectedIssue=INTEG-3785)

## Deployment

No

[INTEG-3785]: https://contentful.atlassian.net/browse/INTEG-3785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/google-docs/src/hooks/useWorkflowAgent.ts
+++ b/apps/google-docs/src/hooks/useWorkflowAgent.ts
@@ -26,7 +26,6 @@ interface UseWorkflowParams {
 
 interface WorkflowHook {
   isAnalyzing: boolean;
-  error: string | null;
   startWorkflow: (contentTypeIds: string[]) => Promise<WorkflowRunResult>;
   resumeWorkflow: (runId: string, resumePayload: ResumePayload) => Promise<WorkflowRunResult>;
 }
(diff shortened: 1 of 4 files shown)


ryunsong-contentful (approved):
Nice cleanup

PR #10870, opened 2026-04-14
Cancel workflow when stopping flow [INTEG-3805]

joaquincasal:
Call the workflow resume endpoint with a cancellation payload when users leave the Google Docs frontend flow after a workflow suspension.

On cancel from suspended steps, frontend now sends resumePayload: { cancelled: true } to close the backend run.
Applies to both document-scope flow and mapping-review flow.
Keeps existing UX behavior (exit/reset) even if the cancel-resume call fails.
Updates tests to cover the new cancel-resume behavior.


apps/google-docs/src/hooks/useWorkflowAgent.ts
```diff
@@ -62,6 +62,35 @@ const previewPayloadFromCompletedRun = (runData: AgentRunData): PreviewPayload =
     throw new Error('Workflow completed but result payload was missing.');
   }
 
+  if (
+    typeof raw === 'object' &&
```
ryunsong-contentful: Can we rename `raw` to `googleDocPayload` or `agentPayload` or something like that?

ryunsong-contentful (approved):
Looks good, just a rename of the raw variable since raw isn't the best representation of what it is

ryunsong-contentful (approved):

PR #10872, opened 2026-04-15
feat(google-docs): Adding exclude modal view [INTEG-3798]

FBanfi:
## Purpose

- Build the new `exclude-content modal` UI for the Google Docs mapping review flow and reshape it around the future selection/highlight interaction from the preview page.
- This update also makes the modal reusable for future edition flows, like the upcoming `move modal`, since both share the same location-selection experience.

## Approach

- The modal expects already-selected content plus candidate locations from the preview layer.
- The shared UI was extracted into a reusable `EditionModal`, with the exclude flow using it directly. The modal supports frontend-only location selection, keeps the primary action non-functional for now, and includes a temporary mock trigger so the UI can be reviewed before the real highlight wiring exists.

**IMP**: The mock will be removed in future versions.

## Testing steps

Automated tests added.

https://github.com/user-attachments/assets/5720994e-b754-4d51-bcbc-0746fa4b31a1

## Breaking Changes

No

[INTEG-3798](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?issueParent=466416&selectedIssue=INTEG-3798)

## Deployment

No

[INTEG-3798]: https://contentful.atlassian.net/browse/INTEG-3798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/google-docs/src/types/workflow.ts
```diff
@@ -69,6 +69,22 @@ export interface WorkflowContentType {
...
+  fieldType: string;
+  sourceRef: SourceRef;
+  isSelected?: boolean;
+}
+
+export interface EditionModalContent {
+  selectedText: string;
+  locations: EditionLocationOption[];
+}
+
```
harikakondur: let's move this to a new types file? maybe call it `editModal.ts`

apps/google-docs/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -25,11 +28,63 @@ interface MappingViewProps {
   selectedEntryIndex: number | null;
 }
 
+const buildMockExcludeSelection: EditionModalContent = {
```
harikakondur: can you move to a different file?

ryunsong-contentful: I assume the next PR will actually grab the entry and build out the contents inside the modal?

ryunsong-contentful (approved):
Looks good, next PR will be to hook the data into the modal I assume

apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/EditionModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+  viewModel: EditionModalContent;
+  title: string;
+  locationSectionDescription: string;
+  primaryButtonLabel: string;
+  additionalContent?: ReactNode;
+}
+
+export const EditionModal = ({
```
harikakondur: [nit] for consistency, can we rename the components and types using `edition` -> `edit`?

harikakondur (approved):

apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/EditionModal.tsx
harikakondur: left a comment on slack about this, but if this modal is just for excluding content lets rename to `ExcludeModal` if not no worries looks good!

FBanfi: The first comment: implemented ✅ .
The second one: yes this was intended because this component will be reused since the assign and reassign modals share the same logic but with one more section.

apps/google-docs/src/locations/Page/components/review/mapping/MappingView.tsx
FBanfi: Yes, correct! I wasn't sure if Juli was going to include that connection in her PR or no, so I didn't add it for now.

harikakondur (approved):

PR #10873, opened 2026-04-15
feat: Link Checker app canonical release []

zacharyyankiver:
## Summary
- port the Link Checker app history onto a branch based on current master
- include the production deploy wiring with app definition ID 55HqLhaYLjQWU5zPfID6mX
- replace the previous side-branch release path with a canonical PR target

## Context
- the earlier production deploy PR merged into zachyankiver/link-checker-clean, not master
- this PR gives the app and its production deploy script a direct, mergeable path onto the canonical branch

## Testing
- not run in this cleanup pass


apps/link-checker/contentful-app-manifest.json
```diff
@@ -0,0 +1,43 @@
...
+    {
+      "id": "checkLink",
+      "name": "Check Link",
+      "description": "Checks a URL and returns its HTTP status (used by the Link Checker sidebar).",
+      "path": "functions/checkLink.js",
+      "entryFile": "functions/checkLink.ts",
+      "accepts": [
+        "appaction.call"
+      ],
+      "allowNetworks": [
```
t-col, line 12: This function is currently limited to this hard-coded `allowNetworks` list. Because the UI and README describe arbitrary link checking, any URL outside these hosts, for example `https://example.com`, will be rejected by the platform before `checkLink.ts` can run. In practice that means an editor sees a healthy external link as broken or unreachable just because it was not pre-listed here.

Why this matters: for a link checker, this is the core feature path, so the app silently fails for normal customer content.

How to reproduce: install the app with the action configured, add `https://example.com` to an entry, and run the scan. The request never reaches the handler because the function cannot egress to that host.

Suggested fixes:
- If the product is meant to check arbitrary editorial links, widen the network policy to match that contract.
- If the product is only meant to check a known host set, make that explicit in the config/UI and derive the function network policy from installer-provided hosts instead of shipping sample domains.
- Add an end-to-end or smoke-style test case using a non-sample host so this mismatch cannot regress silently.

apps/link-checker/utils/extractUrls.ts
```diff
@@ -0,0 +1,148 @@
...
+        ? (value as unknown[]).filter((v): v is string => typeof v === 'string')
+        : typeof value === 'string' && STRING_FIELD_TYPES.includes(field.type)
+        ? [value]
+        : [];
+
+      for (const text of stringValues) {
+        URL_REGEX.lastIndex = 0;
+        let match: RegExpExecArray | null;
+        while ((match = URL_REGEX.exec(text)) !== null) {
+          const url = match[0].trim();
```
t-col: The extractor stores the raw regex match here, which means sentence-ending punctuation stays attached to the URL. In editorial text, `Visit https://example.com.` becomes `https://example.com.`, and `See /about.` becomes `/about.`. The checker then requests the wrong address and reports a false broken link.

Why this matters: this will happen in very normal prose, so editors lose trust in the results quickly.

How to reproduce: put `See https://example.com.` in a Text field and run the sidebar checker. You will see the request target include the trailing period.

Suggested fix:
- Normalize the matched URL before deduping/storing it.
- Strip terminal punctuation like `.`, `,`, `;`, `:`, `!`, `?`, and unmatched `)` when it is outside the actual URL.
- Keep the displayed source text separate if you still want to show the original field content verbatim.
- Add regression tests for both absolute and relative links with trailing punctuation.

apps/link-checker/components/locations/Sidebar.tsx
```diff
@@ -0,0 +1,468 @@
...
+
+/** 2xx status codes (e.g. 200 OK, 204 No Content) indicate success. */
+function isSuccessStatus(status: number): boolean {
+  return status >= 200 && status < 300;
+}
+
+function isUrlOnDenyList(url: string, patterns: string[]): boolean {
+  if (!patterns.length) return false;
+  const normalized = url.toLowerCase();
+  return patterns.some((p) => p.trim() && normalized.includes(p.trim().toLowerCase()));
```
t-col: This allow/deny logic is host-spoofable because it checks whether the configured hostname appears *anywhere* in the full URL string. For example, an allow-list entry of `contentful.com` incorrectly accepts `https://contentful.com.evil.test` and even `https://evil.test/?next=contentful.com`.

Why this matters: the configuration looks like a hostname policy, but the implementation behaves like a substring search, which is much weaker and can be bypassed.

How to reproduce: configure `contentful.com` as an allowed domain, then scan a link like `https://contentful.com.evil.test/path`. It passes the allow-list check even though the hostname is not actually `contentful.com`.

Suggested fix:
- Parse the candidate with `new URL(...)`.
- Compare `url.hostname` against the configured hostname using exact match or a safe subdomain rule such as `hostname === pattern || hostname.endsWith('.' + pattern)`.
- Move that logic into a shared helper and reuse it in both the sidebar and page flows so the rules stay consistent.
- Add tests for spoofed hosts and query-string bait.

apps/link-checker/components/locations/Page.tsx
```diff
@@ -0,0 +1,843 @@
...
+
+        requestCache.set(urlToCheck, request);
+        return request;
+      };
+
+      let skip = 0;
+      let hasLoadedFirstBatch = false;
+      let entriesScanned = 0;
+      let linksFound = 0;
+      const entryQueryBase =
```
t-col: If `contentTypeMap` ends up empty, this query base becomes `{}` and the scan falls back to paging through *every* entry in the environment. That can happen when the assigned content types exist, but none of them contain Symbol/Text/RichText fields that this app can actually scan.

Why this matters: on a large space, this turns a no-op configuration mistake into an expensive full-space crawl that still cannot produce useful results.

How to reproduce:
1. Assign the app to a content type that only has references, assets, or other unsupported field types.
2. Open the page view and run a scan in a large space.
3. The app starts crawling entries even though no assigned content type can yield scanable links.

Suggested fix:
- After building `contentTypeMap`, short-circuit when `contentTypeMap.size === 0`.
- Show a note explaining that the assigned content types do not contain supported field types.
- Skip the `entry.getMany()` loop entirely in that case.

apps/link-checker/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "upsert-actions": "node scripts/upsert-actions-with-env.js",
+    "build-and-upload": "npm run build:all && npm run upload && npm run upsert-actions"
+  },
+  "dependencies": {
+    "@contentful/app-sdk": "4.22.0",
+    "@contentful/f36-components": "4.45.0",
+    "@contentful/f36-multiselect": "^4.81.1",
+    "@contentful/f36-tokens": "4.0.2",
+    "@contentful/react-apps-toolkit": "1.2.16",
+    "next": "15.5.12",
```
t-col: This PR imports a second frontend/tooling stack into `/apps`: Next.js for the app runtime/build, Jest for tests, and an app-local lockfile. In this monorepo that already has real costs: `next lint` is deprecated, Next warns about an ambiguous workspace root because of multiple lockfiles, `npm audit` reports direct advisories on `next@15.5.12`, and `npm ci` currently fails because `package.json` and `package-lock.json` are out of sync.

Why this matters: even if the feature code works, we now own a one-off stack with different maintenance, CI, and security upgrade rules than the rest of the marketplace apps. That increases long-term support cost for the whole team.

How to reproduce: in this PR worktree, run `npm ci`, `npm run lint`, and `npm audit` inside `apps/link-checker`. `npm ci` fails on the lock mismatch, and the lint/audit output already shows the divergence.

Suggested fixes:
- Preferred: port the app to the repo's standard Vite + Vitest shape before merge.
- If Next is truly required, document that exception, patch to a non-vulnerable Next release, and regenerate a lockfile that passes `npm ci`.
- Either way, align lint/test/build scripts with the monorepo conventions so future maintainers do not have to reverse-engineer a special case.

t-col (commented):
Left five inline comments. The first four are blockers from a correctness/security/release-readiness perspective; the fifth is a repo-fit issue I want us to resolve before we normalize this app inside `/apps`.

apps/link-checker/contentful-app-manifest.json
```diff
@@ -0,0 +1,33 @@
...
+    {
+      "id": "checkLink",
+      "name": "Check Link",
+      "description": "Checks a URL and returns its HTTP status (used by the Link Checker sidebar).",
+      "path": "functions/checkLink.js",
+      "entryFile": "functions/checkLink.ts",
+      "accepts": [
+        "appaction.call"
+      ],
+      "allowNetworks": []
```
t-col: This still looks blocking to me. An empty `allowNetworks` array means the function has *no* outbound network permission, not "all networks". Since `checkLink` works by making server-side HTTP requests, this leaves the core feature unable to check any URL at runtime.

Suggested fix:
- Replace `[]` with `[*]`

apps/link-checker/utils/extractUrls.ts
```diff
@@ -0,0 +1,176 @@
...
+
+export interface ExtractedUrl {
+  url: string;
+  fieldId: string;
+  fieldName: string;
+  locale: string;
+}
+
+/** Returns true if the URL is relative (not http/https). */
+export function isRelativeUrl(url: string): boolean {
```
t-col, line 20: One more correctness issue: the extractor intentionally matches `www.example.com/...` URLs, but this helper classifies anything without `http://` or `https://` as relative. Downstream, that means `www.example.com/help` is either skipped when `baseUrl` is unset, or resolved as something like `https://current-site.com/www.example.com/help`, which is the wrong target entirely.

Why this matters: `www.` links are common editorial input, so this will create false results in real content even after the other URL fixes.

How to reproduce: put `Visit www.example.com/help` in a Text field. `extractUrlsFromEntry()` returns `www.example.com/help`, and `isRelativeUrl()` returns `true`, so the sidebar/page flows treat it as a relative path.

Suggested fix:
- Treat `www.` URLs as absolute web links, not relative paths.
- A simple option is to normalize them to `https://...` during extraction before they enter the check pipeline.
- Alternatively, extend the URL classification step so `www.` is recognized as an absolute host form and converted before `new URL(..., baseUrl)` is used.
- Add regression tests for both the sidebar and page flows so `www.example.com/help` is checked as `https://www.example.com/help`, not skipped or rewritten under the configured base URL.

t-col (changes requested):
Thanks for the follow-up pass. The Vite/Vitest migration and the fixes for punctuation stripping, hostname matching, and the empty-content-type crawl all look like good progress. I still see two correctness blockers before I would approve this: the empty allowNetworks change appears to disable all outbound checks, and www. URLs are still classified as relative paths and handled incorrectly.

t-col (approved):

primeinteger (approved):
🚢

PR #10877, opened 2026-04-16
 Exclude, Assign and Reassign highlights [INTEG-3796]

JuliRossi:
## Purpose

Add highlights for exclude assign and reassign

## Approach

- Add exclude , assign and reassign button
- Added hook
- Added tests

## Testing steps

https://github.com/user-attachments/assets/cb4f5ca5-e565-4f28-b3a0-56da3a311a31


apps/google-docs/src/hooks/useReviewTextSelection.tsx
```diff
@@ -0,0 +1,111 @@
+import { useCallback, useEffect, useState, type RefObject } from 'react';
+import type { ActionMenuPosition } from '../locations/Page/components/review/mapping/actionMenuPosition';
+
+const SEGMENT_SURFACE_SELECTOR = '[data-review-segment-surface]';
+
+function elementFromNode(node: Node | null): Element | null {
+  if (!node) return null;
```
ryunsong-contentful: It's probably better to just enforce the type in the input like
```
elementFromNode(node: Node)
```

and then the parent calling this function needs to do null checks first before calling this function.

Then we can clean up all the null types and checks here and make this function cleaner

apps/google-docs/src/hooks/useReviewTextSelection.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const el = elementFromNode(node);
+    if (!el) return false;
+    const surface = el.closest(SEGMENT_SURFACE_SELECTOR);
+    return surface !== null && root.contains(surface);
+  };
+
+  return endpointInSurface(range.startContainer) && endpointInSurface(range.endContainer);
+}
+
+function rangeToActionMenuPosition(range: Range): ActionMenuPosition {
```
ryunsong-contentful: Is there a better name for this?

apps/google-docs/src/hooks/useReviewTextSelection.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export function useReviewTextSelection(
+  rootRef: RefObject<HTMLElement | null>
+): UseReviewTextSelectionResult {
+  const [selectedText, setSelectedText] = useState('');
+  const [selectedRange, setSelectedRange] = useState<Range | null>(null);
+
+  const updateFromSelection = useCallback(() => {
+    const root = rootRef.current;
+    const sel = window.getSelection();
```
ryunsong-contentful: better var name, like windowSelection or something shorter if you can think of something better

apps/google-docs/src/locations/Page/components/review/mapping/TabSegment.tsx
```diff
@@ -0,0 +1,75 @@
...
+  listMarkers: Record<string, ListMarker>;
+  excludedSourceRefs: SourceRef[];
+  selectedEntryIndex: number | null;
+  hoveredMappingKeys: string[];
+  onSetHoveredMappingKeys: (keys: string[]) => void;
+  onAssignImage: (sourceRef: ImageSourceRef, label: string) => void;
+  onExcludeImage: (sourceRef: ImageSourceRef, label: string) => void;
+}
+
+export const TabSegement = ({
```
ryunsong-contentful: What do you think of renaming this to NormalizedDocumentOutline.tsx.

So that it reads `NormalizedDocumentOutline.tsx` for the outline section (left-hand side of the UI) and then `MappingEntryCards.tsx` for the entry card section (right-hand side of the UI)

ryunsong-contentful (dismissed):
Overall looks good. Comments on three things

1. Small code optimizations
2. Renaming a file or two
3. If you think there's a better file organization within this sub folder, lets do it. If you think the way it is right now is fine then that's okay too. I trust your judgement on this particular topic

apps/google-docs/src/hooks/useReviewTextSelection.tsx
JuliRossi: Naming this one was hard. Asked codex for a better name, let's see if AI is correct.

apps/google-docs/src/locations/Page/components/review/mapping/TabSegment.tsx
JuliRossi: Could be misleading, because there are some more elements in the Outline. Went with `NormalizedDocumentSection`

ryunsong-contentful (approved):
Nice improvements on renaming and slight reorg

PR #10879, opened 2026-04-16
fix: bulk edit field menu scrolling [ES-105] 

mgoudy91:
## Summary
Adjust the Bulk Edit field menus to use the available viewport height before becoming scrollable.

## Changes
- apply a shared viewport-aware max-height to the field filter and field visibility menus
- keep menu scrolling enabled only when the list exceeds the available viewport space
- add coverage for the menu sizing behavior in both menu entry points

Updated in:
- apps/bulk-edit/src/locations/Page/components
- apps/bulk-edit/test/locations/Page


primeinteger (approved):
👍

PR #10886, opened 2026-04-17
fix(google-docs): add title to overview entry card [INTEG-3818]

harikakondur:
## Purpose

Adding the entry title to the overview entry card + reducing padding to make more cards visible

Before 

Screenshot 2026-04-16 at 11 19 56 PM

After
Screenshot 2026-04-16 at 11 20 07 PM


ryunsong-contentful (approved):

apps/google-docs/src/locations/Page/components/overview/OverviewEntryList.tsx
```diff
@@ -51,6 +51,7 @@ function OverviewEntryRowCard({
           cursor: isSelected ? 'default' : 'pointer',
           textAlign: 'left',
           width: '100%',
+          padding: `${tokens.spacingXs} ${tokens.spacingS}`,
```
JuliRossi, line 53: 💭 Padding prop is duplicated

apps/google-docs/src/utils/overviewEntryList.ts
```diff
@@ -37,6 +38,24 @@ function resolveContentTypeLabel(
   return name && name.length > 0 ? name : '';
 }
 
+function getTitleFromFieldMappings(
```
JuliRossi: ⚔️ By convention if a title of an entry is undefined then it should be "Untitled"

apps/google-docs/src/utils/overviewEntryList.ts
JuliRossi, line 1: ⛏️  I would move this to getEntryTitle utils

JuliRossi (commented):
Left a few comments.

JuliRossi (approved):
LGTM!

PR #10889, opened 2026-04-17
[] Skip serverless-offline during GA4 lambda deploy

mgoudy91:
## Summary
This avoids loading serverless-offline during GA4 lambda deploys, which currently fail in CI under Node 22 when Serverless tries to require the plugin.

## Changes
- move the GA4 lambda plugin list behind a small helper file
- load serverless-offline only for offline/local commands
- keep deploy and print paths limited to the plugins needed in CI

Updated in:
- apps/google-analytics-4/lambda

## Notes
- serverless-offline remains available for local development
- verified sls print --stage dev succeeds after the change


mgoudy91:
Closing this in favor of #10890, which contains the same GA4 lambda deploy fix on a clean branch based on staging only.

PR #10891, opened 2026-04-17
feat(google-docs): adding integer and decimal edit [INTEG-3814]

FBanfi:
## Purpose

This extends the dropdown field selection logic to allow integer and decimal handling.

## Approach

- Now we check the field type and for the case of integer and decimal we also check if the selected text could be consider as a number.
- The short text and long text are always available, but the integer and decimal are only available if that field type match the selected value.

## Testing steps

Automated tests added.

Manual check:

  - so if we have "5.5 hours" as the selected/highlighted text, that will make short and long text available but no integer and decimal.

> https://github.com/user-attachments/assets/846832ae-d69c-4d0d-abbc-a1d4765ea092

  - if we have just "5.5 ", that will make short text, long text, integer, and decimal available.

> https://github.com/user-attachments/assets/13e482d1-7e82-4e50-98f9-fceee73f84fb

  - if we have just "5", that will make short text, long text and integer available.

> https://github.com/user-attachments/assets/e78f4f91-237e-4bab-b72b-f87bfc814ff9
  

  - if we have just "Hot coffee, haute content, on us.", that will make short text and long text available.

> https://github.com/user-attachments/assets/9fc8e176-7536-41b0-98ae-7f6d0a45d2bc

## Breaking Changes

No

[INTEG-3814](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?issueParent=466416&selectedIssue=INTEG-3814)

## Deployment

No

[INTEG-3814]: https://contentful.atlassian.net/browse/INTEG-3814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/utils.ts
+++ b/apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/utils.ts
@@ -1,0 +1,41 @@
+const enum AllowedContentTypes {
+  INTEGER = 'integer',
+  DECIMAL = 'decimal',
+  TEXT = 'text',
+  SHORT_TEXT = 'short text',
+  LONG_TEXT = 'long text',
+}
+
+export const getSelectedContentMatch = (value: string) => {
+  const trimmedValue = value?.trim() ?? '';
+
+  if (/^[+-]?\d+$/.test(trimmedValue)) {
+    return AllowedContentTypes.INTEGER;
+  }
+
+  if (/^[+-]?(?:\d+\.\d+|\d+\.|\.\d+)$/.test(trimmedValue)) {
+    return AllowedContentTypes.DECIMAL;
+  }
+
+  return AllowedContentTypes.TEXT;
+};
+
+export const isSelectableFieldType = (fieldType: string, selectedText: string) => {
+  const highlightedTextMatch = getSelectedContentMatch(selectedText);
+  const normalizedFieldType = fieldType.trim().toLowerCase();
+
+  switch (normalizedFieldType) {
+    case AllowedContentTypes.SHORT_TEXT:
+    case AllowedContentTypes.LONG_TEXT:
+      return true;
+    case AllowedContentTypes.INTEGER:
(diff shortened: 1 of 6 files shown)


apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+  DECIMAL = 'decimal',
+  TEXT = 'text',
+  SHORT_TEXT = 'short text',
+  LONG_TEXT = 'long text',
+}
+
+export const getSelectedContentMatch = (value: string) => {
+  const trimmedValue = value?.trim() ?? '';
+
+  if (/^[+-]?\d+$/.test(trimmedValue)) {
```
JuliRossi, line 6: ❓ Do we allow signed values? Same question for Decimal

apps/google-docs/test/locations/Page/components/modals/FieldSelectionDropdown.spec.tsx
```diff
@@ -4,11 +4,12 @@ import React from 'react';
 import { FieldSelectionDropdown } from '../../../../../src/locations/Page/components/review/mapping/edit-modals/FieldSelectionDropdown';
 
 describe('FieldSelectionDropdown', () => {
-  it('renders field status from field mappings and disables non-text options', async () => {
+  it('enables numeric field types only when the selected text is numeric', async () => {
     const onSelectedFieldIdsChange = vi.fn();
 
     render(
       <FieldSelectionDropdown
+        selectedText="5 "
```
JuliRossi: 💭 If we allow signed values, could be nice to use 
```suggestion
        selectedText="+5 "
```

JuliRossi (approved):
LGTM

joaquincasal (approved):

harikakondur (approved):

PR #10897, opened 2026-04-19
feat(link-checker): add entry field automation action []

zacharyyankiver:
## Summary
- add a new `Check Entry Field Links` app action for automations
- add a new `Check All Entry Links` app action to scan supported fields across an entry
- scan a target entry field like `body` or scan all supported fields and return a Slack-friendly summary of link results
- reuse the existing single-URL checker and add tests for both automation actions

## Testing
- npm run lint
- npm run build
- npm run test:ci -- __tests__/functions/checkLink.spec.ts __tests__/functions/checkEntryLinks.spec.ts __tests__/functions/checkAllEntryLinks.spec.ts __tests__/utils/extractUrls.spec.ts
- npm run build:functions

### DIFF
--- a/apps/link-checker/functions/checkEntryLinks.ts
+++ b/apps/link-checker/functions/checkEntryLinks.ts
@@ -1,0 +1,454 @@
+import { FunctionEventContext } from '@contentful/node-apps-toolkit';
+import { createClient, PlainClientAPI } from 'contentful-management';
+import { extractUrlsFromEntry, isRelativeUrl, type ExtractedUrl } from '../utils/extractUrls';
+import { normalizeDomainPattern, urlMatchesAnyDomainPattern } from '../utils/domainPatterns';
+import { checkUrl, type CheckLinkResult } from './checkLink';
+
+interface CheckEntryLinksParameters {
+  entryId?: string;
+  fieldId?: string;
+  locale?: string;
+}
+
+interface CheckEntryLinksEvent {
+  body?: CheckEntryLinksParameters;
+}
+
+interface ContentTypeField {
+  id: string;
+  name?: string;
+  type: string;
+}
+
+interface RichEntryField {
+  [locale: string]: unknown;
+}
+
+interface EntryLikeField {
+  id: string;
+  name: string;
+  type: string;
+  locales: string[];
+  getValue: (locale?: string) => unknown;
+}
+
+interface EntryLike {
+  fields: Record<string, EntryLikeField>;
+}
+
+interface EntryLinkResult extends CheckLinkResult {
+  url: string;
+  resolvedUrl?: string;
+  fieldId: string;
+  fieldName: string;
+  locale: string;
+  isValid: boolean;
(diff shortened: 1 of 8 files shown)


apps/link-checker/functions/checkEntryLinks.ts
```diff
@@ -0,0 +1,457 @@
...
+    return trimmed.replace(/^https?:\/\//i, '').toLowerCase();
+  }
+}
+
+function isSuccessStatus(status: number): boolean {
+  return status >= 200 && status < 300;
+}
+
+function isUrlOnDenyList(url: string, patterns: string[]): boolean {
+  if (!patterns.length) return false;
```
mgoudy91, line 91: The new allow/deny helpers are matching by substring against the entire URL, which changes the behavior from hostname-aware domain matching to a much looser check. For example, an allow-list entry like `contentful.com` would now allow `https://notcontentful.com`, and a deny-list entry could be tripped by a path or query string instead of the hostname. Since the existing app already has `urlMatchesAnyDomainPattern` in `utils/domainPatterns.ts`, can we reuse that exact-host-or-subdomain logic here so the automation actions behave the same way as the rest of the app?

apps/link-checker/functions/checkEntryLinks.ts
```diff
@@ -0,0 +1,457 @@
...
+          if (!field) {
+            throw new Error(`Field ${fieldId} was not found on the content type.`);
+          }
+          return field;
+        })
+      : contentTypeFields.filter((field) => supportedFieldTypes.has(field.type));
+
+  const fields = selectedFields.reduce<Record<string, EntryLikeField>>((acc, field) => {
+    const fieldValues = entryFields[field.id];
+    if (!fieldValues || typeof fieldValues !== 'object') {
```
mgoudy91, line 191: This all-fields path throws if the requested locale is missing on any selected field, because `getAvailableLocales` errors instead of letting the reducer skip that field. In practice that means a partially localized entry can make the entire automation fail even when other supported fields do have values in the requested locale. Can we make `checkAllEntryLinks` skip fields that don’t have the requested locale, rather than aborting the whole scan?

mgoudy91 (changes requested):
I found two functional issues that look worth fixing before this ships: the new allow/deny matching no longer uses hostname-aware domain checks, and the all-fields scan can fail outright on partially localized entries instead of skipping unavailable fields.

zacharyyankiver:
Addressed both review issues in the latest push.

What changed:
- restored hostname-aware allow/deny matching in the automation path by reusing `urlMatchesAnyDomainPattern`, so allow/deny rules now match the same exact-host-or-subdomain behavior as the rest of the app
- updated the all-fields scan to skip fields that do not have the requested locale instead of failing the entire scan on partially localized entries

Also added test coverage for both behaviors:
- hostname-aware allow-list matching in the all-fields action
- skipping missing-locale fields during all-fields scans

mgoudy91 (approved):

PR #10899, opened 2026-04-20
Refine GA4 advanced matching configuration UX []

zacharyyankiver:
## Summary
- refine the GA4 advanced matching configuration UX based on staging feedback
- improve grouping, spacing, and contextual validation for advanced configuration rows
- auto-generate advanced patterns from the current setup, update them as page properties are added, and preserve manual edits
- clarify matching terminology and standard-vs-advanced trailing slash behavior

## Validation
- frontend targeted vitest run for AssignContentTypeRow.spec.tsx, AssignContentTypeSection.spec.tsx, and src/utils/getReportSlug.spec.ts passed
- staging branch batched frontend vitest run completed successfully while validating these changes

## Notes
- multiple rules per content type remain allowed intentionally for the customer use case
- advanced rules require a pattern before save, and the validation now renders in context

### DIFF
--- a/apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
+++ b/apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
@@ -11,7 +11,9 @@ import { normalizeContentTypeRules } from 'helpers/contentTypeRules/contentTypeR
 
 const Sidebar = () => {
   const sdk = useSDK<SidebarExtensionSDK>();
-  const installationParameters = sdk.parameters.installation as AppInstallationParameters | undefined;
+  const installationParameters = sdk.parameters.installation as
+    | AppInstallationParameters
+    | undefined;
 
   const { serviceAccountKeyId, propertyId, contentTypes, contentTypeRules } =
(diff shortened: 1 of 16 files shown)


apps/google-analytics-4/frontend/src/locations/Sidebar.tsx
```diff
@@ -42,6 +46,16 @@ const Sidebar = () => {
     return result as CustomRangeDialogResult | undefined;
   };
 
+  if (!installationParameters) {
```
mgoudy91, line 54: [P1] Sidebar skeleton path still dereferences missing installation params

This fallback doesn’t actually protect the render path when `sdk.parameters.installation` is temporarily `undefined`, because React has already executed `useApi(serviceAccountKeyId)` before reaching this branch. `useApi` falls back to `sdk.parameters.installation.serviceAccountKeyId`, so in the exact loading state this is trying to handle, render still throws before the skeleton can appear. Please either move the guard above the `useApi` call or make `useApi` tolerate an absent installation object.

PR #10904, opened 2026-04-21
fix(google-docs): handle AI kill switch access denial [INTEG-3827]

mgoudy91:
## Summary
Surface backend `ai-access-denied` access denials clearly in the Google Docs app so denied orgs and spaces do not hit generic failures during OAuth and AI import flows.

## Changes
- normalize app action and agent API access-denied responses into a dedicated AI access error utility
- move the page into a blocked state when AI access is denied instead of continuing with generic workflow or OAuth errors
- add focused test coverage for error normalization and the blocked page behavior

Updated in:
- `apps/google-docs/src/locations/Page`
- `apps/google-docs/src/services/agents-api.ts`
- `apps/google-docs/src/utils`
- `apps/google-docs/test`

## Test
Captura de pantalla 2026-05-05 a las 11 18 28 a  m

## Notes
- This PR is the frontend UX half of `INTEG-3827`.
- The companion backend enforcement PR is https://github.com/contentful/agents-api/pull/430.
- Rollout plan: ship this apps PR first (or together with the backend PR), then ship the backend PR so kill-switched users get the clearer blocked state as soon as enforcement goes live.
- OAuth is intentionally not the enforcement point; the backend PR owns blocking AI invocation paths, and this PR makes those denials readable in the app.

### DIFF
--- a/apps/google-docs/src/services/agents-api.ts
+++ b/apps/google-docs/src/services/agents-api.ts
@@ -1,5 +1,6 @@
 import { PageAppSDK } from '@contentful/app-sdk';
 import { WORKFLOW_AGENT_ID } from '../utils/constants/agent';
+import { normalizeAiAccessError } from '../utils/aiAccess';
 import {
   AgentRunMessage,
   MappingReviewSuspendPayload,
@@ -75,6 +76,10 @@ export async function getWorkflowRun(
     }
 
     if (!response.ok) {
+      if (response.status === 403) {
+        throw normalizeAiAccessError({ status: response.status });
+      }
+
       throw new Error(`Failed to poll agent run: ${response.status} ${response.statusText}`);
     }
 
@@ -93,7 +98,7 @@ export async function getWorkflowRun(
       return null;
     }
 
-    throw error;
+    throw normalizeAiAccessError(error);
   }
 }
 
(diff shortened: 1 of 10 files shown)


mgoudy91 (commented):
Two issues worth addressing before merge:

1. **`handleReconnectGoogleDrive` swallows 403s** — the catch block calls `console.error` but doesn't route through `handleWorkflowError`, so a reconnect-path denial never reaches the blocked state.
2. **`status === 403` match is too broad** — any generic 403 will surface as "AI features are currently disabled", which is misleading for unrelated permission errors. Should require a recognizable `sys.id`, `code`, or message substring alongside the status check.

apps/google-docs/src/services/agents-api.ts
```diff
@@ -166,14 +175,46 @@ export async function resumeWorkflowRun(
...
 
     return;
   }
 
-  await sdk.cma.agentRun.resumeRun(
-    { spaceId, environmentId, runId },
-    { resumePayload: resumePayload as Record<string, unknown> }
-  );
+  const agentRunApi = sdk.cma.agentRun as {
+    resumeRun?: (
```
ryunsong-contentful: There's probably a way to make this part cleaner. Like define the typing somewhere else and assign it here or use the cma typing if there is one for this

ryunsong-contentful (approved):
One typing comment but otherwise looks good

ryunsong-contentful (approved):
Let me know when/if you need a re-review

joaquincasal (approved):

PR #10911, opened 2026-04-22
fix(google-docs): sync field status in edit modal target location dropdown [INTEG-3820]

harikakondur:
## Purpose

The filled/empty field status wasn't synced with edits made in the review page. This PR fixes the field status signal and improves how unsupported field types are communicated to the user.

## Before
In this example, the content in the `slug` field is excluded . So the field status should be `Empty` after the exclusion but after you reassign, the field status still says `Filled`

https://github.com/user-attachments/assets/f7a96d31-d800-482a-bf46-17402efd25f1

## After
Removed preselected current locations (blue border)

https://github.com/user-attachments/assets/8662395b-411e-481d-8dc1-1bd6bf5fdf07

Edit modal with new badge styles for filled status
Screenshot 2026-04-24 at 1 53 39 PM

Help text to communicate unsupported fields
Screenshot 2026-04-24 at 1 53 44 PM

Filled / Empty status updated with edits

https://github.com/user-attachments/assets/31614a5f-286e-4d97-b649-a5643fba6d93

Allow multi-exclusion

https://github.com/user-attachments/assets/a1556fab-bef8-4f9f-9a4a-f78081b504bc

## Approach

  **Fix field status detection**
  - Added `sourceRefs` to `EditModalFieldMapping` so the dropdown has the data
    it needs to determine fill state
  - Updated `filledFieldIds` in `FieldSelectionDropdown` to filter on
    `sourceRefs.length > 0` instead of mapping presence alone

  **Improve visual communication in the dropdown**
  - Replaced plain "Filled" / "Empty" text with f36 `Badge` components
    (`positive` / `secondary`) pushed to the trailing edge of each option row
  - Added a conditional `FormControl.HelpText` note below the dropdown when
    unsupported field types are present:
    > "Edits aren't supported for: Reference, Boolean, Date & time, Location or
    > JSON fields. Use the entry editor instead."

  **Multi-select current locations for exclusion**
  - Current location cards now support multi-selection — clicking a card toggles
    it; multiple can be selected at once
  - Nothing is pre-selected when the modal opens
  - The CTA stays disabled until at least one location is selected (both assign
    and exclude modes)
  - On confirm, exclusion is applied to each selected location in sequence
  - Removed `isSelected: index === 0` hardcoding in `MappingView` that was
    forcing the first location to always be pre-selected

  **Clean up edit modal**
  - Removed the orange `Note` validation banner — the disabled CTA already
    communicates the blocked state
  - Removed `EditModalDestinationStateKind`, `EditModalDestinationState`,
    `createEditModalDestinationState`, and `DEFAULT_DESTINATION_STATE_MESSAGES`
    — no longer needed
  - Inlined destination state logic directly into `isPrimaryDisabled` with
    inline comments per condition
  - `minHeight` on the modal content is now only applied in assign mode (the
    exclude modal doesn't have the dropdown and doesn't need the extra height)

  **Fix exclusion behaviour**
  - `applyTextExclusionToEntryBlockGraph` now prunes field mappings when all
    source refs are excluded, rather than leaving an empty mapping behind
  - `applyImageExclusionToEntryBlockGraph` inlined with the same pruning
    behaviour for consistency
  ## Testing steps

  - [ ] Open the edit modal on an entry with a mix of text, reference, boolean,
        and media fields
  - [ ] Verify "Filled" (green) / "Empty" (grey) badges reflect actual mapping
        state
  - [ ] Verify the help text note appears below the dropdown
  - [ ] Verify the "Move content" CTA stays disabled until a compatible field is
        selected, with no orange banner
  - [ ] Exclude all text from a mapped field — verify the field mapping is
        removed rather than left empty

  ## Breaking Changes

  None.

  None.

### DIFF
--- a/apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
+++ b/apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
@@ -1,10 +1,6 @@
 import { useEffect, useMemo, useState, type ReactNode } from 'react';
-import { Box, Button, Modal, Note, Flex, Text } from '@contentful/f36-components';
-import {
-  EditModalDestinationStateKind,
-  createEditModalDestinationState,
-  type EditModalContent,
-} from '@types';
+import { Box, Button, Modal, Flex, Text } from '@contentful/f36-components';
+import { type EditModalContent } from '@types';
 
 import {
   locationButton,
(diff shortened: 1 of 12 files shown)


apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
```diff
@@ -101,45 +86,32 @@ export const EditModal = ({
...
-    isAssignMode,
-    destinationFieldState?.hasSelectableOptions,
-    hasNewLocation,
-    selectedDestinationCount,
-    viewModel.newLocation.fieldOptions.length,
-  ]);
-  const isPrimaryDisabled =
-    isAssignMode && destinationState.kind !== EditModalDestinationStateKind.Ready;
+  const isPrimaryDisabled = useMemo(
+    () =>
```
mgoudy91, line 92: I think we need to reset `selectedLocationIds` when the modal opens or when the current locations change. The modal stays mounted between opens, so after selecting a location in one modal the next modal can inherit that stale id. That makes the CTA enabled even though no visible location is selected; assign can then fall back to the first location, while exclude can close without changing anything. Deriving this disabled check from ids that still exist in `viewModel.currentLocations` would also close the gap.

mgoudy91 (commented):
Codex found one potential issue with state tracking between modals that I'm pretty sure is real, lmk what you think

apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
harikakondur: on it 🫡

harikakondur: Updated some things to make sure franco's recent table pr works nicely with my changes
- When you select mutiple cells but not the entire table -> we display a warning that partial table imports in rich text is not supported
- Full table and single cell selections are still supported
https://github.com/user-attachments/assets/a701ef30-8f66-4750-b1ed-3518766f73f0

mgoudy91 (approved):
Fix looks good and I don't see anything else, thanks and nice work!

PR #10918, opened 2026-04-22
Google Docs: Group mapping cards to the same field [INTEG-3826]

joaquincasal:
This PR improves the Google Docs mapping review UI by grouping consecutive document blocks that map to the same Contentful field into a single mapping card, while keeping non-consecutive runs separate and labeled with their position (for example Body (1/2) and Body (2/2)). It also preserves existing HIL actions like reassign, exclude, hover highlighting, image assignment, and partial text selection.

Before

Screenshot 2026-04-22 at 16 04 46

After

Screenshot 2026-04-22 at 16 14 21

Excluding

https://github.com/user-attachments/assets/3182effe-a1f4-4790-aaab-ea450afeabc3

### DIFF
--- a/apps/google-docs/src/locations/Page/components/review/mapping/buildMappingDisplayGroups.ts
+++ b/apps/google-docs/src/locations/Page/components/review/mapping/buildMappingDisplayGroups.ts
@@ -1,3 +1,4 @@
+import { isTextSourceRef } from '@types';
 import type { MappingCardData } from './MappingCard';
 import type { Tab, DocSegment } from './buildDocument';
 import { type MappingHighlight, getMappingCardKey } from './buildHighlights';
@@ -28,6 +29,8 @@ interface DraftMappingDisplayGroup {
   segments: DocSegment[];
   mappingCards: DraftMappingCard[];
   mergeFieldIdentity: string | null;
+  startsAtBoundary: boolean;
+  endsAtBoundary: boolean;
 }
 
 export interface MappingDisplayGroupsResult {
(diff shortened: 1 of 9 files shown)


apps/google-docs/src/locations/Page/components/review/mapping/buildMappingDisplayGroups.ts
```diff
@@ -0,0 +1,225 @@
...
+      return;
+    }
+
+    byFieldIdentity.set(fieldIdentity, {
+      firstHighlight: highlight,
+      mappingKeys: [mappingKey],
+    });
+  });
+
+  return Array.from(byFieldIdentity.entries()).map(([fieldIdentity, value]) => ({
```
mgoudy91, line 91: This dedupes highlights by `fieldIdentity` across the whole segment, which works for block runs but is too coarse for tables because one `DocSegment` can contain many distinct mapped rows. A safer improvement would be to preserve one card per source ref for table segments and only collapse same-field highlights for block segments, where the adjacent-run grouping is the feature we want.

apps/google-docs/src/locations/Page/components/review/mapping/buildMappingDisplayGroups.ts
```diff
@@ -0,0 +1,225 @@
...
+  nextGroupIndex: { current: number },
+  resolveFieldTypeLabel: (highlight: MappingHighlight) => string
+): DraftMappingDisplayGroup[] {
+  const groups: DraftMappingDisplayGroup[] = [];
+
+  tab.segments.forEach((segment) => {
+    const highlights = visibleHighlightsBySegment[segment.id] ?? [];
+    const mappingCards = buildDraftMappingCards(segment, highlights, resolveFieldTypeLabel);
+    const mergeFieldIdentity =
+      segment.kind === 'block' && mappingCards.length === 1 ? mappingCards[0].fieldIdentity : null;
```
mgoudy91: Right now any block with exactly one mapping card gets `mergeFieldIdentity`, which later makes `showGroupedSurface` true even when the highlight only covers part of the paragraph. We could tighten this by only treating a block as grouped when its mapping reaches both block boundaries, or when it actually merged with an adjacent block. That would preserve the new run-grouping behavior without visually implying that an entire partially-mapped paragraph belongs to the field.

mgoudy91 (commented):
I have two suggestions, but I think they should be fairly small to cover. Lmk if you have any questions

mgoudy91 (approved):

PR #10923, opened 2026-04-23
feat: add review-all mapping mode []

zacharyyankiver:
## Summary
- add a read-only `Review all` mode for mapping review with a dedicated action in the overview area
- default the review stage into `Review all`, add clearer instructional messaging, and let users jump into per-entry edits from the overview list
- refine read-only mapping highlights so field boxes, table values, image/media mappings, and hover states match the intended review UX without nested or overlapping boxes

## Context
While testing the Drive Integration review experience, the team wanted a faster way to validate all generated entry mappings at a glance before editing. This update adds an all-entries review state that makes the mapping coverage easier to scan, while still keeping the existing per-entry editing flow available from the overview list.

## Validation
- npx vitest run test/locations/Page/components/review/mapping/MappingView.spec.tsx
- npm run build

### DIFF
--- a/apps/google-docs/src/locations/Page/components/review/mapping/MappingCard.tsx
+++ b/apps/google-docs/src/locations/Page/components/review/mapping/MappingCard.tsx
@@ -4,6 +4,7 @@ import tokens from '@contentful/f36-tokens';
 
 export interface MappingCardData {
   key: string;
+  contentTypeName: string;
   fieldName: string;
   fieldType: string;
   displayLabel: string;
@@ -13,6 +14,7 @@ interface MappingCardProps {
   card: MappingCardData;
   top: number;
   wrapperRef: Ref<HTMLDivElement>;
+  showContentTypeName?: boolean;
   isHovered?: boolean;
   onMouseEnter?: () => void;
   onMouseLeave?: () => void;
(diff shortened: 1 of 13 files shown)


apps/google-docs/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -840,6 +910,14 @@ export const MappingView = ({
...
                   card.mappingKeys.some((key) => hoveredMappingKeys.includes(key))
                 );
+                const mediaLikePattern = /media|image|asset/i;
+                const prefersImageOnlyHighlight =
+                  isReadOnlyAllMappings &&
+                  group.mappingCards.length === 1 &&
+                  (mediaLikePattern.test(group.mappingCards[0].fieldType) ||
+                    mediaLikePattern.test(group.mappingCards[0].fieldName) ||
+                    mediaLikePattern.test(group.mappingCards[0].displayLabel));
+                const showGroupedSurface = group.showGroupedSurface && !prefersImageOnlyHighlight;
```
mgoudy91, line 920: This `prefersImageOnlyHighlight` decision is based on display strings, which can misclassify normal text mappings for fields named things like `Image caption` or `Media credit`. A safer approach would be to derive group/card metadata from the underlying highlights or source refs instead: track whether the group contains any text source refs and any image source refs, then only use the image-only treatment when the mapped sources are actually image-only.

apps/google-docs/src/locations/Page/components/review/mapping/MappingCard.tsx
```diff
@@ -71,6 +74,16 @@ export const MappingCard = ({
...
       }}>
+      {showContentTypeName ? (
+        <Box marginBottom="spacing2Xs">
+          <Text as="span" fontColor="gray600" style={labelTextStyle} marginRight="spacingXs">
+            Content type:
+          </Text>
+          <Text as="span" fontWeight="fontWeightDemiBold" style={valueTextStyle}>
+            {contentTypeName}
+          </Text>
+        </Box>
```
mgoudy91, line 85: In `Review all`, showing only `Content type` is not enough once there are multiple generated entries of the same type, because the cards become indistinguishable. Consider adding a second line for the entry identity here, using the computed review title when available and falling back to something stable like `Article #2`, with truncation + tooltip if space is tight.

mgoudy91:
duplicate

PR #10926, opened 2026-04-24
feat(google-docs): Add inlined images and tables in rich text [INTEG-3836]

FBanfi:
## Purpose

Add Rich Text handling to the Google Docs mapping review modal so selected document content can be assigned or reassigned to `RichText` fields using backend-compatible indexed source refs. This includes mixed-content handling for formatted text, inline/block images, and table content, while improving the modal preview so reviewers can understand what will move before confirming.

## Approach

- We extended the assign/reassign flow to generate `RichText` field mappings from ordered indexed source refs instead of materializing final Rich Text values in the frontend. 
- The flow now preserves rich text-compatible content through indexed refs such as `paragraph`, `heading`, `listItem`, `image`, `tableText`, and `tableImage`, which allows the backend to build the final CMA Rich Text structure. 
- On the UI side, the modal preview was updated to show compact placeholders for images and tables, table mappings are grouped into a single mapping card per field, and grouped table reassign behavior was adjusted so table selections can be moved correctly.

## Testing steps

Normal text, table and image selection:

https://github.com/user-attachments/assets/92e773ba-353b-4556-b872-048e6ccafe7b

Partial table selection:

https://github.com/user-attachments/assets/e360ca5d-0f0b-4547-aa82-9fb6a6c08c95

## Breaking Changes

No.

[INTEG-3836](https://contentful.atlassian.net/browse/INTEG-3836)

## Deployment

No

[INTEG-3836]: https://contentful.atlassian.net/browse/INTEG-3836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/google-docs/test/locations/Page/components/review/RichTextSelectionPreview.spec.tsx
mgoudy91, line 1: I think this file was put in the wrong place :)

apps/google-docs/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -684,38 +735,96 @@ export const MappingView = ({
...
+              nextGraph,
+              document,
+              from,
+              richTextRefs,
+              richTextTargets
+            );
+          }
+        }
+
+        if (!nonRichTextTargets.length) {
```
mgoudy91, line 779: For rich-text reassigns, this branch only collects `mappedState: "mapped"` refs for the selected current mapping. If the selection also contains unmapped content, those refs appear in the modal preview but never get appended to the rich-text target; when there are no non-rich-text targets we return here, dropping the unmapped part of the mixed selection.

apps/google-docs/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -553,40 +577,62 @@ export const MappingView = ({
...
     const mappedPreview = collectMappedExclusionPreviewText(
       textSelectionRootRef.current,
-      selectedRange
+      selectionRange
     ).trim();
-    openExcludeModal(trimmed, getLocationsForSelectedText(), {
-      contentPreview: mappedPreview || trimmed,
-      previewSectionTitle: 'Text to exclude',
-    });
+    const previewSourceRefs = collectRichTextSourceRefsFromSelection(
```
mgoudy91, line 625: This preview source-ref collection can include mapped image refs from the selection, so the exclude modal can show an image as content to exclude. On confirm, though, the selection-exclude path only uses `pendingTextExclusionRanges` and calls `applyTextExclusionToEntryBlockGraph`; `pendingImageSourceRef` is only set for the single-image button path. That leaves mapped images shown in this preview still mapped after confirmation.

apps/google-docs/src/types/editModal.ts
```diff
@@ -12,13 +12,15 @@ export interface EditLocationOption {
...
   sourceRefs?: SourceRef[];
+  mappingKeys?: string[];
   isSelected?: boolean;
 }
 
 export interface EditModalFieldOption {
   id: string;
   fieldName: string;
   fieldDisplayType: string;
+  fieldType?: string;
```
JuliRossi: 💭 This should be required

apps/google-docs/src/locations/Page/components/review/mapping/entryBlockGraphExclusion.ts
JuliRossi, line 1: 💭 Perhaps we can move the rich text logic to its own file, since it's complex

apps/google-docs/test/locations/Page/components/review/RichTextSelectionPreview.spec.tsx
JuliRossi, line 1: 💭 Move this to edit-modals

apps/google-docs/src/locations/Page/components/review/mapping/entryBlockGraphExclusion.ts
JuliRossi, line 1: ⛏️ Perhaps we can use enums for repeated types

apps/google-docs/src/locations/Page/components/review/mapping/edit-modals/RichTextSelectionPreview.tsx
JuliRossi, line 1: Move file to edit-modals

mgoudy91 (approved):
thanks for the changes, this looks good to me!

PR #10928, opened 2026-04-24
feat(google-docs): add review all mapping mode []

zacharyyankiver:
## Summary
- add a read-only `Review all` mode for mapping review and make it the default entry point for the review step
- introduce `View only` / `Edit mode` controls that better match the latest design direction
- improve right-rail mapping cards with entry identity labels, source-aware grouping, and alignment that stays anchored to the related content on the left
- refine grouped mapping rendering so consecutive same-field content is bundled more naturally across table cells, rich text chunks, and image-adjacent content
- make read-only rendering cleaner by removing inner text-level highlight boxes when a grouped outline already exists
- make edit mode clearer by using filled grouped surfaces for assigned content, while preserving assign/exclude behavior for partial selections
- restore expected behavior for partial mappings so assigned portions still render as boxed targets and corresponding field cards continue to appear in the sidebar
- address review feedback by basing media-only decisions on underlying source metadata instead of display strings, and by making repeated same-type entries distinguishable in the right rail

## Demo
- Loom walkthrough: https://www.loom.com/share/e3eb1a39ab474300b62ad323036ef678

## Validation
- `./.circleci/prettier-check.sh`
- `npx vitest run test/locations/Page/components/review/mapping/MappingView.spec.tsx`
- `npm run build`

This PR is intentionally scoped to the validated review-mode improvements that are ready for merge. More experimental edit-mode behavior remains out of scope for this branch.

### DIFF
--- a/apps/google-docs/src/locations/Page/components/overview/OverviewSection.tsx
+++ b/apps/google-docs/src/locations/Page/components/overview/OverviewSection.tsx
@@ -51,8 +51,8 @@ const OverviewSection = ({
     borderRadius: isFirst
       ? `${tokens.borderRadiusMedium} 0 0 ${tokens.borderRadiusMedium}`
       : isLast
-        ? `0 ${tokens.borderRadiusMedium} ${tokens.borderRadiusMedium} 0`
-        : 0,
+      ? `0 ${tokens.borderRadiusMedium} ${tokens.borderRadiusMedium} 0`
+      : 0,
     backgroundColor: isActive ? tokens.colorWhite : 'transparent',
     color: isActive ? tokens.gray900 : tokens.gray700,
     boxShadow: isActive
--- a/apps/google-docs/src/locations/Page/components/overview/OverviewSection.tsx
+++ b/apps/google-docs/src/locations/Page/components/overview/OverviewSection.tsx
@@ -15,6 +15,7 @@ interface OverviewProps {
   onViewAllMappings: () => void;
   onEditMode: () => void;
   isViewingAllMappings: boolean;
+  canEditMappings?: boolean;
   ctaLabel: string;
   onCtaClick: () => void;
   isCtaLoading?: boolean;
@@ -27,6 +28,7 @@ const OverviewSection = ({
   onViewAllMappings,
   onEditMode,
   isViewingAllMappings,
+  canEditMappings = true,
   ctaLabel,
   onCtaClick,
   isCtaLoading = false,
@@ -107,20 +109,22 @@ const OverviewSection = ({
                   size="small"
                   aria-pressed={isViewingAllMappings}
                   onClick={onViewAllMappings}
-                  isDisabled={areModeButtonsDisabled}
-                  style={toggleButtonStyle(isViewingAllMappings, true, false)}>
+                  isDisabled={areModeButtonsDisabled || !canEditMappings}
+                  style={toggleButtonStyle(isViewingAllMappings, true, !canEditMappings)}>
                   View only
                 </Button>
-                <Button
-                  variant="transparent"
-                  startIcon={<PencilSimpleIcon />}
-                  size="small"
-                  aria-pressed={!isViewingAllMappings}
-                  onClick={onEditMode}
-                  isDisabled={areModeButtonsDisabled}
-                  style={toggleButtonStyle(!isViewingAllMappings, false, true)}>
-                  Edit mode
-                </Button>
+                {canEditMappings ? (
+                  <Button
+                    variant="transparent"
+                    startIcon={<PencilSimpleIcon />}
+                    size="small"
+                    aria-pressed={!isViewingAllMappings}
+                    onClick={onEditMode}
+                    isDisabled={areModeButtonsDisabled}
+                    style={toggleButtonStyle(!isViewingAllMappings, false, true)}>
+                    Edit mode
+                  </Button>
+                ) : null}
               </Box>
               <Button
                 variant="primary"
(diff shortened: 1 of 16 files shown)


apps/google-docs/src/locations/Page/components/review/mapping/buildMappingDisplayGroups.ts
```diff
@@ -45,14 +53,246 @@ function uniqueStrings(values: string[]): string[] {
...
+                  entryLabel: resolveEntryLabel(highlight),
+                  hasTextSourceRefs: isTextSourceRef(highlight.sourceRef),
+                  hasImageSourceRefs:
+                    isBlockImageSourceRef(highlight.sourceRef) ||
+                    isTableImageSourceRef(highlight.sourceRef),
+                });
+              });
+            } else {
+              const firstHighlight = partHighlights[0];
+              cards.push({
```
mgoudy91, line 230: When a table text part can't be grouped, this fallback collapses every mapping in the part onto `firstHighlight`. If the same part contains mappings for multiple fields or entries, the later mappings disappear from the rail and hover/reassign/exclude flows get attributed to the wrong field. Can we split this path by field identity, like the non-table branch does?

apps/google-docs/src/locations/Page/components/review/mapping/documentRenderers.tsx
```diff
@@ -188,13 +273,40 @@ export const BlockRenderer = ({
           rangeStart={seg.start}
           rangeEnd={seg.end}
           blockId={block.id}
+          readOnly={readOnly}
+          suppressInlineHighlight={
```
mgoudy91, line 277: In read-only all mode, this suppresses the inner mapped spans at the same time the surrounding block gets a full-surface outline. For partially mapped content that makes the entire paragraph/cell look assigned, which is pretty misleading in `View only`. Can we keep partial mappings boxed inline here instead of promoting them to a full-surface highlight?

mgoudy91 (changes requested):
Thanks for pushing this forward. I found two regressions in the new review-all rendering/grouping path that look worth fixing before merge.

apps/google-docs/src/locations/Page/components/review/mapping/buildMappingDisplayGroups.ts
zacharyyankiver: Discussed with @mgoudy91 and determined this is not needed.

apps/google-docs/src/locations/Page/components/review/mapping/documentRenderers.tsx
zacharyyankiver: Discussed with @mgoudy91 and determined this is not needed.

apps/google-docs/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -120,14 +155,36 @@ function rangeIntersectsNode(range: Range, node: Node): boolean {
   }
 }
 
+function hasPositionalDisplayLabel(label: string): boolean {
```
mgoudy91: This helper looks unused in the current diff. If nothing else calls it, I'd remove it to keep the review-mode refactor a little tighter.

apps/google-docs/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -933,75 +1070,193 @@ export const MappingView = ({
...
+                      {showGroupedSurface ? (
+                        <Box
+                          data-testid={`mapping-group-surface-${group.id}`}
+                          data-hovered={isGroupHovered ? 'true' : 'false'}
+                          style={{
+                            border: `1px solid ${
+                              isGroupHovered ? tokens.green600 : tokens.green500
+                            }`,
+                            borderRadius: tokens.borderRadiusMedium,
+                            backgroundColor: isReadOnlyAllMappings
```
mgoudy91, line 1246: `isReadOnlyAllMappings` is always false in this branch, so this conditional can be simplified to `tokens.green100`.

apps/google-docs/src/locations/Page/components/overview/OverviewSection.tsx
```diff
@@ -59,13 +90,50 @@ const OverviewSection = ({
...
+                  overflow: 'hidden',
+                  boxShadow: '0 1px 2px rgba(17, 24, 39, 0.05)',
+                }}>
+                <Button
+                  variant="transparent"
+                  startIcon={<EyeIcon />}
+                  size="small"
+                  aria-pressed={isViewingAllMappings}
+                  onClick={onViewAllMappings}
+                  isDisabled={areModeButtonsDisabled || !canEditMappings}
```
mgoudy91: When `canEditMappings` is false, this is the only visible mode button and it becomes disabled. That leaves the toggle looking broken after entry creation. I'd either hide the toggle entirely in that state or keep `View only` enabled.

mgoudy91 (commented):
Left a few small inline comments on cleanup and UX details.

mgoudy91 (approved):

apps/google-docs/src/locations/Page/components/review/mapping/MappingView.tsx
zacharyyankiver: This one is outdated relative to the later UX changes in this branch. We intentionally keep View only outline-only now, and only use the green fill in Edit mode, so I left the conditional in place to preserve that behavior.

apps/google-docs/src/locations/Page/components/overview/OverviewSection.tsx
zacharyyankiver: Done

mgoudy91 (approved):

PR #10951, opened 2026-04-29
[INTEG-3858] fix(google-docs): show all entries label in view mode, entry name in edit mode 

ryunsong-contentful:
View mode shows "All entries" under "Currently viewing:" since it renders all mappings regardless of selection. Edit mode continues to show the selected entry's content type name and title as before.


mgoudy91 (approved):
clean, good stuff

PR #10957, opened 2026-05-05
fix(locale-field-populator): display referenced entries in the preview step [ES-138]

primeinteger:
## Purpose

Fixes a bug where the Locale Field Populator app can display wrong referenced entries in the preview step. When a reference field is localized (i.e., different entries linked per locale), the app was resolving references from an arbitrary first-available locale instead of the user's selected source locale.

I noticed this bug while troubleshooting ES-138, although it was not the main issue that was being reported.

## Approach

The root cause was `Object.values(fieldValues)[0]` in `collectReferences`, which picks whichever locale happens to be first in the JSON object rather than the selected source locale. Since reference resolution previously ran on mount -- before the user selects a locale -- the source locale wasn't available.

**Fix: two-phase data loading**

1. **On mount:** fetch only the main entry and its content type
2. **On "Next" click:** resolve references using the selected source locale, then transition to the preview step

Changes:
- `entry.ts`: split `fetchEntries` into `fetchEntryAndContentType` and `fetchReferencesForLocale`; `collectReferences` now accepts a `sourceLocale` param
- `Dialog.tsx`: mount effect fetches entry/content type only; `handleNext` is async and fetches references with the correct locale; added loading state on the Next button
- Added a regression test verifying references resolve from the selected source locale

## Test plan

- [x] All 21 existing + new tests pass (`npx vitest run`)
- [x] Manual test: open the app on an entry with a localized reference field that links to different entries per locale; verify the preview shows entries matching the selected source locale ("home" entry on dev space for this app)
- [x] Verify non-localized reference fields still resolve correctly (fallback for single-key fields)
- [x] Verify back/next cycle re-fetches references when source locale changes

### DIFF
--- a/apps/locale-field-populator/src/locations/Dialog.tsx
+++ b/apps/locale-field-populator/src/locations/Dialog.tsx
@@ -8,9 +8,13 @@ import LocaleSelectionStep from '../components/steps/LocaleSelectionStep';
 import PreviewStep from '../components/steps/PreviewStep';
 import { AdoptedFieldsMap, hasAnyAdoptedFields, ReferencedEntryData } from '../utils/adoptedFields';
 import { SimplifiedLocale, mapLocaleNamesToSimplifiedLocales } from '../utils/locales';
-import { updateEntries, UpdateResult } from '../utils/entry';
+import {
+  fetchEntryAndContentType,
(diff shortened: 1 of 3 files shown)


mgoudy91 (approved):
LGTM

PR #10960, opened 2026-05-05
feat(link-checker): find links and run scan as two steps [MAPS-254]

primeinteger:
## Context

Customers want to use Link Checker as a content search tool (finding which entries contain which links) without running the expensive HTTP status scan.

Tickets: MAPS-254, CCS-3425

## Approach

Split the Page component's single `loadAuditResults` into two user-controlled phases:

1. Find links -- extracts URLs from all entries across configured content types, populates the table with `unchecked` status
2. Run scan -- checks HTTP status of each found link via the App Action

## Testing steps

1. Open the App's Page location
2. Click the "Find links" button
3. Click the "Run scan" button (enabled only after links are found)

https://github.com/user-attachments/assets/2e4919fb-cbfb-4175-8b69-4e004d54b9a8

[MAPS-254]: https://contentful.atlassian.net/browse/MAPS-254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCS-3425]: https://contentful.atlassian.net/browse/CCS-3425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/link-checker/__tests__/locations/Page.spec.tsx
+++ b/apps/link-checker/__tests__/locations/Page.spec.tsx
@@ -73,6 +73,10 @@ describe('Page component', () => {
 
   it('renders scanned links in the page table', async () => {
     render(<Page />);
+    fireEvent.click(screen.getByRole('button', { name: 'Find links' }));
+
+    await screen.findByText('https://example.invalid/not-found');
+
     fireEvent.click(screen.getByRole('button', { name: 'Run scan' }));
 
     await screen.findByText('https://example.invalid/not-found');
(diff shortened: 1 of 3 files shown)


mgoudy91 (approved):
After a scan error, `hasStartedScan` remains `true`, which keeps the "Run scan" button permanently disabled. The user has to click "Find links" again to retry even though `pendingChecks` is still intact. Consider resetting `hasStartedScan` to `false` in the catch block so a retry is possible without re-extracting. Not blocking — just worth a follow-up.

PR #10963, opened 2026-05-06
feat: support nested references in Locale Field Populator [MAPS-255]

primeinteger:
### Summary

- Locale Field Populator now recursively traverses nested entry references (not just direct refs from the main entry)
- Adds configurable max depth (default 3) in the app config screen
- Includes cycle detection and a hard cap of 50 entries to prevent runaway API calls
- Nested references render with indentation in the preview step so users can see the hierarchy

Tickets: MAPS-255, ES-183

### Context

While triaging a support issue, we discovered the app only populated localized fields on direct references. Entries linked deeper in the content graph were silently skipped. This is a common pattern for customers with modular content models (e.g., Page -> Section -> Item).

### Changes

| File | What |
|------|------|
| `utils/entry.ts` | Replaced flat reference fetch with recursive `collectReferencesRecursive` |
| `utils/adoptedFields.ts` | Added `depth` and `parentEntryId` to `ReferencedEntryData` |
| `locations/ConfigScreen.tsx` | Added "Reference depth" dropdown (0-5) |
| `locations/Dialog.tsx` | Reads `maxReferenceDepth` from install params |
| `components/steps/PreviewStep.tsx` | Renders references in traversal order with depth prop |
| `components/preview/ReferenceEntrySection.tsx` | Accepts `depth`, applies indentation |
| `components/preview/ReferenceEntrySection.styles.ts` | `depthIndent` utility |

### Test plan

- [x] Install app on a space with nested content (Page -> Section -> Item pattern)
- [x] Verify depth-1 references still render as before
- [x] Verify depth-2+ references appear indented in preview
- [x] Confirm circular references are detected and shown with the self-reference warning
- [x] Change max depth in config screen, verify it limits traversal
- [x] Confirm "Adopt all fields" and per-field toggles work for nested entries
- [x] Run `tsc --noEmit`, `vite build`, `eslint .` -- all pass locally

### Screenshots

New setting in the app configuration page:

> Screenshot 2026-05-06 at 9 20 33 AM

Nested references showing at the bottom of the modal:

> Screenshot 2026-05-06 at 8 55 14 AM

Reference accordion items expanded in the modal:

> Screenshot 2026-05-06 at 8 55 32 AM

### DIFF
--- a/apps/locale-field-populator/src/components/preview/ReferenceEntrySection.styles.ts
+++ b/apps/locale-field-populator/src/components/preview/ReferenceEntrySection.styles.ts
@@ -1,6 +1,11 @@
 import { css } from 'emotion';
 import tokens from '@contentful/f36-tokens';
 
+export const depthIndent = (depth: number) =>
+  css({
+    marginLeft: depth > 1 ? `${(depth - 1) * 24}px` : undefined,
+  });
+
 export const styles = {
   accordionContainer: css({
     borderLeft: `1px solid ${tokens.gray300}`,
--- a/apps/locale-field-populator/src/components/preview/ReferenceEntrySection.tsx
+++ b/apps/locale-field-populator/src/components/preview/ReferenceEntrySection.tsx
@@ -4,7 +4,7 @@ import { ContentTypeProps, EntryProps } from 'contentful-management';
 import { useMemo } from 'react';
 import { isEntryArrayField, isEntryField } from '../../utils/fieldTypes';
 import PreviewFieldRow from './PreviewFieldRow';
-import { styles } from './ReferenceEntrySection.styles';
+import { depthIndent, styles } from './ReferenceEntrySection.styles';
 import { ArrowSquareOutIcon } from '@contentful/f36-icons';
 
 interface ReferenceEntrySectionProps {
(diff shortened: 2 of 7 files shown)


mgoudy91 (changes requested):
**Issue 1:** `maxDepth=0` ("None - main entry only") doesn't work as expected. `collectReferencesRecursive` is called with `currentDepth=1`, and the `currentDepth < maxDepth` guard only prevents further recursion - it doesn't stop the initial collect-and-push loop. So depth=0 still surfaces all direct references. Fix: add an early return at the top of `collectReferencesRecursive`: `if (currentDepth > maxDepth) return;`

**Issue 2:** When the 50-entry hard cap is hit, traversal silently stops. Users have no indication that references are missing - they'll assume everything was processed and may populate only a partial content graph. Worth adding a notice when `referencedEntries.length >= 50`, e.g. "Some entries were not shown. Try reducing the reference depth."

mgoudy91:
**`apps/locale-field-populator/src/utils/adoptedFields.ts`** — `parentEntryId` is added to `ReferencedEntryData` but nothing in the changed components reads it. Either document its intended use (breadcrumb navigation, grouping by parent, etc.) or remove it to avoid dead interface surface.

mgoudy91 (approved):
perfect

PR #10966, opened 2026-05-07
feat(google-docs): Selec entries to create [INTEG-3894]

joaquincasal:
## Summary

Adds per-entry selection to the Google Docs import review flow, so users can choose which previewed entries are created.

## Changes

- Added entry checkboxes, selected by default.
- Updated CTA to `Create selected entries`.
- Sends only selected entries in the resumed `entryBlockGraph`.
- Prunes references to unselected entries.

## Testing

- Added focused frontend and backend tests.

https://github.com/user-attachments/assets/e363721c-6259-45e5-ab9a-c2345579bfbf


apps/google-docs/src/locations/Page/components/overview/OverviewEntryList.tsx
```diff
@@ -13,56 +13,85 @@ import { truncateLabel } from '../../../../utils/utils';
...
-        </Paragraph>
+        <Flex alignItems="center" gap="spacingXs">
+          <Checkbox
+            aria-label={`Create entry ${row.contentTypeName || 'Untitled'}${
+              row.entryTitle ? ` (${truncateLabel(row.entryTitle, 150)})` : ''
+            }`}
+            isChecked={isEntrySelectedForCreation}
+            isDisabled={areEntrySelectionsDisabled}
+            onChange={(event) => onToggleEntrySelection(row.id, event.target.checked)}
+          />
```
JuliRossi, line 66: 🤔  There seem to be an issue with the checkbox (saw this on the video). Instead of showing the hand pointer ([default from forma) ](https://f36.contentful.com/components/checkbox) is showing the arrow pointer. Perhaps it has to do with the missing props in Card component that were setting up the cursor

apps/google-docs/src/locations/Page/components/review/ReviewPage.tsx
```diff
@@ -74,20 +84,45 @@ export const ReviewPage = ({
...
   const isMappingDisabled = isCreatePending || hasCreatedEntries;
+  const selectedEntryCount = useMemo(
+    () => countSelectedEntries(entryBlockGraph.entries, selectedEntryKeys),
+    [entryBlockGraph.entries, selectedEntryKeys]
+  );
+  const hasSelectedEntries = selectedEntryCount > 0;
 
   const { resumeWorkflow } = useWorkflowAgent({ sdk, documentId: '', oauthToken: '' });
 
+  const handleToggleEntrySelection = useCallback((entryKey: string, isSelected: boolean) => {
```
JuliRossi: ⛏️  Does this need to be a callback?

JuliRossi (approved):
LGTM! Left a few comments

apps/google-docs/src/locations/Page/components/review/ReviewPage.tsx
joaquincasal: Hmm probably not, I'll remove it

apps/google-docs/src/locations/Page/components/overview/OverviewEntryList.tsx
joaquincasal: It's weird, but that only happens when I record the video, when I use the app without recording, it shows the hand pointer. I cannot prove it as taking a screenshot also removes the cursor, but I checked and it works as expected

mgoudy91 (changes requested):
A few issues worth addressing before merge — two correctness/behavioral risks and one quick perf fix, plus one UX suggestion.

- **Issue 1**: `getEntrySelectionKey` index fallback can silently map keys to wrong entries if entries are reordered, and risks collision with numeric `tempId` values.
- **Issue 2**: `filterEntryBlockGraphBySelection` doesn't prune the `referenceGraph` — stale edges to deselected entries may cause backend errors depending on whether the server rebuilds it.
- **Issue 3**: `handleToggleEntrySelection` is missing `useCallback`, causing unnecessary re-renders of the full entry tree on every selection change.
- **Issue 5** (suggestion): The removed UX hint about clicking entry rows to preview content left users with no affordance for that interaction.

joaquincasal:
> A few issues worth addressing before merge — two correctness/behavioral risks and one quick perf fix, plus one UX suggestion.
> 
>     * **Issue 1**: `getEntrySelectionKey` index fallback can silently map keys to wrong entries if entries are reordered, and risks collision with numeric `tempId` values.
> 
>     * **Issue 2**: `filterEntryBlockGraphBySelection` doesn't prune the `referenceGraph` — stale edges to deselected entries may cause backend errors depending on whether the server rebuilds it.
> 
>     * **Issue 3**: `handleToggleEntrySelection` is missing `useCallback`, causing unnecessary re-renders of the full entry tree on every selection change.
> 
>     * **Issue 5** (suggestion): The removed UX hint about clicking entry rows to preview content left users with no affordance for that interaction.

Issue 1: I addressed the key collision risk by namespacing selection keys. Regarding reordering, I don’t think entries can be reordered in the current UI, so it shouldn’t be a problem unless I’m missing something.

Issue 2: The existing `referenceGraph` is not sent after filtering. We only send the selected-only `entryBlockGraph` to the backend, and the backend recreates `referenceGraph`, so I don’t think we’re at risk of errors caused by stale edges. Let me know if you think otherwise.

Issue 3: I think adding `useCallback` wouldn’t save us from re-renders, as selection changes still create a new `selectedEntryKeys` set and that prop flows through the entry tree. The child rows are also not memoized, so stabilizing only the handler identity doesn’t change much.

Issue 5: That’s how it is in the designs, and I checked with Erin.

mgoudy91 (approved):
Thanks for thinking through that, lgtm

PR #10973, opened 2026-05-12
feat(content-insights): performance improvements [MAPS-256]

primeinteger:
## Purpose

Fix Content Insights load times on large spaces by replacing serial offset pagination with parallel cursor streams.

## Approach

- Replace the single `entry.getMany` skip/limit loop with `entry.getManyWithCursor`.
- Fan out one cursor stream per content type with bounded concurrency (`FETCH_CONFIG.CONCURRENCY = 5`, headroom under the ~7 req/s CMA limit) via a small `concurrentMap` worker pool.
- Extract `cursorPaginate` to drain any cursor-paginated CMA endpoint, preserving the existing adaptive batch-shrinking on `Response size too big` (halve down to `MIN_BATCH_SIZE`).
- Lift `useContentTypes` ahead of `useAllEntries` in `Dashboard` and pass content type ids in -- entries fetch is now keyed on the resolved id list.
- Bump `contentful-management` 11.63.1 -> 11.76.0 for `getManyWithCursor`.

### DIFF
--- a/apps/content-insights/package-lock.json
+++ b/apps/content-insights/package-lock.json
@@ -16,7 +16,7 @@
         "@tanstack/react-query": "^5.90.12",
         "@tanstack/react-query-devtools": "^5.91.1",
         "@testing-library/user-event": "^14.6.1",
-        "contentful-management": "11.63.1",
+        "contentful-management": "11.76.0",
         "dayjs": "^1.11.19",
         "emotion": "10.0.27",
         "happy-dom": "^20.0.11",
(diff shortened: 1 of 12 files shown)


mgoudy91 (changes requested):
When one runner throws and `Promise.all` rejects, the other in-flight runners continue executing. If any of those subsequent runners also throw, the exception becomes an unhandled promise rejection -- no one is awaiting them after `Promise.all` has already settled. On large spaces with many concurrent streams this could surface as `UnhandledPromiseRejection` warnings or trigger `window.onerror`.

A simple fix is a shared `aborted` flag checked at the top of the loop:

```ts
let aborted = false;
const runners = Array.from({ length: Math.min(concurrency, items.length) }, async () => {
  while (!aborted) {
    const i = nextIndex++;
    if (i >= items.length) return;
    try {
      results[i] = await worker(items[i], i);
    } catch (err) {
      aborted = true;
      throw err;
    }
  }
});
```

primeinteger:
> When one runner throws and `Promise.all` rejects, the other in-flight runners continue executing.

Good catch - I applied the `aborted` flag as suggested, and added a test that asserts no further items are picked up once a worker throws

mgoudy91 (approved):

PR #10974, opened 2026-05-13
fix(locale-field-populator): handle duplicate references [MAPS-256]

primeinteger:
## Purpose

When an entry is referenced from multiple paths in a content graph (e.g. directly on the main entry and again under a nested entry), the second occurrence rendered with a misleading "This entry has no localized fields" warning, making users think the app was broken.

## Approach

- Treat the highest-level occurrence as the canonical one (full localization UI) and emit subsequent occurrences as explicit "already included above" pointer records, rather than dropping them silently or letting them appear as empty sections.
- Pre-claim every direct reference of a parent before recursing into any sibling, so a sibling's subtree can't steal a later sibling and invert the tree depth.
- Recurse inline so each entry's nested children appear immediately after it in the result list, which is what the depth-based indentation in the preview relies on.
- Skip duplicate records in `initializeAdoptedFields`, the field-count summary, and the headline entry count so the "N referenced entries with M localizable fields" note matches what is actually translatable.
- Unify accordion title font size across linked and unlinked title rows.

## Screenshot

Duplicate referenced entries will show a new "This entry is already included above and will be updated once." banner

Screenshot 2026-05-13 at 11 54 26 AM

### DIFF
--- a/apps/locale-field-populator/src/components/preview/ReferenceEntrySection.styles.ts
+++ b/apps/locale-field-populator/src/components/preview/ReferenceEntrySection.styles.ts
@@ -21,6 +21,11 @@ export const styles = {
     borderRadius: tokens.borderRadiusSmall,
     marginTop: tokens.spacingS,
   }),
+  noteNeutral: css({
+    padding: tokens.spacingM,
+    borderRadius: tokens.borderRadiusSmall,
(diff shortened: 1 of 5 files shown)


mgoudy91 (approved):
nice, good catch 😄

PR #10975, opened 2026-05-13
fix(google-docs): register OAuth functions as actions in manifest []

JuliRossi:
## Purpose
The four OAuth functions (`initiateGdocOauth`, `completeGdocOauth`, `revokeGdocOauthToken`, `checkGdocOauthTokenStatus`) were defined only in the `functions` section of the manifest but not registered as app actions. This caused `getManyForEnvironment` to return an empty list, resulting in a `NotFound` error when the client tried to invoke them.

## Summary
- Added `actions` section to `contentful-app-manifest.json` with all four OAuth functions registered as `function-invocation` type actions linked via `functionId`

## Test plan
- [ ] OAuth connect flow works without `NotFound` error
- [ ] `checkGdocOauthTokenStatus` action is found and invoked successfully on page load
- [ ] Disconnect flow works via `revokeGdocOauthToken`

### DIFF
--- a/apps/drive-integration/contentful-app-manifest.json
+++ b/apps/drive-integration/contentful-app-manifest.json
@@ -17,12 +17,8 @@
       "description": "Initiates the OAuth flow for Drive Integration",
       "path": "functions/oauth/initiateOauth.js",
       "entryFile": "functions/oauth/initiateOauth.ts",
-      "allowNetworks": [
-        "oauth2.googleapis.com"
-      ],
-      "accepts": [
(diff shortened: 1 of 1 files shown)


FBanfi (approved):
Looks good! Just a heads up that the PR description mentions two changes that aren't in the diff:
  - Updated getAppActionId in appAction.ts to match by action.sys.id instead of action.name
  - Removed the upsertAppActions script

FBanfi (approved):

PR #10983, opened 2026-05-15
feat(google-docs): allow editing content mappings across all entries [INTEG-3885]

joaquincasal:
## Summary

  - The Edit Content Mapping modal now shows **all proposed entries** with individual field selectors, allowing users to assign or remove
  selected content to fields across multiple entries in a single action
  - Each entry card displays its content type, title, and a multi-select field dropdown
  - The Save button is disabled until at least one field assignment change is made across any entry
  - Also included some minors UI changes related to the "Edit content mapping" button and the space between the control buttons of the modal and the actual content

  ## Test plan

  - [ ] Select text in the review document → open Edit Content Mapping → verify all proposed entries appear with their field dropdowns
  - [ ] Assign the same content to fields on multiple entries simultaneously → Save → verify mappings applied correctly
  - [ ] Remove a field mapping from an entry that is NOT currently selected in the sidebar → verify the removal takes effect
  - [ ] Verify the Save button stays disabled until a change is made
  - [ ] Repeat the above with image content (edit icon on an image card)

https://github.com/user-attachments/assets/9aeb7efe-61b2-4452-b9d8-d3cb3ef3bcef


mgoudy91 (approved):
LGTM

PR #10986, opened 2026-05-15
feat(drive-integration): wire up CI deploy scripts for lerna []

mgoudy91:
## Summary

- Adds `deploy` and `deploy:test` scripts so lerna picks up the drive-integration app in the standard CI pipeline on master merge
- `deploy` uploads to prod app definition `3EaGZUMKRKVZUyrcoNJ4o4` via `contentful-app-scripts` using vault-injected `${DEFINITIONS_ORG_ID}` / `${CONTENTFUL_CMA_TOKEN}`, then syncs to `${STATIC_S3_BASE}/google-docs` and invalidates the prod CloudFront distribution
- `deploy:test` does the same for the staging definition (`4i0mp5lQtgNsHYVrD5jIkj`) using `${TEST_ORG_ID}` / `${TEST_CMA_TOKEN}` and the staging S3 path
- No build step in either script — CI runs `npm run build` before calling deploy, matching the pattern used by all other apps

## Test plan

- [ ] Confirm `lerna run deploy --since=HEAD^` on master includes drive-integration
- [ ] Confirm `lerna run deploy:test --since=HEAD^` on staging includes drive-integration
- [ ] Verify the staging deployment hits the correct app definition after merge to master


joaquincasal (approved):

apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
```diff
@@ -64,7 +64,17 @@ export const EditModal = ({
 
   const handleSelectableStateChangeForEntry = useCallback(
     (entryId: string) => (state: { hasFieldOptions: boolean; hasSelectableOptions: boolean }) => {
-      setDestinationFieldStateByEntry((prev) => ({ ...prev, [entryId]: state }));
```
joaquincasal, line 67: This change was necessary because it was causing a test to fail:

EditModal was triggering an infinite render loop whenever the modal opened with at least one entry. The chain:

  1. EditModal renders. Inside the entries .map() it calls handleSelectableStateChangeForEntry(loc.id) to build the
  onSelectableStateChange prop for each FieldSelectionDropdown. Because that's a curried useCallback, calling it
  returns a brand-new inner function reference on every render.
  2. FieldSelectionDropdown has a useEffect that fires onSelectableStateChange(...) and includes the prop in its dep
  array — so the new ref makes the effect re-run on every render.
  3. The handler ran:
  setDestinationFieldStateByEntry((prev) => ({ ...prev, [entryId]: state }));
  3. That spread always returns a new object, so React never bails out via Object.is — even when hasFieldOptions /
  hasSelectableOptions haven't changed.
  4. New state object → re-render → new function ref → effect re-fires → goto 3. Infinite loop.

The fix: short-circuit the updater when the value for entryId is unchanged. When existing matches the incoming state, return prev (same reference), React bails, no re-render, the function ref stays stable, the dropdown's effect settles.

apps/drive-integration/package.json
```diff
@@ -22,27 +22,25 @@
...
-    "deploy:staging": "npm run build && npm run upload:app-staging && npm run deploy:sync-staging",
     "upload:app-flinkly": "contentful-app-scripts upload --bundle-dir ./build --organization-id 3m1Cqy5lzALpWNY9GjP2Zm --definition-id 7r10omUtVNoq8X3Yy4YfBL --token $FLINKLY_CONTENTFUL_ACCESS_TOKEN",
     "deploy:sync-flinkly": "aws s3 sync ./build ${STATIC_S3_BASE}/google-docs-dev/flinkly && aws cloudfront create-invalidation --distribution-id $GOOGLE_DOCS_TEST_CLOUDFRONT_DIST_ID --paths \"/*\" > /dev/null 2>&1",
     "deploy:flinkly": "npm run build && npm run upload:app-flinkly && npm run deploy:sync-flinkly",
+    "upload:app-prod": "contentful-app-scripts upload --ci --bundle-dir ./build --organization-id 5EJGHo8tYJcjnEhYWDxivp --definition-id 3EaGZUMKRKVZUyrcoNJ4o4 --token $CONTENTFUL_ACCESS_TOKEN",
     "deploy:sync-prod": "aws s3 sync ./build ${PROD_STATIC_S3_BASE}/google-docs && aws cloudfront create-invalidation --distribution-id $GOOGLE_DOCS_PROD_CLOUDFRONT_DIST_ID --paths \"/*\" > /dev/null 2>&1",
```
joaquincasal: @mgoudy91 is `PROD_STATIC_S3_BASE` correct here? Should it be `STATIC_S3_BASE`? I don't see `PROD_STATIC_S3_BASE` in https://github.com/contentful/apps/blob/master/.circleci/config.yml

mgoudy91: I think you are right, copy paste error

joaquincasal:
```suggestion
    "deploy:sync-prod": "aws s3 sync ./build ${STATIC_S3_BASE}/google-docs && aws cloudfront create-invalidation --distribution-id $GOOGLE_DOCS_PROD_CLOUDFRONT_DIST_ID --paths \"/*\" > /dev/null 2>&1",
```

FBanfi (approved):

PR #10987, opened 2026-05-15
fix(rich-text-versioning): inject reference recommendations into entity picker [ES-262]

primeinteger:
## Summary

An issue was reported that inside the Rich Text Versioning marketplace app, the entity picker's "Suggested" tab loads but never returns results. The same picker on the native rich text editor populates correctly.

Root cause: `recommendations.entityId` and `recommendations.referenceFieldId` are required for the picker's `/semantic/reference-suggestions` call, but nothing in this app's path injects them. On the host, `AiEnabledSingleLinkActions` does the injection for plain reference fields, but for marketplace apps in a RichText location, the host has no equivalent. `field-editor-rich-text`'s `HyperlinkModal` only sets `searchQuery`.

## Fix

`Field.tsx` wraps the iframe SDK with a thin decorator that injects `recommendations: { entityId, referenceFieldId }` into `dialogs.selectSingleEntry` and `dialogs.selectMultipleEntries`. Caller-supplied `recommendations` (e.g. `searchQuery` from `HyperlinkModal`) merge on top, so existing behaviour is preserved. The decorated SDK is passed to `<RichTextEditor>` so the embed-block, embed-inline, and hyperlink picker call sites all participate.

The `recommendations` key isn't declared on `@contentful/app-sdk`'s dialog options type, but the host's dialog channel forwards arbitrary option keys to the picker, which reads `recommendations` directly. The wrapper casts at the boundary.

## Required dep bumps

The decorator only takes effect against `field-editor-rich-text >= 4.10.0` (when `HyperlinkModal` started passing `recommendations` through). To stay current and pick up other RT improvements, this PR also bumps:

- `@contentful/field-editor-rich-text`: `4.16.0` -> `6.3.5`
- `@contentful/app-sdk`: `4.42.0` -> `4.54.0`
- `@contentful/f36-components` / `f36-multiselect` / `f36-tokens`: `4.x` -> `6.x`
- `@contentful/react-apps-toolkit`: `1.2.16` -> `2.0.0`
- `@contentful/rich-text-html-renderer` / `react-renderer` / `types`: minor bumps
- `contentful-management`: `11.54.4` -> `^12.3.0`
- `emotion@10` (deprecated single-package style import) -> `@emotion/css@^11`

The emotion package rename touches four `*.styles.ts` files mechanically (only the import statement changes).

`@tanstack/react-query@^5.0.0` is added as a transitive expectation of the newer field-editor packages, and `@lingui/core@^5.3.0` is added because the newer `@contentful/react-apps-toolkit` requires it as a peer.

## Companion / dependent fixes

The decorator only fully unblocks the customer scenario when paired with two upstream fixes that this work surfaced:

- **contentful/experience-packages#3480**: switches `entity-search`'s semantic fetch from `cmaClient.raw.post(...)` to typed CMA methods (`cmaClient.semanticSearch.get`, `cmaClient.semanticReferenceSuggestions.get`). Without this, the marketplace-app CMA adapter rejects the picker's request with "You can not make raw cma requests from an app." See PR for full root cause.
- **contentful/field-editors#2159**: handles react-query v5's `'pending'` status in the `FetchingWrapped*` cards. Surfaces only in vite-bundled marketplace apps that hoist v5 over field-editor-reference's nested v4. Without this, picking an entry from Suggested crashes `WrappedEntryCard`.

This PR is independently shippable: the decorator is the right thing in this app regardless of when the upstream fixes land. End-to-end the customer scenario is unblocked once all three are in production.

## Test plan

- [x] Added decorator unit tests in `test/locations/Field.spec.tsx` covering: base-case injection, caller-supplied `recommendations` merge (e.g. `searchQuery` preservation), caller-supplied keys overriding the injected base, and `selectMultipleEntries` parity.
- [x] Existing tests still pass: `npm run test:ci` (36/36 pass).
- [x] `npm run build` clean.
- [x] Verified end-to-end against production space (via local dev server) with locally-patched dependencies (mirroring the upstream fixes): `/semantic/search` and `/semantic/reference-suggestions` fire with `entityId`/`referenceFieldId`/`searchQuery` in the request body, picker populates with entries, and picking-then-inserting works.
- [ ] Once upstream PRs are released, verify against unpatched deps.

### DIFF
--- a/apps/rich-text-versioning/package-lock.json
+++ b/apps/rich-text-versioning/package-lock.json
@@ -9,25 +9,29 @@
       "version": "0.1.0",
       "dependencies": {
         "@ali-tas/htmldiff-js": "^2.0.1",
-        "@contentful/app-sdk": "^4.42.0",
-        "@contentful/f36-components": "4.81.1",
-        "@contentful/f36-multiselect": "^4.81.1",
-        "@contentful/f36-tokens": "4.2.0",
-        "@contentful/field-editor-rich-text": "4.16.0",
-        "@contentful/react-apps-toolkit": "1.2.16",
-        "@contentful/rich-text-html-renderer": "^17.1.0",
-        "@contentful/rich-text-react-renderer": "^16.1.0",
-        "@contentful/rich-text-types": "^16.1.0",
+        "@contentful/app-sdk": "^4.54.0",
+        "@contentful/f36-components": "^6.7.1",
+        "@contentful/f36-multiselect": "^6.7.1",
+        "@contentful/f36-tokens": "^6.1.2",
+        "@contentful/field-editor-rich-text": "6.3.5",
+        "@contentful/react-apps-toolkit": "^2.0.0",
+        "@contentful/rich-text-html-renderer": "^17.2.2",
+        "@contentful/rich-text-react-renderer": "^16.2.1",
+        "@contentful/rich-text-types": "^17.2.5",
+        "@emotion/css": "^11.13.5",
+        "@lingui/core": "^5.3.0",
+        "@tanstack/react-query": "^5.0.0",
         "contentful": "^11.7.15",
-        "contentful-management": "11.54.4",
+        "contentful-management": "^12.3.0",
         "dompurify": "^3.0.8",
-        "emotion": "10.0.27",
         "happy-dom": "^20.0.0",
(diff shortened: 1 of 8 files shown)


mgoudy91 (approved):
LGTM

PR #10988, opened 2026-05-15
fix(drive-integration): fix overscroll-behavior [INTEG-3984]

FBanfi:
## Purpose

Background scroll leak: scrolling the field selection dropdown caused the page behind the modal to scroll simultaneously. That made the field dropdown portal to be closed when it shouldn't.

## Approach

The f36 Multiselect popover renders via FloatingPortal, appending directly to `<body>`. When the dropdown list hits its scroll boundary, the browser chains the wheel event up the DOM to body, scrolling the page.
- The fix locks document.body overflow to hidden while the modal is open and restores it on close, straightforward and browser-consistent.

## Testing steps

Go to the edit modal, open the field dropdown and scroll, hitting bottom. It should close the portal neither scroll the background page.

https://github.com/user-attachments/assets/b0c6d322-5058-4bd9-b9a3-9b7d02ca5a7f

## Breaking Changes

No.

No.

## Deployment

No.

### DIFF
--- a/apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
+++ b/apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
@@ -36,6 +36,15 @@ export const EditModal = ({
 
   useEffect(() => {
     if (!isOpen) return;
+    const prev = document.body.style.overflow;
+    document.body.style.overflow = 'hidden';
+    return () => {
+      document.body.style.overflow = prev;
+    };
+  }, [isOpen]);
+
+  useEffect(() => {
+    if (!isOpen) return;
     setSelectedFieldIdsByEntry(
       Object.fromEntries(viewModel.newLocations.map((loc) => [loc.id, loc.initialFieldIds]))
     );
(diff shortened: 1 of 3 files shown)


apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
```diff
@@ -34,6 +34,15 @@ export const EditModal = ({
...
   const newLocationIdsKey = viewModel.newLocations.map((loc) => loc.id).join('|');
 
+  useEffect(() => {
+    if (!isOpen) return;
+    const prev = document.body.style.overflow;
+    document.body.style.overflow = 'hidden';
+    return () => {
+      document.body.style.overflow = prev;
+    };
+  }, [isOpen]);
```
JuliRossi, line 44: ⛏️  Perhaps this could be a hook with a name to show what it fixes. Or you could add a comment on the why this is necessary (for future reference)

JuliRossi (approved):
⭐ Nice catch ! Leave a small comment to improve maintainability but you can disregard

apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/EditModal.tsx
FBanfi: Will do it in the next PR!

PR #10989, opened 2026-05-18
feat&fix(drive-integration): adding no content mapped badge + general fixes [INTEG-3893]

FBanfi:
## Purpose
  
  Two independent fixes shipped together:

  1. Infinite re-render loop in EditModal – opening the edit modal caused a silent infinite render loop
  that locked the UI and hung vitest workers before any test could run.
  2. "No mapped content" badge on entry list rows – entries with no field mappings now show a No mapped 
  content badge in the overview entry list so users can spot unmapped entries at a glance.
  3. Fixed some Page and MappingView failing tests.

### Infinite re-render fix (FieldSelectionDropdown)

  EditModal passes handleSelectableStateChangeForEntry(loc.id) as the onSelectableStateChange prop — a
  curried call made during render, producing a new function reference every render cycle. That reference
  was listed in the useEffect dependency array inside FieldSelectionDropdown, so every render triggered
  the effect → setState in EditModal → new render → new function → effect again → infinite loop.

  The fix stores the callback in a ref so the effect's deps only contain the actual data values
  (fieldOptions.length, isImageContent, selectableOptions.length), breaking the cycle.

  ### "No mapped content" badge

  - Added hasNoMappedContent flag to EntryListRow, computed in buildEntryListFromEntryBlockGraph by
  checking whether every fieldMapping has an empty sourceRefs array.
  - OverviewEntryRowCard renders a secondary Badge inline with the entry title when the flag is set.

### Infinite loop fix:

  - Before:

> https://github.com/user-attachments/assets/601b48aa-0080-424c-a9b0-d33c99de73bd

  - Now:
  

> https://github.com/user-attachments/assets/6a509eca-d888-4042-a2b5-f67ab23a8bc2

  
  ### "No mapped content" badge:

> https://github.com/user-attachments/assets/ae4fb5b4-6a18-416e-9709-47beba07cada

  
## Breaking Changes

  No.

  Link to [INTEG-3893](https://contentful.atlassian.net/jira/software/c/projects/INTEG/boards/454?issueParent=476933&selectedIssue=INTEG-3893)

## Deployment
  
  No.

[INTEG-3893]: https://contentful.atlassian.net/browse/INTEG-3893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
+++ b/apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
@@ -1,4 +1,4 @@
-import { Box, Card, Checkbox, Flex, Paragraph, Text } from '@contentful/f36-components';
+import { Badge, Box, Card, Checkbox, Flex, Paragraph, Text } from '@contentful/f36-components';
 import tokens from '@contentful/f36-tokens';
 import { cx } from '@emotion/css';
 import type { EntryListRow as OverviewEntryListRow } from '../../../../utils/overviewEntryList';
@@ -90,6 +90,11 @@ function OverviewEntryRowCard({
                 </Text>
               ) : null}
             </Paragraph>
+            {row.hasNoMappedContent ? (
+              <Badge variant="secondary" size="small">
+                No mapped content
+              </Badge>
+            ) : null}
           </button>
         </Flex>
       </Card>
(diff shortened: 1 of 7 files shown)


apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
```diff
@@ -89,6 +90,11 @@ function OverviewEntryRowCard({
                   ({truncateLabel(row.entryTitle, 150)})
                 </Text>
               ) : null}
+              {row.hasNoMappedContent ? (
```
joaquincasal: Nit: we could use `{ row.hasNoMappedContent && ...}` here, instead of the ternary

joaquincasal (approved):
Looks good!

mgoudy91 (approved):
LGTM

PR #11005, opened 2026-05-22
Add eslint script [INTEG-3947]

joaquincasal:
Adds ESLint to the Drive Integration app, with the existing errors fixed

### DIFF
--- a/apps/drive-integration/src/setupTests.ts
+++ b/apps/drive-integration/src/setupTests.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
 // jest-dom adds custom jest matchers for asserting on DOM nodes.
 // allows you to do things like:
 // expect(element).toHaveTextContent(/react/i)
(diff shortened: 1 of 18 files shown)


apps/drive-integration/src/setupTests.ts
JuliRossi, line 1: 💭 We could import these
```
 const { configure } = require('@testing-library/react');
    const { expect } = require('vitest'); 
```
the same way we do in other apps, instead of using require

apps/drive-integration/test/locations/Page/Page.spec.tsx
```diff
@@ -77,6 +77,7 @@ vi.mock('../../../src/locations/Page/components/review/ReviewPage', () => ({
 }));
 
 vi.mock('../../../src/locations/Page/components/mainpage/ModalOrchestrator', () => ({
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
   ModalOrchestrator: require('react').forwardRef(
```
JuliRossi, line 81: 💭 Same here, instead of using require, I think is best to simply import React

JuliRossi (approved):
Nice work! Left a few non-blocking commnets

apps/drive-integration/src/setupTests.ts
joaquincasal: I tried removing the eslint-disable but the changes were not straightforward, so I'd rather leave them for now so we don't introduce too many changes with this

PR #11007, opened 2026-05-29
fix: bump rich-text-editor to latest version [ES-319]

chrishelgert:
## Purpose

Bumping rich-text editor to latest version to include bug fixes for RTL

- [x] check that it doesnt break


chrishelgert:
Everything seems to work except resource links - I think this was expected as it's not supported yet?
`props.sdk.dialogs.selectSingleResourceEntity is not a function`

colomolo (approved):

joseliacosta (approved):

PR #11016, opened 2026-06-08
feat(content-insights): add content type filter for Needs Update metric [INTEG-4147]

harikakondur:
## Summary

- New **Content types included in "Needs update"** multiselect in the config screen sets a persistent default
- Inline **Filter by content type** in the "Needs update" tab lets users adjust the view at runtime
- Empty selection includes all content types (preserves existing behavior)
- Table page size increased from 5 → 10 rows across all tabs

## Context

The "Needs update" metric was reporting 54,000+ entries because it counted all content types, including reusable components (Links, CTAs, Nav items) that don't need periodic review.

### Before
Before

### After
After config
After tab

https://github.com/user-attachments/assets/1018e13b-e49f-48f0-9e4a-9ae655f1e510

## Test plan

- [ ] All tests pass (`npx vitest --run`)
- [ ] Config screen filter persists and scopes the metric card and table
- [ ] Inline tab filter updates the table in real time
- [ ] Empty selection shows all content types
- [ ] All tabs show 10 rows per page

### DIFF
--- a/apps/content-insights/src/components/ScheduledContentTabs.tsx
+++ b/apps/content-insights/src/components/ScheduledContentTabs.tsx
@@ -86,12 +86,18 @@ export const ScheduledContentTabs = ({
                   initialSelectedIds={installation.needsUpdateContentTypes}
                   disablePills={false}
                 />
-                <FormControl.HelpText>Leave empty to show all content types.</FormControl.HelpText>
+                <FormControl.HelpText>
+                  Leave empty to use the configured default.
+                </FormControl.HelpText>
               </FormControl>
               <NeedsUpdateTable
                 entries={entries}
                 contentTypes={contentTypes}
-                selectedContentTypeIds={selectedNeedsUpdateContentTypes.map((ct) => ct.id)}
+                selectedContentTypeIds={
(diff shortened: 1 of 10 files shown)


apps/content-insights/src/components/ScheduledContentTabs.tsx
```diff
@@ -70,7 +77,26 @@ export const ScheduledContentTabs = ({
               description={`Content older than ${needsUpdateMonths} ${
                 needsUpdateMonths === 1 ? 'month' : 'months'
```
mgoudy91, line 78: The help text says "Leave empty to show all content types," but clearing the tab filter passes `undefined` to `selectedContentTypeIds`, which makes the hook fall back to `installation.needsUpdateContentTypes` rather than showing all entries. So if an admin configured a default set, clearing the inline filter still scopes to that default, not all types. Either change the fallback so an empty selection means "all" (pass an empty array instead of `undefined`), or update the help text to say "Leave empty to use the configured default."

mgoudy91 (approved):
LGTM

mgoudy91 (approved):

PR #11012, opened 2026-06-04
fix(drive-integration): show actionable error when app not installed in environment []

mgoudy91:
## Problem

When a Google Drive import fails due to the app not being installed in a space, users see a generic "This preview could not be completed. Please start again." message with no actionable guidance.

## Fix

Adds frontend handling for two new failure codes surfaced by agents-api:

- `app-not-installed` — shown when the app is not installed in the space (after the master env fallback in agents-api also finds nothing). Message: "The Drive Integration app is not installed in this space. Install it via Apps > Manage apps and try again."
- `ai-service-unavailable` — shown when the AI service is temporarily unavailable. Message: "The AI service is temporarily unavailable. Please try again in a few minutes."

Both codes are handled in `useWorkflowAgent`, `WorkflowFailureReason`, and `ModalOrchestrator`.

## Test plan

- [ ] Verify `APP_NOT_INSTALLED` error modal appears with correct title and message
- [ ] Verify `AI_SERVICE_UNAVAILABLE` error modal appears with correct title and message
- [ ] Verify generic error still shows for unrecognized failure codes

**Related:** contentful/agents-api#615

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11020, opened 2026-06-10
fix(drive-integration): resolve environment alias for agents API calls [INTEG-4151]

JuliRossi:
## Purpose
`sdk.ids.environment` always returns the raw environment ID (e.g. `master-2026-06-09`), even when the app is opened via an alias like \`master\`. This caused the agents API path to use the raw env ID, which doesn't match the app installation scoped to the alias — resulting in a 403. The issue also affected environments accessed directly by their raw ID when an alias exists.

Also ports `APP_NOT_INSTALLED`  error handling from #11012.

## Summary
- Added a `resolveEnvironmentId` helper in `useWorkflowAgent.ts` that returns `sdk.ids.environmentAlias` when available (the fast path for alias-based access)
- Falls back to fetching the environment via CMA and reading \`sys.aliases[0]\` to find the alias name when the app is accessed via the raw env ID
- Applied to both `startWorkflow` and `resumeWorkflow` call sites
- Added `APP_NOT_INSTALLED` to `WorkflowFailureReason`, error messages, failure reason resolution, and `ModalOrchestrator` error handling

## Test plan
- [ ] Open the drive-integration app on an aliased environment (e.g. `master`) — workflow starts normally
- [ ] Open the app on the underlying raw environment (e.g. `master-2026-06-09`) — workflow no longer returns 403
- [ ] Verify the agents API request URL uses the alias name in the path
- [ ] When the agent returns `app-not-installed`, the UI shows the "App not installed in this environment" error message
- [ ] When the agent returns `ai-service-unavailable`, the UI shows the "AI service temporarily unavailable" error message


apps/drive-integration/src/hooks/useWorkflowAgent.ts
```diff
@@ -253,6 +261,21 @@ const pollAgentRun = async (
...
 
+export async function resolveEnvironmentId(sdk: PageAppSDK): Promise<string> {
+  if (sdk.ids.environmentAlias) {
+    return sdk.ids.environmentAlias;
+  }
+  const spaceId = sdk.ids.space;
+  const rawEnvId = sdk.ids.environment;
+  try {
+    const env = await sdk.cma.environment.get({ spaceId, environmentId: rawEnvId });
+    const alias = env.sys.aliases?.[0]?.sys.id;
```
FBanfi: 🤔 Is it possible to have more than one alias per environment? Just curios due to this `env.sys.aliases?.[0]?.sys.id;`, or is it just because of the returned sdk shape?

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
```diff
@@ -187,6 +187,18 @@ export const ModalOrchestrator = forwardRef<ModalOrchestratorHandle, ModalOrches
...
+        error instanceof WorkflowRunError &&
+        error.reason === WorkflowFailureReason.APP_NOT_INSTALLED
+      ) {
+        setPreviewErrorState({
+          reason: WorkflowFailureReason.APP_NOT_INSTALLED,
+          title: 'App not installed in this environment',
+          message: ERROR_MESSAGES.APP_NOT_INSTALLED,
+        });
+        return;
+      }
```
FBanfi, line 200: I find this a bit confusing, how would the user run the workflow if the app was not installed in that environment?
Or this error refers to cases where: in the middle of a workflow run, the app was uninstalled from a different browser tab?

FBanfi (commented):
Looks good to me! Just left a small question regarding the purpose of the error!

apps/drive-integration/src/hooks/useWorkflowAgent.ts
JuliRossi: Mmm you are correct. it might be risky. The idea was to also solve the issue that affected environments accessed directly by their raw ID when an alias exists (this resulted in error in dev). But there seems to be something funky with the dev installation in particular. Went with a simpler fix that fixes the original issue (we are doing what we do in other app) and will test the second issue in prod to see how it works for that cases

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
JuliRossi: The idea is to have better error handling. If by any chance the request fail in the BE, we want to be as clear as possible.  Better a clear error message than a generic "Unable to generate preview" for the user. Also we want to know if the app is failing because of this

FBanfi: Makes sense! 💯

FBanfi (approved):
💯

PR #11022, opened 2026-06-11
feat: support localized GA4 path patterns []

zacharyyankiver:
## Summary
- add optional locale-specific path pattern overrides for GA4 advanced matching rules
- gate the override UI behind a “Use locale-specific patterns” toggle and widen the locale pattern inputs
- resolve sidebar report paths and wildcard match types from the selected locale override
- validate localized pattern tokens and preserve legacy configs with existing overrides

## Validation
- npm run build
- npm run test:ci
- ./node_modules/.bin/vitest run src/utils/getReportSlug.spec.ts src/hooks/useSidebarRules/useSidebarRules.spec.tsx src/components/config-screen/assign-content-type/AssignContentTypeRow.spec.tsx src/components/config-screen/assign-content-type/AssignContentTypeSection.spec.tsx src/helpers/contentTypeRules/contentTypeRules.spec.ts --minWorkers=1 --maxWorkers=1
- git diff --check -- apps/google-analytics-4/frontend/src

Note: the first local commit attempt hit a missing Husky helper in the fresh worktree (`.husky/_/husky.sh`), so the final commit was created with hooks disabled after the checks above passed.

### DIFF
--- a/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentType.styles.ts
+++ b/apps/google-analytics-4/frontend/src/components/config-screen/assign-content-type/AssignContentType.styles.ts
@@ -68,6 +68,33 @@ export const styles = {
     visibility: 'hidden',
     width: '100%',
   }),
+  localizedPatternInput: css({
+    flex: '1 1 420px',
+    minWidth: '280px',
+  }),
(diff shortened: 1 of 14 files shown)


mgoudy91 (approved):
LGTM

PR #11027, opened 2026-06-17
feat(drive-integration): add entry wizard for new entry design updates [INTEG-4136]

FBanfi:
## Purpose

Implements the \"Add entry\" wizard in the edit content modal as part of the new entry design updates. Users can now create a new entry directly from the mapping review screen, optionally wiring it as a reference to an existing entry.

## Approach

A 4-step (up to 5-step) semi-wizard is rendered inside the existing `EditModal` using the existing `Modal.Controls` buttons for navigation — no new internal buttons:

1. **Select content type** — pick the type for the new entry
2. **Is it a reference?** — yes/no radio
3. **Select the referenced entry (child)** — dropdown of existing entries
4. **Which field holds the reference?** — only shown when the content type has more than one `Link/Entry` or `Array/Entry` field; auto-skipped otherwise
5. **Select fields** — map document content to fields on the new entry (reuses `FieldSelectionDropdown`)

Key implementation details:
- `WizardStep` is an enum; `handleWizardNext/Back` use `switch` statements
- `referenceGraph` is lifted into `useState` in `ReviewPage` (same pattern as `entryBlockGraph`) so new reference edges can be appended client-side
- On save, `{ __ref: childTempId }` is written into the new entry's reference field — the backend merges `mappingEntry.fields` verbatim in step 10, and `entryService.ts` resolves `__ref` values in pass 2
- `isEntryReferenceField` and `buildFieldOptionsForContentType` extracted to `fieldFormatting.ts` to avoid duplication

## Testing steps

Automated tests added for the wizzard.

https://github.com/user-attachments/assets/ada75305-251a-464e-b40d-b7d88757cc43

## Breaking Changes

No.

Link to [INTEG-4136](https://contentful.atlassian.net/browse/INTEG-4136)

🤖 Co-authored with [Claude Code](https://claude.ai/code)

[INTEG-4136]: https://contentful.atlassian.net/browse/INTEG-4136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
     [entryBlockGraph.entries, payload.contentTypes]
   );
 
+  const buildNewLocationForContentType = (contentTypeId: string): EditModalNewLocation => {
```
JuliRossi: 🤔 See if we can move this to editmodal

apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/AddEntryWizard.tsx
```diff
@@ -0,0 +1,160 @@
...
+}: AddEntryWizardProps) => {
+  const newLocation = state.contentTypeId ? buildNewLocation(state.contentTypeId) : null;
+
+  return (
+    <Flex flexDirection="column" gap="spacingS">
+      <Text as="p" fontWeight="fontWeightDemiBold">
+        Add entry
+      </Text>
+
+      {state.step === WizardStep.ContentType && (
```
JuliRossi: ⛏️  You could use a switch for these steps

apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/AddEntryWizard.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+export enum WizardStep {
+  ContentType = 'content-type',
+  IsReference = 'is-reference',
+  SelectReference = 'select-reference',
+  SelectReferenceField = 'select-reference-field',
+  SelectFields = 'select-fields',
+}
+
+export interface WizardState {
```
JuliRossi: ⛏️  Rename this to wizard

apps/drive-integration/src/locations/Page/components/review/mapping/fieldFormatting.ts
```diff
@@ -59,6 +66,21 @@ export function isAssetFieldForImageAssign(field: WorkflowContentTypeField): boo
   }
 }
 
+export function buildFieldOptionsForContentType(
+  contentType: WorkflowContentType | undefined
+): EditModalFieldOption[] {
+  return (contentType?.fields ?? []).filter(isWorkflowContentTypeFieldWithId).map((field) => {
```
JuliRossi: ⛏️  Rename isWorkflowContentTypeFieldWithId or leave comment. Let use `hasFieldId`

apps/drive-integration/src/locations/Page/components/review/mapping/fieldFormatting.ts
```diff
@@ -59,6 +66,21 @@ export function isAssetFieldForImageAssign(field: WorkflowContentTypeField): boo
...
+    const fieldType = typeof field.type === 'string' ? field.type : 'Text';
```
JuliRossi: ⛏️  I think we should not need to check this

apps/drive-integration/src/locations/Page/components/review/mapping/fieldFormatting.ts
```diff
@@ -48,6 +48,13 @@ function isAssetLinkField(
   );
 }
 
+export function isEntryReferenceField(field: WorkflowContentTypeField): boolean {
```
JuliRossi, line 57: 🤔  I think this is already somewhere else

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
...
+          label: `${contentTypeName} (${entryTitle})`,
+        };
+      }),
+    [entryBlockGraph.entries, payload.contentTypes]
+  );
+
+  const handleAddEntry = (params: AddEntryWizardParams) => {
+    const { contentTypeId, fieldIds } = params;
+    const contentType = payload.contentTypes.find((ct) => ct.sys.id === contentTypeId);
+    const newEntryIndex = entryBlockGraph.entries.length;
```
JuliRossi, line 445: ⛏️ Could use crypto uuid

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
...
+    let next: EntryBlockGraph = {
+      ...entryBlockGraph,
+      entries: [...entryBlockGraph.entries, newEntry],
+    };
+
+    if (fieldIds.length > 0) {
+      const resolvedTargets = fieldIds.flatMap((fieldId) => {
+        const field = contentType?.fields?.find((f) => 'id' in f && f.id === fieldId);
+        const fieldType =
+          field && 'type' in field && typeof field.type === 'string' ? field.type : 'Text';
```
JuliRossi: 🤔  Why use Text as fallback

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
...
+    };
+
+    let next: EntryBlockGraph = {
+      ...entryBlockGraph,
+      entries: [...entryBlockGraph.entries, newEntry],
+    };
+
+    if (fieldIds.length > 0) {
+      const resolvedTargets = fieldIds.flatMap((fieldId) => {
+        const field = contentType?.fields?.find((f) => 'id' in f && f.id === fieldId);
```
JuliRossi: ⛏️  We could use hasFieldId

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
...
+        const fieldType =
+          field && 'type' in field && typeof field.type === 'string' ? field.type : 'Text';
+        return [{ entryIndex: newEntryIndex, fieldId, fieldType }];
+      });
+
+      const richTextTargets = resolvedTargets.filter((t) => t.fieldType === 'RichText');
+      const nonRichTextTargets = resolvedTargets.filter((t) => t.fieldType !== 'RichText');
```
JuliRossi, line 487: ⛏️ Could use reduce

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
...
+            next = applyRichTextAssignToEntryBlockGraph(
+              next,
+              document,
+              [imageRef],
+              richTextTargets
+            );
+          }
+        }
+      } else {
+        if (richTextTargets.length && pendingPreviewSourceRefs.length) {
```
JuliRossi, line 505: 🤔 What happens if one is empty but the other isn't?

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
...
+            document,
+            pendingPreviewSourceRefs,
+            richTextTargets
+          );
+        }
+        const allRangesForAssign = [
+          ...(pendingTextExclusionRanges ?? []),
+          ...pendingTextAssignRanges,
+        ];
+        if (nonRichTextTargets.length && allRangesForAssign.length) {
```
JuliRossi, line 517: 🤔 What happens if one is empty but the other isn't?

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
...
+  );
+
+  const handleAddEntry = (params: AddEntryWizardParams) => {
+    const { contentTypeId, fieldIds } = params;
+    const contentType = payload.contentTypes.find((ct) => ct.sys.id === contentTypeId);
+    const newEntryIndex = entryBlockGraph.entries.length;
+    const tempId = `new-entry-${contentTypeId}-${newEntryIndex}`;
+
+    const refField =
+      params.isReference && params.referenceEntryId
```
JuliRossi: ⛏️  Create condition

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
JuliRossi, line 1: 💡  I think this file is too large. We should probably refactor it a bit

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
```diff
@@ -426,6 +406,135 @@ export const MappingView = ({
...
+    }
+
+    onEntryBlockGraphChange(next);
+
+    if (params.isReference && params.referenceEntryId && onReferenceGraphChange) {
+      onReferenceGraphChange({
+        ...referenceGraph,
+        edges: [
+          ...(referenceGraph.edges ?? []),
+          { from: tempId, to: params.referenceEntryId, fieldId: '' },
```
JuliRossi: ⛏️  Check if '' is correct for fieldId

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
FBanfi: This is Intentional, if either side is empty (no source refs / no text ranges), there's nothing to assign so we skip. The entry still gets created, fields can be mapped manually afterward.

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
FBanfi: This is Intentional, if either side is empty (no source refs / no text ranges), there's nothing to assign so we skip. The entry still gets created, fields can be mapped manually afterward.

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
FBanfi: `fieldId` on an edge is never read, only from and to are consumed in `overviewEntryList.ts` and `ReviewPage.tsx`.  The '' is fine as a placeholder to satisfy the type, but refField?.id would be the real field id.
So used `refField?.id ?? ''`

apps/drive-integration/src/locations/Page/components/review/mapping/fieldFormatting.ts
FBanfi: We have a `isReference` and `isReferenceArray` in `entry.ts`, but those check if a value is a `{ __ref: string } `object (used during entry creation to resolve references). On the other hand,`isEntryReferenceField` checks if a field definition is of type `Link/Entry`.
I think the proper place for `isEntryReferenceField` still is `fieldFormatting.ts` because it already owns all field definition predicates.

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
FBanfi: Good point! A field without a type is malformed and shouldn't be processed. Instead of falling back to `Text`, we now skip it entirely. Added a `hasFieldType` guard in `fieldFormatting.ts` and reused it in the two spots in `MappingView.tsx` as well.

apps/drive-integration/src/locations/Page/components/review/mapping/fieldFormatting.ts
FBanfi: Good point! A field without a type is malformed and shouldn't be processed. Instead of falling back to `Text`, we now skip it entirely. Added a `hasFieldType` guard in `fieldFormatting.ts` and reused it in the two spots in `MappingView.tsx` as well.

apps/drive-integration/src/locations/Page/components/review/mapping/MappingView.tsx
FBanfi: Leaving this for another PR

JuliRossi (approved):

PR #11029, opened 2026-06-17
fix(drive-integration): flaky modal tests caused by react-modal timer leak [INTEG-4213]

FBanfi:
## Purpose

\`react-modal\` internally schedules a \`setTimeout\` to remove its portal from the DOM when a modal closes. In CI, this timer sometimes fires after Vitest tears down the jsdom environment, causing \`ReferenceError: document is not defined\`. The failure is intermittent because it's a race between the timer and environment teardown — faster/slower CI runs tip it either way.

## Approach

Added \`vi.useFakeTimers({ shouldAdvanceTime: true })\` / \`vi.runAllTimers()\` / \`vi.useRealTimers()\`  each test file that uses react modal.

- \`useFakeTimers\` gives Vitest control over pending timers
- \`shouldAdvanceTime: true\` keeps the real wall clock ticking so \`waitFor\` and async assertions still resolve normally
- \`runAllTimers()\` flushes react-modal's cleanup timeout synchronously before jsdom is torn down

## Breaking Changes

No.

[INTEG-4213](https://contentful.atlassian.net/browse/INTEG-4213)

## Deployment

No.

🤖 Co-authored with [Claude Code](https://claude.ai/code)

[INTEG-4213]: https://contentful.atlassian.net/browse/INTEG-4213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/drive-integration/src/setupTests.ts
+++ b/apps/drive-integration/src/setupTests.ts
@@ -4,6 +4,20 @@
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
 
+import { beforeEach, afterEach, vi } from 'vitest';
+
+// react-modal schedules a setTimeout to remove its portal on close. Without fake
+// timers that timeout can fire after jsdom is torn down → "document is not defined".
+// shouldAdvanceTime keeps the wall clock running so waitFor still resolves normally.
+beforeEach(() => {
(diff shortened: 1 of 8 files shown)


JuliRossi (approved):
thanks 🙏

PR #11030, opened 2026-06-17
fix(link-checker): reload page when installation params change on visibility restored [ES-322]

david-shibley-contentful:
[ES-322](https://contentful.atlassian.net/browse/ES-322)

## Summary
- `sdk.parameters.installation` is a static snapshot from when the app iframe loaded. If a user updates the allow list in config and returns to the page app without a hard refresh, scans run against stale params — causing valid links to show as "Not on allow list"
- Add a `visibilitychange` listener that fetches current installation parameters from the CMA when the tab becomes visible again and calls `window.location.reload()` if they differ — the in-app equivalent of the hard-refresh workaround given to the customer
- Errors from the CMA fetch are silently swallowed so a failed check never blocks the user

## Test plan
- [ ] Update the allow list in Link Checker config, then navigate back to the page app — it should automatically reload and show the updated results without requiring a manual hard refresh
- [ ] Verify that navigating away and back without changing config does NOT trigger a reload
- [ ] Run `npm test` in `apps/link-checker` — all 60 tests pass including 2 new cases covering the reload and no-reload paths

[ES-322]: https://contentful.atlassian.net/browse/ES-322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/link-checker/components/locations/Page.tsx
+++ b/apps/link-checker/components/locations/Page.tsx
@@ -174,13 +174,17 @@ export default function Page() {
       if (document.visibilityState !== 'visible') return;
       try {
         const appDefinitionId = sdk.ids.app;
-        if (!appDefinitionId || !sdk.cma?.appInstallation?.get) return;
-        const current = await sdk.cma.appInstallation.get({
-          spaceId: sdk.ids.space,
-          environmentId: sdk.ids.environment,
+        if (!appDefinitionId || !sdk.cma?.appInstallation?.getForOrganization) return;
+        const result = await sdk.cma.appInstallation.getForOrganization({
+          organizationId: sdk.ids.organization,
           appDefinitionId,
         });
-        if (JSON.stringify(current?.parameters) !== JSON.stringify(sdk.parameters.installation)) {
+        const match = result.items.find(
+          (item) =>
+            item.sys.space.sys.id === sdk.ids.space &&
+            item.sys.environment.sys.id === sdk.ids.environment
+        );
+        if (JSON.stringify(match?.parameters) !== JSON.stringify(sdk.parameters.installation)) {
           window.location.reload();
         }
       } catch {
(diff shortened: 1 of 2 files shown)


apps/link-checker/components/locations/Page.tsx
```diff
@@ -168,6 +168,33 @@ export default function Page() {
...
+        const result = await sdk.cma.appInstallation.getForOrganization({
+          organizationId: sdk.ids.organization,
+          appDefinitionId,
+        });
+        const match = result.items.find(
+          (item) =>
+            item.sys.space.sys.id === sdk.ids.space &&
+            item.sys.environment.sys.id === sdk.ids.environment
+        );
+        if (JSON.stringify(match?.parameters) !== JSON.stringify(sdk.parameters.installation)) {
```
mgoudy91: `JSON.stringify` comparison is key-order-sensitive — if the CMA response serializes object keys in a different order than the original SDK snapshot (common when two independent serialization paths are involved), this will produce a false mismatch and trigger a spurious reload on every tab restore. Consider a stable approach, e.g. sorting keys before stringifying or using a lightweight deep-equal utility:

```ts
const stableStringify = (obj: unknown) => JSON.stringify(obj, Object.keys(obj as object).sort());
if (stableStringify(match?.parameters) !== stableStringify(sdk.parameters.installation)) {
```

Or import a `deepEqual` helper if one is already available in the project.

mgoudy91 (changes requested):
Raised one issue around the `JSON.stringify` comparison being key-order-sensitive, which risks spurious reloads on every tab restore.

david-shibley-contentful (commented):
Good catch — `JSON.stringify` is key-order-sensitive. Replaced it with a recursive `deepEqual` helper that sorts keys before comparing. Also added a dedicated test that passes the same params in reverse key order and asserts no reload. While I was in there I also fixed the existing no-reload test — its CMA mock was returning `{ selectedContentTypeIds: ['article'] }` while `sdk.parameters.installation` was `{}`, so it would have triggered a reload under any correct implementation.

primeinteger (approved):
This looks solid. I like that you added the deepEqual function directly instead of pulling in lodash or something similar. 👏

david-shibley-contentful:
Thanks! Yeah I actually talked through that particular point with the AI. Decided it was better to do it this way

mgoudy91 (approved):

PR #11034, opened 2026-06-24
fix(deep-clone): surface entry creation failures instead of silently producing broken clones []

david-shibley-contentful:
## Summary
- Failed entry creations during a deep clone were silently swallowed with `console.warn`, leaving users with a partial clone invisibly coupled to the original
- `EntryCloner` now tracks failed entry IDs and throws after `createClones()` if any failed, aborting before `updateReferenceTree()` so no broken structure is produced
- `Sidebar` catches the error, shows a `<Note variant="negative">` with the failure message, and fires `sdk.notifier.error()` — no more false "Clone successful" toast on failure

## Test plan
- [ ] Trigger a clone where an entry creation fails (e.g. mock a CMA error or use an entry with a restricted content type) — verify error `Note` appears in the sidebar and no success toast fires
- [ ] Happy path: full clone with all entries selected still completes and redirects as expected
- [ ] Partial selection (some entries deselected): selected entries clone successfully, deselected remain as shared links, success toast fires
- [ ] Retry after failure: clicking "Clone entry" again resets state including error message

### DIFF
--- a/apps/deep-clone/src/locations/Sidebar.tsx
+++ b/apps/deep-clone/src/locations/Sidebar.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react';
-import { Text, Button, Stack } from '@contentful/f36-components';
+import { Text, Button, Stack, Note } from '@contentful/f36-components';
 import { useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
 import { SidebarAppSDK } from '@contentful/app-sdk';
 import EntryCloner from '../utils/EntryCloner';
@@ -21,6 +21,7 @@ function Sidebar() {
   const [isCloning, setIsCloning] = useState<boolean>(false);
   const [isFinished, setIsFinished] = useState<boolean>(false);
   const [isRedirecting, setIsRedirecting] = useState<boolean>(false);
+  const [cloneError, setCloneError] = useState<string | null>(null);
 
   useEffect(() => {
     if (countdown === 0) return;
@@ -45,6 +46,7 @@ function Sidebar() {
     setReferencesCount(0);
     setClonesCount(0);
     setUpdatesCount(0);
+    setCloneError(null);
   };
 
   const clone = async (): Promise<void> => {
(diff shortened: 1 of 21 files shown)


apps/deep-clone/src/utils/EntryCloner.ts
```diff
@@ -148,6 +158,7 @@ class EntryCloner {
               latestClone = await this.cma.entry.get({ entryId: clone.sys.id });
             } else {
```
mgoudy91, line 163: `failedUpdateIds` is populated here but never checked anywhere — it's effectively dead code. If reference-tree updates fail, `cloneEntry()` still returns the entry normally and the Sidebar shows "Clone successful" despite a broken reference tree. This is the same silent-failure class as the bug being fixed for creation failures. Worth either throwing (like `failedCloneIds` does) or at minimum surfacing the partial-update case to the user.

apps/deep-clone/src/locations/Sidebar.tsx
```diff
@@ -78,7 +80,15 @@ function Sidebar() {
     setIsConfirming(false);
     setIsCloning(true);
 
-    const clonedEntry = await cloner.cloneEntry(selectedEntryIds);
+    let clonedEntry;
```
mgoudy91, line 85: `catch (error: any)` bypasses TypeScript's type checking. Prefer:
```ts
setCloneError(error instanceof Error ? error.message : 'An unexpected error occurred during cloning.');
```

mgoudy91 (changes requested):
Three things worth addressing:

1. `failedUpdateIds` is populated on reference-tree update failure but never checked — see inline comment.
2. `catch (error: any)` bypasses TypeScript's type checking — see inline comment.
3. The new error path has no automated test coverage. A unit test that mocks a CMA creation failure, asserts the error `<Note>` renders, and confirms the success toast does not fire would lock this behavior in against regressions.

mgoudy91 (approved):
All three issues addressed. LGTM.

PR #11043, opened 2026-06-30
feat(drive-integration): fast-fail with actionable errors for known failure modes [INTEG-4331]

david-shibley-contentful:
[INTEG-4331](https://contentful.atlassian.net/browse/INTEG-4331)

## Summary
- Add `DOCUMENT_TOO_COMPLEX` and `PROCESSING_TIMEOUT` failure reasons with actionable user-facing error messages — imports no longer hang silently or fall back to the generic "Please start again" message for these cases
- Enforce a hard cap of 20 tabs in `SelectTabsModal` with an inline validation message; polling timeout now throws a typed `WorkflowRunError(PROCESSING_TIMEOUT)` instead of a plain `Error`
- Add pre-dispatch auth check in `ModalOrchestrator` — if `isOAuthConnected` is false when the user submits content types, the reconnect modal surfaces immediately before any workflow is dispatched

## Test plan
- [ ] Start an import with an expired/disconnected Google account — verify reconnect modal appears immediately at content type submission, not after a failed workflow run
- [ ] In `SelectTabsModal`, select more than 20 tabs — verify "Next" is blocked and the validation message appears with the limit and guidance
- [ ] Select exactly 20 tabs — verify "Next" is enabled and flow proceeds normally
- [ ] Simulate a polling timeout (e.g. by lowering `MAX_POLL_ATTEMPTS` locally) — verify the error modal shows "Import timed out" with the `PROCESSING_TIMEOUT` message, not the generic error
- [ ] Verify existing failure paths (auth expired mid-workflow, doc not found, AI unavailable) still surface their correct error modals unchanged

[INTEG-4331]: https://contentful.atlassian.net/browse/INTEG-4331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/drive-integration/src/hooks/useWorkflowAgent.ts
+++ b/apps/drive-integration/src/hooks/useWorkflowAgent.ts
@@ -135,6 +135,10 @@ const getBackendWorkflowFailureReason = (runData: AgentRunData): WorkflowFailure
     return WorkflowFailureReason.APP_NOT_INSTALLED;
   }
 
+  if (workflowFailure.code === WorkflowFailureReason.DOCUMENT_TOO_COMPLEX) {
+    return WorkflowFailureReason.DOCUMENT_TOO_COMPLEX;
+  }
+
   if (workflowFailure.code === WorkflowFailureReason.GENERIC) {
     return WorkflowFailureReason.GENERIC;
   }
@@ -160,6 +164,14 @@ const getWorkflowFailureMessage = (
 
   if (failureReason === WorkflowFailureReason.APP_NOT_INSTALLED) {
     return ERROR_MESSAGES.APP_NOT_INSTALLED;
+  }
+
+  if (failureReason === WorkflowFailureReason.DOCUMENT_TOO_COMPLEX) {
+    return ERROR_MESSAGES.DOCUMENT_TOO_COMPLEX;
+  }
+
+  if (failureReason === WorkflowFailureReason.PROCESSING_TIMEOUT) {
+    return ERROR_MESSAGES.PROCESSING_TIMEOUT;
   }
 
   return getRunErrorMessage(runData);
@@ -258,7 +270,10 @@ const pollAgentRun = async (
   }
 
   console.error(`✗ Run [${runId}] timed out after ${elapsedSec(startMs)}`);
-  throw new Error('Workflow polling timeout');
+  throw new WorkflowRunError(
+    ERROR_MESSAGES.PROCESSING_TIMEOUT,
+    WorkflowFailureReason.PROCESSING_TIMEOUT
+  );
 };
 
 export const useWorkflowAgent = ({
--- a/apps/drive-integration/src/hooks/useWorkflowAgent.ts
+++ b/apps/drive-integration/src/hooks/useWorkflowAgent.ts
@@ -135,6 +135,7 @@ const getBackendWorkflowFailureReason = (runData: AgentRunData): WorkflowFailure
     return WorkflowFailureReason.APP_NOT_INSTALLED;
   }
 
+  // document-too-complex is not yet emitted by the backend; handler is in place for when it ships.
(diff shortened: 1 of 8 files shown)


apps/drive-integration/src/hooks/useWorkflowAgent.ts
```diff
@@ -135,6 +135,10 @@ const getBackendWorkflowFailureReason = (runData: AgentRunData): WorkflowFailure
     return WorkflowFailureReason.APP_NOT_INSTALLED;
   }
 
+  if (workflowFailure.code === WorkflowFailureReason.DOCUMENT_TOO_COMPLEX) {
+    return WorkflowFailureReason.DOCUMENT_TOO_COMPLEX;
+  }
```
joaquincasal, line 141: Does this failure reason exist in the backend?

apps/drive-integration/src/utils/constants/agent.ts
```diff
@@ -11,3 +11,7 @@ export const CONTENT_TYPE_SUBMIT_LOADING_DELAY_MS = 30000; // 30 seconds to wait
 // Agents-api writes PENDING_REVIEW status before the suspendPayload metadata flushes.
 // Allow this many consecutive PENDING_REVIEW polls with no suspendPayload before giving up.
 export const MAX_PENDING_REVIEW_MISSING_PAYLOAD_RETRIES = 5; // 5 × 10s = 50s max wait
+
+// Hard cap on tab selection to prevent oversized payloads and single-LLM-call timeouts.
+// Revisit once parallel per-entry mappers (M7) ship.
+export const MAX_TABS_SELECTION = 20;
```
joaquincasal, line 17: I think it's possible to have a doc with >20 tabs but with the doc not being too complex/big. Is there other way we can check for complexity? Agreed to revisit later, since ideally we'll find a way to handle big docs as well

apps/drive-integration/src/utils/constants/messages.ts
```diff
@@ -14,6 +14,10 @@ export const ERROR_MESSAGES = {
     'The AI service is temporarily unavailable. Please try again in a few minutes.',
   APP_NOT_INSTALLED:
     'The Drive Integration app is not installed in this environment. Install it via Apps > Manage apps and try again.',
+  DOCUMENT_TOO_COMPLEX:
+    'This document is too large to import in one go. Try splitting it into smaller sections or reducing the number of tabs selected.',
+  PROCESSING_TIMEOUT:
+    'The import took too long to complete. Try a simpler document, reduce the number of tabs selected, or split the document into smaller sections.',
```
joaquincasal: nit: the wording may be confusing for a user that didn't get the tab selection step (when the selected doc does not have tabs)

joaquincasal (approved):

apps/drive-integration/src/hooks/useWorkflowAgent.ts
david-shibley-contentful: Not yet. I added a comment to make that explicit. It's wired up ahead of the backend shipping so the UI is ready when it does.

apps/drive-integration/src/utils/constants/agent.ts
david-shibley-contentful: Agreed. The 20-tab cap is a proxy for payload/timeout risk, not a true complexity measure. The comment already flags M7 as the right time to revisit. Happy to track a follow-up ticket if useful.

apps/drive-integration/src/utils/constants/messages.ts
david-shibley-contentful: Good catch. I stripped the tab-selection guidance from that message so it reads cleanly for tabless docs too. Updated in the latest commit.

joaquincasal (approved):

PR #11045, opened 2026-06-30
feat: add OUT_OF_DOMAIN typed failure reason [INTEG-4349]

david-shibley-contentful:
[INTEG-4349](https://contentful.atlassian.net/browse/INTEG-4349)

> **Stacked on:** #11043 (INTEG-4331 fast-fail errors). Merge that first, then this.

## Summary
- Adds `OUT_OF_DOMAIN = 'out-of-domain'` to `WorkflowFailureReason` enum in `src/types/workflow.ts`
- Wires the new reason through `getBackendWorkflowFailureReason()` and `getWorkflowFailureMessage()` in `useWorkflowAgent.ts`
- Adds `showWorkflowError` branch in `ModalOrchestrator.tsx` with title "Document not supported" and a user-facing message directing users to try a different document or content type
- Adds `OUT_OF_DOMAIN` message constant to `messages.ts`

## Test plan
- [ ] agents-api returns `workflowFailure.code = 'out-of-domain'` → UI shows "Document not supported" error modal with the correct message
- [ ] All other failure reasons (`GOOGLE_DRIVE_AUTH_EXPIRED`, `GOOGLE_DOCS_NOT_FOUND`, etc.) still show their own modals unchanged
- [ ] Generic fallback still triggers for unknown failure codes

[INTEG-4349]: https://contentful.atlassian.net/browse/INTEG-4349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/drive-integration/src/hooks/useWorkflowAgent.ts
```diff
@@ -139,6 +139,10 @@ const getBackendWorkflowFailureReason = (runData: AgentRunData): WorkflowFailure
     return WorkflowFailureReason.DOCUMENT_TOO_COMPLEX;
   }
 
+  if (workflowFailure.code === WorkflowFailureReason.OUT_OF_DOMAIN) {
+    return WorkflowFailureReason.OUT_OF_DOMAIN;
+  }
+
```
joaquincasal: Same comment as the other PR, does this exist in the backend or do we have plans to add it? Not sure in which cases it would be returned

apps/drive-integration/src/hooks/useWorkflowAgent.ts
```diff
@@ -174,6 +178,10 @@ const getWorkflowFailureMessage = (
     return ERROR_MESSAGES.PROCESSING_TIMEOUT;
   }
 
+  if (failureReason === WorkflowFailureReason.OUT_OF_DOMAIN) {
+    return ERROR_MESSAGES.OUT_OF_DOMAIN;
+  }
+
```
joaquincasal: We could consider using a lookup table here to clean up these if statements a bit

apps/drive-integration/src/hooks/useWorkflowAgent.ts
david-shibley-contentful: Not yet; same situation as DOCUMENT_TOO_COMPLEX on the parent PR. It's wired up now so the UI is ready when the backend ships it. Added a comment to both.

apps/drive-integration/src/hooks/useWorkflowAgent.ts
david-shibley-contentful: Good call! I went ahead and did it. Both functions are now lookup-table based: getBackendWorkflowFailureReason uses a Set of known enum values, and getWorkflowFailureMessage uses a Partial Record with a ?? fallback. 64 lines → 17. Updated in the latest commit.

joaquincasal (dismissed):

joaquincasal (approved):

PR #11048, opened 2026-07-01
fix(rich-text-versioning): pin react-query v4 to fix editor load crash []

david-shibley-contentful:
## Summary
- Pin `@tanstack/react-query@4.44.0` as a direct dependency so Vite bundles v4 instead of hoisting v5 from `@contentful/field-editor-shared`
- Fixes `Cannot read properties of undefined (reading 'sys')` when loading the rich text editor on fields with embedded entry/asset references
- Harden the version comparison dialog against missing link targets and fix stale state when fetching content types for embedded entries

## Root cause
With react-query v5 hoisted, `FetchingWrappedInlineEntryCard` in `@contentful/field-editor-rich-text@6.3.9` does not treat `'pending'` as a loading state. The success branch renders before entity data is available and accesses `entry.sys`.

## Test plan
- [x] `npm run build` passes in `apps/rich-text-versioning`
- [x] `npm test` passes (32 tests)
- [x] Manual: open an entry with embedded inline/block entry references — editor loads without console errors
- [ ] Manual: Compare versions dialog renders embedded references correctly

### DIFF
--- a/apps/rich-text-versioning/src/locations/Dialog.tsx
+++ b/apps/rich-text-versioning/src/locations/Dialog.tsx
@@ -13,7 +13,7 @@ import {
 } from '@contentful/f36-components';
 import { useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
 import { documentToReactComponents } from '@contentful/rich-text-react-renderer';
-import { BLOCKS, Document, INLINES } from '@contentful/rich-text-types';
+import { BLOCKS, Document, INLINES, TopLevelBlock } from '@contentful/rich-text-types';
 import { useState, useEffect } from 'react';
 import { styles } from './Dialog.styles';
 import HtmlDiffViewer from '../components/HtmlDiffViewer';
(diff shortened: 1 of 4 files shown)


apps/rich-text-versioning/src/locations/Dialog.tsx
```diff
@@ -116,13 +119,23 @@ const Dialog = () => {
...
+              allEntries
+                .map((entry) => entry.sys.contentType?.sys?.id)
+                .filter((contentTypeId): contentTypeId is string => Boolean(contentTypeId))
+            ),
+          ];
+
+          if (contentTypeIds.length > 0) {
+            const fetchedContentTypes = await sdk.cma.contentType.getMany({
+              query: {
+                'sys.id[in]': contentTypeIds.join(','),
```
joaquincasal, line 134: Just in case: this is changing from an array to a comma-separated string, but I'd expect either work

joaquincasal (approved):

apps/rich-text-versioning/src/locations/Dialog.tsx
david-shibley-contentful: Good call. Either should work with the CMA client. I used .join(',') here to match getEntries / getAssets in this file, not because the array form was failing.

The meaningful change in this block is using the freshly fetched entries (instead of stale state), deduping IDs, and guarding missing contentType refs. Happy to switch back to passing the array if you prefer consistency with the old line rather than with the other queries in Dialog.tsx.

joaquincasal: No need! As long as the CMA accepts both, it's good

ryunsong-contentful (approved):
consider if the `.join` is needed for efficiency purposes

PR #11051, opened 2026-07-03
fix(drive-integration): disable confirm cancel button while cancellation is in-flight []

JuliRossi:
## Purpose
When a user clicked "Cancel without creating" in the confirm cancel modal, the button remained active during the async cancellation, allowing multiple clicks. The `isConfirming` state was also dead code in `ModalOrchestrator` because `activeRunId` is never set at the pre-review stage.

Found this while working on [INTEG-4233](https://contentful.atlassian.net/browse/INTEG-4233)

## Summary
- Added `isConfirming` prop to `ConfirmCancelModal` — disables both buttons and shows a loading spinner on the confirm button while in-flight
- Removed dead `async`/`resumeWorkflow` block from `ModalOrchestrator.handleConfirmCancel` (`activeRunId` is always `null` at that point; it's only set after the workflow reaches `PENDING_REVIEW` and the user is already in `ReviewPage`)
- Wired `isConfirming={isCancelling}` in `ReviewPage`, where the async work actually happens via `onCancelReview`
- Added two tests covering the disabled state and click-blocking behaviour when `isConfirming` is true

## Test plan
 Sync Cancel

https://github.com/user-attachments/assets/cbfc9d3a-0897-4c3e-bdef-84beaf6438db

Async Cancel

https://github.com/user-attachments/assets/04d36b74-b749-497a-b601-1ca56ae23065

[INTEG-4233]: https://contentful.atlassian.net/browse/INTEG-4233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
+++ b/apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
@@ -134,17 +134,8 @@ export const ModalOrchestrator = forwardRef<ModalOrchestratorHandle, ModalOrches
       onResetToMain();
     }, [onResetToMain]);
 
-    const handleConfirmCancel = async () => {
+    const handleConfirmCancel = () => {
       setIsConfirmCancelModalOpen(false);
-
-      if (activeRunId) {
-        try {
-          await resumeWorkflow(activeRunId, { cancelled: true });
-        } catch (error) {
-          console.error(error);
-        }
-      }
-
       resetProgress();
       onResetToMain();
     };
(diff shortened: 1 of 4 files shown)


apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
```diff
@@ -134,17 +134,8 @@ export const ModalOrchestrator = forwardRef<ModalOrchestratorHandle, ModalOrches
       onResetToMain();
     }, [onResetToMain]);
 
-    const handleConfirmCancel = async () => {
+    const handleConfirmCancel = () => {
```
JuliRossi, line 137: We'll only want to cancel workflow if the cancelation comes from review mapping page (logic in page)

FBanfi (approved):
Nice catch and refactor! Looks clean to me!

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
FBanfi: Thanks for the context!

PR #11063, opened 2026-07-09
feat: add bulk-exporter app []

erinkrsandoval:
## Summary
- Adds the Bulk Exporter app to the monorepo under `apps/bulk-exporter`
- Allows users to search, filter, and export Contentful entries across one content type or the whole space
- Supports CSV, JSON, XLSX, XML, and YAML export formats
- Built with Forma 36 v6 components, F36 Datepicker, Multiselect, Pagination, and Modal

## Test plan
- [ ] Run `npm install` in `apps/bulk-exporter`
- [ ] Run `npm run dev` and verify app loads at localhost
- [ ] Install app in a Contentful space and verify the page location loads
- [ ] Generate a list with various filter combinations (content type, status, tags, date ranges)
- [ ] Select entries and export — verify file downloads in selected format
- [ ] Verify locale selector is disabled when "All content types" is selected

### DIFF
--- a/apps/bulk-exporter/.gitignore
+++ b/apps/bulk-exporter/.gitignore
@@ -1,0 +1,44 @@
+# Dependencies
+node_modules/
+.pnp
+.pnp.js
+
+# Testing
+coverage/
+*.lcov
+.nyc_output
+
+# Production
+dist/
+build/
+
+# Misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+# Logs
+npm-debug.log*
+yarn-debug.log*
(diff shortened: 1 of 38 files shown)


ryunsong-contentful (changes requested):
First change the local port to be 3000 so that's it's universally in line to test with other apps locally

ryunsong-contentful:
Outside the comment, the app looks good. Also in the future link the space that the app should be installed in so that I know it was at least tested by the developer and is an easier access point for the person reviewing and testing

ryunsong-contentful (approved):

mgoudy91 (approved):

PR #11065, opened 2026-07-10
fix(ga4): GA4 sidebar shows generic error for all customers [ES-433]

tylerwashington888:
## Problem

Since June 30, every customer using the Google Analytics 4 sidebar sees **\"We couldn't load Google Analytics data\"** even though the lambda is returning valid data.

### What's happening end-to-end

1. The GA4 sidebar calls the lambda's `/api/run_report` endpoint via a signed request
2. The lambda fetches data from Google's Analytics Data API (gRPC/Protobuf) and returns a **200 with valid rows**
3. The frontend runs the response through a Zod schema (`ZRunReportData`) before rendering
4. The schema fails → throws `MalformedApiResponse` error type → `ErrorDisplay` has no handler for it → falls to `default` → shows the generic message

### Why the schema fails

PR #11042 introduced `ZRunReportData` with this metadata shape:

```typescript
metadata: z.object({
  currencyCode: z.string(),      // required
  dataLossFromOtherRow: z.boolean(),
  timeZone: z.string(),          // required
  _currencyCode: z.string(),     // required — THIS IS THE PROBLEM
  _timeZone: z.string(),         // required — THIS IS THE PROBLEM
}),
propertyQuota: z.null(),
```

`_currencyCode` and `_timeZone` are **internal protobuf/protobufjs implementation details** — they exist on the in-memory JS object that protobufjs produces when decoding the binary gRPC response, but they are `undefined` and get silently dropped when the lambda calls `res.json()`. They are never part of the actual JSON payload the frontend receives. The schema was written by inspecting the raw JS object rather than the actual API contract.

The actual JSON the lambda sends looks like:
```json
"metadata": {
  "dataLossFromOtherRow": false,
  "currencyCode": "USD",
  "timeZone": "America/New_York"
}
```

`_currencyCode` and `_timeZone` are never there. `z.string()` on a missing field fails for every single customer.

Additionally, `propertyQuota: z.null()` would fail for any high-traffic property where Google returns a `QuotaStatus` object instead of null.

### Why only reported for some customers

The Zod failure affects all customers — the "3 customers" in the report are just the ones who surfaced it. The bug has been live since June 30.

## Fix

Make the fields that are never actually present optional, and widen `propertyQuota` to accept any value:

```typescript
metadata: z.object({
  currencyCode: z.string().optional(),     // absent for non-ecommerce properties
  dataLossFromOtherRow: z.boolean(),
  timeZone: z.string().optional(),
  _currencyCode: z.string().optional(),    // protobuf internal — never in JSON
  _timeZone: z.string().optional(),        // protobuf internal — never in JSON
}),
propertyQuota: z.unknown(),               // null normally, object when quota is tracked
```

The frontend only reads `rows` and `rowCount` from the response — none of the loosened fields are accessed anywhere in UI code.

## Test plan
- [ ] Open the GA4 sidebar on any entry — analytics data should render instead of the generic error
- [ ] Confirm properties with and without currency tracking both work
- [ ] `npm run test:ci` in `apps/google-analytics-4/frontend` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Tyler

### DIFF
--- a/apps/google-analytics-4/frontend/src/apis/apiTypes.ts
+++ b/apps/google-analytics-4/frontend/src/apis/apiTypes.ts
@@ -52,13 +52,16 @@ export const ZRunReportData = z.object({
   minimums: z.array(ZRow),
   rowCount: z.number(),
   metadata: z.object({
-    currencyCode: z.string(),
+    // currencyCode is an optional proto3 field — absent from JSON when not set (non-ecommerce properties)
+    currencyCode: z.string().optional(),
     dataLossFromOtherRow: z.boolean(),
-    timeZone: z.string(),
-    _currencyCode: z.string(),
-    _timeZone: z.string(),
+    timeZone: z.string().optional(),
+    // _currencyCode/_timeZone are protobuf oneof discriminators — absent from JSON when field is not set
+    _currencyCode: z.string().optional(),
+    _timeZone: z.string().optional(),
   }),
-  propertyQuota: z.null(),
+  // null when quota tracking is disabled; object when property has quota data
+  propertyQuota: z.unknown(),
   kind: z.string(),
 });
 
--- a/apps/google-analytics-4/frontend/src/apis/apiTypes.ts
+++ b/apps/google-analytics-4/frontend/src/apis/apiTypes.ts
@@ -52,15 +52,12 @@ export const ZRunReportData = z.object({
   minimums: z.array(ZRow),
   rowCount: z.number(),
   metadata: z.object({
-    // currencyCode is an optional proto3 field — absent from JSON when not set (non-ecommerce properties)
     currencyCode: z.string().optional(),
     dataLossFromOtherRow: z.boolean(),
     timeZone: z.string().optional(),
-    // _currencyCode/_timeZone are protobuf oneof discriminators — absent from JSON when field is not set
     _currencyCode: z.string().optional(),
     _timeZone: z.string().optional(),
   }),
-  // null when quota tracking is disabled; object when property has quota data
   propertyQuota: z.unknown(),
   kind: z.string(),
 });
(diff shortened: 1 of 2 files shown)


apps/google-analytics-4/frontend/src/apis/apiTypes.ts
```diff
@@ -52,13 +52,16 @@ export const ZRunReportData = z.object({
   minimums: z.array(ZRow),
   rowCount: z.number(),
   metadata: z.object({
-    currencyCode: z.string(),
+    // currencyCode is an optional proto3 field — absent from JSON when not set (non-ecommerce properties)
```
ryunsong-contentful: remove comment

apps/google-analytics-4/frontend/src/apis/apiTypes.ts
```diff
@@ -52,13 +52,16 @@ export const ZRunReportData = z.object({
...
+    currencyCode: z.string().optional(),
     dataLossFromOtherRow: z.boolean(),
-    timeZone: z.string(),
-    _currencyCode: z.string(),
-    _timeZone: z.string(),
+    timeZone: z.string().optional(),
+    // _currencyCode/_timeZone are protobuf oneof discriminators — absent from JSON when field is not set
```
ryunsong-contentful: remove comment

apps/google-analytics-4/frontend/src/apis/apiTypes.ts
```diff
@@ -52,13 +52,16 @@ export const ZRunReportData = z.object({
...
+    _currencyCode: z.string().optional(),
+    _timeZone: z.string().optional(),
   }),
-  propertyQuota: z.null(),
+  // null when quota tracking is disabled; object when property has quota data
```
ryunsong-contentful: remove comment

ryunsong-contentful (approved):
Please make code changes in the comments before merging

tylerwashington888 (commented):
Removed all three inline comments per your feedback. No other changes.

ryunsong-contentful (approved):
Ship it

PR #11087, opened 2026-07-15
feat(drive-integration): display validationFindings in mapping-review UI [INTEG-4383]

david-shibley-contentful:
[INTEG-4383](https://contentful.atlassian.net/browse/INTEG-4383)

## Summary
- Adds `ValidationFinding` interface and `validationFindings?: ValidationFinding[]` to `MappingReviewSuspendPayload` so the UI can consume findings from the validate-payload step
- Renders per-entry badges in the overview list: `negative` ("Needs attention") for `block` findings, `warning` ("Warning") for `warn`-only entries
- Gates the "Create selected entries" button behind an explicit acknowledgement checkbox when block findings are present; warn findings annotate without blocking

Screenshot 2026-07-15 at 12 13 29 PM

## Test plan
- [ ] With `google-docs-agent-improvements` flag on: import a document, verify entries with block findings show "Needs attention" badge, warn-only entries show "Warning" badge, and Create button is disabled until the acknowledgement checkbox is checked
- [ ] With flag off (no `validationFindings` in payload): no badges rendered, Create button enabled as before
- [ ] Deselect all entries → Create button still disabled (existing behaviour preserved)
- [ ] `npm run test:ci` passes (166 tests, 9 new)

[INTEG-4383]: https://contentful.atlassian.net/browse/INTEG-4383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
+++ b/apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
@@ -2,6 +2,7 @@ import { Badge, Box, Card, Checkbox, Flex, Paragraph, Text } from '@contentful/f
 import tokens from '@contentful/f36-tokens';
 import { cx } from '@emotion/css';
 import type { EntryListRow as OverviewEntryListRow } from '../../../../utils/overviewEntryList';
+import { ValidationFindingSeverity } from '@types';
 import type { ValidationFinding } from '@types';
 import {
   noMappedContentBadge,
@@ -49,8 +50,8 @@ function OverviewEntryRowCard({
   const isSelected = row.entryIndex === selectedEntryIndex;
   const isEntrySelectedForCreation = selectedEntryKeys.has(row.id);
   const entryFindings = findingsByEntryIndex?.get(row.entryIndex) ?? [];
-  const hasBlockFindings = entryFindings.some((f) => f.severity === 'block');
-  const hasWarnFindings = entryFindings.some((f) => f.severity === 'warn');
+  const hasBlockFindings = entryFindings.some((f) => f.severity === ValidationFindingSeverity.Block);
+  const hasWarnFindings = entryFindings.some((f) => f.severity === ValidationFindingSeverity.Warn);
 
   const treeLineClass =
     showTreeLines && cx(treeChildRowBase, isLastRow ? treeChildRowLast : treeChildRowNotLast);
--- a/apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
+++ b/apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
@@ -2,6 +2,7 @@ import { Badge, Box, Card, Checkbox, Flex, Paragraph, Text } from '@contentful/f
 import tokens from '@contentful/f36-tokens';
 import { cx } from '@emotion/css';
 import type { EntryListRow as OverviewEntryListRow } from '../../../../utils/overviewEntryList';
+import type { ValidationFinding } from '@types';
 import {
   noMappedContentBadge,
   treeChildRowBase,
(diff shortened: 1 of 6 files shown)


ryunsong-contentful (approved):

apps/drive-integration/src/types/workflow.ts
```diff
@@ -134,6 +151,8 @@ export interface MappingReviewSuspendPayload {
   entryBlockGraph: EntryBlockGraph;
   referenceGraph: ReviewedReferenceGraph;
   contentTypes: WorkflowContentType[];
+  /** Present when the google-docs-agent-improvements flag is on; absent otherwise. */
+  validationFindings?: ValidationFinding[];
```
JuliRossi, line 156: ⛏️Nit: I would remove the comment

apps/drive-integration/src/types/workflow.ts
```diff
@@ -125,6 +125,23 @@ export interface TabsImagesSuspendPayload {
   tabs?: DocTabOption[];
 }
 
+/** Severity levels for validation findings produced by the validate-payload step. */
+export type ValidationFindingSeverity = 'block' | 'warn';
```
JuliRossi: ⛏️ nit: this could be an enum, as we are using it in several places

JuliRossi (approved):
LGTM

apps/drive-integration/src/locations/Page/components/overview/OverviewSection.tsx
```diff
@@ -42,6 +48,19 @@ const OverviewSection = ({
...
+    for (const finding of payload.validationFindings ?? []) {
+      if (finding.entryIndex === undefined) continue;
+      const list = map.get(finding.entryIndex) ?? [];
+      list.push(finding);
+      map.set(finding.entryIndex, list);
+    }
+    return map;
+  }, [payload.validationFindings]);
+
+  const hasBlockFindings = (payload.validationFindings ?? []).some((f) => f.severity === 'block');
```
FBanfi: We could reuse the one that is computed in the `ReviewPage.tsx` to avoid having this two times.

apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
```diff
@@ -90,14 +95,24 @@ function OverviewEntryRowCard({
...
+                <Badge variant="negative" size="small" className={noMappedContentBadge}>
+                  Needs attention
+                </Badge>
+              )}
+              {!hasBlockFindings && hasWarnFindings && (
+                <Badge variant="warning" size="small" className={noMappedContentBadge}>
+                  Warning
+                </Badge>
+              )}
             </Paragraph>
```
FBanfi, line 119: `"No mapped content"` and `"Needs attention"/"Warning"` are rendered as sibling `<Badge>` elements in a `<Paragraph>`. If an entry simultaneously has no mapped content type AND a block finding, both badges appear.

Perhaps this is an expected behavior but since the PR didn't mention about this, just wanted to double check it with you!

FBanfi (approved):
Lgtm! Just left two minor comments!

apps/drive-integration/src/locations/Page/components/overview/OverviewEntryList.tsx
david-shibley-contentful: Good catch — this is intentional. An entry with no mapped content fields that also has a block finding (e.g. a required field is missing) would legitimately show both. The badges serve different purposes: "No mapped content" means none of the field mappings have any source refs, while "Needs attention" means the validate-payload step flagged a structural issue. They're not mutually exclusive so showing both makes sense here.

PR #11092, opened 2026-07-16
feat(drive-integration): async import runs rearchitecture [INTEG-4526]

david-shibley-contentful:
[INTEG-4526](https://contentful.atlassian.net/browse/INTEG-4526)

## Summary
- Rearchitects the import flow from synchronous (20-min blocking spinner) to fully async: wizard fires off the agent run, saves to localStorage, redirects immediately to a new Runs page
- New **Runs page** becomes the app home screen — lists all imports with live status badges (Running / Needs Review / Completed / Failed / Expired), polls every 10s while any run is in-flight, supports multiple concurrent runs
- **ReviewPage** is now reachable directly from the Runs page via a Review CTA (no longer locked inside the wizard flow); entry creation stays synchronous

image

## Key files
- `src/types/runs.ts` — `RunRecord`, `DisplayStatus`, `RunWithStatus`, `AppView` types
- `src/hooks/useRunStorage.ts` — localStorage-backed run history (max 50, scoped by space+env)
- `src/hooks/useRunsPolling.ts` — parallel `Promise.all` polling, 10s interval while running
- `src/services/workflowService.ts` — `resumeAndPollWorkflow` extracted from the hook
- `src/locations/Page/components/runs/` — `RunsPage` + `RunRow` components
- `src/locations/Page/Page.tsx` — `AppView` discriminated union state machine replaces binary toggle
- `src/locations/Page/components/mainpage/ModalOrchestrator.tsx` — fire-and-forget wizard exit

## Test plan
- [ ] 104 unit/component tests pass (`npm test` across all modified files)
- [ ] TypeScript compiles clean (`tsc --noEmit`)
- [ ] Vite production build succeeds (`npm run build`)
- [ ] Start an import → wizard closes immediately, Runs page shows "Running" status
- [ ] Wait for AI analysis → run transitions to "Needs Review", Review CTA appears
- [ ] Click Review → ReviewPage loads with correct mapping payload
- [ ] Create entries → run transitions to "Completed" with entry links
- [ ] Start two concurrent imports → both rows appear independently with correct statuses
- [ ] Dismiss a failed/expired run → row removed from list and localStorage
- [ ] Clear localStorage → empty state with "Start your first import" CTA

[INTEG-4526]: https://contentful.atlassian.net/browse/INTEG-4526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/drive-integration/src/locations/Page/Page.tsx
```diff
@@ -44,32 +80,33 @@ const Page = () => {
...
-    try {
-      await resumeWorkflow(mappingReviewState.runId, { cancelled: true, entryBlockGraph: graph });
-    } catch (error) {
-      console.error(error);
-    } finally {
-      resetFlowAndReturnToMainPage();
+  const handleCancelReview = async (runId?: string) => {
+    if (runId) {
+      try {
+        await resumeAndPollWorkflow(sdk, runId, { cancelled: true });
```
JuliRossi: ⚔️  This is missing the entryBlockGraph. I think we want it for logging purposes

apps/drive-integration/src/locations/Page/Page.tsx
```diff
@@ -143,8 +209,10 @@ const Page = () => {
         isOAuthBusy={isOAuthBusy}
         onReconnectGoogleDrive={startOAuth}
         onAiAccessDenied={handleAiAccessDenied}
-        onMappingReviewReady={handleMappingReviewReady}
-        onResetToMain={handleReturnToMainPage}
+        onRunStarted={handleRunStarted}
+        onResetToMain={() => setAppView({ view: 'runs' })}
```
JuliRossi: 💭 These props do the same thing, `setAppView({ view: 'runs' })`

apps/drive-integration/src/locations/Page/Page.tsx
```diff
@@ -44,32 +80,33 @@ const Page = () => {
...
-    setMappingReviewState(null);
+  const handleReviewRun = (runId: string) => {
+    setAppView({ view: 'review', runId });
   };
 
-  const resetFlowAndReturnToMainPage = () => {
+  const handleExitReview = () => {
     modalOrchestratorRef.current?.resetFlow();
-    handleReturnToMainPage();
+    setAppView({ view: 'runs' });
```
JuliRossi: 💭 Same logic as handleRunStarted(), would replace with that

apps/drive-integration/src/locations/Page/components/runs/RunRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+          {/* Entry links for completed runs */}
+          {run.displayStatus === 'completed' &&
+            run.createdEntryIds &&
+            run.createdEntryIds.length > 0 && (
+              <Flex gap="spacingXs" flexWrap="wrap">
+                {run.createdEntryIds.map((entryId) => (
+                  <TextLink
+                    key={entryId}
+                    href={`https://app.contentful.com/spaces/${spaceId}/entries/${entryId}`}
```
JuliRossi: ❓ Should this use : `sdk.hostnames.webapp` instead of https://app.contentful.com ?

apps/drive-integration/src/hooks/useRunsPolling.ts
```diff
@@ -0,0 +1,109 @@
...
+    case RunStatus.PENDING_REVIEW:
+      return 'needs-review';
+    case RunStatus.COMPLETED:
+      return 'completed';
+    case RunStatus.FAILED:
+      return 'failed';
+    default:
+      return 'expired';
+  }
+}
```
JuliRossi, line 32: 🤔 This could be a map, instead of s switch. Similar to what we do for FAILURE_REASON_MESSAGES

apps/drive-integration/src/types/runs.ts
```diff
@@ -0,0 +1,23 @@
+export interface RunRecord {
+  runId: string;
+  documentTitle: string;
+  documentId: string;
+  contentTypeIds: string[];
+  startedAt: string;
+  createdEntryIds?: string[];
+}
+
+export type DisplayStatus =
```
JuliRossi: 🤔 This could be an enum, as we are using these states in multiple places

apps/drive-integration/src/hooks/useRunsPolling.ts
```diff
@@ -0,0 +1,109 @@
...
+
+function extractErrorMessage(runData: AgentRunData | null): string | undefined {
+  return runData?.metadata?.workflowFailure?.message ?? undefined;
```
JuliRossi:
```suggestion
  return runData?.metadata?.workflowFailure?.message;
```

apps/drive-integration/src/hooks/useRunsPolling.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export interface UseRunsPollingResult {
+  statusMap: Map<string, DisplayStatus>;
+  errorMap: Map<string, string>;
+}
+
+export function useRunsPolling(runs: RunRecord[], sdk: PageAppSDK): UseRunsPollingResult {
+  const [statusMap, setStatusMap] = useState<Map<string, DisplayStatus>>(
+    () => new Map(runs.map((r) => [r.runId, 'loading' as DisplayStatus]))
```
JuliRossi: 🤔 I think we can avoid seeding `loading` here, because in  RunsPage where runsWithStatus is built:
` displayStatus: statusMap[r.runId] ?? 'loading',`

apps/drive-integration/src/hooks/useRunsPolling.ts
```diff
@@ -0,0 +1,109 @@
...
+  // Initial fetch + reactive refetch when run list changes
+  useEffect(() => {
+    // Reset new runs to 'loading'
+    setStatusMap((prev) => {
+      const next = new Map(prev);
+      for (const r of runs) {
+        if (!next.has(r.runId)) next.set(r.runId, 'loading');
+      }
+      return next;
+    });
```
JuliRossi: 🤔  Do we need this reset right after setting it to loading?

apps/drive-integration/src/hooks/useWorkflowAgent.ts
```diff
@@ -279,30 +68,8 @@ export const useWorkflowAgent = ({
...
-  return {
-    isAnalyzing,
-    startWorkflow,
-    resumeWorkflow,
-  };
+  return { isAnalyzing, startWorkflow };
 };
+
+// Re-export WorkflowFailureReason for backward compatibility with callers that imported it here
+export { WorkflowFailureReason };
```
JuliRossi: ⚔️  If we are changing where WorkflowFailureReason is, then we should change the imports

apps/drive-integration/src/hooks/useWorkflowAgent.ts
```diff
@@ -279,30 +68,8 @@ export const useWorkflowAgent = ({
...
-    },
-    [sdk]
-  );
-
-  return {
-    isAnalyzing,
-    startWorkflow,
-    resumeWorkflow,
-  };
+  return { isAnalyzing, startWorkflow };
```
JuliRossi: ⛏️  I think we are no longer using `isAnalyzing`

apps/drive-integration/src/hooks/useRunStorage.ts
```diff
@@ -0,0 +1,101 @@
...
+          writeToStorage(key, next);
+          setStorageError(null);
+        } catch (err) {
+          const msg =
+            err instanceof Error
+              ? err.message
+              : 'Unable to save import history. Storage may be full.';
+          setStorageError(msg);
+        }
+        return next;
```
JuliRossi: 💭 Same logic as persist

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
```diff
@@ -291,23 +292,39 @@ export const ModalOrchestrator = forwardRef<ModalOrchestratorHandle, ModalOrches
...
+        if (storageError) {
+          // Storage unavailable — surface error before proceeding
+          showWorkflowError(
+            new WorkflowRunError(
+              'Unable to track this import: browser storage is unavailable or full.',
+              WorkflowFailureReason.GENERIC
+            )
+          );
+          return;
+        }
```
JuliRossi: ❓ Do we want to start the workflow even if we've got no storage? How is the user going to be able to see the run?

apps/drive-integration/src/locations/Page/components/runs/RunRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+import type { RunWithStatus } from '../../../../types/runs';
+
+interface RunRowProps {
+  run: RunWithStatus;
+  spaceId: string;
+  onReview: (runId: string) => void;
+  onDismiss: (runId: string) => void;
+}
+
+function formatDate(iso: string): string {
```
JuliRossi, line 23: ⛏️ This manual format handling could be avoided if use a library like Dayjs

apps/drive-integration/src/locations/Page/Page.tsx
```diff
@@ -1,41 +1,77 @@
...
 const Page = () => {
   const sdk = useSDK<PageAppSDK>();
   const modalOrchestratorRef = useRef<ModalOrchestratorHandle>(null);
+
   const [aiAccessDeniedMessage, setAiAccessDeniedMessage] = useState<string | null>(null);
-  const [mappingReviewState, setMappingReviewState] = useState<{
-    payload: MappingReviewSuspendPayload;
-    runId?: string;
-  } | null>(null);
+  const [appView, setAppView] = useState<AppView>({ view: 'runs' });
```
JuliRossi: 💭  Could we move runId outside AppView type? It makes AppView type a bit inconsistent. This could also make simpler with UseEffect in line 41

apps/drive-integration/src/locations/Page/Page.tsx
```diff
@@ -1,41 +1,77 @@
...
-    sdk,
-    documentId: '',
-    oauthToken: '',
-  });
+
+  // When navigating to the review view, fetch the suspend payload from the backend
+  useEffect(() => {
+    if (appView.view !== 'review') {
+      setPendingReviewPayload(null);
+      return;
```
JuliRossi:
```suggestion
    if (appView.view !== 'review') {
      setPendingReviewPayload(null);
       setIsLoadingReviewPayload(false);
      return;
```

apps/drive-integration/src/types/runs.ts
```diff
@@ -0,0 +1,23 @@
...
+  | 'completed'
+  | 'failed'
+  | 'expired';
+
+export interface RunWithStatus extends RunRecord {
+  displayStatus: DisplayStatus;
+  errorMessage?: string;
+}
+
+export type AppView = { view: 'runs' } | { view: 'import' } | { view: 'review'; runId: string };
```
JuliRossi: 💭 runs, import and review could be an enum

apps/drive-integration/src/locations/Page/components/runs/RunsPage.tsx
JuliRossi, line 1: ❓ Should we allow the user to a way clear the runs they have on the page? So they can free space on the localstorage as well/ Also if a  run stuck in 'running' has no action button. The user can't dismiss it or can't cancel it

JuliRossi (commented):
Nice work! They way I review PRs:
⛏️  Nitpicks, non blockers
💭 / 🤔  Highly suggested, but non blockers
❓ Question, may be blocking or not depending on the answer
⚔️  Blockers

After testing it locally: There seems to be a bug with the document title that keep it as Untitled document instead of the actual title. We are also showing the entry id instead of the name in the RunPage. Got an edge case, a workflow stucked in running
Captura de pantalla 2026-07-20 a las 1 07 36 p  m

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
JuliRossi, line 1: 💭 I think we no longer need the loading step for the ModalOrchestrator

apps/drive-integration/src/locations/Page/Page.tsx
david-shibley-contentful: Fixed — `handleCancelReview` now accepts `(runId: string, entryBlockGraph: EntryBlockGraph)` and passes the graph to `resumeAndPollWorkflow`. The `onCancelReview` prop on `ReviewPage` threads it through: `onCancelReview={(graph) => handleCancelReview(appView.runId, graph)}`.

apps/drive-integration/src/locations/Page/Page.tsx
david-shibley-contentful: Fixed — both `onResetToMain` and `onRunStarted` in `ModalOrchestrator` now point to `handleRunStarted`, which does the single `setAppView({ view: AppViewKind.RUNS })`. The duplicate prop is gone.

apps/drive-integration/src/locations/Page/Page.tsx
david-shibley-contentful: Fixed — `handleExitReview` now calls `handleRunStarted()` instead of inlining the same `setAppView` call.

apps/drive-integration/src/locations/Page/components/runs/RunRow.tsx
david-shibley-contentful: Fixed — now uses `sdk.hostnames.webapp ?? 'app.contentful.com'` as `webappHost`, passed down from `RunsPage` to `RunRow` as a prop. All `href` values build off that host.

apps/drive-integration/src/hooks/useRunsPolling.ts
david-shibley-contentful: Done — replaced the switch with a `RUN_STATUS_TO_DISPLAY` record map and a `toDisplayStatus` helper that looks up the map and falls back to `DisplayStatus.EXPIRED`.

apps/drive-integration/src/types/runs.ts
david-shibley-contentful: Done — `DisplayStatus` is now a TypeScript enum. All string literal comparisons across `RunRow`, `RunsPage`, and `useRunsPolling` have been updated to use the enum values.

apps/drive-integration/src/hooks/useRunsPolling.ts
david-shibley-contentful: Applied the suggestion — `extractErrorMessage` now returns `runData?.metadata?.workflowFailure?.message` directly with no trailing `?? undefined`.

apps/drive-integration/src/hooks/useRunsPolling.ts
david-shibley-contentful: Fixed — `useState` is now initialized with `new Map()` (no seeding). The `loading` fallback lives only in `RunsPage` as `statusMap.get(r.runId) ?? DisplayStatus.LOADING`.

apps/drive-integration/src/hooks/useRunsPolling.ts
david-shibley-contentful: Removed — the redundant status map reset inside the `useEffect` before the fetch is gone. The map is only updated when fetch results come back.

apps/drive-integration/src/hooks/useWorkflowAgent.ts
david-shibley-contentful: Fixed — removed the `WorkflowFailureReason` re-export and its import from `useWorkflowAgent`. All callers already import it directly from `@types`.

apps/drive-integration/src/hooks/useWorkflowAgent.ts
david-shibley-contentful: Removed — `isAnalyzing` state and its return value are gone from the hook. The return type is now just `{ startWorkflow }`.

apps/drive-integration/src/hooks/useRunStorage.ts
david-shibley-contentful: Fixed — `addRun` now calls `persist(next)` instead of duplicating the localStorage write logic inline.

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
david-shibley-contentful: Good catch — storage check now happens before `startWorkflow` is called. If storage is unavailable the flow shows an error modal and never kicks off the agent run.

apps/drive-integration/src/locations/Page/components/runs/RunRow.tsx
david-shibley-contentful: Noted — left as-is for now to avoid adding a dependency mid-PR, but happy to add dayjs in a follow-up if the team prefers it consistently across the app.

apps/drive-integration/src/locations/Page/Page.tsx
david-shibley-contentful: Good suggestion. Kept the discriminated union for now since it makes the `AppViewKind.REVIEW` case narrowing ergonomic (TypeScript knows `appView.runId` is defined inside that branch). Moving `runId` out would require a separate `useState<string | null>` and a null check everywhere it's used. Worth revisiting, but didn't want to change the shape mid-review.

apps/drive-integration/src/locations/Page/Page.tsx
david-shibley-contentful: Applied — the early return now uses `AppViewKind.REVIEW` enum comparison and also calls `setIsLoadingReviewPayload(false)` so loading state doesn't leak when navigating away.

apps/drive-integration/src/types/runs.ts
david-shibley-contentful: Done — `AppViewKind` is now a TypeScript enum with `RUNS = 'runs'` and `REVIEW = 'review'`. All `AppView` discriminant comparisons and `setAppView` calls use the enum values.

apps/drive-integration/src/locations/Page/components/runs/RunsPage.tsx
david-shibley-contentful: Two open questions here: (1) **Clearing/dismissing runs** — we don't currently expose a way to remove individual rows (the `removeRun` hook exists but isn't wired to any UI). Worth adding a dismiss/clear button, especially for completed/expired rows. Happy to add that in this PR or a follow-up. (2) **Stuck 'running' rows** — agreed, there's no cancel action for a run that gets stuck. The retry button only appears on `failed`/`expired`. One option is to also show a 'Cancel' button on `running` rows that calls the resume endpoint with `{ cancelled: true }`. Wanted to flag this as a conscious gap — what would you prefer?

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
david-shibley-contentful: The `FlowStep.LOADING` step is still needed — it renders the `LoadingModal` while `startAgentRun` is in flight (the network call to kick off the agent). Without it the modal would close immediately before the run is confirmed started. The step goes away as soon as `startWorkflow` resolves and the flow resets.

apps/drive-integration/src/hooks/useWorkflowAgent.ts
JuliRossi: It's not removed. Could be a regression when merging

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestrator.tsx
JuliRossi: What I meant is that we no longer need the LoadingModal. We can go straight to the RunsPage. If needed, the last step of the flow can have a loading state for the button

apps/drive-integration/src/locations/Page/Page.tsx
JuliRossi: We still have duplicated props as onRunStarted and  onResetToMain do the same

apps/drive-integration/src/hooks/useRunsPolling.ts
JuliRossi: This is still doing the same in the current version

apps/drive-integration/src/hooks/useRunsPolling.ts
```diff
@@ -0,0 +1,104 @@
...
+const RUN_STATUS_TO_DISPLAY: Partial<Record<RunStatus, DisplayStatus>> = {
+  [RunStatus.IN_PROGRESS]: DisplayStatus.RUNNING,
+  [RunStatus.DRAFT]: DisplayStatus.RUNNING,
+  [RunStatus.PENDING_REVIEW]: DisplayStatus.NEEDS_REVIEW,
+  [RunStatus.COMPLETED]: DisplayStatus.COMPLETED,
+  [RunStatus.FAILED]: DisplayStatus.FAILED,
+};
+
+function toDisplayStatus(runData: AgentRunData | null): DisplayStatus {
+  if (!runData) return DisplayStatus.EXPIRED;
```
joaquincasal: 🔴 Two linked issues rooted here. `if (!runData) return DisplayStatus.EXPIRED` treats every "no data" as **Expired**, but `getWorkflowRun` returns `null` for a 404 *and* the `.catch(() => null)` on line 51 folds network blips / 500s into the same `null`. Consequences:
- A freshly-started run can render as **"Expired"** before the backend makes it queryable.
- Because the interval only re-arms while something is `RUNNING` (line 81), a transient error on the *only* running run flips it to `EXPIRED`, `stillRunning` goes false, the interval clears, and **polling never restarts** until `runs` changes or the component remounts. One dropped request can strand a healthy import permanently.

`pollAgentRun` (`workflowService.ts:148`) hits the same 404 and handles it as *"not found yet — keep polling."* Can we bring that distinction here? Treat `null` for a recently-`startedAt` run as still `LOADING`, decaying to `EXPIRED` only after a grace window / N consecutive misses (like `MAX_PENDING_REVIEW_MISSING_PAYLOAD_RETRIES`). That fixes the false-Expired *and* keeps the interval alive so it self-heals.

apps/drive-integration/src/locations/Page/Page.tsx
```diff
@@ -1,41 +1,80 @@
...
+    if (appView.view !== AppViewKind.REVIEW) {
+      setPendingReviewPayload(null);
+      setIsLoadingReviewPayload(false);
+      return;
+    }
+
+    let isCancelled = false;
+    setIsLoadingReviewPayload(true);
+
+    void getWorkflowRun(sdk, spaceId, environmentId, appView.runId)
```
joaquincasal: 🔴 The review-payload fetch here is single-shot with no retry, and the effect re-runs only when `runId` changes (dep array, line 70). Pair that with the note in `constants/agent.ts` — *"Agents-api writes PENDING_REVIEW status before the suspendPayload metadata flushes"* (the reason `MAX_PENDING_REVIEW_MISSING_PAYLOAD_RETRIES = 5` exists) — and there's a documented ~50s window where the badge reads "Ready for review" but `suspendPayload` is still `null`.

A user who clicks Review in that window gets `pendingReviewPayload = null`, and the render guard on line 207 (`isLoadingReviewPayload || !pendingReviewPayload`) then shows a **bare spinner with no error state and no way back**. It doesn't self-heal: the effect is keyed only on `runId`, and `RunsPage`'s poller is unmounted in this view, so nothing re-fetches once the payload flushes — only a manual app reload escapes it.

Two changes make this safe: (1) give the fetch the same bounded retry as `pollAgentRun`, and (2) add an error/empty branch with a **"Back to runs"** action. Even just the back button turns a dead-end into a recoverable one.

apps/drive-integration/src/hooks/useRunsPolling.ts
```diff
@@ -0,0 +1,104 @@
...
+  const intervalRef = useRef<ReturnType<typeof setInterval> | null>(null);
+
+  const fetchAllStatuses = useCallback(async () => {
+    if (runs.length === 0) return;
+
+    const spaceId = sdk.ids.space;
+    const environmentId = sdk.ids.environmentAlias ?? sdk.ids.environment;
+
+    const results = await Promise.all(
+      runs.map((r) => getWorkflowRun(sdk, spaceId, environmentId, r.runId).catch(() => null))
```
joaquincasal: 🟡 `runs.map(...)` re-fetches *every* run each tick, including terminal ones. At the 50-run cap, 1 running + 49 completed = 50 parallel requests every 10s that can't change. Wondering if we scope the poll to non-terminal runs and keep the last-known status for the rest — cuts request volume and stops completed runs from being re-exposed to the null→`EXPIRED` path in the comment above.

apps/drive-integration/src/hooks/useRunStorage.ts
```diff
@@ -0,0 +1,104 @@
...
+    [key]
+  );
+
+  const addRun = useCallback(
+    (record: RunRecord) => {
+      setRuns((current) => {
+        if (current.some((r) => r.runId === record.runId)) return current;
+        const next = [record, ...current];
+        if (next.length > MAX_RUNS) next.splice(MAX_RUNS);
+        persist(next);
```
joaquincasal, line 61: 🟡 `addRun`/`removeRun`/`retryRun`/`markCompleted` all call `persist` *inside* the `setRuns(current => …)` updater, and `persist` itself calls `setRuns` + `setStorageError` + writes localStorage. Updaters should be pure — this nests a `setState` and does I/O in the reducer, so StrictMode's double-invoke will dedup/write twice and the second `setRuns` is redundant. Suggest computing `next` and persisting outside the updater (or in a `useEffect` keyed on `runs`).

Related quota edge case (`ModalOrchestrator.tsx:312`): on a failed write, `persist` sets `storageError` but still updates in-memory `runs`, and the guard only trips on the *next* import — so the overflowing run fires on the backend, isn't persisted, and vanishes on reload. Worth a pre-write capacity check or surfacing it for the current run.

joaquincasal (changes requested):
**Requesting changes — architecture is 👍, but two edge cases will make healthy imports look broken.**

Really like the direction: fire-and-forget + a polled Runs page is a much better UX than the 20-min blocking spinner, and the `workflowService` extraction reads cleanly. My hold is on two spots where the new paths treat *transient* backend states as *terminal* — and both are triggered by eventual-consistency we already defend against in `pollAgentRun`. The good news is the fix template already lives in this repo. Details inline.

apps/drive-integration/src/locations/Page/Page.tsx
JuliRossi: Yes, I don't mind having an additional state, because it makes the type a bit clearer

apps/drive-integration/src/locations/Page/components/runs/RunsPage.tsx
JuliRossi: I don't mind we address it in a different PR as this one is a large, but I'd suggest creating a ticket for it. This is a blocking issue

joaquincasal (approved):
Looks good!
One last thing I noticed is that there's no warning for users around the limitations of using local storage (they won't see the runs if they change devices, other users can't see their runs, etc.). I'd that and maybe also a warning about the 50 runs limit.

david-shibley-contentful:
great point @joaquincasal! I will reach out to Erin and see if we can't come up with a solution for that

david-shibley-contentful:
I spoke with Zach and Erin both and they decided they don't want any warnings for now

apps/drive-integration/src/hooks/useWorkflowAgent.ts
```diff
@@ -35,211 +10,19 @@ interface UseWorkflowParams {
...
-  );
-};
-
 export const useWorkflowAgent = ({
   sdk,
   documentId,
   oauthToken,
 }: UseWorkflowParams): WorkflowHook => {
-  const [isAnalyzing, setIsAnalyzing] = useState(false);
-  const { 'google-docs-agent-improvements': extendedTimeout } = useGoogleDocsAgentFlags();
```
FBanfi, line 236: Just a heads up that we remove this `extendedTimeout` but don't add it in the `useWorkflowAgentLegacy.ts`, not sure if that's intended.

Context: `useWorkflowAgentLegacy` hardcodes `MAX_POLL_ATTEMPTS`. So if both flags are off simultaneously, 
  `google-docs-async-runs=false` and `google-docs-agent-improvements=true`, users on the legacy path lose
  the extended timeout and will hit timeouts earlier than before.

apps/drive-integration/src/hooks/useWorkflowAgent.ts
```diff
@@ -35,211 +10,19 @@ interface UseWorkflowParams {
...
-// document-too-complex and out-of-domain are not yet emitted by the backend; handlers are in place for when they ship.
-const FAILURE_REASON_MESSAGES: Partial<Record<WorkflowFailureReason, string>> = {
-  [WorkflowFailureReason.GOOGLE_DRIVE_AUTH_EXPIRED]: ERROR_MESSAGES.GOOGLE_DRIVE_AUTH_ERROR,
-  [WorkflowFailureReason.GOOGLE_DOCS_NOT_FOUND]: ERROR_MESSAGES.GOOGLE_DOCS_NOT_FOUND,
-  [WorkflowFailureReason.AI_SERVICE_UNAVAILABLE]: ERROR_MESSAGES.AI_SERVICE_UNAVAILABLE,
-  [WorkflowFailureReason.APP_NOT_INSTALLED]: ERROR_MESSAGES.APP_NOT_INSTALLED,
-  [WorkflowFailureReason.DOCUMENT_TOO_COMPLEX]: ERROR_MESSAGES.DOCUMENT_TOO_COMPLEX,
-  [WorkflowFailureReason.PROCESSING_TIMEOUT]: ERROR_MESSAGES.PROCESSING_TIMEOUT,
-  [WorkflowFailureReason.OUT_OF_DOMAIN]: ERROR_MESSAGES.OUT_OF_DOMAIN,
-  [WorkflowFailureReason.MISSING_PARAMETER]: ERROR_MESSAGES.MISSING_PARAMETER,
```
FBanfi, line 128: Similar to the previous comment: `useWorkflowAgentLegacy` doesn't have it. If the backend returns that failure code on the legacy path, users will see a generic "Workflow failed" instead of the specific error message.

apps/drive-integration/src/locations/Page/components/runs/RunRow.tsx
```diff
@@ -0,0 +1,199 @@
...
+  status: RunWithStatus['displayStatus'];
+  errorMessage?: string;
+  entryCount?: number;
+}) {
+  switch (status) {
+    case DisplayStatus.LOADING:
+      return (
+        <Badge
+          variant="secondary"
+          style={{ background: '#F0F2F5', color: '#536171', border: 'none' }}>
```
FBanfi, line 42: We could use Forma for the colors, font weight and some of the styling here!

apps/drive-integration/src/locations/Page/components/runs/RunsPage.tsx
```diff
@@ -0,0 +1,308 @@
...
+              </Popover>
+            </Flex>
+          </Flex>
+
+          <style>{`
+            [data-test-id="cf-ui-table"] { border: none; box-shadow: none; }
+            [data-test-id="cf-ui-table-row"] { box-shadow: none; }
+            [data-test-id="cf-ui-table-row"]:not(:last-child) [data-test-id="cf-ui-table-cell"] { border-bottom: 1px solid #E5E8ED; }
+            [data-test-id="cf-ui-table-head"] [data-test-id="cf-ui-table-row"] [data-test-id="cf-ui-table-cell"] { border-bottom: 1px solid #E5E8ED; }
+          `}</style>
```
FBanfi, line 261: Nit: targeting data-test-id attributes for styling is a bit fragile — those are meant for testing and f36 could rename/remove them without warning. also since this is a global `<style>` injection it'd clobber any other table on the same page.
Wondering if wrapping the table in a scoped container and using `@emotion/css` (which f36 already uses
  under the hood) would be cleaner here?

apps/drive-integration/src/locations/Page/components/runs/RunsPage.tsx
```diff
@@ -0,0 +1,308 @@
...
+                      <SortAscendingIcon size="small" />
+                    )}
+                    <span>Sort by</span>
+                    <CaretDownIcon size="small" />
+                  </button>
+                </Popover.Trigger>
+                <Popover.Content>
+                  <Flex flexDirection="column" style={{ minWidth: '160px' }}>
+                    {(['newest', 'oldest'] as SortOrder[]).map((value) => (
+                      <button
```
FBanfi, line 231: We could use Forma's button

FBanfi (approved):

PR #11094, opened 2026-07-17
[] fix(security): validate postMessage origins and fix slack middleware path

mgoudy91:
## Summary

**Slack lambda — missing leading slash (APPS-1)**
- \`apps/slack/lambda/lib/app.ts\`: \`'api/tokens'\` → \`'/api/tokens'\` in middleware mount. The missing slash meant \`path-to-regexp@0.1.x\` never matched, silently bypassing request verification on \`POST /api/tokens\`.

**Wildcard postMessage — senders (APPS-2)**
All four sender files replaced \`'*'\` with \`document.referrer ? new URL(document.referrer).origin : window.location.origin\`:
- \`apps/slack/frontend/src/index.tsx\`
- \`apps/smartling/frontend/src/standalone.ts\`
- \`apps/jira/jira-app/src/standalone.ts\`

**Wildcard postMessage — receivers (APPS-2)**
Added \`event.origin\` checks to all receivers:
- Slack: \`message.origin !== new URL(BACKEND_BASE_URL).origin\`
- Smartling: \`origin !== window.location.origin\`
- Jira: \`e.origin !== new URL(constants.OAUTH_REDIRECT_URI).origin\`

## Test plan
- [x] Slack OAuth connect flow works end-to-end
- [x] \`POST /api/tokens\` now has request verification applied
- [x] Smartling, Jira OAuth flows still deliver tokens correctly
- [x] \`postMessage\` from a foreign origin is ignored by each receiver


ryunsong-contentful (approved):

harikakondur:
removed typeform changes which were fixed in this pr: https://github.com/contentful/apps/pull/11168

jjolton-contentful (changes requested):
Thanks for picking this up! The security intent is right, but the origin pinning lands on the wrong host for Jira + Slack and will break OAuth in prod. Requesting changes.

### Pattern to mirror

[#11168 (Typeform)](https://github.com/contentful/apps/pull/11168) already solved this class of bug the right way:

- **Sender:** `postMessage(data, origin)` where `origin` comes from the callback page URL
- **Receiver:** `event.origin === window.location.origin`

Callback and opener share the **app** origin. The API/OAuth-redirect host is only a bounce — it is not `event.origin` on the message that carries the token.

### Blocking

1. **Jira receiver** (`OAuth.tsx`) — `expectedOrigin` is derived from `OAUTH_REDIRECT_URI` (`https://api.jira.ctfapps.net`). After token exchange the lambda 302s to `FRONTEND_URL` (`https://app.jira.ctfapps.net`), and `standalone` posts from there. Legitimate messages have origin `app.jira…` and get dropped.

   **Fix:** require `e.origin === window.location.origin` (Typeform pattern). Update the specs to fire `MessageEvent` with `https://app.jira.ctfapps.net`, not `api.`.

   Note: the **sender** change in `standalone.ts` (callback URL `origin`) is already correct — leave it.

2. **Slack receiver** (`useConnect.ts`) — same mismatch. `postMessage` runs on `https://slack.ctfapps.net` after the API HTML bounce; the listener requires `BACKEND_BASE_URL` origin (`slack-api.ctfapps.net`). Every successful OAuth message is rejected.

   **Fix:** `message.origin === window.location.origin`.

3. **Slack + Smartling senders** — `document.referrer ? … : window.location.origin` is the wrong target. Referrer can be the API bounce (Slack) or the IdP (Smartling SSO), so `targetOrigin` may not match the opener. Even when referrer is stripped and the fallback works, Slack still dies on finding 2.

   **Fix:** use the callback page origin (Jira/Typeform style), not `document.referrer`.

### Suggested shape

| App | Sender `targetOrigin` | Receiver check |
|-----|----------------------|----------------|
| Jira | callback `origin` ✅ already | `window.location.origin` (not `OAUTH_REDIRECT_URI`) |
| Slack | callback `origin` (drop referrer) | `window.location.origin` (not `BACKEND_BASE_URL`) |
| Smartling | callback `origin` (drop referrer) | `window.location.origin` ✅ already |

Happy to pair on the Jira/Slack receiver bits if useful — once those two are flipped to app origin, the scary path is unblocked.

jjolton-contentful (approved):

PR #11099, opened 2026-07-20
fix(drive-integration): treat Add Entry references as child of existing parent [INTEG-4524]

JuliRossi:
[INTEG-4524](https://contentful.atlassian.net/browse/INTEG-4524)

## Summary

- Collapse Add Entry into a single-screen form (`AddEntryForm`) with conditional reference controls.
- Flip reference semantics: the **new entry is a child** of an existing parent (not a parent pointing at an existing entry).
- On Yes: pick parent entry → pick parent reference field when the parent has more than one → write `__ref` / `sourceEntryIds` on the **parent** and edge `from=parent → to=child`.
- Single Link fields **replace** any prior child; Array fields **append**.
- Disable **Yes** only when no existing parent content type has an Entry reference field.

## New reference logic

```mermaid
flowchart TD
  A[Select content type for new entry] --> B{Any existing parent<br/>has Entry reference fields?}
  B -->|No| C[Yes disabled — only No allowed]
  B -->|Yes| D{Should this new entry be a<br/>reference of an existing entry?}
  D -->|No| E[Map fields on new entry only]
  D -->|Yes| F[Select parent entry]
  F --> G{Parent has how many<br/>Entry reference fields?}
  G -->|0| H[Save disabled]
  G -->|1| I[Auto-pick that field]
  G -->|2+| J[Select parent field]
  I --> K[Save]
  J --> K
  K --> L[Create new child entry]
  L --> M{Parent field type}
  M -->|Link| N[Replace parent field value,<br/>sourceEntryIds, and edges]
  M -->|Array| O[Append child to parent field,<br/>sourceEntryIds, and edges]
  N --> P["Edge: from=parent → to=child"]
  O --> P
```

### Before vs after

| | Before (wrong) | After |
|---|---|---|
| Who holds `__ref` | New entry | Parent entry |
| Reference fields shown | New content type | Parent content type |
| Graph edge | `from=new → to=existing` | `from=parent → to=child` |
| Yes disabled when | New CT has no ref fields | No parent can accept a ref |

## Test plan

- [ ] Add entry **No** → saves standalone entry with mapped fields
- [ ] Add entry **Yes** with parent that has one Link field → links as child without showing field select
- [ ] Add entry **Yes** with parent that has multiple ref fields → field select appears; Save requires a choice
- [ ] Parent already has a filled single Link → creating a new child **replaces** the old reference on create (old id gone from `sourceEntryIds` / payload)
- [ ] Parent Array reference → new child is **appended**
- [ ] When every existing entry’s CT has no Entry refs → **Yes** is disabled, **No** still works
- [ ] New entry CT with no ref fields (e.g. Component/Tag) but parents can accept refs → **Yes** stays enabled

[INTEG-4524]: https://contentful.atlassian.net/browse/INTEG-4524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JuliRossi:
Dividing this into 2 PRs, so it is easier to review

PR #11110, opened 2026-07-21
feat(drive-integration): Replace Add Entry wizard with single-screen form [INTEG-4524]

JuliRossi:
## Summary
- Replace multi-step `AddEntryWizard` with a single-screen `AddEntryForm` (content type, Yes/No reference radios, conditional entry/field selects, field multiselect).
- Simplify `EditModal` footer to Back + Save (no Next) while the add-entry form is open.
- Disable **Yes** when the selected **new** content type has no Entry reference fields (pre–child-reference semantics).

> **Note:** This PR does **not** change who owns the reference link. The new entry still holds `__ref` and edges are still `from=new → to=existing`. Child-of-parent linking lands in the follow-up PR stacked on this branch.

https://github.com/user-attachments/assets/5042e655-e943-48c6-8ec7-932e55638b5f

Update: When a parent does not have a reference field, it doesn't appear in the multiselect or lets you select radio button

https://github.com/user-attachments/assets/8acbc70e-61e4-485a-b243-8f196372dfcb

## Test plan
- [ ] Open Add entry from Edit content mapping → single form (no wizard steps / no Next)
- [ ] Before content type: only content type select is shown
- [ ] After content type: radios + field multiselect appear together
- [ ] Yes → entry select (+ reference field select when the new CT has >1 ref fields)
- [ ] No → no entry/ref-field selects; field multiselect remains
- [ ] Content type with no Entry ref fields → Yes disabled, No still works
- [ ] Save creates entry and maps selected fields


apps/drive-integration/test/locations/Page/components/modals/EditModal.spec.tsx
david-shibley-contentful, line 1: EditModal.spec.tsx doesn't have a test that fills out the form and asserts onAddEntry is called. The unit tests for toAddEntryFormParams cover the mapping logic, but the wire from "click Save" → handleAddEntrySave → onAddEntry is untested at the integration level

david-shibley-contentful:
Back now wipes the entire form state (setAddEntryFormState(null)). In the wizard, Back preserved what you'd filled in. If a user is 4 fields deep and misclicks Back, they start over. Quick "is this intentional?" comment. If yes, fine, just leave it; if no, it needs a fix.

apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/AddEntryForm.tsx
```diff
@@ -0,0 +1,229 @@
...
+  return {
+    contentTypeId: state.contentTypeId,
+    isReference: state.isReference ?? false,
+    referenceEntryId: state.isReference ? state.referenceEntryId || null : null,
+    referenceFieldId: state.isReference
+      ? state.referenceFieldId || referenceFieldOptions[0]?.id || null
+      : null,
+    fieldIds: state.selectedFieldIds,
+  };
+};
```
FBanfi, line 103: Nit: we could make a utils file and move all this logic there to clean a bit the component.

FBanfi (approved):
Looks good to me! Leaving the approval on my side!

JuliRossi:
> Back now wipes the entire form state (setAddEntryFormState(null)). In the wizard, Back preserved what you'd filled in. If a user is 4 fields deep and misclicks Back, they start over. Quick "is this intentional?" comment. If yes, fine, just leave it; if no, it needs a fix.

Since the form is now a single screen, Back is effectively a cancel, there's no previous step to return to. The state wipe is intentional. Also you can't go 4 fields deep, since it only allows you to add an entry at a time.

apps/drive-integration/src/locations/Page/components/review/mapping/edit-modals/AddEntryForm.tsx
JuliRossi: Will do it on a different PR so it doesn't colide with #11111

FBanfi (approved):
🔥

PR #11111, opened 2026-07-21
fix(drive-integration): link new Add Entry as child of selected parent [INTEG-4524]

JuliRossi:
## Summary
- Flip Add Entry reference semantics: the **new entry is a child** of an existing parent (not a parent pointing at an existing entry).
- Reference fields come from the **parent** content type; `__ref` / `sourceEntryIds` are written on the **parent**; edge is `from=parent → to=child`.
- Single **Link** fields **replace** any prior child; **Array** fields **append**.
- Disable **Yes** only when no existing parent content type has an Entry reference field.
- Stacked on #11110 (UI form refactor).

## New reference logic

```mermaid
flowchart TD
  A[Select content type for new entry] --> B{Any existing parent<br/>has Entry reference fields?}
  B -->|No| C[Yes disabled — only No allowed]
  B -->|Yes| D{Should this new entry be a<br/>reference of an existing entry?}
  D -->|No| E[Map fields on new entry only]
  D -->|Yes| F[Select parent entry]
  F --> G{Parent has how many<br/>Entry reference fields?}
  G -->|0| H[Save disabled]
  G -->|1| I[Auto-pick that field]
  G -->|2+| J[Select parent field]
  I --> K[Save]
  J --> K
  K --> L[Create new child entry]
  L --> M{Parent field type}
  M -->|Link| N[Replace parent field value,<br/>sourceEntryIds, and edges]
  M -->|Array| O[Append child to parent field,<br/>sourceEntryIds, and edges]
  N --> P["Edge: from=parent → to=child"]
  O --> P
```

### Before vs after

| | Before (wrong) | After |
|---|---|---|
| Who holds `__ref` | New entry | Parent entry |
| Reference fields shown | New content type | Parent content type |
| Graph edge | `from=new → to=existing` | `from=parent → to=child` |
| Yes disabled when | New CT has no ref fields | No parent can accept a ref |

## Test plan
- [ ] Add entry **No** → saves standalone entry with mapped fields
- [ ] Add entry **Yes** with parent that has one Link field → links as child without showing field select
- [ ] Add entry **Yes** with parent that has multiple ref fields → field select appears; Save requires a choice
- [ ] Parent already has a filled single Link → creating a new child **replaces** the old reference on create
- [ ] Parent Array reference → new child is **appended**
- [ ] When every existing entry’s CT has no Entry refs → **Yes** is disabled, **No** still works
- [ ] New entry CT with no ref fields (e.g. Component/Tag) but parents can accept refs → **Yes** stays enabled

### DIFF
--- a/apps/drive-integration/src/locations/Page/components/review/mapping/linkChildToParent.ts
+++ b/apps/drive-integration/src/locations/Page/components/review/mapping/linkChildToParent.ts
@@ -1,0 +1,101 @@
+import type {
+  EntryBlockGraphEntry,
+  ReviewedReferenceGraph,
+  ReviewedReferenceGraphEdge,
+  WorkflowContentTypeField,
+} from '@types';
+
+type ParentRefField = Pick<WorkflowContentTypeField, 'id' | 'type'>;
+
+/**
+ * Links a new child entry onto a parent for a given reference field.
+ * - Single Link: replaces existing field value, sourceEntryIds, and graph edges for that field
+ * - Array: appends the new child
+ */
+export function linkChildToParentEntry(args: {
+  parentEntry: EntryBlockGraphEntry;
+  childTempId: string;
+  refField: ParentRefField & { id: string };
+  defaultLocale: string;
+  previousEdges: ReviewedReferenceGraphEdge[];
+}): {
+  parentEntry: EntryBlockGraphEntry;
+  edges: ReviewedReferenceGraphEdge[];
+} {
+  const { parentEntry, childTempId, refField, defaultLocale, previousEdges } = args;
+  const isArrayRef = refField.type === 'Array';
+  const childRef = { __ref: childTempId };
+
+  const existingLocalized = parentEntry.fields?.[refField.id] ?? {};
+  const localesToWrite =
+    Object.keys(existingLocalized).length > 0 ? Object.keys(existingLocalized) : [defaultLocale];
+  const nextLocalized: Record<string, unknown> = { ...existingLocalized };
+  for (const locale of localesToWrite) {
+    if (isArrayRef) {
+      const existingRefs = Array.isArray(nextLocalized[locale])
+        ? (nextLocalized[locale] as unknown[])
+        : [];
+      nextLocalized[locale] = [...existingRefs, childRef];
+    } else {
+      nextLocalized[locale] = childRef;
+    }
+  }
+
+  const fmIdx = parentEntry.fieldMappings.findIndex((fm) => fm.fieldId === refField.id);
+  const nextFieldMappings = [...parentEntry.fieldMappings];
+  if (fmIdx >= 0) {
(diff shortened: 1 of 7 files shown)


apps/drive-integration/src/locations/Page/components/review/mapping/linkChildToParent.ts
```diff
@@ -0,0 +1,101 @@
...
+      const existingRefs = Array.isArray(nextLocalized[locale])
+        ? (nextLocalized[locale] as unknown[])
+        : [];
+      nextLocalized[locale] = [...existingRefs, childRef];
+    } else {
+      nextLocalized[locale] = childRef;
+    }
+  }
+
+  const fmIdx = parentEntry.fieldMappings.findIndex((fm) => fm.fieldId === refField.id);
```
david-shibley-contentful: Can we add a test for the fmIdx === -1 push path?

apps/drive-integration/src/locations/Page/components/review/mapping/linkChildToParent.ts
```diff
@@ -0,0 +1,101 @@
...
+  const nextFieldMappings = [...parentEntry.fieldMappings];
+  if (fmIdx >= 0) {
```
FBanfi: Nit: would be better to rename this to fieldMappingIndex to understand better what does it mean

apps/drive-integration/src/locations/Page/components/review/mapping/linkChildToParent.ts
```diff
@@ -0,0 +1,101 @@
...
+  const nextEdges: ReviewedReferenceGraphEdge[] = parentId
+    ? [
+        ...keptEdges,
+        {
+          from: parentId,
+          to: childTempId,
+          fieldId: refField.id,
+        },
+      ]
+    : keptEdges;
```
FBanfi: If the `parentId` comes null we maintain the previous edges but the field value and fieldMappings are updated correctly, so there is a mismatch there.

A posible solution could be handle the parent checks at the start to avoid this.

apps/drive-integration/test/locations/Page/components/review/mapping/linkChildToParent.spec.ts
FBanfi, line 1: We could add a case for `parentEntry.tempId === undefined`. To verify that a Link-field replacement doesn't leave a stale edge.

FBanfi (commented):
Looks great! Left a comment that I don't think would end up happing in practice but that may be worth addressing just in case!

apps/drive-integration/src/locations/Page/components/review/mapping/linkChildToParent.ts
JuliRossi: Great catch!

PR #11075, opened 2026-07-14
feat: drive-integration v1.1 — UI redesigns, keyboard accessibility, pre-flight doc scope []

JuliRossi:
## Purpose
Ships all accumulated drive-integration v1.1 work: modal and button redesigns, keyboard accessibility improvements and a batch of UX bug fixes

## Summary
- **Modal redesign** — Move modal (INTEG-4116), Remove button (INTEG-4135), and review-page UI adjustments (INTEG-4179) with new \`AddEntryWizard\` for the new-entry design flow (INTEG-4136)
- **Keyboard accessibility** — Edit action buttons and multiselect keyboard nav in review view (INTEG-3889); new \`keyboard.ts\` utility
- **Tab selection UX** — \`SelectTabsModal\` now uses \`truncateMiddle\` and respects \`MAX_TABS_SELECTION\`
- **Bug fixes** — field dropdown closes on selection (INTEG-4040), multiselect popover misalignment (INTEG-4038), root entry selection on edit mode entry (INTEG-3881)
- **Cleanup** — removed dev fixtures/debug logs/mock singletons (INTEG-3806); added ESLint config; mock test infrastructure aligned with master (\`mockSdk\`/\`mockCma\` singletons)

## Composed from
| PR | Description |
|----|-------------|
| #10991 | chore: remove dev fixtures, debug logs, and mock singletons [INTEG-3806] |
| #10994 | fix: select first root entry when entering edit mode [INTEG-3881] |
| #10996 | fix: multiselect popover misalignment when list is not scrollable [INTEG-4038] |
| #11003 | fix: fix field dropdowns closing on selection [INTEG-4040] |
| #11005 | chore: add eslint script [INTEG-3947] |
| #11006 | feat: keyboard accessibility for edit action buttons in review view [INTEG-3889] |
| #11019 | refactor: Move modal redesign [INTEG-4116] |
| #11023 | feat: adding remove button redesign [INTEG-4135] |
| #11027 | feat: add entry wizard for new entry design updates [INTEG-4136] |
| #11028 | fix: review page UI adjustments [INTEG-4179] |
| #11029 | fix: flaky modal tests caused by react-modal timer leak [INTEG-4036] |
| #11032 | test: fix react-modal timer leak in RemoveContentModal spec |

## Test plan
- [ ] Run \`npm test\` in \`apps/drive-integration\` — all specs pass
- [ ] Manually: open a multi-tab Google Doc → tab selection modal appears, keyboard nav works in multiselect
[...]

PR #11046, opened 2026-07-01
feat(drive-integration): collect document scope pre-flight to eliminate Inngest suspend race [INTEG-4233]

JuliRossi:
[INTEG-4233](https://contentful.atlassian.net/browse/INTEG-4233)

## Purpose
Eliminates the first Inngest suspend (`select-tabs-images-step`) for new workflow runs. Under Inngest, a suspend event and a concurrent resume signal can arrive out of order — removing the suspend removes the race condition entirely.

## Summary
- **Pre-flight scope collection**: after content type selection, the FE calls `docs.googleapis.com` directly with the already-held OAuth token, runs `getDocumentScope()` locally, shows tab/image selection UI if needed, then starts the workflow with `documentScope` in metadata
- **`loadPickerApi` fix**: replaced broken `<script onload>` callback pattern (second script injection whose `onload` never fires on cached `api.js`) with `loadGapi().then(() => gapi.load('picker', ...))` — incidental but it was broken

## Test plan 
  - [ ] Single-tab doc, no images → skips directly to loading
  - [ ] Multi-tab doc → shows tab selection pre-flight
  - [ ] Doc with images → shows image inclusion modal pre-flight
  - [ ] Multi-tab doc with images → tab selection → image modal → workflow start
  - [ ] Auth-expired error during pre-flight → "Reconnect Drive" modal
  - [ ] 404 / PERMISSION_DENIED during pre-flight → "Document not found" modal
 
### Local testing
https://github.com/user-attachments/assets/68bfd188-0801-4ec9-a46f-979ae6be9bcf

## Deployment
This will be merged once https://github.com/contentful/agents-api/pull/755 is deployed in prod :D

Generated with [Claude Code](https://claude.com/claude-code)

[INTEG-4233]: https://contentful.atlassian.net/browse/INTEG-4233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11144, opened 2026-07-27
fix(drive-integration): restore pre-flight tab/image selection in legacy flow []

david-shibley-contentful:
## Summary
- The async-runs rearchitecture (PR #11092) introduced `ModalOrchestratorLegacy` as the flag-off fallback, but it was wired incorrectly: it called `startWorkflow` immediately with hardcoded `{ selectedTabIds: [], includeImages: false }` and tried to show the tab/image selection UI by routing the backend's first `PENDING_REVIEW` suspend after the fact
- This caused two bugs:
  - **Single-tab docs with no images**: the backend suspend payload has `requiresTabSelection: false` and `requiresImageSelection: false`, so `showDocumentScopeReview` fell through to `setFlowStep(null)` — the flow silently stopped without ever resuming, leaving the run stuck in `PENDING_REVIEW`
  - **Multi-tab docs**: the backend likely used the empty `selectedTabIds: []` from the initial request metadata to process only the first tab before suspending, causing very little content to be mapped
- Fix: restore the pre-flight `fetchDocumentSelection` pattern from the original `ModalOrchestrator` — fetch document scope first (tabs + image count), show the tabs/images selection UI, then call `startWorkflow` with the user's actual selection baked in

## Test plan
- [x] With flag OFF (legacy flow): import a multi-tab Google Doc — tabs selection modal appears, selecting specific tabs processes only those tabs
- [x] With flag OFF: import a single-tab doc with images — images selection modal appears
- [x] With flag OFF: import a single-tab doc with no images — flow skips directly to loading, processes correctly
- [x] With flag ON (async flow): no change in behavior

### DIFF
--- a/apps/drive-integration/src/hooks/useWorkflowAgentLegacy.ts
+++ b/apps/drive-integration/src/hooks/useWorkflowAgentLegacy.ts
@@ -104,6 +104,7 @@ const FAILURE_REASON_MESSAGES: Partial<Record<WorkflowFailureReason, string>> = 
   [WorkflowFailureReason.DOCUMENT_TOO_COMPLEX]: ERROR_MESSAGES.DOCUMENT_TOO_COMPLEX,
   [WorkflowFailureReason.PROCESSING_TIMEOUT]: ERROR_MESSAGES.PROCESSING_TIMEOUT,
   [WorkflowFailureReason.OUT_OF_DOMAIN]: ERROR_MESSAGES.OUT_OF_DOMAIN,
+  [WorkflowFailureReason.MISSING_PARAMETER]: ERROR_MESSAGES.MISSING_PARAMETER,
 };
 
 const getWorkflowFailureMessage = (
--- a/apps/drive-integration/src/hooks/useWorkflowAgentLegacy.ts
+++ b/apps/drive-integration/src/hooks/useWorkflowAgentLegacy.ts
@@ -9,7 +9,6 @@ import {
 import {
   MappingReviewSuspendPayload,
   ResumePayload,
-  TabsImagesSuspendPayload,
   CompletedWorkflowPayload,
   WorkflowRunResult,
   RunStatus,
@@ -112,9 +111,7 @@ const getWorkflowFailureMessage = (
   failureReason: WorkflowFailureReason
 ): string => FAILURE_REASON_MESSAGES[failureReason] ?? getRunErrorMessage(runData);
 
-const getSuspendPayload = (
-  runData: AgentRunData
-): TabsImagesSuspendPayload | MappingReviewSuspendPayload | undefined =>
+const getSuspendPayload = (runData: AgentRunData): MappingReviewSuspendPayload | undefined =>
   runData.metadata?.suspendPayload;
 
 const getWorkflowRunResult = (
@@ -166,17 +163,13 @@ const pollAgentRun = async (
 ): Promise<WorkflowRunResult> => {
   const startMs = Date.now();
   let pendingReviewMissingPayloadCount = 0;
-  console.log(`⏳ Polling run [${runId}]`);
-
   for (let attempt = 0; attempt < MAX_POLL_ATTEMPTS; attempt++) {
     const runData = await getWorkflowRun(sdk, spaceId, environmentId, runId);
     if (!runData) {
-      console.log(`  #${attempt + 1} — not found yet (${elapsedSec(startMs)})`);
       await wait(POLL_INTERVAL_MS);
       continue;
     }
     const status = getRunStatus(runData);
-    console.log(`  #${attempt + 1} — status: ${status} (${elapsedSec(startMs)})`);
     if (status === RunStatus.PENDING_REVIEW && !getSuspendPayload(runData)) {
       pendingReviewMissingPayloadCount++;
     } else {
@@ -184,7 +177,6 @@ const pollAgentRun = async (
     }
     const workflowRun = getWorkflowRunResult(runData, runId, pendingReviewMissingPayloadCount);
     if (workflowRun) {
-      console.log(`✓ Run [${runId}] settled: ${status} in ${elapsedSec(startMs)}`);
       return workflowRun;
     }
     await wait(POLL_INTERVAL_MS);
--- a/apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestratorLegacy.tsx
+++ b/apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestratorLegacy.tsx
@@ -12,8 +12,6 @@ import { SelectTabsModal } from '../modals/step_3/SelectTabsModal';
 import {
   DocumentTabProps,
   MappingReviewSuspendPayload,
-  ResumePayload,
-  TabsImagesSuspendPayload,
   RunStatus,
   WorkflowRunResult,
   WorkflowFailureReason,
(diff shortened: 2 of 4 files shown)


JuliRossi (changes requested):
We need to also re-add the changes for `useWorkflowAgentLegacy.ts`.  

⚔️ **`MISSING_PARAMETER` is dropped.** `showWorkflowError` here has no
`MISSING_PARAMETER` branch, so that backend failure shows the generic "Unable to generate
preview". Two spots to match master:
- `ModalOrchestratorLegacy.tsx` → add the `MISSING_PARAMETER` case in `showWorkflowError`.
- `useWorkflowAgentLegacy.ts` → add `[WorkflowFailureReason.MISSING_PARAMETER]: ERROR_MESSAGES.MISSING_PARAMETER` to `FAILURE_REASON_MESSAGES`.

JuliRossi (approved):

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestratorLegacy.tsx
joaquincasal: This name was changed in the original PR but it's not changed here

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestratorLegacy.tsx
```diff
@@ -275,68 +289,50 @@ export const ModalOrchestratorLegacy = forwardRef<
...
-        }))
-      );
+    const showDocumentScopeReview = (
+      selectionConfig: DocumentSelectionConfig,
+      contentTypeIds: string[]
+    ) => {
+      setAvailableTabs(selectionConfig.tabs.map((tab) => ({ tabId: tab.id, tabTitle: tab.title })));
       setSelectedTabs([]);
       setUseAllTabs(null);
       setIncludeImages(null);
```
joaquincasal: These were removed in the original PR but they're present here

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestratorLegacy.tsx
```diff
@@ -275,68 +289,50 @@ export const ModalOrchestratorLegacy = forwardRef<
...
       }
 
-      if (suspendPayload?.requiresImageSelection) {
+      if (requiresImages) {
         setFlowStep(FlowStep.INCLUDE_IMAGES);
         return;
       }
 
-      setFlowStep(null);
+      void startWorkflowWithDelayedLoading(contentTypeIds, {
```
joaquincasal: This used to be named `startWorkflowWithScope`

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestratorLegacy.tsx
```diff
@@ -275,68 +289,50 @@ export const ModalOrchestratorLegacy = forwardRef<
...
-      setFlowStep(FlowStep.LOADING);
-
-      const workflowRun = await resumeWorkflow(activeRunId, resumePayload);
-      handleWorkflowResult(workflowRun);
-    };
-
-    const startWorkflowWithDelayedLoading = async (contentTypeIds: string[]) => {
+    const startWorkflowWithDelayedLoading = async (
+      contentTypeIds: string[],
+      documentSelection: { selectedTabIds: string[]; includeImages: boolean }
```
joaquincasal: Looks like we lost the type `DocumentSelection` here

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestratorLegacy.tsx
joaquincasal, line 355: This section is really similar to what we had before the PR

apps/drive-integration/src/hooks/useWorkflowAgentLegacy.ts
joaquincasal: This is still old, as the TabsImagesSuspendPayload is no longer used

joaquincasal (changes requested):
Just reviewed manually and left some comments, there are a few changes that are still not recovered in this PR, so we may be missing things here. (Whenever I say "original PR" in my comments I mean #11046.
Claude's review is still ongoing, I'll add it when it finishes

apps/drive-integration/src/locations/Page/components/mainpage/ModalOrchestratorLegacy.tsx
```diff
@@ -275,68 +289,50 @@ export const ModalOrchestratorLegacy = forwardRef<
...
+      if (requiresImages) {
         setFlowStep(FlowStep.INCLUDE_IMAGES);
         return;
       }
 
-      setFlowStep(null);
+      void startWorkflowWithDelayedLoading(contentTypeIds, {
+        selectedTabIds: [],
+        includeImages: false,
+      }).catch(handleWorkflowError);
```
joaquincasal: I think this branch drops the workflow result and leaves single-tab/no-image docs stuck on the loading modal — which looks like the exact bug this PR is meant to fix.

Unlike the async `startWorkflowWithScope` (which self-handles its result via `addRun`/`onRunStarted`), the legacy `startWorkflowWithDelayedLoading` just **returns** the `WorkflowRunResult` — the caller is responsible for routing it through `handleWorkflowResult`. The other two callers do exactly that (`handleWorkflowResult(await startWorkflowWithDelayedLoading(...))` in `handleSelectTabsContinue`/`handleIncludeImagesContinue`), but this branch `void`s it. So for a single-tab/no-image doc: the run polls to `PENDING_REVIEW`, the result is discarded, `handleWorkflowResult` never runs → `onMappingReviewReady` never fires, `activeRunId` stays null, and `flowStep` is stuck at `LOADING` (close is suppressed and escape is disabled while loading, so it's a hard stall — only a reload escapes).

Looks like the `void ...catch()` shape was carried over from the async version, where it's safe. Wondering if this should mirror the sibling handlers:
```ts
try {
  handleWorkflowResult(
    await startWorkflowWithDelayedLoading(contentTypeIds, {
      selectedTabIds: [],
      includeImages: false,
    })
  );
} catch (error) {
  handleWorkflowError(error);
}
```
(`showDocumentScopeReview` would need to be `async` for this, or wrap it in an async IIFE.) Same root cause also means `activeRunId` never gets set, so a subsequent cancel can't reach the backend run — worth confirming that's covered by the same fix.

joaquincasal (changes requested):
Thanks for chasing this down — the git archaeology backs up the theory. For context on *why* the code was lost: this wasn't a merge conflict. #11092 was based on Jul-21 master, which already had #11046 (and #11068/#11075/#11089/#11109). The legacy files (`ModalOrchestratorLegacy.tsx`, `useWorkflowAgentLegacy.ts`, `PageLegacy.tsx`) were hand-written fresh from a pre-#11046 snapshot of the old single orchestrator, so the legacy path can be missing **any** improvement that landed before the split — not just the pre-flight pattern. And since the legacy path is the default (`googleDocsAgentImprovements ?? false`), these regressions ship to most users until the flag flips on.

The pre-flight restoration itself looks structurally complete, but I think there are **two blockers** before this is safe to merge:

**1. The no-selection branch still leaves single-tab/no-image runs stuck** (inline comment below). This is the exact scenario the PR targets, so worth a close look.

**2. `entryBlockGraph` is still dropped on mapping-review cancel** — a second casualty from #11089. `PageLegacy.tsx`'s `handleCancelMappingReview` takes no argument and sends `{ cancelled: true }`, while `PageAsyncRuns.tsx:132` correctly forwards `{ cancelled: true, entryBlockGraph }`. `ReviewPage` calls `onCancelReview(entryBlockGraph)`, but TS parameter-bivariance lets the zero-arg handler satisfy the prop and compile clean, so the edited graph is silently discarded. #11089 specifically changed this exact function to accept the graph; the legacy rewrite reverted it. Net effect: `hil_edited`/`rejected` analytics are wrong on cancel-from-review for default-path users. `ResumePayload.entryBlockGraph?` already exists, so it's a one-line wiring fix.

**Smaller stuff:**
- **Lost logging cleanup:** `useWorkflowAgentLegacy.ts` re-introduces the 4 polling `console.log`s that #11075 removed — more evidence the file was copied from a pre-#11075 snapshot. Worth dropping them again.

Everything else checks out: MISSING_PARAMETER wiring matches async, flag/prop/handle wiring is sound, and #11068/#11075(UI)/#11077/#11051/#11109 are all accounted for in the legacy path.

joaquincasal (approved):

PR #11201, opened 2026-07-29
fix(drive-integration): disable Next button and reduce loading delay after tab/image selection []

david-shibley-contentful:
## Summary
- Reduce `CONTENT_TYPE_SUBMIT_LOADING_DELAY_MS` from 30s to 500ms — the 30s value was a holdover from the old suspend-routing flow and caused the UI to appear frozen after clicking Next on the tabs/images modals
- Add `isLoading` prop to `SelectTabsModal` and `IncludeImagesModal` — disables and shows spinner on the Next button while the workflow is starting, preventing repeated submits
- Track `isWorkflowStarting` state in both `ModalOrchestrator` (async) and `ModalOrchestratorLegacy` (flag-off), passed down as `isLoading`

## Test plan
- [ ] Select a multi-tab document with images, click through tabs → images → Next: button should immediately show spinner and be disabled
- [ ] Verify the loading modal appears after ~500ms (instead of 30s) if the backend is slow
- [ ] Verify fast responses skip the loading modal entirely (no flash)
- [ ] Confirm same behaviour on the async flow (flag ON)

### DIFF
--- a/apps/drive-integration/src/utils/constants/agent.ts
+++ b/apps/drive-integration/src/utils/constants/agent.ts
@@ -8,8 +8,6 @@ const EXTENDED_POLL_TIME_MS = 60 * 60 * 1000; // 60 minutes (google-docs-agent-i
 export const MAX_POLL_ATTEMPTS = Math.floor(MAX_POLL_TIME_MS / POLL_INTERVAL_MS);
 export const EXTENDED_MAX_POLL_ATTEMPTS = Math.floor(EXTENDED_POLL_TIME_MS / POLL_INTERVAL_MS);
 
-export const CONTENT_TYPE_SUBMIT_LOADING_DELAY_MS = 500; // brief delay before showing loading modal to avoid flash on fast responses
-
 // Agents-api writes PENDING_REVIEW status before the suspendPayload metadata flushes.
 // Allow this many consecutive PENDING_REVIEW polls with no suspendPayload before giving up.
 export const MAX_PENDING_REVIEW_MISSING_PAYLOAD_RETRIES = 5; // 5 × 10s = 50s max wait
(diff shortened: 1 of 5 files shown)


apps/drive-integration/src/utils/constants/agent.ts
```diff
@@ -8,7 +8,7 @@ const EXTENDED_POLL_TIME_MS = 60 * 60 * 1000; // 60 minutes (google-docs-agent-i
 export const MAX_POLL_ATTEMPTS = Math.floor(MAX_POLL_TIME_MS / POLL_INTERVAL_MS);
 export const EXTENDED_MAX_POLL_ATTEMPTS = Math.floor(EXTENDED_POLL_TIME_MS / POLL_INTERVAL_MS);
 
-export const CONTENT_TYPE_SUBMIT_LOADING_DELAY_MS = 30000; // 30 seconds to wait for suspend payload
+export const CONTENT_TYPE_SUBMIT_LOADING_DELAY_MS = 500; // brief delay before showing loading modal to avoid flash on fast responses
```
FBanfi: In the old path (`ModalOrchestratorLegacy.tsx`): the setTimeout in startWorkflowWithScope...

We should be able to drop the `setTimeout` safely here now. Since `startWorkflowWithScope` is always going to block for a meaningful amount of time because we start the workflow after the `selectTabs` and `includeImages` modals. There's no fast-path where the loading modal would flash and disappear. 

Because of that we could just call `setFlowStep(FlowStep.LOADING)` directly after `setIsWorkflowStarting(true)` and skip the debounce entirely. 
The isLoading button spinner still covers the brief moment before the modal renders.

Something like this change (removing the timeout):
```
const startWorkflowWithScope = async (
      contentTypeIds: string[],
      documentSelection: DocumentSelection
  ) => {
      setIsWorkflowStarting(true);
      setFlowStep(FlowStep.LOADING);

      try {
        return await startWorkflow(contentTypeIds, documentSelection);
      } finally {
        setIsWorkflowStarting(false);
      }
};
```

FBanfi: This will also let us remove entirely this `CONTENT_TYPE_SUBMIT_LOADING_DELAY_MS` const since it shouldn't be needed any more neither in the old and new flow.

FBanfi (commented):
Nice safe guard for the loading part of the ModalOrchestrator! Left a comment in the `CONTENT_TYPE_SUBMIT_LOADING_DELAY_MS` const change!

apps/drive-integration/src/utils/constants/agent.ts
david-shibley-contentful: okay great! I was curious if we actually needed this or not

FBanfi (approved):
Looks good to me!

PR #11202, opened 2026-07-30
fix: move OpenAI key to secret installation parameter [AIS-295]

tylerpina:
[AIS-295](https://contentful.atlassian.net/browse/AIS-295)

## Summary
- Adds a Contentful Function (`functions/openai-proxy.ts`) that reads the OpenAI API key server-side via `context.appInstallationParameters` and proxies chat completion requests to OpenAI — the key never touches browser-side code
- Wires up a `contentful-app-manifest.json` with an App Action (`openaiProxyAction`) so the Dialog location calls the function via `sdk.cma.appActionCall.createWithResult` instead of calling OpenAI directly
- Removes `key` from `AppInstallationParameters` and all browser-readable state; config screen `APIKey` component becomes write-only (admin can set a new key but can't read the current one back); `Model` dropdown uses a hardcoded list instead of fetching via the key; sidebar key-validation removed

## Test plan
- [ ] `npm test` passes (58/58)
- [ ] `npm run lint` passes (0 errors, 0 warnings)
- [ ] `npx tsc --noEmit` clean
- [ ] `npm run build:functions` emits `build/functions/openai-proxy.js`
- [ ] **Manual — Task 6:** Update AppDefinition (`3iszK8Gl7aaxLvxGyCOhgA` prod, `7sjBFUzBxj1fRrI0eUBwLc` test) to declare `key` as `"type": "Secret"` via CMA or Contentful UI
- [ ] **Manual:** After AppDefinition update, verify `sdk.parameters.installation.key` returns `***` in browser console
- [ ] **Manual:** Deploy function bundle (`npm run deploy:test`) and confirm App Action invocation succeeds end-to-end in a test space
- [ ] **Manual — Secret preservation:** With a key already stored, leave the API-key field **blank**, change the model / brand profile, and save. Confirm generation still works afterward — the omitted key must preserve the stored Secret, not clear it.
- [ ] **Manual — nested→flat migration:** On a space installed *before* this PR (nested `brandProfile`), open config without re-saving and confirm brand-profile fields render populated, and that generation still includes brand context (dual-read fallback). Then re-save and confirm the params flatten.

## Notes
- This PR is the code half of the fix. The AppDefinition `Secret` type declaration (Task 6 in the plan) must be applied separately before deploy — without it, the key is still plaintext in the browser even after this code lands.
- UX change: generation is no longer streamed token-by-token; the full response is revealed at once. Streaming was architecturally incompatible with Contentful Functions' single-response model. During the wait, the generating state now shows a Forma 36 skeleton loader (grey pulsing text lines) in place of the old empty textarea.
- The **Stop Generating** button now cancels an in-flight generation client-side: the App Action is raced against an `AbortController` signal, any late-arriving response is discarded, and the UI returns to the input view (a user stop is treated as a cancellation, not an error). The proxy call itself can't be terminated server-side — App Actions can't signal a running invocation and OpenAI's chat-completions API has no cancel endpoint — so the OpenAI request still completes/bills in the background.
- Migration: brand-profile fields moved from a nested `brandProfile` object to flat top-level params (required for the `additionalProperties: false` Secret schema). A shared `toProfileType(installation)` helper dual-reads — flat first, nested fallback — so installs from before this PR keep their brand profile in both config and generation until an admin re-saves, which flattens them in place. Top-level `profile` was already flat.
- Cleanup: removed the unreachable 429 error branch in `GeneratedTextPanel` (the proxy collapses all OpenAI failures into a generic `Error` with no status code, so the branch could never match — 429s now show the default generate error) and its orphaned deps, plus the unused `ParametersMissingWarning` component.

[AIS-295]: https://contentful.atlassian.net/browse/AIS-295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ai-content-generator/src/hooks/dialog/useAI.tsx
```diff
@@ -1,124 +1,129 @@
...
+        [ProfileFields.VALUES]: installation.values,
+        [ProfileFields.TONE]: installation.tone,
+        [ProfileFields.EXCLUDE]: installation.exclude,
+        [ProfileFields.INCLUDE]: installation.include,
+        [ProfileFields.AUDIENCE]: installation.audience,
+        [ProfileFields.ADDITIONAL]: installation.additional,
+      };
+      const messages = createGPTPayload(prompt, profile, targetLocale);
+
+      const actionCall = sdk.cma.appActionCall.createWithResponse(
```
jjolton-contentful: `createWithResponse` is a legacy mechanism for making app action calls that generates a bunch of red-herring 404s that can often mislead users poking around the dev tools console. `createWithResult` is the more modern version that comes with a few niceties (like better result/error typing) in addition to doing away with the nasty 404 polling loop (see https://github.com/contentful/apps/blob/0849cf51c2b472330e54e39dcbc228926b1d1623/apps/drive-integration/src/utils/appAction.ts#L29 for example usage)

apps/ai-content-generator/src/components/config/parameterReducer.ts
```diff
@@ -114,46 +87,26 @@ const parameterReducer = (
...
-          },
-          audience: {
-            value: parameter.brandProfile?.audience || '',
-            isValid: true,
-          },
-          additional: {
-            value: parameter.brandProfile?.additional || '',
-            isValid: true,
-          },
+          values: { value: parameter.values || '', isValid: true },
```
jjolton-contentful: **blocker:** this only rehydrates from the new flat shape (`parameter.values`, `parameter.tone`, …).

Existing installs still have nested `brandProfile: { values, tone, … }`. Until an admin re-saves under the flattened schema, the config UI and dialog prompts will see empty brand-profile fields. (Top-level `profile` was already flat and is fine.)

When Secret defs land with `additionalProperties: false`, every flat field in the save payload also needs a matching AppDefinition Symbol — otherwise that save will fail even after dual-read is in place.

Please dual-read on load — prefer flat fields, fall back to `parameter.brandProfile?.…` — and flatten on the next save so we migrate in place.

apps/ai-content-generator/src/hooks/dialog/useAI.tsx
```diff
@@ -1,124 +1,129 @@
...
     try {
-      const payload = createGPTPayload(
-        prompt,
+      const installation = sdk.parameters.installation;
+      const model = installation.model ?? defaultModelId;
+      // Installation params are persisted flat; reassemble the ProfileType the
+      // prompt builder expects.
+      const profile: ProfileType = {
+        [ProfileFields.PROFILE]: installation.profile,
+        [ProfileFields.VALUES]: installation.values,
```
jjolton-contentful: Same migration concern as the config reducer: `installation.values` / `tone` / etc. won't exist on currently-installed spaces (they're under `installation.brandProfile`). Generation will quietly drop brand-profile context for existing customers until they re-save config. Top-level `profile` is already fine.

Dual-read here (or a shared `toProfileType(installation)` helper used by config + dialog + sidebar) would close the gap.

apps/ai-content-generator/package.json
```diff
@@ -32,13 +34,14 @@
...
-    "build": "tsc && vite build --mode ${NODE_ENV:='production'}",
+    "build": "tsc && vite build --mode ${NODE_ENV:='production'} && npm run build:functions && cp -r build/functions dist/functions",
+    "build:functions": "contentful-app-scripts build-functions --ci",
     "test": "vitest --logHeapUsage --coverage",
     "test:ci": "vitest",
     "lint": "eslint src --max-warnings 0",
     "create-app-definition": "contentful-app-scripts create-app-definition",
-    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEFINITIONS_ORG_ID} --definition-id 3iszK8Gl7aaxLvxGyCOhgA --token ${CONTENTFUL_CMA_TOKEN}",
-    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEV_TESTING_ORG_ID} --definition-id 7sjBFUzBxj1fRrI0eUBwLc --token ${TEST_CMA_TOKEN}"
+    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEFINITIONS_ORG_ID} --definition-id ${DEFINITIONS_APP_DEF_ID} --token ${CONTENTFUL_CMA_TOKEN} && npm run upsert-actions",
```
jjolton-contentful: **blocker for deploy:**

1. `deploy:test` was removed. Root CI runs `lerna run deploy:test`, so this package will just be skipped on staging deploys.
2. Prod `deploy` swapped the hardcoded def ID (`3iszK8Gl7aaxLvxGyCOhgA`) for `${DEFINITIONS_APP_DEF_ID}`. That env var isn't used elsewhere in this monorepo (peer apps hardcode their definition IDs), so unless CircleCI already injects it, upload/upsert-actions will get an empty definition id.

Please restore `deploy:test` and keep the known def IDs (`3iszK8Gl7aaxLvxGyCOhgA` prod, `7sjBFUzBxj1fRrI0eUBwLc` test), or land the matching CI/env wiring in the same PR.

apps/ai-content-generator/src/components/config/config-page/ConfigPage.tsx
```diff
@@ -37,45 +29,43 @@ const initialContentTypes: Set<string> = new Set();
...
 
-  const validateParams = (): string[] => {
-    const notifierErrors = [];
-
-    if (!parameters.key.isValid) {
-      notifierErrors.push(`${ConfigErrors.failedToSave} ${ConfigErrors.missingApiKey}`);
+    // The key is write-only: only persist it when the admin enters a new value,
+    // otherwise omit it so we don't clobber the stored Secret with an empty
+    // string (the field placeholder promises "leave blank to keep existing").
+    if (keyInput) {
```
jjolton-contentful, line 61: Omit-when-blank is the right instinct for Secret params. Docs don't spell out "omit = preserve" as explicitly as I'd like though — can you add a manual test plan item that confirms: leave key blank, change model/brand profile, save → generation still works with the previously stored secret?

If omit actually clears, every non-key config save becomes a silent credential wipe.

jjolton-contentful (changes requested):
Nice direction on the SAST fix — getting the OpenAI key out of the browser and behind a Function/App Action is the right shape.

Already left a note on `createWithResponse` → `createWithResult`. A few more things I'd want addressed (or explicitly confirmed) before this ships:

### Blockers
1. **Nested → flat install-param migration** — existing installs still have `brandProfile: { … }` nested. New read paths only look at top-level flat fields, so brand profile silently empties until an admin re-saves. (Top-level `profile` was already flat.) Needs dual-read (nested **or** flat) on load + flatten on save. When Secret defs land with `additionalProperties: false`, every flat field in the save payload also needs a matching AppDefinition Symbol.
2. **`deploy:test` removed + `${DEFINITIONS_APP_DEF_ID}`** — root CI is `lerna run deploy:test` / `lerna run deploy`. Dropping `deploy:test` means this app gets skipped on staging; prod now depends on an env var that used to be a hardcoded def ID (`3iszK8Gl7aaxLvxGyCOhgA` prod / `7sjBFUzBxj1fRrI0eUBwLc` test). Please restore `deploy:test` (and keep/prod-wire the def IDs) or land the matching CI/env change in the same PR.
3. **Secret omit-on-save** — omitting blank `key` so we don't clobber Secrets looks right, but worth an explicit manual AC: save model/brand profile with the key field blank → prior secret still works for generation.

### Non-blocking
- First install can now save with no key at all (credential validation was removed) — generation just fails later. Curious if we want a soft "key required on first save" check when no secret is present yet.
- No unit coverage on `openai-proxy.ts` itself.

Happy to pair on the dual-read path if useful.

tylerpina (commented):
Fixed — bumped `@contentful/app-sdk` to 4.68.1 (the version that added `createWithResult` to the CMA surface) and swapped the call. Response is now read from `result.sys.result` instead of `JSON.parse(response.response.body)`.

apps/ai-content-generator/src/components/config/parameterReducer.ts
tylerpina: Fixed. Added a shared `toProfileType(installation)` helper in `appInstallationParameters.ts` that dual-reads — flat top-level fields first, legacy nested `brandProfile` fallback. The reducer's `APPLY_CONTENTFUL_PARAMETERS` now rehydrates through it, so existing installs keep their brand profile until an admin re-saves (which flattens in place). Re: `additionalProperties: false` — every flat field in the save payload already has a matching Symbol on the AppDefinition (that's Task 6 in the plan); noted the requirement in the type's doc comment. Unit coverage added in `appInstallationParameters.spec.ts`.

apps/ai-content-generator/src/hooks/dialog/useAI.tsx
tylerpina: Fixed via the same `toProfileType(installation)` helper you suggested — `useAI` now builds the `ProfileType` through it instead of reading top-level fields directly, so pre-migration installs keep brand context in generated prompts. Config and dialog now read through the one helper.

apps/ai-content-generator/package.json
tylerpina: Fixed. Restored `deploy:test` and put the hardcoded def IDs back (`3iszK8Gl7aaxLvxGyCOhgA` prod, `7sjBFUzBxj1fRrI0eUBwLc` test) — dropped `${DEFINITIONS_APP_DEF_ID}` since it isn't wired in CI. Both deploy targets now chain `upsert-actions` (prod → `upsert-actions`, test → `upsert-actions:test`) so the App Action gets upserted in staging and prod.

apps/ai-content-generator/src/components/config/config-page/ConfigPage.tsx
tylerpina: Added a manual AC to the test plan: with a key already stored, leave the field blank, change model/brand profile, save → confirm generation still works (omit must preserve, not clear, the Secret). Also added a migration AC covering the dual-read fallback on pre-PR installs.

tylerpina:
Thanks for the thorough pass, @jjolton-contentful — pushed fixes for all three blockers plus the two flagged items. Ready for another look.

### Blockers
1. **Nested → flat migration** — added a shared `toProfileType(installation)` helper that dual-reads: flat top-level fields first, legacy nested `brandProfile` fallback. Wired into both `parameterReducer` (`APPLY_CONTENTFUL_PARAMETERS`) and `useAI`, so config and dialog read the same way and existing installs keep brand context until an admin re-saves (which flattens in place). Every flat field already has a matching Symbol on the AppDefinition (Task 6), noted in the type's doc comment.
2. **`deploy:test` + `${DEFINITIONS_APP_DEF_ID}`** — restored `deploy:test` and put the hardcoded def IDs back (`3iszK8Gl7aaxLvxGyCOhgA` prod / `7sjBFUzBxj1fRrI0eUBwLc` test); dropped the unwired env var. Both deploy targets now chain `upsert-actions` (prod → `upsert-actions`, test → `upsert-actions:test`) so the App Action upserts in staging and prod. Verified end-to-end against the test def.
3. **Secret omit-on-save** — added a manual AC to the test plan confirming a blank key preserves the stored Secret, plus a migration AC for the dual-read fallback.

### Non-blocking (both now done, commit `3f2f7175a`)
- **First-install "key required" check** — `ConfigPage` reads `sdk.parameters.installation?.key` to detect whether a Secret is already stored. `validateParams` blocks a save only when there's no stored key *and* no entered key, so first installs must supply a key while re-saves with a blank field still preserve the existing Secret (the key is write-only — we check the masked param's presence, never its value).
- **`openai-proxy.ts` unit coverage** — new `functions/openai-proxy.spec.ts` (6 tests): missing key throws, invalid `messages` JSON throws, success returns `{ text }`, empty completion returns `{ text: '' }`, non-ok response surfaces the OpenAI error-body message, and falls back to `statusText` when the body has none. Added 2 `ConfigPage` tests for the soft-check via the captured `onConfigure` callback. 66/66 tests + lint + tsc green.

Also confirmed via the App Framework docs + extensibility-api that stored plaintext values survive the flip to `Secret`, Functions receive them unmasked via `context.appInstallationParameters`, and `additionalProperties: false` is write-only (reads of pre-migration installs are unaffected) — so existing customers keep working with no re-entry.

jjolton-contentful (approved):

PR #11203, opened 2026-07-31
fix: move AWS Bedrock IAM credentials to secret installation parameters [AIS-296]

tylerpina:
[AIS-296](https://contentful.atlassian.net/browse/AIS-296)

## Summary
- Adds a Contentful Function (`functions/bedrock-proxy.ts`) that reads `accessKeyId`, `secretAccessKey`, and `region` server-side via `context.appInstallationParameters` and proxies InvokeModel requests to AWS Bedrock using AWS SigV4 signing — credentials never touch browser-side code
- Wires up `contentful-app-manifest.json` with an App Action (`bedrockProxyAction`) so the Dialog location calls the function via `sdk.cma.appActionCall.createWithResult` instead of instantiating `BedrockClient`/`BedrockRuntimeClient` in the browser
- Deletes the browser-side `src/utils/aiApi` client (the actual SAST finding) and removes `accessKeyId`/`secretAccessKey` from browser-readable state; the config `AccessKey` component becomes write-only (admin can set new credentials but can't read stored values back); `Model` becomes a static list rather than live-validating against AWS with the credentials
- Flattens the installation parameters to top-level scalars so a `Secret`-typed param validates (a `Secret` opts the whole params object into strict scalar-only validation — nested `brandProfile` is flattened, `enabledFeatures` is JSON-encoded into a Symbol string and rehydrated in the reducer/hooks)
- Migration is backward-compatible via two shared dual-read helpers in `appInstallationParameters.ts`, used by the reducer, dialog (`useAI`), and sidebar so all three read the same way: `toProfileType(installation)` reads the brand profile flat-first with a nested `brandProfile` fallback, and `parseEnabledFeatures(value)` accepts both a JSON-encoded Symbol string and a real `AIFeature[]` array. Existing installs keep their brand profile and feature selection until an admin re-saves (which flattens in place); top-level `profile` was already flat

## Notes
- This PR is the code half of the fix. The AppDefinition `Secret` type declaration for `accessKeyId` and `secretAccessKey` must be applied separately before deploy — without it, the credentials are still plaintext in the browser even after this code lands.
- **SigV4 signing service name is `bedrock`, not `bedrock-runtime`** — the latter is only the endpoint host prefix. Signing with the wrong name yields `InvalidSignatureException` on every call. Covered by a regression test in `functions/bedrock-proxy.spec.ts`.
- UX change: generation is no longer streamed token-by-token; the full response is revealed at once (a skeleton loader shows while generating). Streaming was architecturally incompatible with Contentful Functions' single-response model.
- Model availability checking (which called AWS `ListFoundationModels` + `InvokeModelCommand` using the browser-side credentials) is removed. The model list is now static.
- **On merge:** update the status column for row 273 in the Cypress SAST findings spreadsheet and keep the Jira ticket in sync.

## Test plan
- [ ] `npx vitest run` passes (65/65)
- [ ] `npm run lint` passes (0 errors, 0 warnings)
- [ ] `npx tsc --noEmit` clean
- [ ] `npm run build:functions` emits `build/functions/bedrock-proxy.js`
- [ ] **Manual — AppDefinition update**: Declare `accessKeyId` and `secretAccessKey` as `"type": "Secret"` (and remaining installation fields as `Symbol`) via CMA or Contentful UI for the Bedrock app definition
- [ ] **Manual**: After AppDefinition update, verify `sdk.parameters.installation.accessKeyId` returns `***` in browser console
- [ ] **Manual**: Deploy function bundle (`npm run deploy:test`) and confirm App Action invocation succeeds end-to-end in a test space
- [ ] **Manual — Secret preservation**: With credentials already stored, leave both credential fields **blank**, change model / region / brand profile, save. Confirm generation still works afterward — the omitted credentials must preserve the stored Secrets, not clear them.
- [ ] **Manual — nested→flat migration**: On a space installed *before* this PR (nested `brandProfile`, `enabledFeatures` as an array), open config without re-saving and confirm brand-profile fields render populated and the correct features are selected (not defaulted to "all"); confirm generation still includes brand context. Then re-save and confirm the params flatten.

[AIS-296]: https://contentful.atlassian.net/browse/AIS-296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/bedrock-content-generator/src/hooks/dialog/useAI.tsx
```diff
@@ -1,92 +1,68 @@
...
+      const installation = sdk.parameters.installation;
+      const { model: modelId, region } = installation;
+      const featuredModel = featuredModels.find((m) => m.id === modelId);
+      const invokeId = featuredModel?.getInvokeId
+        ? featuredModel.getInvokeId(region)
+        : modelId ?? featuredModels[0].id;
+
+      const systemPrompt = baseSystemPrompt(installation, targetLocale);
+
+      const response = await sdk.cma.appActionCall.createWithResponse(
```
jjolton-contentful: `createWithResponse` is a legacy mechanism for making app action calls that generates a bunch of red-herring 404s that can often mislead users poking around the dev tools console. `createWithResult` is the more modern version that comes with a few niceties (like better result/error typing) in addition to doing away with the nasty 404 polling loop (see https://github.com/contentful/apps/blob/0849cf51c2b472330e54e39dcbc228926b1d1623/apps/drive-integration/src/utils/appAction.ts#L29 for example usage)

apps/bedrock-content-generator/src/components/config/parameterReducer.ts
```diff
@@ -147,53 +149,46 @@ const parameterReducer = (
...
-          },
-          audience: {
-            value: parameter.brandProfile?.audience || '',
-            isValid: true,
-          },
-          additional: {
-            value: parameter.brandProfile?.additional || '',
-            isValid: true,
-          },
+          values: { value: parameter.values || '', isValid: true },
```
jjolton-contentful: **blocker:** this only rehydrates from the new flat shape (`parameter.values`, `parameter.tone`, …) and JSON-encoded `enabledFeatures`.

Existing installs still have:
- nested `brandProfile: { values, tone, … }`
- `enabledFeatures` as a real array (not a JSON string)

Until an admin re-saves, the config UI loses brand-profile fields and feature selection falls back to "all features". (Top-level `profile` was already flat and is fine.)

When Secret defs land with `additionalProperties: false`, every flat field in the save payload also needs a matching AppDefinition Symbol — otherwise that save will fail even after dual-read is in place.

Please dual-read on load (flat first, fall back to nested/`Array.isArray`) and flatten on save so we migrate in place.

apps/bedrock-content-generator/src/hooks/dialog/useAI.tsx
```diff
@@ -1,92 +1,68 @@
...
     try {
-      const systemPrompt = baseSystemPrompt(
+      const installation = sdk.parameters.installation;
+      const { model: modelId, region } = installation;
+      const featuredModel = featuredModels.find((m) => m.id === modelId);
+      const invokeId = featuredModel?.getInvokeId
+        ? featuredModel.getInvokeId(region)
+        : modelId ?? featuredModels[0].id;
+
+      const systemPrompt = baseSystemPrompt(installation, targetLocale);
```
jjolton-contentful: `baseSystemPrompt` expects a `ProfileType` (`profile` / `values` / `tone` / …). Passing the whole flat `installation` works **after** migration, but on currently-installed spaces the brand-profile fields still live under `installation.brandProfile` — so existing customers will quietly lose brand context in generated prompts. (Top-level `profile` is already fine.)

A shared `toProfileType(installation)` that dual-reads nested vs flat would keep dialog/config/sidebar in sync.

apps/bedrock-content-generator/src/hooks/sidebar/useSidebarParameters.ts
```diff
@@ -1,29 +1,39 @@
...
-  const sdk = useSDK<SidebarAppSDK<AppInstallationParameters>>();
-  const { profile, enabledFeatures } = sdk.parameters.installation;
+  const sdk = useSDK<SidebarAppSDK<PersistedInstallationParameters>>();
+  const installation = sdk.parameters.installation;
+  const profile = installation.profile;
+  // enabledFeatures is JSON-encoded as a Symbol string in the persisted shape.
+  let parsedFeatures: AIFeature[] | undefined;
+  try {
+    parsedFeatures = installation.enabledFeatures
+      ? JSON.parse(installation.enabledFeatures)
```
jjolton-contentful: Existing installs store `enabledFeatures` as `AIFeature[]`, not a JSON string. `JSON.parse` on an array will throw (caught → `undefined`) and you'll fall through to "all features", which silently ignores the customer's actual selection until they re-save.

Please accept both shapes: `Array.isArray(x) ? x : JSON.parse(x)`.

apps/bedrock-content-generator/package.json
```diff
@@ -27,13 +28,14 @@
...
-    "build": "tsc && vite build --mode ${NODE_ENV:='production'}",
+    "build": "tsc && vite build --mode ${NODE_ENV:='production'} && npm run build:functions && cp -r build/functions dist/functions",
     "test": "vitest --logHeapUsage --coverage",
     "test:ci": "vitest",
     "lint": "eslint src --max-warnings 0",
+    "build:functions": "contentful-app-scripts build-functions --ci",
     "create-app-definition": "contentful-app-scripts create-app-definition",
-    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEFINITIONS_ORG_ID} --definition-id jTed08LvhkveeUoBpJOim --token ${CONTENTFUL_CMA_TOKEN}",
-    "deploy:test": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEV_TESTING_ORG_ID} --definition-id 63AGu5jLZaruxYGLmPrWSm --token ${TEST_CMA_TOKEN}"
+    "deploy": "contentful-app-scripts upload --ci --bundle-dir ./dist --organization-id ${DEFINITIONS_ORG_ID} --definition-id ${DEFINITIONS_APP_DEF_ID} --token ${CONTENTFUL_CMA_TOKEN} && npm run upsert-actions",
```
jjolton-contentful: **blocker for deploy:**

1. `deploy:test` was removed. Root CI runs `lerna run deploy:test`, so this package will be skipped on staging.
2. Prod `deploy` swapped the hardcoded def ID (`jTed08LvhkveeUoBpJOim`) for `${DEFINITIONS_APP_DEF_ID}`. That env var isn't used elsewhere in this monorepo (peer apps hardcode their definition IDs), so unless CircleCI already injects it, upload/upsert-actions get an empty definition id.

Please restore `deploy:test` / keep the known IDs (`jTed08LvhkveeUoBpJOim` prod, `63AGu5jLZaruxYGLmPrWSm` test), or land the matching CI/env wiring in the same PR.

apps/bedrock-content-generator/contentful-app-manifest.json
```diff
@@ -0,0 +1,46 @@
+{
+  "functions": [
+    {
+      "id": "bedrockProxyFunction",
+      "name": "Bedrock proxy function",
+      "description": "Calls AWS Bedrock using the secret IAM credentials stored server-side.",
+      "path": "functions/bedrock-proxy.js",
+      "entryFile": "functions/bedrock-proxy.ts",
+      "allowNetworks": [
+        "https://*.amazonaws.com"
```
jjolton-contentful: Non-blocking hardening: this allowlist is broader than the proxy needs — we only call `bedrock-runtime.${region}.amazonaws.com`.

Caveat from `functions-api` outbound matching (`assertUrlIsAllowed`): a leading `*.` means "any host under this registrable domain," not a mid-label glob. So `https://bedrock-runtime.*.amazonaws.com` still matches all of `amazonaws.com` (same as `https://*.amazonaws.com`).

To actually tighten, list exact region hosts (hostname equality), e.g. `bedrock-runtime.us-east-1.amazonaws.com`, … for the regions we support. Otherwise leaving `*.amazonaws.com` is fine and honest about the matcher.

apps/bedrock-content-generator/src/components/config/config-page/ConfigPage.tsx
```diff
@@ -50,46 +52,49 @@ const initialContentTypes: Set<string> = new Set();
...
+      values: parameters.brandProfile.values?.value,
+      // JSON-encode the array since Symbol params are scalar-only.
+      enabledFeatures: JSON.stringify(
+        parameters.enabledFeatures?.value || (Object.keys(featureConfig) as AIFeature[])
+      ),
     };
+
+    // Credentials are write-only: only persist when the admin enters new values,
+    // otherwise omit so we don't clobber stored Secrets with empty strings.
+    if (accessKeyIdInput) params.accessKeyId = accessKeyIdInput;
```
jjolton-contentful, line 81: Omit-when-blank is the right instinct for Secret params. Can you add a manual AC that confirms: leave both credential fields blank, change model/region/brand profile, save → generation still works with the previously stored Secrets?

If omit actually clears, every non-credential config save becomes a silent wipe.

jjolton-contentful (changes requested):
Nice work getting the IAM creds out of the browser — Function + Secret install params is the right fix for the SAST finding, and the SigV4 `bedrock` vs `bedrock-runtime` regression test is a great catch.

Already left a note on `createWithResponse` → `createWithResult`. A few more things before this ships:

### Blockers
1. **Nested → flat install-param migration** — existing installs still have nested `brandProfile` and array `enabledFeatures`. New read paths only look at the flat/JSON-string shape, so brand profile empties and feature toggles fall back to defaults until an admin re-saves. (Top-level `profile` was already flat.) Dual-read on load + flatten on save please. When Secret defs land with `additionalProperties: false`, every flat field in the save payload also needs a matching AppDefinition Symbol.
2. **`deploy:test` removed + `${DEFINITIONS_APP_DEF_ID}`** — same as the OpenAI twin: root CI is `lerna run deploy:test` / `deploy`, so staging will skip this app and prod now depends on an env var that used to be a hardcoded def ID (`jTed08LvhkveeUoBpJOim` prod / `63AGu5jLZaruxYGLmPrWSm` test).
3. **Secret omit-on-save** — same ask as OpenAI: explicit manual AC that blank credential fields preserve the stored Secrets across an unrelated config save.

### Non-blocking
- **`allowNetworks: https://*.amazonaws.com`** — wider than the proxy needs. `functions-api` wildcards are domain-scoped (`*.amazonaws.com` ≈ all AWS), not mid-label; real tightening means listing exact `bedrock-runtime.<region>.amazonaws.com` hosts (or leave as-is).
- First install can save with no credentials (validation removed).
- `@aws-sdk/client-bedrock*` look unused now that the browser client is gone — nice cleanup if you're already in `package.json`.

Happy to pair on the dual-read helper if useful.

tylerpina (commented):
Fixed — bumped  to 4.68.1 (the version that added `createWithResult` to the CMA surface) and swapped the call. Response is now read from `result.sys.result` instead of `JSON.parse(response.response.body)`.

tylerpina:
Thanks @jjolton-contentful — pushed fixes for both blockers plus the two hardening notes and the Secret AC. Ready for another look.

### Blockers
1. **Nested → flat migration (brand profile + `enabledFeatures`)** — added two shared dual-read helpers in `appInstallationParameters.ts`, used by the reducer, dialog (`useAI`), and sidebar so all three read the same way:
   - `toProfileType(installation)` — brand profile flat-first, nested `brandProfile` fallback.
   - `parseEnabledFeatures(value)` — accepts both a JSON-encoded Symbol string (new schema) and a real `AIFeature[]` (pre-migration), per your `Array.isArray(x) ? x : JSON.parse(x)` suggestion, falling back to all features only when absent/unparseable.

   Existing installs keep brand context and feature selection until an admin re-saves.
2. **`deploy:test` + `${DEFINITIONS_APP_DEF_ID}`** — restored `deploy:test`, put the hardcoded def IDs back (`jTed08LvhkveeUoBpJOim` prod / `63AGu5jLZaruxYGLmPrWSm` test), and chained `upsert-actions`/`upsert-actions:test` into both deploy targets. Verified end-to-end against the test def.

### Hardening / non-blocking
- **`allowNetworks`** — tightened `https://*.amazonaws.com` to the exact `bedrock-runtime.<region>.amazonaws.com` hosts for the supported regions (`us-east-1`, `us-west-2`, `eu-central-1`), matching the host the proxy actually builds.
- **Secret omit-on-save** — added a manual AC confirming blank credentials preserve the stored Secrets, plus a migration AC for the dual-read fallback.

Added unit coverage for both helpers; full suite green (65/65).

jjolton-contentful (approved):

PR #11261, opened 2026-08-04
fix: fixes repositoryId issue and adds selection type of single or multiple []

MikeDReynolds68:
## Purpose

This update adds following capabilities:

- The `repositoryId` property was missing from the Adobe Content Advisor config properties and has been added so it can be used for restricting user access to a single repository
- The `selectedAssets` property has been added to the Content Advisor config properties to persist previously selected assets and uses the asset ids from the current field value if the `prefillSelectedAssets` config value is 'Yes'
- The `CustomUpdateStateValueFn` has been added to modify the save functionality so the app replaces the existing field values with the new selections if the `prefillSelectedAssets` config value is 'Yes' and adds the new selections to the previously selected assets if the value is 'No'
- The `prefillSelectedAssets` config has been added to support the `selectedAssets` property in the Content Advisor props and the `CustomUpdateStateValueFn`
- The `selectionType` property has been added to the Content Advisor props to support the creation of asset fields that contain a 'single' asset or 'multiple' assets (the selectionType property is populated from the value of a new instance config property of the same name with the following options: single, multiple
- The `aemTierType` config property has been changed from text to a dropdown to prevent invalid values and contains the following options: delivery, author
- The `env` config property has been changed from text to a dropdown to prevent invalid values and contains the following options: prod, stage

## Approach

The changes were made in accordance with the capabilities provided by the `contentful/dam-app-base` package, the Adobe Content Advisor front-end-as-a-service app, and best practices for installation and instance config properties. 

## Testing steps

The following tests should be passed:
- Providing the `repositoryId` in the application's installation config restricts users to the identified repository and no others
- Omitting the `repositoryId` in the application's installation config allows users to see all repositories available to them based on their AEM Assets permissions
- Setting `prefillSelectedAssets` to 'Yes' selects all previously selected assets when opening the Adobe Content Advisor modal
- Setting `prefillSelectedAssets` to 'Yes' causes the previously selected assets to be replaced with the assets selected in the Adobe Content Advisor modal
- Setting `prefillSelectedAssets` to 'No' leaves the previously selected assets unselected when opening the Adobe Content Advisor modal
- Setting `prefillSelectedAssets` to 'No' cause the assets selected in the Adobe Content Advisor modal to be added to the previously selected assets
- Setting `selectionType` to 'single' restricts the user to the selection of a single asset in the Adobe Content Advisor modal
- Setting `selectionType` to 'multiple' allows the user to select multiple assets in the Adobe Content Advisor modal
- Setting `aemTierType` to 'delivery' or 'stage' restricts the user to repositories of the selected type
- Setting `env` to 'prod' or 'stage' restricts the user to repositories in the selected environment

## Breaking Changes

None

All changes are based on information collected in zoom calls, email or slack

## Deployment

None that I am aware of

### DIFF
--- a/apps/aem-assets/src/index.jsx
+++ b/apps/aem-assets/src/index.jsx
@@ -40,8 +40,8 @@ export function makeThumbnail(asset) {
 
 async function openDialog(sdk, _currentValue, _config) {
   const parameters = { ..._config, ...sdk.parameters.instance };
-  // const assetIds = _currentValue.map((asset) => ({ id: asset.id }));
-  // parameters.selectedAssets = assetIds;
+  const assetIds = _currentValue.map((asset) => ({ id: asset.id }));
+  parameters.selectedAssets = assetIds;
 
   const result = await sdk.dialogs.openCurrentApp({
     position: 'center',
@@ -189,8 +189,18 @@ if (ENABLE_POPUP_AUTH_REDIRECT_FIX && !isEmbeddedInIframe()) {
 
 async function renderDialog(sdk) {
   const config = sdk.parameters.invocation;
-  const { imsClientId, imsOrg, repositoryId, aemTierType, env, hideUploadButton, hideTreeNav } =
-    config;
+  const {
+    imsClientId,
+    imsOrg,
+    repositoryId,
+    aemTierType,
+    env,
+    hideUploadButton,
+    prefillSelectedAssets,
+    selectedAssets,
+    hideTreeNav,
+    selectionType,
+  } = config;
 
   if (ENABLE_POPUP_AUTH_REDIRECT_FIX) {
     persistImsConfig({ imsClientId, imsOrg });
@@ -250,10 +260,12 @@ async function renderDialog(sdk) {
 
   const contentAdvisorProps = {
     imsOrg,
+    repositoryId,
     aemTierType: aemTierType ? aemTierType.split(',') : ['delivery', 'author'],
-    env,
+    env: env === 'stage' ? 'stage' : undefined,
     hideTreeNav,
-    selectionType: 'multiple',
+    selectedAssets: selectedAssets && Array.isArray(selectedAssets) ? selectedAssets : [],
+    selectionType,
     uploadConfig: {
       hideUploadButton: hideUploadButton === 'Yes' ? true : false,
     },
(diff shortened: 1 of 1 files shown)


apps/aem-assets/src/index.jsx
```diff
@@ -250,10 +260,15 @@ async function renderDialog(sdk) {
 
   const contentAdvisorProps = {
     imsOrg,
+    repositoryId,
     aemTierType: aemTierType ? aemTierType.split(',') : ['delivery', 'author'],
-    env,
+    env: env === 'stage' ? 'stage' : undefined,
     hideTreeNav,
-    selectionType: 'multiple',
```
whitelisab, line 256: It looks like `selectionType` was previously hardcoded to `multiple`, but now we are accessing it from invocation parameters, however I don't see `selectionType` as a defined parameter. Where is the value for this being set now?

apps/aem-assets/src/index.jsx
```diff
@@ -342,17 +373,30 @@ setup({
...
-      description: 'Specifies the AEM repository environment [prod,stage] for the app',
-      required: false,
+      type: 'List',
+      value: 'prod,stage',
+      default: 'prod',
+      description:
+        'Specifies the AEM repository environment for the app (defaults to prod if no selection made)',
+    },
+    {
+      id: 'prefillSelectedAssets',
```
whitelisab, line 393: It might be helpful to add this to the configuration section of the README where the other configuration parameters are mentioned.

apps/aem-assets/src/index.jsx
```diff
@@ -342,17 +373,30 @@ setup({
     {
       id: 'aemTierType',
       name: 'AEM Tier',
-      type: 'Symbol',
-      description: 'Specifies the tier type [delivery, author] for the app (defaults to both)',
+      type: 'List',
```
whitelisab, line 376: It looks like this changes the behavior of this field from being able to define both `delivery,author` to being a single select, but the description says that it still defaults to delivery and author. Can you confirm that this change matches the behavior you are seeking?

whitelisab (commented):
The `repositoryId` fix looks good, I just had a few small questions on this PR

apps/aem-assets/src/index.jsx
MikeDReynolds68: @whitelisab The selectionType is an instance parameter that needs to be set on the fields on which the connector is being used.

apps/aem-assets/src/index.jsx
MikeDReynolds68: @whitelisab I will add to the README and submit a separate PR.

apps/aem-assets/src/index.jsx
MikeDReynolds68: @whitelisab I will update this with the README update.

whitelisab (approved):

PR #11271, opened 2026-08-12
feat: add audit-log-viewer example app []

andrewwestgard-tech:
Adds the Audit Log Viewer as a self-hosted reference app under examples/audit-log-viewer. The app reads Contentful audit log files from a customer-owned cloud storage destination (AWS S3, Azure Blob Storage, or Google Cloud Storage) and renders them in a filterable, paginated table with charts — all from inside the Contentful web app.

A Contentful-hosted App Action Function holds cloud credentials as Secret installation parameters and generates short-lived signed URLs; the browser never sees the credentials directly.


mgoudy91 (approved):
Code looks good. Acknowledging that @zacharyyankiver and I chatted about the information shown in the log and that it's acceptable with the disclosure

PR #11274, opened 2026-08-13
fix: move audit-log-viewer from examples/ to apps/, wire prod deploy []

mgoudy91:
## Summary
- Moves `audit-log-viewer` from `examples/` to `apps/` — lerna's package glob is `apps/**/*` (see `lerna.json`), so the app was invisible to build/test/lint/deploy in #11271.
- Renames the package from `audit-log-viewer-example` to `audit-log-viewer` since it's no longer an example.
- Adds a `deploy` script wired to the real prod app definition id (`7xmKcOcvWM7l2vVhjUpmwB`), matching the pattern used by `bulk-exporter` and `locale-field-populator`, so the existing lerna/CircleCI pipeline can publish it.

## Test plan
- [ ] CI build/test/lint picks up `apps/audit-log-viewer` (previously skipped)
- [ ] `npm run deploy` (lerna) publishes the bundle to app definition `7xmKcOcvWM7l2vVhjUpmwB` on merge to master


primeinteger (approved):
👍

PR #11277, opened 2026-08-19
feat(bulk-exporter): rename to Content Exporter, add select-all-matching []

zacharyyankiver:
## Summary
- Rename the app from Entry Exporter to Content Exporter
- Add Gmail-style "select all matching this search" banner/flow to the results table, allowing selection beyond the current page and export via the existing filtered-query path
- Fix a font-size mismatch in the selection banner so the sentence text matches the inline action link's size

## Test plan
- [ ] `npx tsc --noEmit` passes
- [ ] Verify selection banner appears after selecting all rows on a page, with correct singular/plural wording
- [ ] Click "Select all N entries matching this search" and confirm banner updates to show total count with "Clear selection"
- [ ] Export while in select-all-matching mode and confirm all matching entries are exported (not just the current page)
- [ ] Confirm selection resets on search/page change
- [ ] Visually confirm banner sentence text and "Clear selection" / "Select all..." link now render at the same font size

### DIFF
--- a/apps/bulk-exporter/README.md
+++ b/apps/bulk-exporter/README.md
@@ -1,4 +1,4 @@
-# Entry Exporter for Contentful
+# Content Exporter for Contentful
 
 A Contentful App that allows you to export unlimited entries from any content type to **5 different formats** (CSV, JSON, XLSX, XML, YAML), bypassing the 40-entry limitation of the Contentful web interface.
 
@@ -144,7 +144,7 @@ npm run deploy -- --organization-id YOUR_ORG_ID --definition-id YOUR_APP_DEF_ID 
 ### Search & Preview
 
 1. Navigate to **Apps** in the Contentful web UI main menu
-2. Select **Entry Exporter**
+2. Select **Content Exporter**
 3. Use the tabbed interface to build your query:
 
 #### Filter Tab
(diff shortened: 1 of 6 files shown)


jjolton-contentful:
I did a fresh pass and found one behavioral mismatch in the new select-all flow:

In `ResultsList`, when `selectAllMatching` is active, the banner says "All N entries matching this search are selected" and the header checkbox is forced checked (around lines 656 and 735), but each row checkbox still renders from `selectedIds.includes(entry.sys.id)` (around line 846).

That means in select-all-matching mode, the UI can claim all entries are selected while row checkboxes appear unchecked, which is confusing and makes single-row deselection behavior non-obvious.

Would you be open to making row checkbox checked state include `selectAllMatching` as well (or otherwise visually indicating full-page selection in that mode)?

primeinteger (dismissed):
👍

jjolton-contentful:
Follow-up: I’m treating my checkbox-state finding as non-blocking UX consistency (not export correctness). The select-all-matching export path is correct. I’m approving so auto-merge can proceed; we can optionally clean up row-checkbox visual consistency in a follow-up.

jjolton-contentful (approved):

PR #11280, opened 2026-08-20
feat: scaffold PIC-1321 App Framework storage for Klaviyo field mappings

david-shibley-contentful:
## Summary
- Adds the app-side half of the PIC-1321 App Framework storage RFC to the Klaviyo app: a `storage` block in `contentful-app-manifest.json`, a local type shim for the not-yet-shipped `context.storage` DSL, and two new Functions (`getFieldMappings`, `setFieldMappings`) written against that contract.
- Purely additive/forward-looking: nothing is wired in yet. `entrySyncFunction.ts` and the existing frontend field-mapping UI (which currently store mappings as a stringified JSON blob in a `klaviyoFieldMappings` CMA entry, per the drift from ADR-0003) are untouched.
- **Known risk**: the manifest change is confirmed safe against the local `@contentful/app-scripts@2.3.0` build validation, but it's unverified whether the live Extensibility API's bundle-upload validation accepts the new top-level `storage` key. Do not run `upload-ci`/`upload-staging`/`deploy` with this change until that's confirmed against staging.

## Test plan
- [x] `npm run build:functions` succeeds locally with the new manifest `storage` block and both new Functions bundle cleanly
- [x] `npx tsc --noEmit` passes with no type errors against the installed `@contentful/node-apps-toolkit@3.15.0`
- [x] `npm test` run (no existing test files touch these new files; pre-existing suite has zero test files, so this is a no-op check)
- [ ] Confirm live Extensibility API bundle-upload validation accepts the new manifest `storage` key before ever running `upload-ci`/`deploy` on this branch

### DIFF
--- a/apps/klaviyo/functions/getFieldMappings.ts
+++ b/apps/klaviyo/functions/getFieldMappings.ts
@@ -1,0 +1,103 @@
+import type {
+  FunctionEventContext,
+  FunctionEventHandler,
+  FunctionTypeEnum,
+  AppActionRequest,
+  AppActionResponse,
+} from '@contentful/node-apps-toolkit';
+import type { WithStorage } from './types/storage';
+import {
+  getAllKlaviyoFieldMappings,
+  getEntryKlaviyoFieldMappings,
+} from '../src/utils/field-mappings-legacy';
+
+interface GetFieldMappingsParams {
+  entryId?: string;
+}
+
+interface FieldMappingRow {
+  entry_id: string;
+  content_type_id: string | null;
+  contentful_field_id: string;
+  klaviyo_block_name: string;
+  field_type: string | null;
+  locale: string | null;
+  is_asset_field: boolean | null;
+}
+
+// Legacy mapping objects use camelCase keys, and older ones fall back to
+// aliases (`id`, `name`, `type`) instead of the canonical field names — see
+// the same fallback logic in entrySyncFunction.ts.
+function toStorageRow(mapping: any): FieldMappingRow | null {
+  const contentfulFieldId = mapping.contentfulFieldId || mapping.id;
+  const klaviyoBlockName = mapping.klaviyoBlockName || mapping.name || contentfulFieldId;
+  if (!contentfulFieldId || !klaviyoBlockName) return null;
+  return {
+    entry_id: mapping.entryId ?? '',
+    content_type_id: mapping.contentTypeId ?? null,
+    contentful_field_id: contentfulFieldId,
+    klaviyo_block_name: klaviyoBlockName,
+    field_type: mapping.fieldType || mapping.type || null,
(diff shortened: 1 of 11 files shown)


apps/klaviyo/docs/ADRs/0007-app-framework-storage-migration.md
JuliRossi, line 1: ⚔️ I think we are exposing a bit too much about the innerworkings of app storage in this ADR. The idea was to use it for internal apps only.

apps/klaviyo/functions/getFieldMappings.ts
```diff
@@ -0,0 +1,103 @@
...
+  locale: string | null;
+  is_asset_field: boolean | null;
+}
+
+// Legacy mapping objects use camelCase keys, and older ones fall back to
+// aliases (`id`, `name`, `type`) instead of the canonical field names — see
+// the same fallback logic in entrySyncFunction.ts.
+function toStorageRow(mapping: any): FieldMappingRow | null {
+  const contentfulFieldId = mapping.contentfulFieldId || mapping.id;
+  const klaviyoBlockName = mapping.klaviyoBlockName || mapping.name || contentfulFieldId;
```
JuliRossi, line 33: 💭  I think we are doing this in `entrySyncFunction.ts` as well the `contentfulFieldId`/`klaviyoBlockName`/`fieldType` alias fallback in `toStorageRow` is basically the same logic. Perhaps this could be a shared helper both files import?

apps/klaviyo/src/utils/field-mappings-client.ts
```diff
@@ -0,0 +1,79 @@
...
+  mappings: any[]
+): Promise<void> => {
+  try {
+    const appActionId = await getAppActionIdByName(sdk, 'Set Field Mappings');
+    await sdk.cma.appActionCall.createWithResponse(
+      { appActionId, appDefinitionId: sdk.ids.app || '' },
+      { parameters: { entryId, mappings } }
+    );
+  } catch (error) {
+    logger.error('Error writing field mappings via Set Field Mappings action:', error);
```
JuliRossi, line 77: 💭 Do we want to surface this failure to the caller somehow? Right now if the App Action call fails we just log and return none of the four call sites (`ConfigScreen`, `FieldMappingScreen`, `FieldSelectDialog`, `Sidebar`) can tell the save didn't go through.

JuliRossi (changes requested):

apps/klaviyo/docs/ADRs/0007-app-framework-storage-migration.md
david-shibley-contentful: Good call — removed the ADR entirely in eab158857, nothing else in the codebase referenced it. Updated the PR description to drop the dangling link too.

PR #11281, opened 2026-08-20
Fix/add asset url root []

MikeDReynolds68:
## Purpose

This PR includes the following fixes:
`index.jsx`
 - adds comments to `selectedAssets` and `hideTreeNav` variables from exploded `config` in `renderDialog()` to denote their origination from instance parameters.
 - changes `hideAuthError()` to `hideAlert()` and updates all uses as non-error messages use this method.
 - changes `showAuthError()` to `showAlert()` and updates all uses as non-error messages use this method.
 - updates `customUpdateStateValue()` to prevent duplicates in result per Adobe recommendations.
 - updates `handleSelection()` to support error handling per Adobe recommendations.
 - updates `makeThumbnail()` to use `getRenditions()` and `getSafeRenditionUrl()` from `utils.js` for consistency.
 - updates `openDialog()` for improved stability per Adobe recommendations.
 - updates multiple installation parameters in `setup()` to include `required: false` and reordered List options so the default value is the first option.
 - removes `alwaysUseDMDelivery` parameter from `contentAdvisorProps` as it is not listed in the Content Advisor properties table in the online documentation and was originally added as a workaround based on feedback from RL. The `assetsUrlRoot` installation parameter solves for the issue this was originally added for. 
 - removes `expiryOptions` and `showToast` from `contentAdvisorProps` per Adobe recommendations.

`utils.js`
 - adds `getAssetId()` helper method to reduce logic duplication.
 - adds `getSafeRendition()` to retrieve the closest available rendition when no rendition matches the preferred size per Adobe PS recommendations.
 - adds `/as/<asset-name>` onto the `thumbUrl` in `getThumbUrl()` when the `assetsUrlRoot` installation parameter is provided.
 - updates `getMetadata()` for improved performance and stability per Adobe recommendations.
 - updates `getRenditions()` for improved performance and stability per Adobe recommendations.
 - updates `getThumbUrl()` to use `getSafeRendition()` when no `assetsUrlRoot` is provided.
 - updates `getThumbUrl()` to use `https://experience.adobe.com` as the thumbnail URL root when `assetsUrlRoot` is not provided and `asset.renditions` is not an array.
 - updates `pick()` for improved stability per Adobe recommendations.
 - updates `transformAssets()` for improved stability per Adobe recommendations.

## Approach

The fixes were applied based on customer requirements and recommendations from Adobe based on code review.

## Testing steps

Test 1
 - Navigate to the Adobe AEM Assets connector configuration, enter a value for the **Assets URL Root** parameter and save the configuration. 
 - Navigate to a content entry containing an Adobe Assets field and click the `Select assets from AEM` button.
 - In the Adobe Content Advisor modal, select an image and then click the `Select` button in the top right corner.
 - Inspect the thumbnail image displayed in the Contentful UI and confirm that the source matches the pattern `{{assetsUrlRoot}}/{{assetId}}/as/{{assetName}}`.

Test 2
 - Navigate to the Adobe AEM Assets connector configuration, remove the value for the **Assets URL Root** parameter and save the configuration.
- Navigate to a content entry containing an Adobe Assets field and click the `Select assets from AEM` button.
 - In the Adobe Content Advisor modal, select an image and then click the `Select` button in the top right corner.
 - Inspect the thumbnail image displayed in the Contentful UI and confirm that the source doe not match the pattern `{{assetsUrlRoot}}/{{assetId}}/as/{{assetName}}`.

## Breaking Changes

none

none

## Deployment

none

### DIFF
--- a/apps/aem-assets/src/utils.js
+++ b/apps/aem-assets/src/utils.js
@@ -30,13 +30,18 @@ export function getRenditions(asset) {
   return renditions.sort((a, b) => a.width - b.width);
 }
 
-export function transformAssets(assets) {
+export function transformAssets(assets, config) {
   const transformedAssets = assets.map((asset) => {
+    const assetId = asset['repo:assetId'] || asset['repo:id'] || asset.id;
     const renditions = getRenditions(asset);
-    const thumbUrl = renditions.find((r) => r.width >= 150 && r.width <= 400).href;
+    const urlRoot = config.assetsUrlRoot;
+
+    const thumbUrl = urlRoot
+      ? `${urlRoot}${!urlRoot.endsWith('/') ? '/' : ''}${assetId}`
+      : renditions.find((r) => r.width >= 150 && r.width <= 400).href;
 
     const transformedAsset = {
-      id: asset['repo:assetId'] || asset['repo:id'] || asset.id,
+      id: assetId,
       name: asset['repo:name'] || asset.name || '',
       url: thumbUrl,
       metadata: getMetadata(asset, renditions),
--- a/apps/aem-assets/src/utils.js
+++ b/apps/aem-assets/src/utils.js
@@ -1,62 +1,74 @@
 const ASSET_RENDITIONS_KEY = 'http://ns.adobe.com/adobecloud/rel/rendition';
+const ADOBE_EXPERIENCE_URL = 'https://experience.adobe.com/';
 
 /**
  * Source: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_pick
  */
 export function pick(object, keys) {
-  return keys.reduce((obj, key) => {
-    if (object && object.hasOwnProperty(key)) {
-      obj[key] = object[key];
+  return keys.reduce((result, key) => {
+    if (object && object.hasOwnProperty.call(object, key)) {
+      result[key] = object[key];
     }
-    return obj;
+    return result;
   }, {});
 }
 
+function getAssetId(asset) {
+  return asset['repo:assetId'] || asset['repo:id'] || asset.id;
+}
+
 export function getMetadata(asset, renditions) {
-  const { _embedded, _links, ...copy } = asset.computedMetadata;
-  const metadata = { ...copy, renditions };
-  return metadata;
+  const computed = asset?.computedMetadata || {};
+  const { embedded, links, ...copy } = computed;
+  return { ...copy, renditions };
 }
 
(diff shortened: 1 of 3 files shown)


apps/aem-assets/src/utils.js
```diff
@@ -1,57 +1,74 @@
...
+function getAssetId(asset) {
+  return asset['repo:assetId'] || asset['repo:id'] || asset.id;
+}
+
 export function getMetadata(asset, renditions) {
-  const { _embedded, _links, ...copy } = asset.computedMetadata;
-  const metadata = { ...copy, renditions };
-  return metadata;
+  const computed = asset?.computedMetadata || {};
+  const { embedded, links, ...copy } = computed;
```
mgoudy91: `getMetadata()`'s destructuring was renamed from `_embedded, _links` to `embedded, links` (no underscore), but `computedMetadata`'s actual keys are `_embedded`/`_links` (see the `_links[ASSET_RENDITIONS_KEY]` usage in `getRenditions` above). This means the rename no longer strips those fields — `_links`/`_embedded` raw HAL data will now leak into the `metadata` field that gets persisted on the entry. This looks like an unintentional regression from the rename rather than an intended behavior change. Can we restore the underscore-prefixed keys here so the exclusion actually works?

mgoudy91 (changes requested):
Two things flagged below:
1. `getMetadata()`'s destructuring rename looks like it broke the `_links`/`_embedded` exclusion — likely an unintentional regression, requesting a fix before merge.
2. A question on whether dropping `alwaysUseDMDelivery` from `contentAdvisorProps` was intentional (not mentioned in the PR description).

mgoudy91:
Quick question on `index.jsx`:`alwaysUseDMDelivery: aemTierType !== 'author'` was dropped from `contentAdvisorProps` in this PR and isn't mentioned in the description (which only calls out removing `expiryOptions`/`showToast` and making `repositoryId`/`aemTierType`/`env` conditional). Was removing `alwaysUseDMDelivery` intentional/part of Adobe's recommendation, or did it get swept up accidentally while restructuring that block?

MikeDReynolds68:
Thanks @mgoudy91 for catching and flagging the issue in `getMetadata()` and lack of documentation on the removal of the `alwaysUseDMDelivery` parameter. I have updated the notes in the PR for the removal of the `alwaysUseDMDelivery` parameter and pushed an update for the missing underscores.

mgoudy91 (approved):

PR #11291, opened 2026-09-08
fix: exclude archived-folder forms from Marketo picker [ES-642]

michaelpineirocontentful:
## Summary
The Adobe Marketo Form Selector was showing archived Marketo forms in its picker with no way to filter them out, flooding the dropdown for orgs with a long-lived Marketo instance. Marketo archives at the folder level (`folder.isArchive`), never on the form itself, so an archived form can still report `status: approved` — the picker's fetch had no way to distinguish it from a live form.

## Solution
- Cross-reference each form returned by `forms.json` against its folder's `isArchive` flag via `GET /rest/asset/v1/folder/{id}.json?type=Folder`
- Dedupe folder lookups so forms sharing a folder cost one request, not N
- Drop forms whose folder is archived; keep forms with no folder info as a defensive fallback for older/malformed data
- Extracted a small `marketoGet` helper to remove duplicated auth-header/JSON-parse boilerplate between the forms and folder requests

## Context
Marketo's REST API does not expose an archive flag on the form/asset itself — archiving is strictly a folder-level UI concept (confirmed against Adobe's own Assets/Folders API docs, and independently corroborated by an open Adobe feature request describing the identical gap for Programs). A form sitting in an archived folder keeps its normal `approved`/`draft` status, so the previous unfiltered fetch had no signal to act on.

Verified end-to-end against a live Marketo test instance by invoking the real, unmodified handler directly (not a reimplementation of its HTTP calls): before this fix, 25 of 95 forms returned (26.3%) were archived-folder leakage — all manually prefixed `z_archived_` by the org, since Marketo gives them no other way to hide them from integrations. After the fix, 0 leak through.

## Test plan
- [x] Added failing tests first, confirmed they failed against the pre-fix handler for the right reason
- [x] `getMarketoForms.spec.ts`: 6/6 passing, including new cases for archived-folder exclusion, deduped folder lookups, and the no-folder-info fallback
- [x] Full app suite: 31/31 passing
- [x] Typecheck and lint clean (one pre-existing, unrelated type gap in the test file's `HandlerResponse` helper type predates this change — confirmed via `git stash` — left untouched)
- [x] Verified against a live Marketo test instance using the real handler: 25/95 archived leakage before, 0/70 after

### DIFF
--- a/apps/marketo/functions/getMarketoForms.ts
+++ b/apps/marketo/functions/getMarketoForms.ts
@@ -6,16 +6,19 @@ import type {
 } from '@contentful/node-apps-toolkit';
 import { MarketoApiError } from './exceptions';
 import { getMarketoToken } from './getMarketoToken';
-import type { AppInstallationParameters, MarketoFormsResponse } from '../src/types';
+import type {
+  AppInstallationParameters,
+  MarketoApiResponse,
+  MarketoFolderRecord,
+  MarketoFormRecord,
+  MarketoFormsResponse,
+} from '../src/types';
 
-type MarketoForm = {
-  id: string;
-  url: string;
-  name: string;
-  folder?: { value: number };
-};
-
-const marketoGet = async (baseUrl: string, path: string, accessToken: string) => {
+const marketoGet = async <T>(
+  baseUrl: string,
+  path: string,
+  accessToken: string
+): Promise<{ response: Response; body: MarketoApiResponse<T> }> => {
   const response = await fetch(`${baseUrl}${path}`, {
     method: 'GET',
     headers: {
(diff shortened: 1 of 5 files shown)


jjolton-contentful (approved):
Solid fix — the folder-level archive cross-reference is the right call given Marketo doesn't expose archive state on the form itself, and the live-instance verification (25/95 leakage → 0) is convincing.

**What I liked:**
- TDD flow (failing tests first) and the typed API contract + fixture factories
- Deduped folder lookups — nice touch for orgs with shared folders
- Fail-open on missing folder info is a sensible defensive fallback

**Minor notes (non-blocking):**
- Fail-open when a folder lookup returns `success: false` is reasonable but implicit — a one-liner comment in `isFolderArchived` would help the next reader
- `Promise.all` on folder fetches means one transient network failure blanks the whole picker vs. partial results — probably fine in practice but worth knowing

LGTM 🚀

PR #11293, opened 2026-09-09
fix: validate origin and source on Klaviyo OAuth message handler []

michaelpineirocontentful:
## Summary
The window message listener that completes the Klaviyo OAuth flow accepts any `message` event delivered to the window, without checking where it came from or which window sent it. Combined with a same-origin sender, this closes the loop on validating the full message channel in both directions.

## Solution
- Reject any message in the OAuth completion handler that doesn't come from the app's own origin, or that doesn't come from the specific popup window this instance opened.

## Context
Stacked on top of the popup-side origin fix — this PR is the receiving half of the same channel. See that PR for the broader context on this stack.

## Test plan
- [x] `tsc --noEmit` passes
- [x] Frontend build (`vite build`) succeeds
- [ ] Manual OAuth connect flow verified in a staging environment


david-shibley-contentful (dismissed):
good call

ryunsong-contentful (dismissed):

ryunsong-contentful (approved):

PR #11294, opened 2026-09-09
fix: anchor Klaviyo proxy endpoint allowlist to the full path shape []

michaelpineirocontentful:
## Summary
The Klaviyo proxy function's endpoint allowlist only checked the path segment before the first slash. A value like `template-universal-content/../lists` passed that check, and the traversal segment let the request reach a Klaviyo API endpoint that was never allowlisted.

## Solution
- Replace the prefix-only check with a pattern anchored to the exact shape the app actually sends: an allowed endpoint name, optionally followed by a single id segment. Anything else is rejected outright.

## Context
Third in a small stack of hardening fixes to the Klaviyo app found during a routine security review. Verified the new pattern still matches every endpoint value the frontend currently sends (bare endpoint names and `endpoint/{id}`), while rejecting traversal and encoded-traversal variants.

## Test plan
- [x] `tsc --noEmit` passes
- [x] Functions build (`build-functions`) succeeds
- [x] Verified the allowlist pattern against real call sites plus traversal payloads

### DIFF
--- a/apps/klaviyo/functions/proxyRequest.ts
+++ b/apps/klaviyo/functions/proxyRequest.ts
@@ -7,8 +7,7 @@ import type {
 
 const KLAVIYO_API_URL = 'https://a.klaviyo.com/api';
 const KLAVIYO_API_REVISION = '2025-04-15';
-const ALLOWED_ENDPOINTS = ['template-universal-content', 'images'];
-const ALLOWED_ENDPOINT_PATTERN = new RegExp(`^(${ALLOWED_ENDPOINTS.join('|')})(/[A-Za-z0-9_-]+)?$`);
+const ALLOWED_ENDPOINT_PATTERN = /^(template-universal-content|images)(\/[A-Za-z0-9_-]+)?$/;
 
 type AppActionParameters = {
   endpoint: string;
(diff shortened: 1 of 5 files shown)


apps/klaviyo/functions/proxyRequest.ts
```diff
@@ -8,6 +8,7 @@ import type {
 const KLAVIYO_API_URL = 'https://a.klaviyo.com/api';
 const KLAVIYO_API_REVISION = '2025-04-15';
 const ALLOWED_ENDPOINTS = ['template-universal-content', 'images'];
```
david-shibley-contentful: Looks like this is no longer used, we can get rid of it

david-shibley-contentful (approved):
one comment, none blocking

ryunsong-contentful (approved):

apps/klaviyo/functions/proxyRequest.ts
michaelpineirocontentful: Good catch — inlined it into the pattern literal directly, no separate binding needed. Fixed in fa7f85058.

ryunsong-contentful (approved):

PR #11295, opened 2026-09-09
fix: stop writing the Klaviyo bearer token to function logs []

michaelpineirocontentful:
## Summary
The Klaviyo request helper logged the full outgoing request headers — including the `Authorization` bearer token it had just set — whenever the Klaviyo API returned a non-OK response. It also logged the raw token object on the invalid-token-format error path.

## Solution
- Strip the `Authorization` header before logging request headers on API errors.
- Log only the token's type (not its value) on the invalid-token-format path.

## Context
Last in a small stack of hardening fixes to the Klaviyo app found during a routine security review.

## Test plan
- [x] `tsc --noEmit` passes
- [x] Functions build (`build-functions`) succeeds

### DIFF
--- a/apps/klaviyo/functions/__tests__/klaviyoService.test.ts
+++ b/apps/klaviyo/functions/__tests__/klaviyoService.test.ts
@@ -23,7 +23,9 @@ describe('KlaviyoService.makeRequest token handling', () => {
       text: vi.fn().mockResolvedValue('{"errors":[{"detail":"boom"}]}'),
     });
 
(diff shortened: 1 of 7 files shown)


david-shibley-contentful (approved):
oh gosh, this one 🤦

ryunsong-contentful (approved):

david-shibley-contentful (approved):

ryunsong-contentful (approved):

PR #11305, opened 2026-09-15
CSS truncate status [ES-713]

ethan-ozelius-contentful:
## Purpose
Fix a CSS issue, where if the reason for an invalid link is "too many redirects, <redirect link>,<redirect link>,<redirect link>..." the table cell expanded to fit all the text.

**Solution:** add a max width and overflow css rule.

**BEFORE**
Screenshot 2026-09-15 at 3 10 21 PM

**AFTER**
Screenshot 2026-09-15 at 3 41 42 PM


primeinteger (approved):
👍

